PR #1692, opened 2019-01-24
fix: only repo collaborators can assign to self

mikeal:
When we open up the repo we can no longer assume that everyone in the repo is a collaborator.


phritz:
i think the new text is reasonably clear and this pr is getting really stale so am going to close.

PR #1755, opened 2019-02-01
Allow the chain LS command to consume a block height

mslipper:
This PR adds a `--since` parameter to `go-filecoin chain ls`. It allows users to surface only blocks after and including a specific block height. I wrote this for the network stats dashboard, but I think it'll be helpful to other users as well because `chain ls` currently returns the entire blockchain on every invocation.


mishmosh:
blocked by the chain API refactor in #1847

mslipper:
We likely don't need this anymore - piping `chain ls` to `head -n <num>` works just fine, and cancels the underlying HTTP connection.

mishmosh:
@mslipper In that case, is it OK to close this?

mslipper:
@mishmosh done :)

PR #1793, opened 2019-02-04
exclude test from PR limit

acruikshank:
### Problem

We limit the the length of PRs to 400 lines in our contributing guide. This is good because reviewing long PRs is burdensome to reviewers and slows progress. It is also bad because it disincentivizes testing. If a PR requires say 350 lines worth of non-test changes, I can either pull out and figure out how to break up the work, or undertest and get it through under the limit.

### Solution

Exclude new tests from this limit. New test code is easier to review and we don't want any barriers to contributing to our test suite.


phritz:
this pr is going nowhere fast, please reopen if you want to make a change

Issue #1054, opened 2018-10-02
use a signature for the ticket


phritz:
In https://github.com/filecoin-project/go-filecoin/pull/1048/ we introduced per-miner randomness into the ticket in order to fix a bug where all miners would win the challenge simultaneously because the ticket was the same for each. (The ticket at this point is not a signature, it's a hash of the input bytes, so not unique to each miner.)

Per the spec the ticket needs to be a signature: https://github.com/filecoin-project/specs/blob/master/expected-consensus.md#tickets

Make it so.

- [x] Part 1/3 Foundational work (#1795)
- [x] Part 2/3 Disable ticket checking, make tickets cryptographically signed (#2115)
- [ ] Part 3/3 Enable ticket checking (protocol breaking)

phritz:
@shannonwells it looks like 3/3 (enable ticket checking) is done https://github.com/filecoin-project/go-filecoin/blob/675344fd36517ba6ee06990784afc30129378daf/consensus/expected.go#L285 and this issue can be closed. Am I wrong? 

I do see some TODO's in here, eg https://github.com/filecoin-project/go-filecoin/blob/675344fd36517ba6ee06990784afc30129378daf/consensus/expected.go#L278, perhaps that's a reason to keep this open, to ensure we do those TODOs?

shannonwells:
No, it isn't done; the signer pubkey needs to be made part of the block, and checking whether it's a winning ticket currently doesn't include validating that it has been signed.  This is a protocol breaking change.

Once we put the pubkey into the block, we can implement signing blocks in the block mining worker and checking the block signature in consensus/expected,  but that should be a separate issue.

PR #1795, opened 2019-02-04
Part 1/3 Tickets --> Signed tickets #1054

shannonwells:
Part 1 of 3 for #1054 

1. In node.CreateMiner, generate a public/private key pair.  This will be the keypair for signing blocks and tickets. Store it in node.Wallet.  Alter the node's Mining config to store the block signer address so the keyinfo can be used.

    This assumes that there will be only one block miner;  if a subsequent call to CreateMiner is performed, right now it will simply generate a new key pair and overwrite the old one. I think this should be fine (?) since we will be publishing the minerAddress and its public key for verification in the block -- just not with this PR.

1. Add two parameters for a mining DefaultWorker struct and instantiation:
    * `blockSigningAddr address.Address`  -- this will be the address for the keypair generated above
    * `blockSigner types.Signer` -- this will be node.Wallet

Last, get all the code and tests working again.  This DOES NOT change how tickets are created; only the information available to a worker is changed.  This is to limit the size of the PRs to something manageable.

Additionally there is some clerical type stuff - killing a TODO-snake with a rename, and a small refactor of mining_test.

**PART 2** will consist of generating the ticket as a signature of the hashed Proof.   *Ticket validation will be temporarily disabled* until Part 3, because changing this generation will break a bunch of other things anyway, although not the protocol.

**PART 3** will consist of validating the ticket.  This will reinstate ticket validation, although it will this time do so in accordance with the spec, i.e., check that the ticket is a valid signature over the block proof, using the block public key.  This will break protocol since currently the public key is not part of blocks.  It'll cause some complicated issues with testing too, since now we can't generate a winning ticket so easily.

### DIFF
--- a/api/impl/mining.go
+++ b/api/impl/mining.go
@@ -54,10 +54,13 @@ func (api *nodeMining) Once(ctx context.Context) (*types.Block, error) {
 		}
 		return nd.Consensus.Weight(ctx, ts, pSt)
 	}
+
 	getAncestors := func(ctx context.Context, ts types.TipSet, newBlockHeight *types.BlockHeight) ([]types.TipSet, error) {
 		return chain.GetRecentAncestors(ctx, ts, nd.ChainReader, newBlockHeight, consensus.AncestorRoundsNeeded, consensus.LookBackParameter)
 	}
-	worker := mining.NewDefaultWorker(nd.MsgPool, getState, getWeight, getAncestors, consensus.NewDefaultProcessor(), nd.PowerTable, nd.Blockstore, nd.CborStore(), miningAddr, blockTime)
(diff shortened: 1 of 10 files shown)


shannonwells:
part 1 is done, the rest of the plan is:
**PART 2** will consist of generating the ticket as a signature of the hashed Proof. _Ticket validation will be temporarily disabled_ until Part 3, because changing this generation will break a bunch of other things anyway, although not the protocol.

**PART 3** will consist of validating the ticket. This will reinstate ticket validation, although it will this time do so in accordance with the spec, i.e., check that the ticket is a valid signature over the block proof, using the block public key. This will break protocol since currently the public key is not part of blocks. It'll cause some complicated issues with testing too, since now we can't generate a winning ticket so easily.

PR #1845, opened 2019-02-08
Miner payment management

### LINKED ISSUE #1433: Storage Miner Voucher Management
Every voucher received by the storage miner should be stored locally (ideally in a secure location, but this will come later).  A command should be added that allows the user to view all of the payment channel vouchers the miner has, including information about whether or not they can be redeemed yet.
Whether or not a client has attempted to close the channel should also be tracked. For now, just report this in the voucher view output, but make sure that the code that handles watching for this is easily changeable (in the future, we will want to act on this).

rkowalick:
We coalesce the the storage of storage deals for both storage miners and storage clients into a single datastore with a single namespace. This allows easy querying of storage deals, regardless of whether one is a miner or client. This also easily allows a miner to obtain vouchers for storage deals that have been successfully executed, thus enabling miner payments with minimal hoop-jumping.

Resolves #1433

### DIFF
--- a/api/impl/client.go
+++ b/api/impl/client.go
@@ -21,7 +21,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker"
 	"github.com/filecoin-project/go-filecoin/address"
 	mapi "github.com/filecoin-project/go-filecoin/api"
-	"github.com/filecoin-project/go-filecoin/protocol/storage"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -63,11 +63,11 @@ func (api *nodeClient) ImportData(ctx context.Context, data io.Reader) (ipld.Nod
 	return nd, bufds.Commit()
 }
 
-func (api *nodeClient) ProposeStorageDeal(ctx context.Context, data cid.Cid, miner address.Address, askid uint64, duration uint64, allowDuplicates bool) (*storage.DealResponse, error) {
+func (api *nodeClient) ProposeStorageDeal(ctx context.Context, data cid.Cid, miner address.Address, askid uint64, duration uint64, allowDuplicates bool) (*storagedeal.Response, error) {
 	return api.api.node.StorageMinerClient.ProposeDeal(ctx, miner, data, askid, duration, allowDuplicates)
 }
 
-func (api *nodeClient) QueryStorageDeal(ctx context.Context, prop cid.Cid) (*storage.DealResponse, error) {
+func (api *nodeClient) QueryStorageDeal(ctx context.Context, prop cid.Cid) (*storagedeal.Response, error) {
 	return api.api.node.StorageMinerClient.QueryDeal(ctx, prop)
 }
 
--- a/commands/client.go
+++ b/commands/client.go
@@ -13,7 +13,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/api"
-	"github.com/filecoin-project/go-filecoin/protocol/storage"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 )
 
 var clientCmd = &cmds.Command{
@@ -152,9 +152,9 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 
 		return re.Emit(resp)
 	},
-	Type: storage.DealResponse{},
+	Type: storagedeal.Response{},
 	Encoders: cmds.EncoderMap{
-		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, resp *storage.DealResponse) error {
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, resp *storagedeal.Response) error {
 			fmt.Fprintf(w, "State:   %s\n", resp.State.String())       // nolint: errcheck
 			fmt.Fprintf(w, "Message: %s\n", resp.Message)              // nolint: errcheck
 			fmt.Fprintf(w, "DealID:  %s\n", resp.ProposalCid.String()) // nolint: errcheck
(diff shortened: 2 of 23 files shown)


porcelain/deals.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/filecoin-project/go-filecoin/protocol/storage/deal"
+)
+
+type dlsPlumbing interface {
+	DealsLs() ([]*deal.Deal, error)
+}
+
+// DealByCid returns a single deal matching a given cid or an error
+func DealByCid(api dlsPlumbing, dealCid cid.Cid) (*deal.Deal, error) {
```
rkowalick: call this arg plumbing

node/node.go
```diff
@@ -409,6 +410,7 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 		MsgQueryer: msg.NewQueryer(nc.Repo, fcWallet, chainReader, &cstOffline, bs),
 		MsgSender:  msg.NewSender(nc.Repo, fcWallet, chainReader, msgPool, fsub.Publish),
 		MsgWaiter:  msg.NewWaiter(chainReader, bs, &cstOffline),
+		Deals:      dls.New(nc.Repo.DealsDatastore()),
```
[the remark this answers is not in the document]
rosalinekarr: I think this is a good idea and opens the door to some potentially useful features in the future. Having this datastore exposed without the protocol could allow users to list their active deals, view how much money each deal has made, export deals to a backup and import deals on another client with access to the same sector base.

api/impl/client.go
```diff
@@ -2,6 +2,7 @@ package impl
 
 import (
 	"context"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
```
phritz: out of order

commands/client.go
```diff
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
```
phritz: this one is out of order as well

node/block_propagate_test.go
```diff
@@ -143,7 +143,7 @@ func (r *zeroRewarder) GasReward(ctx context.Context, st state.Tree, minerAddr a
 }
 
 // makeNodes makes at least two nodes, a miner and a client; numNodes is the total wanted
-func makeNodes(ctx context.Context, t *testing.T, assertions *assert.Assertions, numNodes int) (address.Address, []*Node) {
```
rosalinekarr: 🎉 for removing unnecessary contexts!

plumbing/api.go
```diff
@@ -120,7 +125,17 @@ func (api *API) BlockGet(ctx context.Context, id cid.Cid) (*types.Block, error)
...
-// MessagePoolPending lists messages in the pool.
+// DealsLs a slice of all storagedeals in the local datastore and possibly an error
+func (api *API) DealsLs() ([]*storagedeal.Deal, error) {
+	return api.storagedeals.Ls()
+}
+
+// DealPut puts a given deal in the datastore
+func (api *API) DealPut(storageDeal *storagedeal.Deal) error {
+	return api.storagedeals.Put(storageDeal)
+}
```
rosalinekarr: I like where this is going. In the future, we could consider adding a `DealRemove` or `DealDelete` for letting users define their own data retention policies around deals.

We briefly discussed expiration policies around deals when we initially added deal persistence, but we never came up with a great solution for it. With this, we could potentially have an automatic delete after a few days for most users and let larger miners with capacity and interest in holding records for longer define their own policies through manual commands.

protocol/storage/storagedeal/state.go
```diff
@@ -1,13 +1,15 @@
-package storage
+package storagedeal
```
phritz: blocking: why do we have a new package for this? i suspect an import cycle which speaks to a deeper problem. i think the issue is likely that storage protocol depends on deal put/ls plumbing, which depend on storage protocol. this is a strong signal that deal put/ls do not belong in plumbing. the putter and getter should be direct, private dependencies of the storage protocol and should not be in the plumbing interface. commands that need deal ls should get it from the deal lser the storage protocol uses, which can be surfaced to them in storage protocol control api which will be separate from plumbing because it is higher level, which is part of https://github.com/filecoin-project/go-filecoin/issues/2030. for now we can keep it but @shannonwells can you please ack (and @rkowalick please ensure she does) that undoing this -- moving the deal back into the storage package as storage.deal and removing deal put/ls from plumbing -- is part of the scope for 2030? sorry to ramble here, i'm going ooo. and sorry @rkowalick this is not really something you could have forseen, we are underprepared to move protocol specific calls at this point, we'll have to play catch up.

plumbing/strgdls/store.go
```diff
@@ -0,0 +1,60 @@
...
+	return &Store{dealsDs: dealsDatastore}
+}
+
+// Ls returns a slice of deals matching the given query, with a possible error
+func (store *Store) Ls() ([]*storagedeal.Deal, error) {
+	var deals []*storagedeal.Deal
+
+	results, err := store.dealsDs.Query(query.Query{Prefix: "/" + StorageDealPrefix})
+	if err != nil {
+		return deals, errors.Wrap(err, "failed to query deals from datastore")
```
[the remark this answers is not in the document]
rkowalick: This `err` case is distinct from the query returning an empty result set, so I thought it might be wise to point it out.

plumbing/api.go
```diff
@@ -120,7 +125,17 @@ func (api *API) BlockGet(ctx context.Context, id cid.Cid) (*types.Block, error)
 	return api.chain.GetBlock(ctx, id)
 }
 
-// MessagePoolPending lists messages in the pool.
+// DealsLs a slice of all storagedeals in the local datastore and possibly an error
```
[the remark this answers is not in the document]
shannonwells: Yes, if we are going to make storage mining part of protocol then it will need to be moved out of here.

protocol/storage/storagedeal/state.go
shannonwells: Yes it makes sense to have everything that's interacting with storage deals use the upcoming protocol api, if that's what you're asking.

plumbing/strgdls/store.go
```diff
@@ -0,0 +1,60 @@
...
+// StorageDealPrefix is the datastore prefix for storage deals
+const StorageDealPrefix = "storagedeals"
+
+// New returns a new Store.
+func New(dealsDatastore repo.Datastore) *Store {
+	return &Store{dealsDs: dealsDatastore}
+}
+
+// Ls returns a slice of deals matching the given query, with a possible error
+func (store *Store) Ls() ([]*storagedeal.Deal, error) {
```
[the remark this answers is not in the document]
rkowalick: This has been added.

protocol/storage/storagedeal/state.go
rkowalick: @phritz I agree entirely that this is the core of the problem. It seems the protocol refactor stuff that @shannonwells will likely reorganize this stuff, so it seems like wasted effort to pull this stuff out of plumbing right now.

plumbing/api.go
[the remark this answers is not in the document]
rkowalick: ACK'd, comment I made below is heavily related to this thread too.

PR #1864, opened 2019-02-11
Update FAST retrieval test to work with miner payments

travisperson:
This test can now pass.


tools/fast/series/series_sendfilecoindefaults.go
```diff
@@ -0,0 +1,19 @@
+package series
```
phritz: series/series_foo as filename stutters

tools/fast/series/series_sendfilecoindefaults.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+)
+
+// SendFilecoinDefaults sends the `value` amount of fil from the default wallet
+// address of the `from` node to the `to` node's default wallet.
+func SendFilecoinDefaults(ctx context.Context, from, to *fast.Filecoin, value int) error {
+	var toAddr address.Address
+	if err := to.ConfigGet(ctx, "wallet.defaultAddress", &toAddr); err != nil {
```
phritz: we should have a command that returns the default address using a plumbing call that is higher level than config because it might not yet be set. maybe file an issue on the plumbing/porcelain epic to have plumbing that returns the default address.

tools/fast/series/series_sendfilecoindefaults.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+)
+
+// SendFilecoinDefaults sends the `value` amount of fil from the default wallet
+// address of the `from` node to the `to` node's default wallet.
+func SendFilecoinDefaults(ctx context.Context, from, to *fast.Filecoin, value int) error {
```
phritz: note this can only sent integer number of fil.

tools/fast/series/series_sendfilecoindefaults.go
travisperson: I will mass rename these later.

tools/fast/series/series_sendfilecoindefaults.go
travisperson: https://github.com/filecoin-project/go-filecoin/issues/2049

tools/fast/series/series_sendfilecoindefaults.go
travisperson: This is for API consistency with the other series used in this function which was added prior to our understanding of how `message send` actually worked. This will be updated later.

PR #1896, opened 2019-02-13
add default values for gas stuff

whyrusleeping:
This makes it much nicer to use filecoin.

### DIFF
--- a/commands/main.go
+++ b/commands/main.go
@@ -332,7 +332,7 @@ func parseGasOptions(req *cmds.Request) (types.AttoFIL, types.GasUnits, bool, er
 
 		price = p
 	} else {
-		price = types.NewAttoFIL(big.NewInt(10000))
+		price = types.NewAttoFIL(big.NewInt(300))
 	}
 
 	var gasLimit types.GasUnits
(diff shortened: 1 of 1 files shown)


phritz:
@mishmosh can you expand on what you think needs to change re cli and wiki?

mishmosh:
If we move forward with this PR here are the docs that need updating:
- wiki mining tutorial, explicit gas values can be deleted from the example
- wiki commands page, gas values can be in brackets to show they're optional
- CLI helptext for this command, gas values should be in brackets to show they are optional

laser:
@acruikshank @anorth - Is this PR dead? If not: What needs to happen to move it forward?

anorth:
It could move forward if someone commits to addressing the comments and updating documentation after landing it. Alternatively, I don't think it's high priority, so could be abandoned in favour of another go later.

Future client libraries will probably automate some of this with chain-inspection based estimates of the gas market price.

frrist:
Closing as this is now being addressed via https://github.com/filecoin-project/go-filecoin/pull/3118

PR #1899, opened 2019-02-13
use paramfetch before paramcache

sidke:
has `deps` use `paramfetch` to pull any predefined parameters from ipfs before generating any missing ones with `paramcache`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ tools/faucet/faucet
 tools/aggregator/aggregator
 tools/genesis-file-server/genesis-file-server
 proofs/misc/parameters.json
-proofs/bin/paramfetch
 proofs/bin/paramcache
 proofs/bin/paramfetch
 proofs/include/libfilecoin_proofs.h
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -12,6 +12,8 @@ fixtures/genesis.car
 tools/faucet/faucet
 tools/aggregator/aggregator
 tools/genesis-file-server/genesis-file-server
+proofs/misc/parameters.json
+proofs/bin/paramfetch
```
phritz: i think this is already below

ognots:
So long as we're using HTTPS to access the gateway (mitigating MITM attacks) doesn't IPFS content addressing provide suitable integrity checking? I'm still not up to speed on the ins and outs of using the IPFS gateway, so there's definitely attack vectors I could be unaware of.

Otherwise LGTM

travisperson:
> So long as we're using HTTPS to access the gateway (mitigating MITM attacks) doesn't IPFS content addressing provide suitable integrity checking?

It guarantees that the gateway is serving the correct content, but does nothing to ensure that it was downloaded correctly through http.

PR #1943, opened 2019-02-14
don't stop bootstrapping 20s in please

ZenGround0:
Talking with @anacrolix it is a bad thing to kill the bootstrapping context early.  This should only go down when the node goes down.

### DIFF
--- a/filnet/bootstrap.go
+++ b/filnet/bootstrap.go
@@ -39,7 +39,7 @@ type Bootstrapper struct {
 	r routing.IpfsRouting
 	// Does the work. Usually Bootstrapper.bootstrap. Argument is a slice of
 	// currently-connected peers (so it won't attempt to reconnect).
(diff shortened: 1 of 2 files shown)


filnet/bootstrap.go
```diff
@@ -94,28 +94,28 @@ func (b *Bootstrapper) Stop() {
...
+func (b *Bootstrapper) bootstrap(currentPeers []peer.ID) {
 	peersNeeded := b.MinPeerThreshold - len(currentPeers)
 	if peersNeeded < 1 {
 		return
 	}
 
-	ctx, cancel := context.WithTimeout(b.ctx, b.ConnectionTimeout)
 	var wg sync.WaitGroup
 	defer func() {
 		wg.Wait()
```
phritz: we have to wait until we connect to the bootstrap peers before starting dht routing?

PR #1944, opened 2019-02-14
copy groth params to params.out in smartdeps function

ognots:
also remove hydrateParamCache function because it's no longer needed

missed an earlier comment from @laser pointing out we should also be copying the groth params to params.out

https://github.com/filecoin-project/go-filecoin/pull/1928#discussion_r256636904


scripts/copy-groth-params.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 copy_groth_params() {
-  mkdir -p /tmp/filecoin-proof-parameters
-  pushd /tmp/filecoin-proof-parameters
+  local params_dir="${FILECOIN_PARAMETER_CACHE:-"/tmp/filecoin-proof-parameters/"}"
```
ognots: this was necessary to make ci test pass

phritz:
@ognots does this still make sense in light of #1899 ? can you rebase and advise?

PR #1948, opened 2019-02-14
docs: Add dual license

RichardLitt:
Missing it in the manifest. Not sure if this is how GX reads it (cc @whyrusleeping), but I'm going off of [this example from @kemitchell](https://www.npmjs.com/package/dual-licensed-package).

### DIFF
--- a/package.json
+++ b/package.json
@@ -198,7 +198,7 @@
   ],
   "gxVersion": "0.12.1",
   "language": "go",
-  "license": "(MIT OR APACHE)",
+  "license": "(MIT OR APACHE-2.0)",
   "name": "go-filecoin",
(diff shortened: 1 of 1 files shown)


phritz:
@dignifiedquire

package.json
```diff
@@ -198,7 +198,7 @@
   ],
   "gxVersion": "0.12.1",
   "language": "go",
-  "license": "",
+  "license": "(MIT OR APACHE)",
```
dignifiedquire: should be `Apache 2` I believe

RichardLitt: https://spdx.org/licenses/ says Apache-2.0. Changed!

dignifiedquire: perfect, thank you

Issue #1950, opened 2019-02-14
Build deps leaves junk in working directory


anorth:
**Describe the bug**
After building deps on a clean checkout, some files are created in the working directory. These files are not ignored by git and so complicate making clean commits.

**To Reproduce**
1. Checkout fresh working directory
2. `go run ./build/*.go deps`
3. `git status`

**Expected behavior**
No untracked or modified files.

**Screenshots**
```
(anorth/mpool+)$ git status
On branch anorth/mpool
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	proofs/bin/paramfetch
	proofs/misc/

nothing added to commit but untracked files present (use "git add" to track)
```

**Additional context**
This regression happened a couple of days ago when the new proofs integration landed.

ognots:
No longer necessary, it's been since remedied by:
https://github.com/filecoin-project/go-filecoin/blob/master/.gitignore#L16
https://github.com/filecoin-project/go-filecoin/blob/master/.gitignore#L20

PR #1953, opened 2019-02-14
#1950 cleanup artifacts left by build deps

ognots:
leaves behind artifacts not listed in .gitignore or referenced by any code so delete them


phritz:
@sidke can you take a look?

ognots:
No longer necessary, it's been since remedied by:
https://github.com/filecoin-project/go-filecoin/blob/master/.gitignore#L16
https://github.com/filecoin-project/go-filecoin/blob/master/.gitignore#L20

Issue #1933, opened 2019-02-14
Add command for viewing a message in the mpool


ognots:
### Description

There is no way to get information about messages listed in `go-filecoin mpool ls`

Add a command to view message details

### Acceptance criteria

As a go-filecoin user, I should be able to view the serialized json output of a message

### Risks + pitfalls

TBD

### Where to begin

Consider using the Message `String` method https://github.com/filecoin-project/go-filecoin/blob/master/types/message.go#L62-L73

PR #1954, opened 2019-02-14
Move mpool from node API to plumbing

### LINKED ISSUE #2017: Move message pool to plumbing
### Description
The message pool is a core component that should be available to porcelain and protocols.

As a policy, we should move things off the Node interface before adding more functionality, so this is a prereq to #1933.

### Acceptance criteria
- Message pool is removed from the node interface/object
- Message pool commands use plumbing/porcelain

anorth:
Closes #2017. A prefactor for #1933


plumbing/api.go
```diff
@@ -60,7 +63,7 @@ func New(deps *APIDeps) *API {
 
 		chain:        deps.Chain,
 		config:       deps.Config,
-		messagePool:  deps.MessagePool,
+		mpool:        mpool.New(deps.MessagePool, deps.Pubsub),
```
rosalinekarr: Can we call it `msgPool` for consistency?

plumbing/api.go
```diff
@@ -29,7 +31,7 @@ type API struct {
 
 	chain        *chn.Reader
 	config       *cfg.Config
-	messagePool  *core.MessagePool
+	mpool        *mpool.Pool
```
phritz: mpl is idiomaticer, but up to you

plumbing/mpool/pool.go
```diff
@@ -0,0 +1,49 @@
...
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	"gx/ipfs/QmVRxA4J3UPQpw74dLrQ6NJkfysCA1H4GU28gVpXQt9zMU/go-libp2p-pubsub"
+
+	"github.com/filecoin-project/go-filecoin/core"
+	"github.com/filecoin-project/go-filecoin/plumbing/msg"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// Pool interfaces to the core message pool
+type Pool struct {
```
phritz: mpool.pool stutters pretty bad. we could actually stick this in the msg package and have it be msg.pool. alternatively, remove could just be implemented by having the plumbing api call directly into the core.messagepool. then this could be an mpl.lser or mpl.viewer.

plumbing/mpool/pool.go
```diff
@@ -0,0 +1,49 @@
...
+	pubsub *pubsub.PubSub
+}
+
+// New builds a new Pool.
+func New(pool *core.MessagePool, pubsub *pubsub.PubSub) *Pool {
+	return &Pool{pool: pool, pubsub: pubsub}
+}
+
+// Pending lists un-mined messages in the pool
+func (ls *Pool) Pending(ctx context.Context, messageCount uint) ([]*types.SignedMessage, error) {
```
phritz: should probably say what messagecount does and is for, waits for that many messages and its used for testing. 

actually this is a pretty weird thing to have in the interface. if we didn't have to support it we could just have the core.messagepool in the plumbing and we wouldn't need this wrapper. if pubsub was available in plumbing then a porcelain call could implement this wait functionality by depending on the plumbing pending() and messagepubsub() calls.

plumbing/mpool/pool.go
```diff
@@ -0,0 +1,49 @@
...
+	pending := ls.pool.Pending()
+	if len(pending) < int(messageCount) {
+		subscription, err := ls.pubsub.Subscribe(msg.Topic)
```
phritz: maybe subscribe before calling pending in case a message arrives in between these calls as written?

plumbing/mpool/pool.go
phritz: blocking: we should have a unit test for the message count functionality

phritz:
the more i think about the message counting the more i think we should push it into the tests somehow, and out of the interface...

anorth:
Ok, I'm going to pause this PR and move pubsub into plumbing first. #2018

anorth:
I have rebased this against #2019 and pulled the wait functionality up to porcelain, where it's now tested. I removed the `plumbing/mpl` package in favour of just putting the `core.MessagePool` on the plumbing struct.

Ready for review again.

plumbing/mpool/pool.go
anorth: Gone

plumbing/mpool/pool.go
anorth: Done!

plumbing/mpool/pool.go
anorth: I've added an additional call to pending before the first `Next()`, preserving the fast path that avoids subscribing at all unless necessary.

porcelain/mpool_test.go
```diff
@@ -0,0 +1,95 @@
+package porcelain_test
+
+import (
+	"context"
+	"gx/ipfs/QmVRxA4J3UPQpw74dLrQ6NJkfysCA1H4GU28gVpXQt9zMU/go-libp2p-pubsub"
```
phritz: this one belongs below

plumbing/ps/testing.go
```diff
@@ -0,0 +1,83 @@
...
+	return s.topic
+}
+
+// Next returns the next messages from this subscription.
+func (s *FakeSubscription) Next(ctx context.Context) (*pubsub.Message, error) {
+	if s.err != nil {
+		return nil, s.err
+	}
+	select {
+	case msg := <-s.pending:
```
phritz: pointing this out now not bc i think it is important here bc it doesnt seem so, but because this is one of the sharp edges of go channels: receiving on a channel actually returns two values. the second is a bool signifying the channel is still open. if the second is false then the channel was closed by caller and the first received value is the zero value for the type. i have been bitten by this many times so make a point of mentioning it to anyone who might not have had the pleasure of hitting it.

PR #1992, opened 2019-02-15
don't compile binaries during docker build

ognots:
* add an extra step in circle config to generate faucet and genesis-file-server and persist to workspace
* attach workspace and param cache to build_docker_img
* COPY binaries and params to docker image
* build go-filecoin container last to avoid adding params to faucet and genesis-file-server docker context
* make build_docker_img dependent on build_faucet_and_genesis
* remove rust dependencies from base docker image

the intention of these changes is to speed up the docker build time and make the docker image smaller

-----

this should be tested before merging by removing the branch filtering on `build_faucet_and_genesis` and `build_docker_img` jobs of `test_all` workflow, and commenting out the docker pushing and tagging steps

a second test would be to trigger a test devnet deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,11 +233,14 @@ jobs:
       - run:
           name: Create linux bundle
           command: ./scripts/build-bundle.sh
+
       - store_artifacts:
           path: "/go/src/github.com/filecoin-project/go-filecoin/filecoin-Linux.tar.gz"
           destination: filecoin-Linux.tar.gz
+
       - store_test_results:
           path: test-results
+
       - persist_to_workspace:
           root: "."
           paths:
@@ -266,6 +269,21 @@ jobs:
           command: |
             ./scripts/publish-release.sh
 
+  build_faucet_and_genesis:
+    docker:
+      - image: circleci/golang:1.11.1
+    working_directory: /go/src/github.com/filecoin-project/go-filecoin
+    resource_class: small
+    steps:
+      - command: |
+          go build -o ./faucet ./tools/faucet/main.go
+          go build -o ./genesis-file-server ./tools/genesis-file-server/main.go
+
+      - persist_to_workspace:
+          root: "."
+          paths:
+            - "faucet"
+            - "genesis-file-server"
 
   build_docker_img:
     docker:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -328,6 +345,16 @@ jobs:
             docker tag 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:$SHORT_GIT_SHA 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
             docker push 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
 
+      - run:
+          name: build & push image - filecoin
+          command: |
+            mv $HOME/filecoin-proof-parameters ./filecoin-proof-parameters
```
travisperson: How are the params being generated at `$HOME/filecoin-proof-parameters`. I don't see anything that would change them from their default of `/tmp/filecoin-proof-parameters`. You can use `FILECOIN_PARAMETER_CACHE` to change it though. I believe setting it during `deps` should work.

Dockerfile
```diff
@@ -12,17 +11,6 @@ ENV SRC_DIR /go/src/github.com/filecoin-project/go-filecoin
...
-RUN cd $SRC_DIR \
-&& . $HOME/.cargo/env \
-&& go run ./build/*go deps \
-&& go run ./build/*go build \
-&& go build -o ./faucet ./tools/faucet/main.go \
-&& go build -o ./genesis-file-server ./tools/genesis-file-server/main.go
-
-# Build gengen
-RUN cd
-
```
travisperson: I'm commenting here because I can't below, but on line 47 of this new file there is an `COPY --from=builder /tmp/filecoin-proof-parameters/* /tmp/filecoin-proof-parameters/` that can be removed if we are no longer generating proofs here.

.circleci/config.yml
ognots: @travisperson it gets set here https://github.com/filecoin-project/go-filecoin/blob/master/.circleci/config.yml#L178

I had found that out the hard way when trying to symlink the groth parameters in `smartdeps` and they weren't found in `/tmp` in CI

Dockerfile
ognots: Nice catch, thanks! I'll update on Tuesday and then try testing it

.circleci/config.yml
gmasgras: Echoing @travisperson's observations, while the files in `filecoin-proof-parameters/` will be present in the `docker build` context, I believe that we also need to set `FILECOIN_PARAMETER_CACHE=filecoin-proof-parameters` in  `Dockerfile` or as a `build-arg` so that the build process sees that those files already exist and don't need to be regenerated

travisperson: Ah it was already set prior to this PR see line 15

```
echo 'export FILECOIN_PARAMETER_CACHE="${HOME}/filecoin-proof-parameters/"' >> $BASH_ENV
```

gmasgras: that only sets the ENV for the current shell running on CI, it doesn't automatically get set inside the Docker build AFAIK

.circleci/config.yml
```diff
@@ -292,41 +315,52 @@ jobs:
...
-            docker tag 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:$SHORT_GIT_SHA 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
-            docker push 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
+#            docker push 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:$SHORT_GIT_SHA
+#            docker tag 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:$SHORT_GIT_SHA 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
+#            docker push 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin-faucet:latest
+
+      - run:
+          name: build & push image - filecoin
+          command: |
+            mv $HOME/filecoin-proof-parameters ./filecoin-proof-parameters
```
gmasgras: > Ah it was already set prior to this PR see line 15
> 
> ```
> echo 'export FILECOIN_PARAMETER_CACHE="${HOME}/filecoin-proof-parameters/"' >> $BASH_ENV
> ```

that only sets the ENV for the current shell running on CI, it doesn't automatically get set inside the Docker build context AFAIK

.circleci/config.yml
travisperson: Sorry, I jumped over your comment, I was replying to @ognots and my original.

.circleci/config.yml
```diff
@@ -266,6 +274,22 @@ jobs:
...
 
+  build_faucet_and_genesis:
+    docker:
+      - image: circleci/golang:1.11.1
+    working_directory: /go/src/github.com/filecoin-project/go-filecoin
+    resource_class: small
+    steps:
+      -run:
+         command: |
+           go build -o ./faucet ./tools/faucet/main.go
```
gmasgras: I believe that this won't build without running `go run ./build/*go deps` first.  
Maybe `build_linux` is a better place for this since we run `go deps` there already (and it's slow)

travisperson:
I think this all looks pretty good. I do have one request, which may change some stuff. I would like to make sure we document how to build these containers manually now, as sometimes it's required. It use to be simple enough that we didn't really have any documentation. Checkout the commit, run the docker build commands

Now we have more than just source dependencies for the build. There are some additional steps and more than just `go run build/*.go deps && go run build/*.go build` (mostly around proof params).

Dockerfile
gmasgras: I think that this approach to not building anything in the Dockerfile results in a tight couple with Circle and will break the use-case of someone trying to build a Docker image on their machine. Not sure how important that is for the users but I think it would be useful for devs.

ognots: @gmasgras @travisperson I agree on needing to leaving Dockerfile alone for manual builds.
Before seeing this I already backed out changes to `Dockerfile` and created a new CI specific file `Dockerfile.ci.base` based on debian stretch

ognots:
@travisperson see comment https://github.com/filecoin-project/go-filecoin/pull/1992#discussion_r258692014 on manual builds.

since I've backed out changes to Dockerfile, I'm going to create a separate issue for adding instructions to manually build a Docker image

ognots:
Here's a passing build and push (only commented out pushing `:latest` tag https://circleci.com/workflow-run/02b50237-2046-4aa3-86a0-53f16f10b43d

Working on rebasing and squashing commit history and then will be ready for a final review

.circleci/config.yml
ognots: @gmasgras @travisperson the proof parameters get pulled in via docker `COPY` https://github.com/filecoin-project/go-filecoin/blob/f6a77f7a22199f84b0d2c62ad3b436ebc6ed553d/Dockerfile.ci.filecoin#L6
and the parameters are moved to docker context here https://github.com/filecoin-project/go-filecoin/blob/f6a77f7a22199f84b0d2c62ad3b436ebc6ed553d/.circleci/config.yml#L373

ognots: the files are moved to `/tmp/filecoin-proof-parameters` as expected by default.

it might be better to copy the proof params to `/usr/share/filecoin-proof-parameters` and then set `ENV FILECOIN_PARAMETER_CACHE /usr/share/filecoin-proof-parameters` in Dockerfile

.circleci/config.yml
ognots: instead I added deps from cache

ognots:
@gmasgras @travisperson ready for final review before merging

ognots:
I just accidentally fubared the git history of this branch. Going to create a new feature branch and re-commit more concisely rather than try rebasing further

ognots:
Closing in favour of https://github.com/filecoin-project/go-filecoin/pull/2060

Issue #2003, opened 2019-02-16
[bug] Asset failed to be downloaded


NatoBoram:
**Describe the bug**

Command `FILECOIN_USE_PRECOMPILED_RUST_PROOFS=true go run ./build/*.go deps` fails.

**To Reproduce**

1. `sudo apt install snapd pkg-config`
2. `sudo snap install go --classic`
3. `curl https://sh.rustup.rs -sSf | sh`
4. `go get -u -v -fix github.com/filecoin-project/go-filecoin`
5. `cd ~/go/src/github.com/filecoin-project/go-filecoin`
6. `FILECOIN_USE_PRECOMPILED_RUST_PROOFS=true go run ./build/*.go deps`
7. See error

**Expected behavior**

Dependencies should be installed.

**Screenshots**

```bash
./scripts/install-rust-proofs.sh
using precompiled rust-proofs
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5685  100  5685    0     0  24399      0 --:--:-- --:--:-- --:--:-- 24399
./scripts/install-rust-proofs.sh: line 10: jq: command not found
./scripts/install-rust-proofs.sh: line 18: jq: command not found
curl: (3) <url> malformed
basename: missing operand
Try 'basename --help' for more information.
curl: (3) <url> malformed
asset failed to be downloaded
failed to find or obtain precompiled rust-proofs, falling back to local
~/go/src/github.com/filecoin-project/go-filecoin/proofs/rust-proofs ~/go/src/github.com/filecoin-project/go-filecoin
cargo 1.32.0-nightly (5e85ba14a 2018-12-02)
    Updating crates.io index
    Updating git repository `https://github.com/zcash-hackworks/sapling-crypto`
    Updating git repository `https://github.com/dignifiedquire/rust-gperftools`
    Updating git repository `https://github.com/filecoin-project/merkle_light`
    Updating git repository `https://github.com/gtank/blake2-rfc`
   Compiling filecoin-proofs v0.1.0 (/home/nato/go/src/github.com/filecoin-project/go-filecoin/proofs/rust-proofs/fileco
in-proofs)
error: failed to run custom build command for `filecoin-proofs v0.1.0 (/home/nato/go/src/github.com/filecoin-project/go-
filecoin/proofs/rust-proofs/filecoin-proofs)`
[...]

PR #2006, opened 2019-02-16
rust: stop install script if anything fails

schomatis:
Related to https://github.com/filecoin-project/go-filecoin/issues/2003.

### DIFF
--- a/scripts/install-rust-proofs.sh
+++ b/scripts/install-rust-proofs.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
 install_precompiled() {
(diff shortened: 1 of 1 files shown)


laser:
@schomatis - This change breaks the fallback-to-local-build behavior of the `install-rust-fil-proofs.sh` script.

The script [checks the status code of the `install_precompiled` function](https://github.com/filecoin-project/go-filecoin/pull/2006/files#diff-1ec1662638c9438aa6cbb68a70093190R82) and, if `install_precompiled` returns a non-zero code, falls back to a local build.

Adding the `set -e` flag short-circuits the fallback behavior.

PR #2015, opened 2019-02-17
Feat/payment channel porcelain 1942

rosalinekarr:
# Problem

The code is a mess.

# Solution

Refactor.


phritz:
also fyi @ZenGround0 is out, suggest shannonwells or ac as reviwer

PR #2019, opened 2019-02-18
Move pubsub to plumbing and remove from Node interface/object.

### LINKED ISSUE #2018: Move pubsub into plumbing
### Description
The libp2p pubsub object is a core component that should be made available to porcelain and higher-level calls, without needing a Node. It's needed for a clean movement of message pool functionality into plumbing/porcelain (#2017)

Note that the pubsub object may also be curried in to other plumbing interfaces like the `MsgSender`.

### Acceptance criteria
- Pubsub functionality is available to porcelain and higher level components without reference to the Node

anorth:
Closes #2018.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -53,6 +53,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/plumbing/msg"
 	"github.com/filecoin-project/go-filecoin/plumbing/mthdsig"
 	"github.com/filecoin-project/go-filecoin/plumbing/ntwk"
+	"github.com/filecoin-project/go-filecoin/plumbing/ps"
 	"github.com/filecoin-project/go-filecoin/porcelain"
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
@@ -71,8 +72,6 @@ var filecoinDHTProtocol dhtprotocol.ID = "/fil/kad/1.0.0"
 var log = logging.Logger("node") // nolint: deadcode
 
 var (
-	// ErrNoRepo is returned when the configs repo is nil
-	ErrNoRepo = errors.New("must pass a repo option to the node build process")
 	// ErrNoMinerAddress is returned when the node is not configured to have any miner addresses.
 	ErrNoMinerAddress = errors.New("no miner addresses configured")
 )
@@ -121,7 +120,6 @@ type Node struct {
 	RetrievalMiner  *retrieval.Miner
 
 	// Network Fields
-	PubSub       *pubsub.PubSub
 	BlockSub     *pubsub.Subscription
 	MessageSub   *pubsub.Subscription
 	Ping         *ping.PingService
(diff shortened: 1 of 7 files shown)


node/node.go
```diff
@@ -412,6 +410,8 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 		MsgQueryer:   msg.NewQueryer(nc.Repo, fcWallet, chainReader, &cstOffline, bs),
 		MsgSender:    msg.NewSender(nc.Repo, fcWallet, chainReader, msgPool, fsub.Publish),
 		MsgWaiter:    msg.NewWaiter(chainReader, bs, &cstOffline),
+		Subscriber:   ps.NewSubscriber(fsub),
+		Publisher:    ps.NewPublisher(fsub),
```
anorth: Tell me if this is overkill and the plumbing should just have a `libp2p.pubsub.PubSub` object. Higher up I would assert that publish and subscribe interfaces should be separate, and "third-party" libraries should be abstracted over, but low down it's not such a big deal.

phritz: dont think its overkill, i think we should be doing more of this

plumbing/api.go
```diff
@@ -48,6 +52,8 @@ type APIDeps struct {
 	MsgQueryer   *msg.Queryer
 	MsgSender    *msg.Sender
 	MsgWaiter    *msg.Waiter
+	Subscriber   *ps.Subscriber
```
phritz: previously it had alpha ordering, this is a tossup whether to keep it

plumbing/api.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	"gx/ipfs/QmVRxA4J3UPQpw74dLrQ6NJkfysCA1H4GU28gVpXQt9zMU/go-libp2p-pubsub"
```
phritz: we could and maybe should take this a step further by defining a local pubsub package at the top level and an interface in that package that all consumers in gfc uses and then have a constructor in that package that returns an instance of this pubsub. that way if we change the implementation out it is only a small change. could do as a follow on or not, up to you.

plumbing/api.go
```diff
@@ -145,6 +153,16 @@ func (api *API) MessageWait(ctx context.Context, msgCid cid.Cid, cb func(*types.
 	return api.msgWaiter.Wait(ctx, msgCid, cb)
 }
 
+// PubSubSubscribe subscribes to a libp2p topic
```
phritz: i think it says libp2p topic to indicate to the reader that this is a network pubsub, not something local to this process. i think that is a good idea, though i might say 'subscribes to a topic for notifications from the filecoin network'

plumbing/ps/publisher.go
```diff
@@ -0,0 +1,18 @@
+package ps
```
phritz: a super thin wrapper like this makes me think we should consider not having it, instead just defining in the plumbing.api the publisher and subscriber interfaces we expect a ps dependency passed in to support and then to pass calls directly to that thing rather than through this wrapper

rosalinekarr:
I've been trying to use the `ntwk` plumbing component as a sort of wrapper around everything libp2p we pull off the node. My motivation is to reduce name collisions with stuff like "node" going forward and later on to help us refactor the `buildHost` method which makes heavy use of libp2p in `node/node.go`.

I think it might be a good idea to move these methods into `ntwk` plumbing too, but I realize that's a pretty significant change here. @phritz @anorth, what do you think?

plumbing/api.go
anorth: I'll restore it with a rename in combination with other follow-ups, something like `NetPublisher/NetSubscriber`

anorth:
I like all these larger suggestions. I would like to merge this now since I've worked chained off it, but will follow up to:
- extract top-level pubsub package abstracting over libp2p
- move the `ps` plumbing to the `ntwk` package

phritz:
sounds great, bombs away

PR #2022, opened 2019-02-18
Remove unnecessary plumbing.chn subpackage with pass-through wrapper objects

anorth:
@phritz I think you suggested removing the `Reader` wrapper struct, and following that through removed the whole `chn` package.

I'm in favour because code is evil, but open to the decision to leave this as is.


node/node_test.go
```diff
@@ -166,7 +165,7 @@ func TestNodeStartMining(t *testing.T) {
 		MsgSender:    msg.NewSender(minerNode.Repo, minerNode.Wallet, minerNode.ChainReader, minerNode.MsgPool, minerNode.PubSub.Publish),
 		MsgWaiter:    msg.NewWaiter(minerNode.ChainReader, minerNode.Blockstore, minerNode.CborStore()),
 		Config:       pbConfig.NewConfig(minerNode.Repo),
-		Chain:        chn.New(minerNode.ChainReader),
+		Chain:        minerNode.ChainReader,
```
phritz: alpha would have it above, maybe fix while you are here

plumbing/api.go
```diff
@@ -27,7 +27,7 @@ import (
 type API struct {
 	logger logging.EventLogger
 
-	chain        *chn.Reader
+	chain        chain.ReadStore
```
phritz: even more idiomatic would be to define an interface giving the subset of chain.readstore we need here and have chain's type be that. consumers should generally define the interface they need. but also ok to leave as it given that this is consistent with the way the code works throughout gfc.

anorth: Thanks. I opt to leave this one as is since it's an internal package. Ultimately, plumbing will be the ~only the consumer of this interface, so most methods should be required.

PR #2036, opened 2019-02-20
gx updates

whyrusleeping:
pulled another gx dep update through for y'all. Some nice goodies in this one.


whyrusleeping:
for future reference, I did this with:
```
gx-workspace update start --all
gx-workspace update run
```

It was mostly 'no changes', but there were a couple packages (hamt and keystore) that it had to do releases for. That just requires running an ipfs daemon and pinning the packages to pinbot once theyre created.

Once that process finished, the top level go-filecoin needed a little love:
```
gx-go rw --fix
gx-go rw   # fix import paths with bad references

gx deps check # will show some deps we're importing directly with the wrong versions, manual fix is easy enough
```

phritz:
@ZenGround0 would you mind adding hints of the above commands to https://github.com/filecoin-project/go-filecoin/blob/master/CODEWALK.md#dependencies?

PR #2042, opened 2019-02-20
docs: enhance CODEWALK.md with libp2p colour

### LINKED ISSUE #1769: DHT explainer
### Description
We should have a [codewalks](https://github.com/filecoin-project/go-filecoin/blob/master/CODEWALK.md) level description of how the DHT works in filecoin.  Probably merits 2 sections: 
1. setup: including starting the bootstrap service and explaining what is happening at the level of records to get a node bootstrapped
2. use: right now I think its only used to go from miner peer id (gotten from chain) to miner address to connect directly to miners during deals.  Should explain which call hits the DHT and high level explanation of what is happening with lookup to make things work.  

Might be further sections as our implementation becomes better.  For example when we do chain bootstrapping correctly we will probably do some DHT walking to find random peers for querying chain head.

@anacrolix if interested/if you need some filecoin work, this is something you would probably do well, I could steer you towards entrypoints for setup and use of DHT in go-filecoin.  No pressure though, I'm happy to take this on.

### Where to begin
A little bit of research into kademlia, and in particular libp2p's implementation.

raulk:
Initial pass; please provide feedback on elements that (1) I missed, or (2) are unclear.

Fixes #1769.


CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
 The keystore contains the binary encoded peer key for interacting securely over the network. 
 This data lives in a file at `$HOME/.filecoin/keystore/self`.
 
+## libp2p Networking
+
+As mentioned above, Filecoin relies on [libp2p](https://libp2p.io/) for its
+networking needs. Here we'll explain some of the mechanics that drive the
+network layer, in particular:
```
anorth: Please format 1 sentence per line, breaking only very long sentences at a phrase boundary. It makes code review comments easier to target and reduces manual re-wrapping when things change.

Bonus: add a comment at the top describing this format for future contributors.

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+
+1. Node discovery
+2. NAT topology determination
+3. Autorelay services for solving connectivity issues
+
+### Node discovery
+
+We use the libp2p Kademlia DHT bootstrapping mechanism to find Filecoin peers
+to interface with. libp2p maintains a k-bucket routing table, where peers are
+organized according to the _distance metric proposed_ in the [Kademlia
```
phritz: please say what a routing table is, namely a map from peerid to multiaddr or multiaddrs

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+
+As mentioned above, Filecoin relies on [libp2p](https://libp2p.io/) for its
+networking needs. Here we'll explain some of the mechanics that drive the
+network layer, in particular:
+
+1. Node discovery
+2. NAT topology determination
+3. Autorelay services for solving connectivity issues
+
+### Node discovery
```
phritz: would you mind adding a sentence about:
- how my entry gets added to the routing table and when/how it gets pruned
- what happens if there are multiple entries for my peerid
- whether dht messages flow over an existing or separate physical connection. i think there are many logical streams that occur over the same physical connection

we should probably say what problems the dht solves for us as well. i think it is
1. ensure the network stays well-connect so that if some nodes go down it continues to operate
2. enable us to map from peerid to dialable multiaddr so that a client can connect to a miner whose peerid is found on chain in the miner actor

@whyrusleeping this are the two problems yes?

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+repeat this process iteratively, until either:
+
+a. we find the peer; highly unlikely scenario, as the chance is 1/2^256 with
+   our peer ID length; or
+b. we yield because a timer runs out; the most probable scenario. This is a
+   healthy timeout to timebox discovery iterations.
+
+Throughout this process, the libp2p stack generates `CONNECTED` events for
+every new connection we establish to a peer, for which `go-filecoin` registers
+a callback that triggers the Filecoin `HELLO` protocol negotiation. If the
```
phritz: i think this means that because we'll disconnect from any peers that fail the hello (that dont share the same genesis block as us) our dht is limited to the nodes we can actually exchange blocks with.

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+
+Relay capability is activated by sensing our NAT status, once at start (may
+take a few minutes) and continuously at runtime. Two protocols are involved
+here:
+
+* **Identify** protocol: for every peer we connect to, we use the *Identify*
+  protocol to learn the address that peer observes from us (source
+  ip_address:port). We compile all our observed addresses as we learn them.
+* **AutoNAT** protocol: for every peer we connect to that supports the AutoNAT
+  service, we request a _dialback_ to our observed addresses. We wait 42
```
phritz: interesting timeout, why so long?

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+to enable devices in a private network to share a common public IP address in 
+the global, scarce IPv4 space.
+
+Outbound connections behind a NAT tend to be frictionless, but inbound
+connections are problematic. For a primer on this topic, we recommend reading
+[Peer-to-Peer Communication Across Network Address
+Translators](https://pdos.csail.mit.edu/papers/p2pnat.pdf).
+
+To enable a pair of Filecoin peers behind NATs to communicate with one
+another, we use libp2p relay services: nodes deployed across the globe that
```
phritz: not sure if our 5 nodes running on one box in aws constitute nodes deployed across the globe, though i guess strictly speaking that is true :)

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
 The keystore contains the binary encoded peer key for interacting securely over the network. 
 This data lives in a file at `$HOME/.filecoin/keystore/self`.
 
+## libp2p Networking
```
anorth: Blocking: update the table of contents

Nit: lowercase 'N' for consistency with other titles. Or maybe `Networking (libp2p)`

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+
+As mentioned above, Filecoin relies on [libp2p](https://libp2p.io/) for its
+networking needs. Here we'll explain some of the mechanics that drive the
+network layer, in particular:
+
+1. Node discovery
+2. NAT topology determination
+3. Autorelay services for solving connectivity issues
```
anorth: Remove this unnecessary list.

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+
+### Node discovery
+
+We use the libp2p Kademlia DHT bootstrapping mechanism to find Filecoin peers
+to interface with. libp2p maintains a k-bucket routing table, where peers are
+organized according to the _distance metric proposed_ in the [Kademlia
+paper](http://www.scs.stanford.edu/~dm/home/papers/kpos.pdf).
+
+To populate our routing table, and to keep it healthy over time, we traverse
+the network periodically (and once on startup) by performing random walks. We
```
anacrolix: Unrelated, but this is a problem. Randomly walking when your routing table favours nodes closer to your address means you're getting almost no value to the routing table from this behaviour. Additionally, randomly walking in the bootstrap methods provides no way for the caller to benefit from the nodes seen. The correct approach if the nodes are to be arbitrarily connected to to maintain a mesh, would be to use a FindPeer or related call, and operate on the nodes visited.

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+networking needs. Here we'll explain some of the mechanics that drive the
+network layer, in particular:
+
+1. Node discovery
+2. NAT topology determination
+3. Autorelay services for solving connectivity issues
+
+### Node discovery
+
+We use the libp2p Kademlia DHT bootstrapping mechanism to find Filecoin peers
```
anorth: Link to https://github.com/libp2p/go-libp2p-kad-dht (if that's the right target)

CODEWALK.md
anacrolix: I've read the bit further down, it seems the random walk does generate value as the connections obtained are hooked elsewhere. However I still think the random walking doesn't belong in the bootstrap, and should be an explicit FindPeers wrapper elsewhere. Meaning routing table health is maintained with periodic `FindPeer(self)`, and filecoin "mesh" health generated by periodic `FindPeer(random)`.

CODEWALK.md
```diff
@@ -346,6 +346,118 @@ essentially who is storing what data, for what fee and which sectors have been s
...
+or both, of the parties is/are behind a Network Address Translation (NAT)
+devices, such as routers. NATs are commonplace in different network types: 
+residential, companies, public hotspots, universities, etc. Their purpose is 
+to enable devices in a private network to share a common public IP address in 
+the global, scarce IPv4 space.
+
+Outbound connections behind a NAT tend to be frictionless, but inbound
+connections are problematic. For a primer on this topic, we recommend reading
+[Peer-to-Peer Communication Across Network Address
+Translators](https://pdos.csail.mit.edu/papers/p2pnat.pdf).
```
anorth: Remove these two paragraphs in favour of getting to concrete details, perhaps leaving behind just the link to paper.

CODEWALK.md
anacrolix: > 42

:+1:

CODEWALK.md
anacrolix: I'm not sure on our implementation, but a FindPeers should only return DHT-enabled nodes. Can filecoin nodes operate without DHT at all?

CODEWALK.md
anacrolix: It is not that, so I think that's an extra valuable addition :P

anacrolix:
How much of this is valuable to libp2p as well? Could parts of it be moved there, and linked to from filecoin?

CODEWALK.md
mgoelzer: 😂

raulk:
@anorth @phritz I feel we need to reconcile both your points of view. On the one hand, @phritz requested more detail re: some concepts like a routing table; OTOH, @anorth thinks there's too much detail here.

The way I see it, we have two options:

1. I can take a considerable amount of content away from here and put it somewhere libp2p. For now likely repo READMEs, or an explainer directory in libp2p/libp2p where we compile non-normative high-level explanations that are not specs. In the near future, these will be featured in docs.libp2p.io (under construction).
2. We can keep these explanations somewhere Filecoin if you feel it benefits your audience.

From when I read #1769, it sounded like you wanted as much clarity as possible, not just for FC users but for FC devs. IF that's the case, maybe the `CODEWALK.md` is not the right venue for this? Should we create a `networking.md` file instead?

raulk:
Another thing worth pointing out to make sure we're on the same page: the value the libp2p team delivers here is in providing clarity of how certain mechanisms and services you rely on work, to aid understanding and debugging.

Explaining how Filecoin uses libp2p, what protocols it exposes, and how they behave wrt. to networking, is something best done by a Filecoin dev. Note: I can do it, no problem, but it will require me to review your code. If you want that, let me know.

CODEWALK.md
raulk: Haha they're across the globe from where you are :P I'll tone down this statement.

CODEWALK.md
raulk: Yeah, checking with the author.

CODEWALK.md
raulk: @phritz that's not accurate in the present. The behaviour makes sense, and you might even want to blacklist peers for a period that made it into your DHT but don't expose the `HELLO` protocol, but that's not how Filecoin is behaving now.

CODEWALK.md
raulk: Checking with the source.

CODEWALK.md
raulk: Sure, this format is a bit weird, but I'll reformat. Please make sure to document it so that future contributors don't trip up, and you have an actual style guide to reference.

CODEWALK.md
raulk: Why is it unnecessary? It provides a useful lay of the land to help readability.

CODEWALK.md
raulk: Are all readers expected to be familiar with what a NAT is? I don't think you can count on that. There's a healthy balance between no intro at all, and a 14-page paper.

CODEWALK.md
raulk: Done ✅

CODEWALK.md
raulk: Done.

CODEWALK.md
raulk: Tracking the possibility of speeding up NAT assertion here: https://github.com/libp2p/go-libp2p-autonat/issues/11.

raulk: Removed the number; it's a bit more complex than that. Filed an issue for speeding up the NAT assertion process: https://github.com/libp2p/go-libp2p-autonat/issues/11

CODEWALK.md
raulk: Mind opening an issue in libp2p/specs with your thoughts to continue discussing there? This is likely not the right place for the kind of engagement the conversation requires.

CODEWALK.md
raulk: Not sure we should be pointing to code of third party libs you use.

raulk:
Ok, comments addressed.

mgoelzer:
Meta comment:

libp2p needs to do a better job of providing a primer that explains what it is/does, what are the interfaces to this black box called `libp2p`, define the vocabulary we use, how are its parts are typically used together, etc.  

This should live on docs.libp2p.io (currently a 404, I know).  Then it could be incorporated by reference in a section like this one.  For example:

>_This section assumes basic familiarity with libp2p, the peer-to-peer networking subsystem used by Filecoin.  For a backgrounder, see [link]._

I +1'd [Raul's comment](https://github.com/filecoin-project/go-filecoin/pull/2042#issuecomment-466004584) because I don't think libp2p engineers actually know how Filecoin is using libp2p, and having them try to create that kind of documentation takes away time they could spend on doing the first paragraph's work excellently.  I'm actually not sure that this is what [@anorth's comment](https://github.com/filecoin-project/go-filecoin/pull/2042#pullrequestreview-206027749) was asking for.  I think he's just pointing out that this information is needed from somebody to close that particular FC issue.

cc @yusefnapora

phritz:
yes, how filecoin uses libp2p is out of scope for libp2p folks

@raulk if this information has a good home in one place in libp2p land then host it there and we can link to it from here. if it does not have a good home, or it would be split across several places and people have to piece it together, please move it into a networking.md doc here and link to it from our codewalk. lets get this thing landed.

thank you so much for doing this!

phritz:
also, you have all the approvals you need, please merge as soon as you have placed the networking discussion its proper home, dont gate on more reviewers

CODEWALK.md
```diff
@@ -314,6 +313,111 @@ Protocols (goroutines) communicate through custom channels.
...
+3. Autorelay services for solving connectivity issues.
+
+In addition, it must be noted that Filecoin relies on [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) for 
+pubsub gossip among peers propagating blockchain blocks and unmined messages.
+
+### Peer discovery
+
+We use the libp2p Kademlia Distributed Hashtable (Kad DHT) service to:
+
+1. discover Filecoin peers to interface with upon boot.
```
mishmosh: filecoin usually calls them "nodes" (also imperfect bc this doc talks so much about the node API). maybe give a heads up on the synonym with something like "Filecoin nodes, aka peers"

yusefnapora:
This is awesome stuff, and I'm definitely interested in helping pull things into the libp2p domain where it make sense. I'm hoping to have docs.libp2p.io up and running within a few days, although it's still light on content. Once it's up, I'd like to copy some of this material over with a few tweaks as part of the primer that @mgoelzer is referring to, which is high on the priority list for new content.

My preference would be to merge this in and trim it down later once we can link out to the libp2p docs site, but that's up to you guys.

mishmosh:
@phritz said: 
> also, you have all the approvals you need, please merge as soon as you have placed the networking discussion its proper home

with merge protection enabled, someone from @filecoin-project/go-filecoin-committers will need to merge (i nominate @anorth). So the next steps are:
- move networking discussion to its proper home
- ping @anorth for another look once ready
- @anorth merges once ready

anorth:
Thanks, moving content to `networking.md` that can one day be hosted by `libp2p` SGTM, and yes we still need someone from filecoin team to describe usage. I will edit the PR description so Github doesn't automatically "fix" #1769.

I intend the CODEWALK as high level pointers to where everything is and how components relate, rather than a detailed description of how everything works (which is in the code itself or a spec).

CODEWALK.md
```diff
@@ -24,13 +24,17 @@ It is complemented by specs (link forthcoming) that describe the key concepts im
   - [Plumbing & porcelain](#plumbing--porcelain)
   - [Commands](#commands)
   - [Protocols](#protocols)
-  - [Network layer](#network-layer)
+  - [Actors](#actors)
```
raulk: These existing sections were missing in the TOC.

raulk:
@anorth ready here. I feel this explainer with the latest enhancements does cover the requirements set forth in #1769, but it's your call.

networking.md
```diff
@@ -0,0 +1,122 @@
...
+## Pubsub via Gossipsub
+
+Filecoin relies on [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) for pubsub gossip among peers propagating blockchain blocks and unmined messages.
+Please refer to the linked spec for more infomation.
+
+## Contribute to libp2p
+
+Start exploring the libp2p universe on [our website](https://libp2p.io/) and [entrypoint Github repo](https://github.com/libp2p/libp2p).
+There are multiple language implementations you can get involved in.
+Also join the [#libp2p channel on Freenode](http://webchat.freenode.net/?channels=%23libp2p) to interact with the community.
```
anorth: Cheeky!

PR #2047, opened 2019-02-20
Feat/wallet balance porcelain 1797

### LINKED ISSUE #1797: move wallet balance to use plumbing/porcelain
### Description
background #1469 

the task here is to move the node wallet balance command to use plumbing/porcelain. currently it is implemented in terms of nodeapi, which uses a bunch of core calls: https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/impl/address.go#L37. a more principled approach would be to have an actorget call in plumbing that returns an actor. then it is simply a porcelain call to return that actor.balance. or heck, you dont even need to introduce porcelain for that, the wallet balance command can just get the actor and output actor.balance.

once the address balance command uses plumbing, delete the balance call from nodeaddress and from the nodeapi
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/impl/address.go#L37
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/address.go#L16

obviously, any other commands that use it have to be switched too

actorget can have signature like `actorget(address) *actor.actor` and can use the latest chain state. note that a more maximally factored version of this would require address and block cid or height and use that to get the state tree, and a convenience function would go into porcelain to run actorget on the latest. but i dont think you need to go that far unless you want to. we already have plumbing that is not maximally factored like that.

### Acceptance criteria

- balance command is no longer part of the address node api interface

### Risks + pitfalls

- like #1693 i find the wallet and address commands to be confusing. i'd be all for changing the wallet balance command to be address balance. makes so much more sense as balance is a property of an address, not a wallet. 
- be aware of #1796 and use it if it has landed

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/impl/address.go#L37
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/commands/address.go#L126

rosalinekarr:
# Problem

We don't want the node API to exist, and `Address().Balance()` is a small part of node API.

# Solution

Move `Address().Balance()`'s functionality to a porcelain method so we can move one step closer to killing the node.

Resolves #1797

### DIFF
--- a/plumbing/api.go
+++ b/plumbing/api.go
@@ -113,8 +113,8 @@ func (api *API) ChainLs(ctx context.Context) <-chan interface{} {
 	return api.chain.BlockHistory(ctx, api.chain.Head())
 }
 
-// ChainGetActorFromLatestState returns an actor from the latest state on the chain
-func (api *API) ChainGetActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
+// ActorGet returns an actor from the latest state on the chain
+func (api *API) ActorGet(ctx context.Context, addr address.Address) (*actor.Actor, error) {
 	state, err := api.chain.LatestState(ctx)
 	if err != nil {
 		return nil, err
--- a/plumbing/api.go
+++ b/plumbing/api.go
@@ -8,6 +8,7 @@ import (
 	logging "gx/ipfs/QmbkT7eMTyXfpeyB3ZMxxcxg7XH8t6uXp49jqzz4HB7BGF/go-log"
 	"gx/ipfs/QmepvmmYNM6q4RaUiwEikQFhgMFHXg2PLhx2E9iaRd3jmS/go-libp2p-pubsub"
 
+	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
 	"github.com/filecoin-project/go-filecoin/core"
@@ -17,7 +18,6 @@ import (
 	"github.com/filecoin-project/go-filecoin/plumbing/mthdsig"
 	"github.com/filecoin-project/go-filecoin/plumbing/ntwk"
 	"github.com/filecoin-project/go-filecoin/plumbing/ps"
-	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/wallet"
 )
@@ -113,9 +113,13 @@ func (api *API) ChainLs(ctx context.Context) <-chan interface{} {
 	return api.chain.BlockHistory(ctx, api.chain.Head())
 }
 
-// ChainLatestState returns the latest state tree from the chain reader
-func (api *API) ChainLatestState(ctx context.Context) (state.Tree, error) {
-	return api.chain.LatestState(ctx)
+// ChainGetActorFromLatestState returns an actor from the latest state on the chain
+func (api *API) ChainGetActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
+	state, err := api.chain.LatestState(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return state.GetActor(ctx, addr)
 }
 
 // BlockGet gets a block by CID
--- a/plumbing/api.go
+++ b/plumbing/api.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/plumbing/mthdsig"
 	"github.com/filecoin-project/go-filecoin/plumbing/ntwk"
 	"github.com/filecoin-project/go-filecoin/plumbing/ps"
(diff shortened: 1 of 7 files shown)


plumbing/chn/chain.go
```diff
@@ -41,3 +43,8 @@ func (c *Reader) Ls(ctx context.Context) <-chan interface{} {
 func (c *Reader) BlockGet(ctx context.Context, id cid.Cid) (*types.Block, error) {
 	return c.chainReader.GetBlock(ctx, id)
 }
+
+// LatestState returns the latest state of the chain
+func (c *Reader) LatestState(ctx context.Context) (state.Tree, error) {
+	return c.chainReader.LatestState(ctx)
+}
```
anorth: Apologies that you'll have a conflict here, but I hope it's simple to resolve. I removed this package in favour of a ChainReader on the plumbing object.

porcelain/wallet_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type walletTestPlumbing struct{}
+
+func (wtp *walletTestPlumbing) ChainLatestState(ctx context.Context) (state.Tree, error) {
+	store := hamt.NewCborStore()
+	tree := state.NewEmptyStateTree(store)
+
+	testActor := actor.NewActor(cid.Undef, types.NewAttoFILFromFIL(20))
```
anorth: This would be a bit more flexible for future tests with the balance as a field on the struct, and the constant `20` mentioned only in the test method.

plumbing/api.go
```diff
@@ -104,6 +105,11 @@ func (api *API) ChainLs(ctx context.Context) <-chan interface{} {
 	return api.chain.Ls(ctx)
 }
 
+// ChainLatestState returns the latest state tree from the chain reader
+func (api *API) ChainLatestState(ctx context.Context) (state.Tree, error) {
```
acruikshank: BLOCKING: The state.Tree is a read write interface to the data store. It's not an appropriate return value from the plumbing API. Better to implement a `GetActorFromLatestState`. This will simplify testing.

plumbing/chn/chain_test.go
```diff
@@ -46,6 +47,10 @@ func (mcr *FakeChainer) Head() types.TipSet {
 	return mcr.head
 }
 
+func (mcr *FakeChainer) LatestState(ctx context.Context) (state.Tree, error) {
+	return nil, errors.New("Not defined yet")
```
acruikshank: All plumbing code should be tested.

plumbing/api.go
phritz: but getting the latest tree is sometimes what you want, isn't it? does it make sense to have a read only interface that is defined here?

acruikshank: @phritz A read only interface would expose `GetActor` and possibly `ForEachActor` I don't really have a problem with it so long as it's read-only, but it seems a little clunky to nest the API this way. It's:
``` go
api.ChainGetActor(ctx, addr)
```
vs.
``` go
api.ChainLatestState(ctx).GetActor(addr)
```

acruikshank: Coming at it from another angle, say you are a developer trying to inspect an actor. As you peruse the API, you'd have to be pretty familiar with the system to understand that you should start with chain and then get latest state which would then expose the GetActor method. 

The plumbing API is not supposed to be easy to use, but where we have the option to make it easier at no real cost, we should take it.

plumbing/chn/chain_test.go
rosalinekarr: `ChainGetActorFromLatestState` is just a pass through method to the chain. There's not really anything to test here except "method calls other method," and chain already has it's own tests.

plumbing/api.go
```diff
@@ -112,6 +113,15 @@ func (api *API) ChainLs(ctx context.Context) <-chan interface{} {
 	return api.chain.BlockHistory(ctx, api.chain.Head())
 }
 
+// ChainGetActorFromLatestState returns an actor from the latest state on the chain
+func (api *API) ChainGetActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
```
acruikshank: `ActorGet` would be a more consistent and concise name. Apologies that I suggested this one. Beyond that, I have 3 concerns:

1. This is the only method in API that is not a simple pass-through.
2. This isn't tested. This is probably related to the above.
3. We're importing `chain.ReadStore` instead of an interface off it. This made sense when we were simply exposing its methods, but less so now that we are deriving functionality from it. Writing a test for this method is slightly non-trivial which suggests this is a problem.

The real question is whether this should be made a plumbing/something/Something-er like config and messages. It seems too much structure for such a simple thing, but then this seems a bit too little.

thoughts @phritz?

PR #2051, opened 2019-02-20
Feat/deal propose sigs

frrist:
This PR adds a new type to the storage protocol called `SignedDealProposal`. A `SignedDealProposal` contains a `DealProposal` and a `Signature`, and is sent to the miner by the client when proposing a deal. The `Signature` is created using the clients `from` address in the deal, the miner verifies the deal came from this address before accepting.

### DIFF
--- a/protocol/storage/client.go
+++ b/protocol/storage/client.go
@@ -68,6 +68,7 @@ type clientPorcelainAPI interface {
 	MinerGetAsk(ctx context.Context, minerAddr address.Address, askID uint64) (miner.Ask, error)
 	MinerGetOwnerAddress(ctx context.Context, minerAddr address.Address) (address.Address, error)
 	MinerGetPeerID(ctx context.Context, minerAddr address.Address) (peer.ID, error)
+	types.Signer
 }
 
 type clientDeal struct {
@@ -142,12 +143,12 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data 
 		TotalPrice:   totalPrice,
 		Duration:     duration,
 		MinerAddress: miner,
-		// TODO: Sign this proposal
 	}
 
 	if smc.isMaybeDupDeal(proposal) && !allowDuplicates {
 		return nil, Errors[ErrDupicateDeal]
 	}
+
 	// create payment information
 	cpResp, err := smc.api.CreatePayments(ctx, porcelain.CreatePaymentsParams{
 		From:            fromAddress,
@@ -169,6 +170,11 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data 
 	proposal.Payment.ChannelMsgCid = &cpResp.ChannelMsgCid
 	proposal.Payment.Vouchers = cpResp.Vouchers
 
+	signedProposal, err := proposal.NewSignedProposal(fromAddress, smc.api)
+	if err != nil {
+		return nil, err
+	}
+
 	// send proposal
 	pid, err := smc.api.MinerGetPeerID(ctx, miner)
 	if err != nil {
@@ -176,7 +182,7 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data 
 	}
 
 	var response DealResponse
-	err = smc.node.MakeProtocolRequest(ctx, makeDealProtocol, pid, proposal, &response)
+	err = smc.node.MakeProtocolRequest(ctx, makeDealProtocol, pid, signedProposal, &response)
 	if err != nil {
 		return nil, errors.Wrap(err, "error sending proposal")
 	}
@@ -187,7 +193,7 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data 
 
 	// Note: currently the miner requests the data out of band
 
-	if err := smc.recordResponse(&response, miner, proposal); err != nil {
+	if err := smc.recordResponse(&response, miner, &signedProposal.DealProposal); err != nil {
 		return nil, errors.Wrap(err, "failed to track response")
 	}
 
--- a/protocol/storage/client_test.go
+++ b/protocol/storage/client_test.go
@@ -23,6 +23,8 @@ import (
 	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/require"
 )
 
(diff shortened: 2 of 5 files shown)


protocol/storage/client_test.go
```diff
@@ -58,6 +59,7 @@ func TestProposeDeal(t *testing.T) {
 	askID := uint64(67)
 	duration := uint64(10000)
 	dealResponse, err := client.ProposeDeal(ctx, minerAddr, dataCid, askID, duration, false)
+	fmt.Println(dealResponse)
```
anorth: Remove println here and at 115

protocol/storage/client_test.go
```diff
@@ -208,6 +211,10 @@ func newTestClientNode(responder func(request interface{}) (interface{}, error))
 	}
 }
 
+func (tcn *testClientNode) SignBytes(data []byte, addr address.Address) (types.Signature, error) {
+	return nil, nil
```
anorth: Blocking: set some value for the signature and assert that the signed proposal carries it.

protocol/storage/miner_test.go
```diff
@@ -114,10 +114,26 @@ func TestReceiveStorageProposal(t *testing.T) {
 		assert := assert.New(t)
 		require := require.New(t)
 
-		_, miner, proposal := newMinerTestSetup()
+		porcelainAPI, miner, proposal := newMinerTestSetup()
 		proposal.Payment.Vouchers = []*paymentbroker.PaymentVoucher{}
 
-		res, err := miner.receiveStorageProposal(context.Background(), proposal)
+		// since we are modifying the deal after we create it, we must resign it
```
anorth: Blocking: putting this re-signing inline in the test obscures its purpose (especially since it's duplicated below). Instead, make the number of vouchers a parameter of `newMinerTestSetup` or the methods that it calls here.

protocol/storage/miner_test.go
```diff
@@ -128,10 +144,26 @@ func TestReceiveStorageProposal(t *testing.T) {
 		assert := assert.New(t)
 		require := require.New(t)
 
-		_, miner, proposal := newMinerTestSetup()
+		porcelainAPI, miner, proposal := newMinerTestSetup()
 		proposal.Payment.Vouchers[0].Signature = types.Signature([]byte{})
 
-		res, err := miner.receiveStorageProposal(context.Background(), proposal)
+		// since we are modifying the deal after we create it, we must resign it
```
anorth: Ditto, refactor the testing helper methods to allow this test to construct the input it needs more directly.

If that's too onerous, at least break out this signing logic to a new helper so the test code itself is direct.

protocol/storage/types.go
```diff
@@ -60,8 +61,24 @@ type DealProposal struct {
...
+}
 
-	// Signature types.Signature
+// Marshal the DealProposal into bytes.
+func (dp *DealProposal) Marshal() ([]byte, error) {
+	return cbor.DumpObject(dp)
+}
+
+// SignedDealProposal is the information sent over the wire, when a clinet proposes a deal to a miner.
+// The SignedDealProposal contains the DealProposal and a Signature for the proposal.
```
anorth: Nit: this comment is mostly redundant given the code. Suggest "SignedDealProposal is a deal proposal signed by the proposing client".

protocol/storage/client.go
```diff
@@ -340,17 +357,25 @@ type ClientNodeImpl struct {
...
 		dserv:     ds,
 		host:      host,
 		blockTime: bt,
+		signer:    s,
 	}
 }
 
+// SignBytes implements the types.signer interface. This allows ClientNode to
+// sign data with the address `addr`.
+func (cni *ClientNodeImpl) SignBytes(data []byte, addr address.Address) (types.Signature, error) {
```
acruikshank: BLOCKING: SignBytes is already available from the plumbing API. https://github.com/filecoin-project/go-filecoin/blob/9549501fa9513cc9141426c7e958a2138ff97890/plumbing/api.go#L177 Since the client already has access to the API, there's no need to reimplement it here.

protocol/storage/miner_test.go
```diff
@@ -441,4 +489,20 @@ func testDealProposal(porcelainAPI *minerTestPorcelain, voucherInterval int, amo
 			Vouchers:      vouchers,
 		},
 	}
+
+	bdp, err := proposal.Marshal()
+	if err != nil {
```
acruikshank: These if err != nil panic clauses should be replaced with require.NoErr(err)

frrist:
Thanks for the feedback, I have addressed each comment in the new commits, let me know if this looks good and I'll rebase them out.

anorth:
This appears to be protocol-breaking, so we need to be careful wrt devnet. I'm not sure what our process is here.

protocol/storage/client.go
```diff
@@ -169,14 +170,30 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 	proposal.Payment.ChannelMsgCid = &cpResp.ChannelMsgCid
 	proposal.Payment.Vouchers = cpResp.Vouchers
 
+	// client signs the DealProposal with the address they are sending from
```
anorth: Consider breaking this out to a well-named function, removing the need for this comment.

protocol/storage/miner_test.go
```diff
@@ -100,7 +108,8 @@ func TestReceiveStorageProposal(t *testing.T) {
 		assert := assert.New(t)
 		require := require.New(t)
 
-		porcelainAPI, miner, proposal := newMinerTestSetup()
+		porcelainAPI := newMinerTestPorcelain(require)
+		miner, proposal := newMinerTestSetup(porcelainAPI, VoucherInterval, defaultAmountInc)
```
anorth: Thank you, this flexibility is an improvement. These "common" setups could be factored into a `defaultMinerTestSetup` returning the triplet, just for conciseness: your call.

protocol/storage/types.go
```diff
@@ -60,8 +61,23 @@ type DealProposal struct {
 	// will use to pay the miner. It should be verifiable by the
 	// miner using on-chain information.
 	Payment PaymentInfo
+}
+
+// Unmarshal a DealProposal from the given bytes.
```
anorth: ~the given~

Issue #2013, opened 2019-02-17
[bug] ln: failed to create symbolic link 'params.out': File exists


NatoBoram:
**Describe the bug**

When running `go run ./build/*.go deps`, [`copy-groth-params.sh`](https://github.com/filecoin-project/go-filecoin/blob/master/scripts/copy-groth-params.sh) creates a simlink with the command [`ln -s ${params_file} params.out`](https://github.com/filecoin-project/go-filecoin/blob/master/scripts/copy-groth-params.sh#L11). The command fails if the simlink was previously created.

**To Reproduce**

1. `FILECOIN_USE_PRECOMPILED_RUST_PROOFS=true go run ./build/*.go deps`
2. `FILECOIN_USE_PRECOMPILED_RUST_PROOFS=false go run ./build/*.go deps`
3. See error

**Expected behavior**

The script should override the simlink with the option `--force`.

**Screenshots**

```shell
./scripts/copy-groth-params.sh
/tmp/filecoin-proof-parameters ~/go/src/github.com/filecoin-project/go-filecoin
ln: failed to create symbolic link 'params.out': File exists
~/go/src/github.com/filecoin-project/go-filecoin
```

**Desktop**

```yaml
nato@hp-server
--------------
OS: Ubuntu 18.04.2 LTS x86_64
Host: HP G62 Notebook PC 0598120000202710010020100
Kernel: 4.15.0-46-generic
Uptime: 18 hours, 51 mins
Packages: 663
Shell: bash 4.4.19
Terminal: /dev/pts/0
CPU: Intel i3 M 370 (4) @ 2.399GHz
GPU: Intel Core Processor
Memory: 3071MiB / 3744MiB
```

**Additional context**

https://github.com/filecoin-project/go-filecoin/issues/2000#issuecomment-464416139

rosalinekarr:
PR #1975 should have resolved this, closing.

rosalinekarr:
My bad, this is not resolved according to @ognots

ognots:
Resolved by https://github.com/filecoin-project/go-filecoin/pull/2058

PR #2058, opened 2019-02-21
force symlink in copy_groth_params

ognots:
currently, if the symlink already exists, it throws an error.
see https://github.com/filecoin-project/go-filecoin/issues/2013


ognots:
I've tested on Mac locally successfully when params.out exists and doesn't exist

Issue #2027, opened 2019-02-19
High priority fixes and improvements


mishmosh:
A set of highest-priority bugs and productivity improvements discovered after opening repos. We will start with the problems that will require more time to diagnose/support than to simply fix: things that are incorrect, cause crashes, etc.

phritz:
to be clear what we are looking for here:
- things that are relatively small in scope, <= 3 points
- anything that is incorrect wrt protocol (eg, wrong units)
- anything that is going to eat more helpdesk/slack time to explain/accommodate than it would take to fix
- things that are very confusing/annoying to users or devs
- any crashes, unexpected errors, etc.

anorth:
This epic was created in the wake of the open sourcing and initial release of go-filecoin, and has now served its purpose. We still have outstanding issues, but many are less well defined than this epic calls for, and we are no longer devoting explicit time to these issues, outside consideration of all the other things we could build or fix.

I'll close this issue now.

PR #2063, opened 2019-02-21
add command to track bandwidth usage

whyrusleeping:
In working on figuring out why filecoin is using such an absurd amount of bandwidth (hint: it rhymes with fee age tree), i wrote this to expose libp2p bandwidth metrics.

### DIFF
--- a/commands/main.go
+++ b/commands/main.go
@@ -113,6 +113,7 @@ NETWORK COMMANDS
   go-filecoin id                     - Show info about the network peers
   go-filecoin ping <peer ID>...      - Send echo request packets to p2p network members
   go-filecoin swarm                  - Interact with the swarm
+  go-filecoin stats                  - Monitor statistics on your network usage
 
 ACTOR COMMANDS
   go-filecoin actor                  - Interact with actors. Actors are built-in smart contracts.
@@ -168,6 +169,7 @@ var rootSubcmdsDaemon = map[string]*cmds.Command{
 	"ping":             pingCmd,
 	"retrieval-client": retrievalClientCmd,
 	"show":             showCmd,
+	"stats":            statsCmd,
 	"swarm":            swarmCmd,
 	"version":          versionCmd,
 	"wallet":           walletCmd,
--- a/commands/stats_daemon_test.go
+++ b/commands/stats_daemon_test.go
@@ -1,0 +1,20 @@
+package commands
+
+import (
+	"testing"
+
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/assert"
+)
+
+func TestStatsBandwidth(t *testing.T) {
+	t.Parallel()
+	assert := assert.New(t)
+
+	d := th.NewDaemon(t).Start()
+	defer d.ShutdownSuccess()
+
+	stats := d.RunSuccess("stats", "bandwidth").ReadStdoutTrimNewlines()
+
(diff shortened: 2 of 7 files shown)


anorth:
FYI I think adding things to Node violates current policy (details omitted here). It belongs as a new piece of plumbing instead.

whyrusleeping:
That seems inconvenient, but okay. Someone else can pick that change up if they want.

phritz:
@whyrusleeping this is enshrined in our contributing guide and we ask that everyone play by the same rules: https://github.com/filecoin-project/go-filecoin/blob/master/CONTRIBUTING.md#developer-dos-and-donts. We are doing a lot of work to unwind the node api and this change takes us in the opposite direction.

phritz:
@rosalinekarr how would you feel about coming through and moving this implementation into plumbing as part of the work you are doing?

frrist:
If we don't think this will fit into the plumbing changes we have, this would make a great prometheus (or alike) metric for the node to expose.

phritz:
@frrist that would be a great metric to export, separately from this change

rosalinekarr:
@whyrusleeping where's the issue for this PR? i'll reassign it to myself.

mishmosh:
@rosalinekarr I don't think there is one, and in the future it should be written before PR. For now, can you create and add to epic #2027? Appreciate your taking this through to completion.

commands/main.go
```diff
@@ -113,6 +113,7 @@ NETWORK COMMANDS
   go-filecoin id                     - Show info about the network peers
   go-filecoin ping <peer ID>...      - Send echo request packets to p2p network members
   go-filecoin swarm                  - Interact with the swarm
+  go-filecoin stats                  - Monitor statistics on your network usage
```
ZenGround0: nit: this maybe fits better under the tools heading

gmasgras:
We(infra) are also tracking this here:
https://protocollabs.grafana.net/d/oesf9T1iz/kittyhawk?orgId=1&from=now-7d&to=now&refresh=30s&var-instance_name=user

screen shot 2019-03-06 at 2 44 19 pm

plumbing/ntwk/ntwk.go
```diff
@@ -13,20 +14,33 @@ type Network struct {
 	host host.Host
 	*pubsub.Subscriber
 	*pubsub.Publisher
+	metrics.Reporter
```
ZenGround0: BLOCKING: You should sync up with @anorth about network wrappers before merging this.  I just tried something similar with the router and he hit me with [this comment](https://github.com/filecoin-project/go-filecoin/pull/2158#discussion_r262235741).  It might be that this object is simple enough that you can avoid wrapping but I'm not aligned enough with these criteria to make a call.

commands/stats_daemon_test.go
```diff
@@ -0,0 +1,20 @@
+package commands
+
+import (
+	"testing"
+
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/assert"
+)
+
+func TestStatsBandwidth(t *testing.T) {
```
frrist: lets add a test for the json encoding as well

plumbing/ntwk/ntwk.go
anorth: @ZenGround0 thanks for pinging me. @rosalinekarr consider this un-blocked.

Since this is only being used for a command, rather than any functional integration, and there are no tests (not needed), I think it's pragmatic to leave this as is for now. We should aim to abstract libp2p away from most places, this one doesn't seem worth the trouble today.

PR #2067, opened 2019-02-22
Use bitswap sessions for block fetching

whyrusleeping:
did some refactoring and cleanup on the chain sync/store code to make it use bitswap sessions.

As currently written in master, we request every block from every person we are connected to, resulting in an obscene amount of bandwidth overhead. We need to fix that ASAP.

This PR improves it a bit, further improvements may be needed.

### DIFF
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -9,10 +9,8 @@ import (
 	"gx/ipfs/QmNf3wujpV2Y7Lnj2hy2UrmuX8bhMDStRHbnSLh7Ypf36h/go-hamt-ipld"
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
 	bstore "gx/ipfs/QmRu7tiRnFk9mMPpVECQTBQJqXtmG132jJxA1w9A7TtpBz/go-ipfs-blockstore"
-	"gx/ipfs/QmSz8kAe2JCKp2dWSG8gHSWnwSmne8YfRXTeK5HBmc9L7t/go-ipfs-exchange-offline"
 	"gx/ipfs/QmUadX5EcvrBmxAV9sE7wUWtWSqxns5K84qKJBixmcT1w9/go-datastore"
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
-	bserv "gx/ipfs/QmZsGVGCqMCNzHLNMB6q4F6yyvomqf1VxwhJwSfgo1NGaF/go-blockservice"
 	logging "gx/ipfs/QmbkT7eMTyXfpeyB3ZMxxcxg7XH8t6uXp49jqzz4HB7BGF/go-log"
 	"gx/ipfs/QmdbxjQWogRCHRaxhhGnYdT1oQJzL9GdqSKzCdqWr85AP2/pubsub"
 
@@ -29,19 +27,19 @@ var headKey = datastore.NewKey("/chain/heaviestTipSet")
 // DefaultStore is a generic implementation of the Store interface.
 // It works(tm) for now.
 type DefaultStore struct {
-	// privateStore is the on disk storage for blocks.  This is private to
+	// bsPriv is the on disk storage for blocks.  This is private to
 	// the DefaultStore to keep code that adds blocks to the DefaultStore's
 	// underlying storage isolated to this module.  It is important that only
 	// code with access to a DefaultStore can write to this storage to
 	// simplify checking the security guarantee that only tipsets of a
 	// validated chain are stored in the filecoin node's DefaultStore.
-	privateStore *hamt.CborIpldStore
+	bsPriv bstore.Blockstore
 	// stateStore is the on disk storage used for loading states.  It can be
 	// shared with the rest of the filecoin node.
 	stateStore *hamt.CborIpldStore
-	// ds is the datastore backing the privateStore.  It is also accessed
-	// directly to set and get meta information about the chain, specifically
-	// the tipset cidset to state root mapping, and the heaviest tipset cids.
+	// ds is the datastore backing bsPriv.  It is also accessed directly
+	// to set and get chain meta-data, specifically the tipset cidset to
+	// state root mapping, and the heaviest tipset cids.
 	ds repo.Datastore
 
 	// genesis is the CID of the genesis block.
@@ -61,8 +59,6 @@ type DefaultStore struct {
 
 	// Tracks tipsets by height/parentset for use by expected consensus.
 	tipIndex *TipIndex
-
-	// TODO block cache should go here
 }
 
 // Ensure DefaultStore satisfies the Store interface at compile time.
@@ -70,15 +66,14 @@ var _ Store = (*DefaultStore)(nil)
 
 // NewDefaultStore constructs a new default store.
 func NewDefaultStore(ds repo.Datastore, stateStore *hamt.CborIpldStore, genesisCid cid.Cid) *DefaultStore {
-	bs := bstore.NewBlockstore(ds)
-	priv := hamt.CborIpldStore{Blocks: bserv.New(bs, offline.Exchange(bs))}
+	priv := bstore.NewBlockstore(ds)
 	return &DefaultStore{
-		privateStore: &priv,
-		stateStore:   stateStore,
-		headEvents:   pubsub.New(128),
-		ds:           ds,
-		tipIndex:     NewTipIndex(),
-		genesis:      genesisCid,
+		bsPriv:     priv,
+		stateStore: stateStore,
+		ds:         ds,
+		headEvents: pubsub.New(128),
+		tipIndex:   NewTipIndex(),
+		genesis:    genesisCid,
 	}
 }
 
@@ -201,7 +196,7 @@ func (store *DefaultStore) loadStateRoot(ts types.TipSet) (cid.Cid, error) {
 
 // putBlk persists a block to disk.
 func (store *DefaultStore) putBlk(ctx context.Context, block *types.Block) error {
-	if _, err := store.privateStore.Put(ctx, block); err != nil {
+	if err := store.bsPriv.Put(block.ToNode()); err != nil {
 		return errors.Wrap(err, "failed to put block")
 	}
 	return nil
@@ -255,27 +250,26 @@ func (store *DefaultStore) HasTipSetAndStatesWithParentsAndHeight(ctx context.Co
 }
 
 // GetBlocks retrieves the blocks referenced in the input cid set.
-func (store *DefaultStore) GetBlocks(ctx context.Context, ids types.SortedCidSet) ([]*types.Block, error) {
-	var blks []*types.Block
-	for it := ids.Iter(); !it.Complete(); it.Next() {
+func (store *DefaultStore) GetBlocks(ctx context.Context, cids types.SortedCidSet) ([]*types.Block, error) {
+	var blocks []*types.Block
+	for it := cids.Iter(); !it.Complete(); it.Next() {
 		id := it.Value()
-		blk, err := store.GetBlock(ctx, id)
+		block, err := store.GetBlock(ctx, id)
 		if err != nil {
 			return nil, errors.Wrap(err, "error fetching block")
 		}
-		blks = append(blks, blk)
+		blocks = append(blocks, block)
 	}
-	return blks, nil
-
+	return blocks, nil
 }
 
 // GetBlock retrieves a block by cid.
 func (store *DefaultStore) GetBlock(ctx context.Context, c cid.Cid) (*types.Block, error) {
-	var blk types.Block
-	if err := store.privateStore.Get(ctx, c, &blk); err != nil {
+	data, err := store.bsPriv.Get(c)
+	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get block %s", c.String())
 	}
-	return &blk, nil
+	return types.DecodeBlock(data.RawData())
 }
 
 // HasAllBlocks indicates whether the blocks are in the store.
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -2,6 +2,7 @@ package chain
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -31,12 +32,16 @@ var (
 
 var logSyncer = logging.Logger("chain.syncer")
 
+type syncFetcher interface {
+	GetBlocks(context.Context, []cid.Cid) ([]*types.Block, error)
+}
+
 // DefaultSyncer updates its chain.Store according to the methods of its
 // consensus.Protocol.  It uses a bad tipset cache and a limit on new
 // blocks to traverse during chain collection.  The DefaultSyncer can query the
 // network for blocks.  The DefaultSyncer maintains the following invariant on
 // its store: all tipsets that pass the syncer's validity checks are added to the
-// chain store, and their state is added to cstOffline.
+// chain store, and their state is added to stateStore.
 //
 // Ideally the code that syncs the chain according to consensus rules should
 // be independent of any particular implementation of consensus.  Currently the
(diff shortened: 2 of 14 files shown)


whyrusleeping:
This might need some cleanup before merging, if someone wants to volunteer to help, that would be great.

phritz:
looks like chain tests timeout with this change

ZenGround0:
@whyrusleeping 

I can take this over and get it merged.  Here are a few questions

Looks like I was mistaken but thought that a cbor store started with a bservice would take care of bitswap stuff.  What was wrong with the "onlineCborStore"?  I want to make sure that I understand the problems with it before we replace its intended functionality with this.

> As currently written in master, we request every block from every person we are connected to, resulting in an obscene amount of bandwidth overhead. We need to fix that ASAP.

Could you explain where this is happening on master?  Does this PR fix it?  Unless I'm mistaken the optimization happening here is that we ask for all blocks in a tipset over bitswap (using a session directly) instead of one at a time over bitswap (using cbor store) which doesn't appear to address the problem you are describing.  Is it the case that the networked cborstore asks all peers for all data but a bitswap session does not? Please tell me what you know to help me file an issue for any remaining work.

Any strong reason for adding the network fetch to the default chain store object?  We have been wanted to descope and clean up this guy for a while so ideally we'd put the networked `GetBlocks` on something else like a `BlockFetcher` or something.

Bitswap noob question: should we be using a single session for the lifetime of a node?  (this seems to be what this PR does).  The name implies maybe doing something more short-lived.  It would be simple to have a new session for each chain that needs to be fetched if that is better.

whyrusleeping:
By default, every request made to bitswap will broadcast wants. There are small optimizations made for requesting multiple blocks at a time with bitswap, but it still broadcasts the wants.

> should we be using a single session for the lifetime of a node?

Bitswap sessions (as a concept) is designed to optimize fetches in the cases where we know people who have some bits of the data we are looking for will likely have other bits of the data we are looking for. We happen to be in this weird case where pretty much everyone will have everyone we are looking for. So it is a bit weird that i'm using one 'session' for the lifetime of the node, and there should be better facilities given to us here (i.e. ways of seeding a session with a set of known peers on creation) but for now, its not too bad.

The downside of using multiple sessions is that sessions have a bootstrapping period where they try to find many people who have the content we're looking for by broadcast spamming and doing dht searches. We would have to pay that cost every time we started a new session.

> Any strong reason for adding the network fetch to the default chain store object?

No, but that object is a tangle. It (and the syncer) appears to be doing the same things multiple times in several different overlapping ways. For example, the logic in the syncers 'maybe get blocks from network' is really weird... we first try to get things from the chain store, which pulls from our local disk. then we try to get from the 'offline' cbor store, which pulls from our local disk, then from an offline exchange, which pulls from the local disk (this should get fixed, theres no need for this offline thing, the blockservice should just learn to deal with a nil exchange), *then* it fetches from the online cborstore, which checks the local disk, then asks (everyone) over bitswap.

So, generally, +1 to keeping network fetching logic out of the 'chain store'.

chain/default_store.go
```diff
@@ -61,24 +60,21 @@ type DefaultStore struct {
...
-	priv := hamt.CborIpldStore{Blocks: bserv.New(bs, offline.Exchange(bs))}
+	priv := bstore.NewBlockstore(ds)
 	return &DefaultStore{
-		privateStore: &priv,
-		stateStore:   stateStore,
-		headEvents:   pubsub.New(128),
-		ds:           ds,
-		tipIndex:     NewTipIndex(),
-		genesis:      genesisCid,
+		bsPriv:     priv,
```
ZenGround0: Made sense to make the private store for holding types.Block an ipfs blockstore

chain/default_syncer.go
```diff
@@ -88,82 +94,63 @@ func NewDefaultSyncer(online, offline *hamt.CborIpldStore, c consensus.Protocol,
...
-		}
-		// try the network
-		if err = syncer.cstOnline.Get(ctx, blkCid, &blk); err != nil {
-			return nil, err
-		}
-		blks = append(blks, blk)
-	}
-	return blks, nil
+
+	return syncer.fetcher.GetBlocks(ctx, blkCids)
```
ZenGround0: The syncer reads new blocks from one and only one place now.

chain/default_syncer.go
```diff
@@ -88,82 +94,63 @@ func NewDefaultSyncer(online, offline *hamt.CborIpldStore, c consensus.Protocol,
...
+func (syncer *DefaultSyncer) collectChain(ctx context.Context, blkCids []cid.Cid) ([]types.TipSet, string, error) {
 	var chain []types.TipSet
 	defer logSyncer.Info("chain synced")
 	for {
 		var blks []*types.Block
 		// check the cache for bad tipsets before doing anything
 		tsKey := types.NewSortedCidSet(blkCids...).String()
 
+		// Finish traversal if the tipset made is tracked in the store.
+		if syncer.chainStore.HasTipSetAndState(ctx, tsKey) {
```
ZenGround0: During testing I discovered that the fetcher needed to be able to fetch blocks that were already in the chainStore for the syncer to work right.  This was indirect and unexpected so I moved this check up and changed the function signature to return a tipset id that the caller can use to get the parent tipset from the store.

chain/default_syncer_test.go
```diff
@@ -169,22 +170,31 @@ func requireSetTestChain(require *require.Assertions, con consensus.Protocol, mo
...
 
 // loadSyncerFromRepo creates a store and syncer from an existing repo.
-func loadSyncerFromRepo(require *require.Assertions, r repo.Repo) (*chain.DefaultSyncer, *hamt.CborIpldStore) {
+func loadSyncerFromRepo(require *require.Assertions, r repo.Repo) (*chain.DefaultSyncer, *net.TestFetcher) {
 	powerTable := &testhelpers.TestView{}
 	bs := bstore.NewBlockstore(r.Datastore())
 	cst := hamt.NewCborStore()
 	verifier := proofs.NewFakeVerifier(true, nil)
 	con := consensus.NewExpected(cst, bs, testhelpers.NewTestProcessor(), powerTable, genCid, verifier)
-	syncer, testchain, cst, _ := initSyncTest(require, con, initGenesis, cst, bs, r)
```
ZenGround0: Turns out that this was overwriting the repo's head to the genesis block unknown to anyone.  Hence the new setup code not relying on `initSyncTest`.

net/fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+	logging "gx/ipfs/QmbkT7eMTyXfpeyB3ZMxxcxg7XH8t6uXp49jqzz4HB7BGF/go-log"
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+var logFetcher = logging.Logger("net.fetcher")
+
+// Fetcher is used to fetch data over the network.  It is implemented with
+// a networked blockservice and a persistent bitswap session.
+type Fetcher struct {
```
ZenGround0: Not Block Fetcher because soon we will need a function to fetch messages and I am thinking this will be a good place for it.

ZenGround0:
@phritz @anorth I am still working on some basic fetcher unit tests which will land soon.  It believe this is in a good place for review.

net/fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+// GetBlocks fetches the blocks with the given cids from the network using the
+// Fetcher's bitswap session.
+func (f *Fetcher) GetBlocks(ctx context.Context, cids []cid.Cid) ([]*types.Block, error) {
+	var unsanitized []blocks.Block
+	for b := range f.session.GetBlocks(ctx, cids) {
+		unsanitized = append(unsanitized, b)
+	}
+
+	if len(unsanitized) < len(cids) {
+		return nil, errors.Wrap(ctx.Err(), "failed to fetch all requested blocks")
```
phritz: you wrap nil right? eg if it is a programming error and not a context error

net/fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	var blocks []*types.Block
+	for _, u := range unsanitized {
+		block, err := types.DecodeBlock(u.RawData())
+		if err != nil {
+			return nil, errors.Wrap(err, "returned data was not a block")
```
phritz: fetched data?

net/fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+var logFetcher = logging.Logger("net.fetcher")
+
+// Fetcher is used to fetch data over the network.  It is implemented with
+// a networked blockservice and a persistent bitswap session.
+type Fetcher struct {
+	// blockService is a network enabled blockstore.
+	blockService bserv.BlockService
+
+	// session is a bitswap session that enables efficient transfer.
+	session *bserv.Session
```
phritz: it is always so weird to me to see us keep a handle on a thingy, and then also on an underlying thingy. its like fake layering. these stats have very little to do with fetching so if you wanted you could just have a logstats function that is passed the underlying exhange in the caller, and not have the fetcher implement logstats. i guess we want to get this in, so maybe not now, but look for this kind of thing in the future ie single responsibility principle.

net/fetcher_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/filecoin-project/go-filecoin/net"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+func requireBlockStorePut(require *require.Assertions, bs bstore.Blockstore, data ipld.Node) {
+	err := bs.Put(data)
+	require.NoError(err)
+}
+
+func TestFetchHappyPath(t *testing.T) {
```
phritz: blocking: this is a good test but iirc does it actually test round trip? it doesn't test that two separate services connected to each other can successfully pull from one another, does it?

node/block.go
```diff
@@ -18,7 +18,7 @@ func (node *Node) AddNewBlock(ctx context.Context, b *types.Block) (err error) {
 	// Put block in storage wired to an exchange so this node and other
 	// nodes can fetch it.
 	log.Debugf("putting block in bitswap exchange: %s", b.Cid().String())
-	blkCid, err := node.OnlineStore.Put(ctx, b)
```
phritz: wait, i thought the cborstore was for state tree?

chain/default_store.go
```diff
@@ -29,13 +27,13 @@ var headKey = datastore.NewKey("/chain/heaviestTipSet")
...
 	// the DefaultStore to keep code that adds blocks to the DefaultStore's
 	// underlying storage isolated to this module.  It is important that only
 	// code with access to a DefaultStore can write to this storage to
 	// simplify checking the security guarantee that only tipsets of a
 	// validated chain are stored in the filecoin node's DefaultStore.
-	privateStore *hamt.CborIpldStore
+	bsPriv bstore.Blockstore
 	// stateStore is the on disk storage used for loading states.  It can be
 	// shared with the rest of the filecoin node.
 	stateStore *hamt.CborIpldStore
```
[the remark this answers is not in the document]
ZenGround0: no, the state store is wired to the node's shared datastore (node.cborStore etc), bsPriv is wired to the chain datastore, hence "Priv(ate)"

chain/default_syncer.go
```diff
@@ -54,8 +59,9 @@ type DefaultSyncer struct {
 	// are not run concurrently with other calls to widen to ensure
 	// that the syncer always finds the heaviest existing tipset.
 	mu sync.Mutex
-	// cstOnline is the online storage for fetching blocks.  It should be connected to the network with bitswap.
-	cstOnline *hamt.CborIpldStore
+	// fetcher is the networked block fetching service for fetching blocks
```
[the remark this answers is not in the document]
ZenGround0: It actually does know about them.  If you look at the fetcher implementation it errors if the cid it fetches doesn't marshal to a filecoin block.

node/block.go
ZenGround0: I would gladly change it to call into the node's blockstore.  I could also (1) add a method to the fetcher for local additions to its blockstore (2) change the syncer interface to accept blocks directly and stop being indirect like this.

net/fetcher.go
```diff
@@ -0,0 +1,73 @@
...
+func NewFetcher(ctx context.Context, bsrv bserv.BlockService) *Fetcher {
+	return &Fetcher{
+		blockService: bsrv,
+		session:      bserv.NewSession(ctx, bsrv),
+	}
+}
+
+// GetBlocks fetches the blocks with the given cids from the network using the
+// Fetcher's bitswap session.
+func (f *Fetcher) GetBlocks(ctx context.Context, cids []cid.Cid) ([]*types.Block, error) {
```
anorth: FYI I usually consider `GetXxx` to imply an instantaneous, non-blocking retrieval from local memory. Use a word like `Fetch` to indicate that this could be a slow operation. I see the underlying `session.GetBlocks`, but we can adapt terminology from third-party code at this layer.

net/fetcher_test.go
ZenGround0: Thanks I wasn't aware of this convention.

net/testing.go
```diff
@@ -86,3 +90,40 @@ func (fd *fakeDialer) Conns() []inet.Conn              { panic("not implemented"
 func (fd *fakeDialer) ConnsToPeer(peer.ID) []inet.Conn { panic("not implemented") }
 func (fd *fakeDialer) Notify(inet.Notifiee)            { panic("not implemented") }
 func (fd *fakeDialer) StopNotify(inet.Notifiee)        { panic("not implemented") }
+
+// TestFetcher is an object with the same method set as Fetcher plus a method
+// for adding blocks to the source.  It is used to implement an object that
+// behaves like Fetcher but does not go to the network for use in tests.
```
anorth: 💯

chain/default_syncer.go
ZenGround0: Thanks, that's a great idea.

types/message.go
```diff
@@ -48,10 +49,21 @@ func (msg *Message) Marshal() ([]byte, error) {
 	return cbor.DumpObject(msg)
 }
 
+// ToNode converts the Block to an IPLD node.
```
anorth: Block -> Message

chain/default_syncer.go
ZenGround0: yeah, because of how blockservices work.  I'll try to make the comment clearer.

anorth:
I am also out of my depth, but didn't see any glaring wrongness. @whyrusleeping could you review the new state?

net/fetcher.go
ZenGround0: That makes sense.  Any compatible ideas that avoid stuttering with `Fetcher.FetchBlocks`?

net/fetcher.go
ZenGround0: cc @whyrusleeping I could use your input as I inherited this from your code.

chain/default_syncer.go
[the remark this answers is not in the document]
whyrusleeping: it knows about the [block](https://github.com/ipfs/go-block-format) that was named block when nobody was really talking about blockchains... It does not know about the go-filecoin block structure. It does (as @ZenGround0 pointed out) know how to check that data it gets matches the requested CID though.

chain/default_store.go
ZenGround0: I made a somewhat arbitrary decision based on poor understanding.  There was nothing significant about it.

net/fetcher.go
ZenGround0: I agree, I will do some error checking to only log when its not nil

net/fetcher_test.go
ZenGround0: I want to make sure that the offending terminology is `rtBlocks` and `rtCids`.  If it's something else please advise and I'll fix that too.

chain/default_syncer.go
```diff
@@ -79,91 +89,74 @@ func NewDefaultSyncer(online, offline *hamt.CborIpldStore, c consensus.Protocol,
...
 
 		height, _ := ts.Height()
 		if len(chain)%500 == 0 {
 			logSyncer.Infof("syncing the chain, currently at block height %d", height)
-		}
-
-		// Finish traversal if the tipset made is tracked in the store.
-		if syncer.chainStore.HasTipSetAndState(ctx, tsKey) {
-			return chain, ts, nil
+			syncer.fetcher.LogStat()
```
whyrusleeping: i'd lean toward making this method actually return the stats, and then print it where needed. Me putting this here was really just a debug log to help observe the improvements.

ipfs has this wired through as a command.

net/fetcher.go
whyrusleeping: i'm not sure i grok @phritz's comment relative to what he's commenting on. The session being here is the whole point, and is an optimized 'fetching thingy'. I think having the blockservice here isnt necessary, and that getting the statistics for block sends and receives should be wired through separately (maybe as a command like we do in ipfs)

whyrusleeping: as i was writing that comment I started to understand what phritz is saying. And agree, the blockservice really doesnt belong here.

node/block.go
whyrusleeping: seems reasonable

whyrusleeping:
@ZenGround0 aside from the comment about the blockservice not being needed in the fetcher, and the logstats needing to return the data isntead of just printing it, this LGTM. We should be able to measure the impact this has syncing a new chain before and after the patch.

chain/default_syncer.go
ZenGround0: Because the blockservice is needed to get the stats and we are talking about removing this from the Fetcher I am just going to remove logstats altogether and file a followon issue for wiring bitswap up to a command for viewing stats.

PR #2069, opened 2019-02-22
cache block cids for faster everything

whyrusleeping:
improves chain validation part of sync time by a lot.

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -63,7 +63,7 @@ type Block struct {
 // set this to true to panic if the blocks data differs from the cached cid. This should
 // be obviated by changing the block to have protected construction, private fields, and
 // getters for all the values.
-const paranoid = false
+var paranoid = false
 
 // Cid returns the content id of this block.
 func (b *Block) Cid() cid.Cid {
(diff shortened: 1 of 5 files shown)


whyrusleeping:
The caveat here is that we will need to ensure that once a block is created (and its cid is cached) we never change it. This should probably be done by turning block into a protected type (private fields, controlled creation, getters and such).

For now, i've included the 'paranoid' flag that degraded performance, but will panic if someone changed a block after it was cached.

types/block.go
```diff
@@ -54,13 +54,47 @@ type Block struct {
...
 	Proof proofs.PoStProof `json:"proof"`
+
+	cachedCid cid.Cid
+
+	cachedBytes []byte
 }
 
+// set this to true to panic if the blocks data differs from the cached cid. This should
+// be obviated by changing the block to have protected construction, private fields, and
+// getters for all the values.
```
phritz: can you file an issue for this against testnet and link to it here? hidden technical debt is what sunk the titanic.

types/block.go
```diff
@@ -54,13 +54,47 @@ type Block struct {
...
+const paranoid = false
```
phritz: please add a test that uses assert.panics to verify this behavior

whyrusleeping:
woo! green tests!

PR #2092, opened 2019-02-25
Move plumbing/ps up to top level pubsub package and merge into network plumbing

### LINKED ISSUE #2048: Refactor libp2p pubsub wrapper to top level package and move plumbing.ps to plumbing.ntwk
See discussion in #2019

anorth:
There are two commits here, with most of the work in the first to wrap libp2p/pubsub with our own. This adds a little more code than was existing in order to also wrap Message type, but is then sufficient to remove all imports of `go-libp2p-pubsub` except by `node` and the wrappers themselves.

The second implements @rosalinekarr's suggestion to merge pubsb plumbing into network plumbing. This exposes a weird test where we try to construct a complete plumbing from the outside of a node.

Closes #2048.


pubsub/testing.go
```diff
@@ -59,7 +75,7 @@ func (s *FakeSubscription) Cancel() {
 // Manipulators
 
 // Post posts a new message to this subscription.
-func (s *FakeSubscription) Post(msg *pubsub.Message) {
+func (s *FakeSubscription) Post(msg *FakeMessage) {
```
phritz: shouldnt this take a message, not a fakemessage?

anorth: Thanks, sorry, that was a remnant from before I wrapped Message.

pubsub/subscriber.go
```diff
@@ -0,0 +1,52 @@
...
+	Cancel()
+}
+
+// NewSubscriber builds a new subscriber
+func NewSubscriber(sub *pubsub.PubSub) *Subscriber {
+	return &Subscriber{pubsub: sub}
+}
+
+// Subscribe subscribes to a pubsub topic
+func (s *Subscriber) Subscribe(topic string) (Subscription, error) {
```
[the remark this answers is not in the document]
anorth: First note that I've not created a `Subscriber` interface, only a concrete struct. The plumbing holds references to a concrete type, but one owned by us. The `Subscribe` method on that struct does differ from `libp2p`'s, though, omitting a unused variadic parameter of another `libp2p`-defined type (`SubOpt`, an alias for a function pointer). Defining a `Subscriber` interface is possible but not currently needed in order to abstract libp2p from code, excepting `node.go`. It too would need a wrapper to preserve the simpler `Subscribe`, since Go won't accept a function absent an arg as a subtype of a function with a trailing variadic arg.

One level down, `Subscription` is an interface. It matches `libp2p` as closely as possible while hiding the `libp2p/pubsub.Message` type (a struct 😕). The local `Message` interface is a subset of the `libp2p` one, but it seems the type checker doesn't consider a pointer to a struct `Message` acceptable as a value of an interface `Message`, for reasons that make sense to me.

```pubsub/subscriber.go:41:2: cannot use sub (type *pubsub.Subscription) as type Subscription in return argument:
	*pubsub.Subscription does not implement Subscription (wrong type for Next method)
		have Next(context.Context) (*pubsub.Message, error)
		want Next(context.Context) (Message, error)```

However, writing the seemingly-redundant extension struct and wrapper around `Next` gives Go the hook to do an implicit conversion from struct pointer to interface value.

We could expose the libp2p `Message` struct, but that will add libp2p imports in a few places and make use in testing harder (it's basically a wrapper of a protobuf). Or libp2p could define `Message` as an interface, which we could then subset. I think this is an example of the Go pattern of defining interfaces only at consumers falls down; the producer would do us a favour by defining a (broader) interface.

**Comment by anorth** [comment]
Thanks for that final comment, and I agree. It felt wrong to add another top-level package for this, but that was the existing pattern. I think we could benefit from a level of nesting in some places, like the 'protocol/x' packages. The directory structure could look more like the architecture diagram.

**Comment by phritz** [comment]
i would love to have a small set of top level directories something like

core/
plumbing/
porcelain/
protocols/
tools/

File: pubsub/testing.go [hunk 3]
```diff
@@ -1,16 +1,32 @@
-package ps
+package pubsub
```
rosalinekarr: You may want to rename this file to `pubsub/subscriber_test.go` to match our naming convention

anorth: This is not the tests, this is test support code.

Issue #1751, opened 2019-01-31
Order Messages by Gas Price


acruikshank:
### Description

Filecoin block miners should be able to process messages in a way that generates the most profit. Choosing to process messages with a higher gas prices is the way to do this.

Miners must continue to take message nonce into consideration, so the next message to process must satisfy

1. The message has a lower nonce than any other message from the same actor.
2. The message's gas price is has price greater than or equal to any other message that meets the first condition.

### Acceptance criteria

1. When the messages in the message pool exceed the block's gas limit, only the messages prioritized by the above rules appear in the block.

### Risks + pitfalls

Sending a price of 0 is still valid (although now potentially has consequences), so tests should still probably pass.

### Where to begin

processor.go

anorth:
I've picked this up. My approach is to build per-actor queues ordered by nonce, then a priority-queue of actor queues ordered by gas price of the next message from that actor, and merge these actor queues into a single list.

I believe this is optimal if ignoring the bin-packing aspect of the problem. More squeezing is possible by also considering the gas limit and solving a 0/1 knapsack problem.

PR #2093, opened 2019-02-25
Move core.OrderMessagesByNonce to mining.SelectMessagesForBlock and simplify tests.

anorth:
This is a prefactor for work on #1751 ordering by gas price. I've moved OrderMessagesByNonce next to its only caller.

In the long term, we may want to move sophisticated message queueing inside the messages pool, but until performance requires it, let's keep the pool simple.

### DIFF
--- a/mining/block_generate.go
+++ b/mining/block_generate.go
@@ -6,11 +6,12 @@ package mining
 
 import (
 	"context"
+	"sort"
 	"time"
 
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
 
-	"github.com/filecoin-project/go-filecoin/core"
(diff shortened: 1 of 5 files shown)


mining/block_generate.go
```diff
@@ -91,15 +92,17 @@ func (w *DefaultWorker) Generate(ctx context.Context,
 		Ticket:          ticket,
 	}
 
-	// TODO: Should we really be pruning the message pool here at all? Maybe this should happen elsewhere.
 	for i, msg := range res.PermanentFailures {
 		// We will not be able to apply this message in the future because the error was permanent.
 		// Therefore, we will remove it from the MessagePool now.
+		// There might be better places to do this, such as wherever succesful messages are removed
```
acruikshank: There's a linter error because successful is misspelled here.

PR #2094, opened 2019-02-25
perform lint step before build and test steps

ognots:
run lint step after installing dependencies and saving them to cache.
gometalinter is installed by dependencies step so this step can't be any earlier

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,6 +200,10 @@ jobs:
             - "/home/circleci/.cargo"
 
       - run:
+          name: Lint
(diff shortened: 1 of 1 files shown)


ognots:
Note: Installing `gometalinter` outside `deps` or `smartdeps` could make the linting occur earlier, but would require an additional step for humans, so I did not pursue this option. Feedback welcome on this decision.

ognots:
Thanks for process feedback! Will be sure to assign reviewers - I'm just not always sure are the best candidates sometimes

Issue #1693, opened 2019-01-24
Decouple the `Address` and `Wallet` commands


frrist:
### Description
Currently the `address` and `wallet` commands have overlap, i.e.:
```
USAGE
  go-filecoin wallet - Manage your filecoin wallets

  go-filecoin wallet

SUBCOMMANDS
  go-filecoin wallet addrs                 - Interact with addresses
  go-filecoin wallet balance <address>     - 
  go-filecoin wallet export <addresses>... - 
  go-filecoin wallet import <walletFile>   -
```
```
USAGE
  go-filecoin address - Interact with addresses

  go-filecoin address

SUBCOMMANDS
  go-filecoin address lookup <address> - 
  go-filecoin address ls               - 
  go-filecoin address new              -
```

This is redundant as prefixing the `address` command with `wallet` doesn't give us an additional functionality, e.g. there is no difference between calling `go-filecoin wallet addrs ls` and `go-filecoin address ls` -- it's just confusing. 

### Acceptance criteria
- Remove `addrs` subcommand from the `wallet` command s.t. the `wallet` command and `address` command have zero overlap in functionality.
- Update any affected tests.

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/master/commands/address.go#L14
https://github.com/filecoin-project/go-filecoin/blob/master/commands/address.go#L26

rosalinekarr:
Some of this work is likely going to conflict with some of my work on porcelain/plumbing. Just ping me if you start working on any of this and we can discuss specifics.

rosalinekarr:
Ok, so after reading over this again. I don't think this should conflict since most of this work will be in the `commands` directory, and the porcelain/plumbing work will be in the `api` directory. What few if any conflicts do arise from this should be manageable.

alanshaw:
I'm going to try this one!

PR #2100, opened 2019-02-25
correctly export wallet data

### LINKED ISSUE #2024: [bug] wallet export / import use the different format
when I export my wallet addr:
```sh
go-filecoin wallet export fcqhn.....
```
got
```
Address:	fcqhn.....
PrivateKey:	ceaf5e302....
Curve:		secp256k1
```

or use `--enc=json` get a json output, but
```sh
go-filecoin wallet import <walletFile> 
```
got
```
Error: invalid length, need 256 bits
```

why not wallet export / import use the same format

frrist:
### Whats this?

This PR aligns the type `go-filecoin wallet export --enc=json` returns and the type `go-filecoin wallet import` expects. 

#2024 was caused by the `export` command returning:
`{"privateKey":"xVHXgLfpcJ6kisQONRelKHA5DuY+cpIvTTK9K3utxDI=","curve":"secp256k1"}`

And the `import` command expexting:
`{"KeyInfo":[{"privateKey":"xVHXgLfpcJ6kisQONRelKHA5DuY+cpIvTTK9K3utxDI=","curve":"secp256k1"}]}`

closes #2024

### DIFF
--- a/api/address.go
+++ b/api/address.go
@@ -13,7 +13,7 @@ import (
 // Address is the interface that defines methods to manage Filecoin addresses and wallets.
 type Address interface {
 	Addrs() Addrs
-	Import(ctx context.Context, f files.Directory) ([]address.Address, error)
+	Import(ctx context.Context, f files.File) ([]address.Address, error)
 	Export(ctx context.Context, addrs []address.Address) ([]*types.KeyInfo, error)
 }
 
--- a/api/impl/address.go
+++ b/api/impl/address.go
@@ -57,12 +57,17 @@ func (api *nodeAddrs) Lookup(ctx context.Context, addr address.Address) (peer.ID
 	return id, nil
 }
 
-func (api *nodeAddress) Import(ctx context.Context, d files.Directory) ([]address.Address, error) {
+func (api *nodeAddress) Import(ctx context.Context, f files.File) ([]address.Address, error) {
 	nd := api.api.node
 
-	kinfos, err := parseKeyInfos(d)
+	kinfos, err := parseKeyInfos(f)
 	if err != nil {
 		return nil, err
+	}
+
+	// error if we fail to parse keyinfo from the provided file.
+	if len(kinfos) == 0 {
+		return nil, fmt.Errorf("no keys in wallet file")
 	}
 
 	dsb := nd.Wallet.Backends(wallet.DSBackendType)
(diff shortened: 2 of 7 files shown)


api/address.go
```diff
@@ -13,7 +13,8 @@ import (
 // Address is the interface that defines methods to manage Filecoin addresses and wallets.
 type Address interface {
```
anorth: Aside: this is a confusing name for an interface that contains multiple addresses.

api/impl/address.go
```diff
@@ -57,14 +57,19 @@ func (api *nodeAddrs) Lookup(ctx context.Context, addr address.Address) (peer.ID
...
 
-	kinfos, err := parseKeyInfos(d)
+	kinfos, err := parseKeyInfos(f)
 	if err != nil {
 		return nil, err
 	}
 
+	// error if we fail to parse keyinfo from the provided file.
+	if len(kinfos) == 0 {
+		return nil, fmt.Errorf("failed to parse keys from wallet file")
```
anorth: Be more specific "no keys in wallet file %s"

commands/address.go
```diff
@@ -142,7 +143,17 @@ var walletImportCmd = &cmds.Command{
...
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		addrs, err := GetAPI(env).Address().Import(req.Context, req.Files)
+		iter := req.Files.Entries()
+		if !iter.Next() {
+			return fmt.Errorf("no file given: %s", iter.Err())
+		}
+
+		fi, ok := iter.Node().(files.File)
+		if !ok {
+			return fmt.Errorf("given file was not a files.File")
```
anorth: Since this message will go to the user, perhaps avoid Go terms and say "%s is not a file"

api/impl/address.go
```diff
@@ -110,21 +115,15 @@ func (api *nodeAddress) Export(ctx context.Context, addrs []address.Address) ([]
...
-		fi, ok := iter.Node().(files.File)
-		if !ok {
-			return nil, fmt.Errorf("file passed to import was not a file")
-		}
-
-		var ki types.KeyInfo
-		if err := json.NewDecoder(fi).Decode(&ki); err != nil {
-			return nil, err
-		}
+// WalletImportResult is the structure used to hold imported wallet data.
```
anorth: Do I understand correctly that this type defines the schema expected for imported wallets? If so make that very explicit, and probably raise this higher up in the file, above the funcs.

api/address.go
frrist: Agreed, this bit is even more confusing: https://github.com/filecoin-project/go-filecoin/issues/1693.

api/impl/address.go
frrist: I will state that there are "no keys in the file", but I don't believe the interface for `go-ipfs-files` we are using has a `filename`/`string` methods, meaning I can't return the name of the file :(

frrist: Aforementioned files interface: https://ipfs.io/ipfs/QmQmhotPUzVrMEWNK3x1R5jQ5ZHWyL7tVUrmRPjrBrvyCb/go-ipfs-files/file.go

commands/address.go
frrist: Same issue as stated here: https://github.com/filecoin-project/go-filecoin/pull/2100#discussion_r260047378

api/impl/address.go
frrist: The `WalletImportResult` type and [this type](https://github.com/filecoin-project/go-filecoin/blob/master/commands/address.go#L171) are the same -- I think there should be 1 type both commands use. The type cannot live in `commands/` as that will cause an import cycle with `api/impl`. I suppose I can move it to `api/impl` and import that into `commands/`

commands/address_daemon_test.go
```diff
@@ -115,3 +117,26 @@ func TestWalletLoadFromFile(t *testing.T) {
...
 }
+
+func TestWalletExportImportRoundTrip(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	d := th.NewDaemon(t).Start()
+	defer d.ShutdownSuccess()
+
+	dw := d.RunSuccess("address", "ls").ReadStdoutTrimNewlines()
```
phritz: should probably create a few addresses before calling ls. my guess is there might be ordering issues, but maybe not.

frrist: The ordering on import and export shouldn't matter, the problem was the way the exported structure was serialized to json didn't match what import expected.

PR #2103, opened 2019-02-26
Implement a priority queue for messages to be mined.

### LINKED ISSUE #1751: Order Messages by Gas Price

anorth:
Messages are ordered by decreasing gas price subject to per-address nonces strictly
increasing.

Closes #1751.

### DIFF
--- a/mining/block_generate.go
+++ b/mining/block_generate.go
@@ -6,12 +6,10 @@ package mining
 
 import (
 	"context"
-	"sort"
 	"time"
 
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
 
-	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/vm"
@@ -56,9 +54,8 @@ func (w *DefaultWorker) Generate(ctx context.Context,
 	}
 
 	pending := w.messageSource.Pending()
-	messages := make([]*types.SignedMessage, len(pending))
-
-	copy(messages, SelectMessagesForBlock(pending))
+	mq := NewMessageQueue(pending)
+	messages := mq.Drain()
 
 	vms := vm.NewStorageMap(w.blockstore)
 	res, err := w.processor.ApplyMessagesAndPayRewards(ctx, stateTree, vms, messages, w.minerAddr, types.NewBlockHeight(blockHeight), ancestors)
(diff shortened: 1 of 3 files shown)


mining/mqueue.go
```diff
@@ -0,0 +1,114 @@
...
+	aq[i], aq[j] = aq[j], aq[i]
+}
+
+func (aq *queueHeap) Push(x interface{}) {
+	item := x.(nonceQueue)
+	*aq = append(*aq, item)
+}
+
+func (aq *queueHeap) Pop() interface{} {
+	old := *aq
```
acruikshank: Since you only mutate *aq on the last operation, isn't `old` uneccessary?

mining/mqueue.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// MessageQueue is a priority queue of messages from different actors. Messages are ordered
+// by decreasing gas price, subject to the constraint that messages from a single actor are
+// always in increasing nonce order.
+// All messages for a queue are inserted at construction, after which messages may only
+// be popped.
+// Potential improvements include:
+// - deprioritising messages after a gap in nonce value, which can never be mined (see Ethereum)
+// - attempting to pack messages into a fixed gas limit (i.e. 0/1 knapsack subject to nonce ordering)
```
shannonwells: I'm unfamiliar with the "knapsack" terminology; what does this refer to?

mining/block_generate.go
```diff
@@ -56,9 +54,8 @@ func (w *DefaultWorker) Generate(ctx context.Context,
 	}
 
 	pending := w.messageSource.Pending()
-	messages := make([]*types.SignedMessage, len(pending))
-
-	copy(messages, SelectMessagesForBlock(pending))
+	mq := NewMessageQueue(pending)
+	messages := mq.Drain()
```
shannonwells: yesssss

mining/mqueue.go
```diff
@@ -0,0 +1,114 @@
...
+// A slice of messages ordered by Nonce (for a single sender).
+type nonceQueue []*types.SignedMessage
+
+// Implements heap.Interface to hold a priority queue of queues per sender.
+// Heap priority is given by the gas price of the first (lowest nonce) message for each queue.
+type queueHeap []nonceQueue
+
+func (aq queueHeap) Len() int { return len(aq) }
+
+// Len implements Heap.Interface.Less to compare items on gas price and sender address.
```
shannonwells: >Len implements . . .

Less implements?

mining/mqueue.go
```diff
@@ -0,0 +1,114 @@
...
+	bestQueue := &mq.bySender[0]
+
+	// Pop first message off that actor's queue
+	msg := (*bestQueue)[0]
+	if len(*bestQueue) == 1 {
+		// If the queue will become empty, remove it from the heap.
+		heap.Pop(&mq.bySender)
+	} else {
+		*bestQueue = (*bestQueue)[1:]
+		heap.Fix(&mq.bySender, 0)
```
shannonwells: So this first element will be garbage collected after passing the value back?  Also I'm not sure why you don't call Pop here, since it seems like this is already ordered.  Perhaps I don't understand what Fix does (still somewhat of a go newb).

mining/mqueue.go
anorth: I added a comment: https://en.wikipedia.org/wiki/Knapsack_problem

mining/mqueue.go
anorth: Thanks, I've added more comments.
The actor's queue needs to be re-ordered in the heap based on the gas price of the next message it contains. `Fix` is equivalent to but faster than `x = heap.Pop(); heap.Push(x)`.

Issue #2053, opened 2019-02-21
Add 'dag get' command for messages (or viewable somehow by hash)


whyrusleeping:
I expect to be able to run `go-filecoin dag get <Message CID>` and have it print me out the message. In addition, a nicely formatted version could be made available via `go-filecoin show message <Message CID>`

anorth:
This sounds the same as https://github.com/filecoin-project/go-filecoin/issues/1933. If not, could you please expand on the difference?

whyrusleeping:
@anorth ah, i guess this issue proposes a solution to that issue. otherwise, yes. addressing the same problem.

anorth:
Great, closing in favour of that since I'm already on it.

PR #2109, opened 2019-02-26
Add mpool show command to show an unmined message detail.

### LINKED ISSUE #1933: Add command for viewing a message in the mpool

anorth:
I emulated the output format from `block show`.

```Message Details
To:        fcqp606qfk5gwmq6ac24g4mhv3cr8zzf67vqkpulh
From:      fcq6k3r52n89c0m0nd8sm8cf27srpgephkqkrc9nj
Nonce:     102
Value:     100000
Method:    createChannel
Params:    glYAANrn4y/rt8o+bnah3odFEv661KJWQ+KXAg==
Gas price: 0
Gas limit: 300
Signature: Pzn2NWyHjcO7gc4denBVC64nBb/n7otlEmyyC2g1vHJhM3yufdVl07FmCugC4mQihu5302vxU3SPU3eBDKI5xgA=

Message Details
To:        fcq423mk77skf2tg2542ku6skje9859ecu3z7dmha
From:      fcqunwslgsaa52xuqxlxvpx4km8h2erprz9trc4l3
Nonce:     2
Value:     0
Method:
Params:
Gas price: 1000
Gas limit: 1000
Signature: U3V2tqwZfa5KxH//GdapekmG5icglvddij7S0yR20iVGWGMfbo3hpo8LVbwJcmXe9sj+jpI5bvZSzon427alGgE=
```
Params are "too hard" to decode here without more work. My understanding is that decoding them requires knowledge of the expected schema–it's not self-describing.

Closes #1933 (see also #2053).

### DIFF
--- a/commands/mpool.go
+++ b/commands/mpool.go
@@ -1,8 +1,10 @@
 package commands
 
 import (
+	"encoding/base64"
 	"fmt"
 	"io"
+	"strconv"
 
 	"gx/ipfs/QmQtQrtNioesAWtrx8csBvfY37gTe94d6wQ3VikZUjxD39/go-ipfs-cmds"
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
@@ -17,8 +19,9 @@ var mpoolCmd = &cmds.Command{
 		Tagline: "Manage the message pool",
 	},
 	Subcommands: map[string]*cmds.Command{
-		"ls": mpoolLsCmd,
-		"rm": mpoolRemoveCmd,
+		"ls":   mpoolLsCmd,
+		"show": mpoolShowCmd,
+		"rm":   mpoolRemoveCmd,
 	},
 }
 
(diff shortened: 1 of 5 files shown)


commands/mpool.go
```diff
@@ -47,13 +50,58 @@ var mpoolLsCmd = &cmds.Command{
 				if err != nil {
 					return err
 				}
-				fmt.Fprintln(w, c.String()) // nolint: errcheck
+				_ = PrintString(w, c)
```
ZenGround0: 👍

commands/mpool.go
```diff
@@ -47,13 +50,58 @@ var mpoolLsCmd = &cmds.Command{
...
 		}),
 	},
 }
 
+var mpoolShowCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Show content of an outstanding message",
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("cid", true, false, "The CID of the message to delete"),
```
ZenGround0: BLOCKING: argument message looks like copypasta from the delete command

mishmosh:
@shannonwells please review?

core/message_pool.go
```diff
@@ -56,6 +56,13 @@ func (pool *MessagePool) Pending() []*types.SignedMessage {
 	return out
 }
 
+// Get retrieves a message from the pool by CID.
+func (pool *MessagePool) Get(c cid.Cid) *types.SignedMessage {
```
ZenGround0: I may be in the minority here but my opinion is that this interface would be more intuitive if it returned an error in the case that the pool does not contain the requested cid.

When called later from commands (or elsewhere) this error could then be propagated and returned to the user.

I understand the desire to use the zero value as a more concise indicator that the message isn't found, but for some reason it feels less golang-like.  One nice thing about returning an error when the message isn't found is that we can easily distinguish lack of membership from the should-never-happen bad-state situation where the message pool maps a legitimate cid to nil.

Interested to hear other opinions as I'm not sure if this is a mostly-good, mostly-bad, or neutral opinion from the standard golang style point of view.

commands/mpool.go
```diff
@@ -47,13 +50,58 @@ var mpoolLsCmd = &cmds.Command{
...
+	},
+	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
+		msgCid, err := cid.Parse(req.Arguments[0])
+		if err != nil {
+			return errors.Wrap(err, "invalid message cid")
+		}
+
+		msg := GetPorcelainAPI(env).MessagePoolGet(msgCid)
```
ZenGround0: BLOCKING: Even if you disagree with my comment below on `mpool:Get` returning an error this needs to emit an error in the case the message is not found.  Otherwise the encoder will panic trying to dereference the nil pointer.

core/message_pool.go
shannonwells: I would add that doing it this way breaks a convention in the code base, and so it violates the principle of least surprise.

commands/mpool.go
anorth: Eek, thanks

commands/mpool.go
anorth: Thanks for catching. I had intended to print nothing, since not finding something in the message pool is hardly within the user/caller's control. It might even be the desired outcome.

anorth:
Thanks for the reviews. I have changed `Get` to mimic map, returning a value and boolean indicating found (discussed in Slack).

It proved too hard to wrangle `cmds` to deal with a nil success value, so "not found" is explicitly handled as an error. I believe this addresses everything so I'll merge.

anorth:
I also added tests. Please don't accept new functionality without tests.

PR #2112, opened 2019-02-26
Implement custom wrapper for secp256k1

### LINKED ISSUE #1991: [bug] LGPL code included in liberally licensed repo
Thanks for open sourcing this, looks awesome.

But I noticed the `crypto` pkg copied over from go-ethereum is LGPL licensed, which in my understanding is incompatible with Apache/MIT.

dignifiedquire:
Resolves #1991 

I would also suggest to do rename the package to `secp256k1` given that bls signatures are coming soon. And then as a follow up create the structure like this
- `crypto`
  - `crypto/secp256k1`
  - `crypto/bls`


frrist:
Looks like the failing test is `TestMinerCreate` due to timeout, will investigate.

dignifiedquire:
LGTM

crypto/crypto.go
```diff
@@ -1,128 +1,68 @@
...
-	return BytesToECDSA(b)
-}
 
-// GenerateKey generates an ecdsa private key
-func GenerateKey() (*ecdsa.PrivateKey, error) {
-	return GenerateKeyFromSeed(rand.Reader)
+	privkey := make([]byte, PrivateKeyBytes)
+	blob := key.D.Bytes()
+
+	copy(privkey[PrivateKeyBytes-len(blob):], blob)
```
anorth: Please document the context/intent for this. I don't understand why it's needed. When could it blob not be the right length? What if blob is longer than PrivateKeyBytes?

wallet/dsbackend.go
```diff
@@ -145,18 +143,13 @@ func (backend *DSBackend) SignBytes(data []byte, addr address.Address) (types.Si
...
-	return wutil.Sign(privateKey, data)
+	return wutil.Sign(ki.Key(), data)
 }
 
 // Verify cryptographically verifies that 'sig' is the signed hash of 'data' with
 // the public key `pk`.
-func (backend *DSBackend) Verify(data []byte, pk []byte, sig types.Signature) (bool, error) {
-	return wutil.Verify(pk, data, sig)
+func (backend *DSBackend) Verify(data, pk []byte, sig types.Signature) (bool, error) {
+	return crypto.Verify(pk, data, sig), nil
```
anorth: This appears to be the only implementation of the interface, so drop the error return value.

crypto/crypto.go
frrist: @dignifiedquire can you chime in on this one? I am unsure.

crypto/crypto.go
```diff
@@ -1,128 +1,68 @@
...
 
-// BytesToECDSAPub creates a public key with the given `pub` value.
-func BytesToECDSAPub(pub []byte) *ecdsa.PublicKey {
-	if len(pub) == 0 {
-		return nil
+// Verify checks the given signature and returns true if it is valid.
+func Verify(pk, msg, signature []byte) bool {
+	if len(signature) == 65 {
+		// Drop the V in [R | S | V] style signatures
+		return secp256k1.VerifySignature(pk[:], msg, signature[:len(signature)-2])
```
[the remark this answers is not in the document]
dignifiedquire: thank you for catching, fixed

crypto/crypto.go
dignifiedquire: Unless the elliptic curve implementation is broken this has a fixed size, always as the serialization for these points is defined to be a fixed byte size for this particular curve

anorth: Ok. It would help less familiar reads to leave a comment here to that effect.

crypto/crypto.go
```diff
@@ -1,128 +1,69 @@
...
 }
 
-// BytesToECDSAPub creates a public key with the given `pub` value.
-func BytesToECDSAPub(pub []byte) *ecdsa.PublicKey {
-	if len(pub) == 0 {
-		return nil
+// Verify checks the given signature and returns true if it is valid.
+func Verify(pk, msg, signature []byte) bool {
+	if len(signature) == 65 {
+		// Drop the V (1byte) in [R | S | V] style signatures.
```
phritz: in a brief comment can you say something about what this means for the unfamilar? seems weird to accommodate some other format.

crypto/crypto_test.go
```diff
@@ -1,88 +1,60 @@
...
-	sig, err := Sign(msg[:], key)
-	assert.NoError(err)
-
-	// recover a public key from a message and a signature, it should be
-	// equal to the previously generated public key
-	recPub, err := Ecrecover(msg[:], sig)
-	assert.NoError(err)
-	assert.Equal(expPub, recPub)
+	assert.Equal(len(sk), 32)
+	assert.NotEqual(sk[0], 0)
```
phritz: why not zero? please add a comment for the unfamiliar whose reaction to seeing this line like is that it is going to flake 1/256 runs.

crypto/crypto_test.go
```diff
@@ -1,88 +1,60 @@
...
+	msg2 := make([]byte, 32)
+	copy(msg2, msg)
+	msg2[0] = 1
+	msg2[1] = 255
+	assert.False(crypto.Verify(pk, msg2, digest))
+
+	// invalid signature - different digest
+	digest2 := make([]byte, 65)
+	copy(digest2, digest)
+	digest2[0] = 1
```
whyrusleeping: so does the test fail if `digest[0:2] == []byte{1,255}` ? should happen every 1/2^16 runs

frrist: sure does:
```
frrist@pine ~/s/g/f/g/crypto (switch-crypto)> go test -count=25600 ./...
--- FAIL: TestGenerateKey (0.00s)
    crypto_test.go:46: 
                Error Trace:    crypto_test.go:46
                Error:          Should be false
                Test:           TestGenerateKey
FAIL
FAIL    github.com/filecoin-project/go-filecoin/crypto  30.080s
```
will fix

dignifiedquire: ups

frrist:
@phritz @anorth @whyrusleeping @dignifiedquire want to double check we are good to merge here, after addressing more comments, any objects?

Issue #2032, opened 2019-02-20
[bug] `pubKey` field in `createMiner` messages is null


mslipper:
**Describe the bug**
Take a look at block height 12925, transaction index 5, on the user devnet. You'll see that it's a `createMiner` message with the message params `g0EK9lgiEiCQIEGVW4dOtQy6yuSsBbCCBvtM2Hjrv7z7SqJNAoJsWw==`. When these params are ABI-decoded, the following result pops out:

```
> cbor.decode(Buffer.from('g0EK9lgiEiCQIEGVW4dOtQy6yuSsBbCCBvtM2Hjrv7z7SqJNAoJsWw==', 'base64'))
[ <Buffer 0a>,
  null,
  <Buffer 12 20 90 20 41 95 5b 87 4e b5 0c ba ca e4 ac 05 b0 82 06 fb 4c d8 78 eb bf bc fb 4a a2 4d 02 82 6c 5b> ]
```

The second item of the array is supposed to be the `pubKey` field, but it's null. Is this expected?

**To Reproduce**
1. Run `chain.Ls()` via HTTP to get full JSON responses.
2. Look at the block and transaction index listed above.

**Expected behavior**
The miner's pubkey should never be null - or should it? I'm not sure.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04

**Additional context**
This brought down the stats dashboard. The dashboard now tolerates null pubkeys, but I'd like to get to the bottom of what's happening here.

mslipper:
Any updates all? I'm seeing a _lot_ of null miner addresses in the logs now.

shannonwells:
@mslipper I'm looking at it now - PR #2115  may fix it; I'm currently writing a test to repro the issue.

phritz:
i think this was not reproducible, @mslipper please re open if you are still seeing this

PR #2113, opened 2019-02-26
Feat/miner create porcelain 1161

### LINKED ISSUE #1161: the `miner create` command should be porcelain
### Description

The `miner create` command blocks until the `createMiner` message has been included in a block. This is different than `miner update-peerid` (and other commands), which return the cid of the created message (leaving the user to `message wait <cid>` as a second step).

This is because `miner create` implements a user flow rather than a single action. The user issues a create miner message on the network, the command waits for a block containing the create miner transaction, and then that result  is stored in config.

We will continue to need more and more commands like this that perform higher level actions on behalf of the user, but we should not design the system around these complex tasks. We will introduce the concept of porcelain, to distinguish higher level actions that depend on multiple api calls, from the lower level 'plumbing' commands that mirror the api.

### Acceptance criteria

- create a `porcelain` package subpackage of `commands`.
- miner create (and other porcelain commands) have no unique api calls. All functionality is implemented with api calls representing by existing plumbing commands.

### Where to begin

- message send
- message wait
- config

rosalinekarr:
# Problem

We want to delete the `api` directory, but the `Miner().Create()` method is part of it.

# Solution

Move `Miner().Create()` to porcelain.

Resolves #1161


rosalinekarr:
@acruikshank @phritz There's still some small cleanup and renaming that I want to do, but all the logic is here now to be review.

node/node.go
```diff
@@ -988,81 +985,6 @@ func (node *Node) NewAddress() (address.Address, error) {
...
-	if err != nil {
-		return &minerAddr, err
-	}
-
-	err = node.saveMinerConfig(minerAddr, blockSignerAddr)
-	if err != nil {
-		return &minerAddr, err
-	}
-
-	err = node.setupMining(ctx)
```
rosalinekarr: It turns out we never actually needed to call this on Miner Create. Mining Start handles it.

rosalinekarr:
⚠️ *DO NOT MERGE YET!!!* ⚠️

This work has a high risk of re-introducing bugs that will be resolved by @shannonwells's current work. This PR should not be merged until issue #2032 is finished and this work is rebased on it.

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+) (_ *address.Address, err error) {
+	if accountAddr == (address.Address{}) {
+		accountAddr, err = plumbing.GetAndMaybeSetDefaultSenderAddress()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if pid == "" {
+		pid = plumbing.NetworkGetPeerID()
```
ZenGround0: why not do this all the time and forget the pid argument?

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+}
+
+// MinerCreate creates a new miner actor for the given account and returns its address.
+// It will wait for the the actor to appear on-chain and add set the address to mining.minerAddress in the config.
+// TODO: add ability to pass in a KeyInfo to store for signing blocks.
+//       See https://github.com/filecoin-project/go-filecoin/issues/1843
+func MinerCreate(
+	ctx context.Context,
+	plumbing mcAPI,
+	accountAddr address.Address,
```
ZenGround0: Naming this `ownerAddr` would probably help avoid confusion.

porcelain/miner.go
phritz: or better, eliminate this if and require the caller pass the peerid in? actually why not have the caller do all the set up and pass in what this function needs to do its job, namely the peerid, account addr, and pubkey? would make this call simpler and easier to test.

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if pid == "" {
+		pid = plumbing.NetworkGetPeerID()
+	}
+
+	ctx = log.Start(ctx, "Node.CreateMiner")
```
phritz: i know you didn't introduce this, but do we get any value out of this? i honestly don't know.

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+			return vmErrors.VMExitCodeToError(receipt.ExitCode, storagemarket.Errors)
+		}
+		minerAddr, err = address.NewFromBytes(receipt.Return[0])
+		return err
+	})
+	return minerAddr, err
+}
+
+func persistMinerAddress(ctx context.Context, plumbing mcAPI, minerAddr address.Address) error {
+	// TODO: https://github.com/filecoin-project/go-filecoin/issues/1843
```
phritz: this todo is referenced at the entrypoint above, suggest that is sufficient so remove this reminder

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+	res, _, err := plumbing.MessageQuery(
```
phritz: i guess it is strictly correct to do this but it also seems like it might be unnecessary given that we have the owner addr in the caller and could just pass it in. consider simplifying, we could just move the two configsets into the caller.

node/node.go
```diff
@@ -988,81 +985,6 @@ func (node *Node) NewAddress() (address.Address, error) {
 	return wallet.NewAddress(node.Wallet)
 }
 
-// CreateMiner creates a new miner actor for the given account and returns its address.
```
phritz: omg yay red!!!

porcelain/miner_test.go
```diff
@@ -22,6 +23,103 @@ import (
...
+
+func (mpc *minerCreate) WalletGetPubKeyForAddress(addr address.Address) ([]byte, error) {
+	return mpc.wallet.GetPubKeyForAddress(addr)
+}
+
+func (mpc *minerCreate) WalletHasAddress(addr address.Address) bool {
+	return mpc.wallet.HasAddress(addr)
+}
+
+func TestMinerCreate(t *testing.T) {
```
phritz: blocking: this test doesn't give us much assurance that the call is working. at the very least you could use the fake objects above to assert that things are hooked up properly eg wait receives the cid returned by send and that invoking the wait callback with no a zero exit code results in no error, as well as a test that zero exit code in the receipt results in an error. you could also check the side effects. i think we need to improve the success case and have at least one error.

plumbing/api.go
```diff
@@ -196,6 +196,16 @@ func (api *API) WalletFind(address address.Address) (wallet.Backend, error) {
 	return api.wallet.Find(address)
 }
 
+// WalletGetPubKeyForAddress returns the public key for a given address
+func (api *API) WalletGetPubKeyForAddress(addr address.Address) ([]byte, error) {
+	return api.wallet.GetPubKeyForAddress(addr)
+}
+
+// WalletHasAddress returns true if the wallet has a given address and false otherwise
+func (api *API) WalletHasAddress(addr address.Address) bool {
```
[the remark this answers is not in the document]
acruikshank: This is a tough call. `Wallet` backs the plumbing API, but it also provides an interface for core services. If `HasAddress` on `Wallet` is justified by needs in core services, deciding to only expose the minimal interface in plumbing means code duplication in the porcelain API. On the other hand, exposing `HasAddress` in plumbing adds an constraint on `Wallet` that it maintain `HasAddress` even if it's no longer needed in core.

Both sides are problematic, but neither are problematic enough for me to care. Incidentally, I don't think `HasAddress` is useful in plumbing for other reasons. I'm maybe mildly in favor of removing redundancies in the plumbing API.

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+	ctx = log.Start(ctx, "Node.CreateMiner")
+	defer func() {
+		log.FinishWithErr(ctx, err)
+	}()
+
+	addrInterface, err := plumbing.ConfigGet("mining.minerAddress")
+	if err != nil {
+		return nil, err
+	}
+	addr, success := addrInterface.(address.Address)
```
acruikshank: This is unnecessary. You can compare addrInterface to address.Address.

porcelain/miner.go
```diff
@@ -12,13 +12,146 @@ import (
...
+	if !success {
+		return nil, fmt.Errorf("failed to read miner address")
+	}
+	if (addr != address.Address{}) {
+		return nil, fmt.Errorf("can only have one miner per node")
+	}
+
+	pubKey := []byte{}
+	if plumbing.WalletHasAddress(addr) {
```
acruikshank: BLOCKING: This should be the minerOwner address and it should be an error if the public key doesn't exist.

porcelain/miner.go
acruikshank: `mining.blockSignerAddress` should no longer go in config. This method should reduce to saving the minerAddress in config, and probably doesn't need to be a separate method.

plumbing/api.go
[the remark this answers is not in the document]
phritz: sgtm

phritz:
@rosalinekarr is this ready for another look?

phritz:
@rosalinekarr the best way to get me to take another look is to request re review (the little circle thingy by my name in teh reviewer list once i have already added one review). i often only look at reviews newly assigned to me so i might miss it unless you do this...

porcelain/miner_test.go
```diff
@@ -22,6 +22,102 @@ import (
...
+			plumbing,
+			address.Address{},
+			types.NewGasPrice(0),
+			types.NewGasUnits(100),
+			1,
+			"",
+			collateral,
+		)
+		require.NoError(err)
+		assert.Equal(addr, &address.Address{})
```
phritz: should return some address and assert it here, eg address.NewForTestGetter()()

PR #2114, opened 2019-02-26
Pay miner owner instead of miner

### LINKED ISSUE #1844: Miner address is recipient of block reward, should be owner.
### Description
cc @phritz @shannonwells please verify that this is right and I'm not going crazy.

The intent behind having a separate `ownerAddr` and `minerAddr` was to keep two different accounts for block rewards (`ownerAddr`) and collateral (`minerAddr`).  Currently `minerAddr` is getting both collateral and block reward (see [block reward payout ](https://github.com/filecoin-project/go-filecoin/blob/master/consensus/processor.go#L446) [coming from block.Miner](https://github.com/filecoin-project/go-filecoin/blob/master/consensus/processor.go#L105), [which is set during mining with worker.minerAddr](https://github.com/filecoin-project/go-filecoin/blob/master/mining/block_generate.go#L73), [which is set to the miner, not the miner owner](https://github.com/filecoin-project/go-filecoin/blob/master/node/node.go#L776).

### Acceptance criteria
Owner gets paid not miner actor.

### Risks + pitfalls
This could mess up some tests if they've been set up with faulty assumptions and now rely on the miner address getting the funds.

### Where to begin
Very simple change in node to pass in the miner owner when constructing the default worker.

acruikshank:
closes #1844

### Problem

For some time now, we have been paying block rewards and gas payments to the miner actor. These payments should be going to the account actor listed as the miner actor's owner.

### Solution

This PR switches the address to which the processor directs payments to that of the miner's owner.

The major complication with this is that blocks contain only the miner's address. The miner owner must be queried from the miner actor. This has the additional complication that all test code that attempts to validate a block must have an actor with an owner in state for the miner address specified in the block. Most of the changes in this PR are related to that.

### DIFF
--- a/chain/default_syncer_test.go
+++ b/chain/default_syncer_test.go
@@ -8,6 +8,7 @@ import (
 	"gx/ipfs/QmNf3wujpV2Y7Lnj2hy2UrmuX8bhMDStRHbnSLh7Ypf36h/go-hamt-ipld"
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
 	bstore "gx/ipfs/QmRu7tiRnFk9mMPpVECQTBQJqXtmG132jJxA1w9A7TtpBz/go-ipfs-blockstore"
+	"gx/ipfs/QmTu65MVbemtUxJEWgsTtzv9Zv9P8rvmqNA4eG9TrTRGYc/go-libp2p-peer"
 
 	"github.com/filecoin-project/go-filecoin/actor/builtin"
 	"github.com/filecoin-project/go-filecoin/address"
@@ -43,16 +44,22 @@ var (
 	genTS, link1, link2, link3, link4 types.TipSet
 
 	// utils
-	cidGetter    func() cid.Cid
-	minerAddress address.Address
+	cidGetter         func() cid.Cid
+	minerAddress      address.Address
+	minerOwnerAddress address.Address
+	minerPeerID       peer.ID
 )
 
 func init() {
+	minerAddress = address.MakeTestAddress("miner")
+	minerOwnerAddress = address.MakeTestAddress("minerOwner")
+	minerPeerID = testhelpers.RequireRandomPeerID()
+
 	// Set up the test chain
 	bs := bstore.NewBlockstore(repo.NewInMemoryRepo().Datastore())
 	cst := hamt.NewCborStore()
 	var err error
-	genesis, err = consensus.InitGenesis(cst, bs)
+	genesis, err = initGenesis(cst, bs)
 	if err != nil {
 		panic(err)
 	}
@@ -63,7 +70,6 @@ func init() {
 	cidGetter = types.NewCidForTestGetter()
 
 	genStateRoot = genesis.StateRoot
-	minerAddress = address.MakeTestAddress("miner")
 }
 
 // This function sets global variables according to the tests needs.  The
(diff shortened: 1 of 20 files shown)


shannonwells:
This is related to #1054  and some of the groundwork for solving the test code issues is being done [here in this WIP PR](https://github.com/filecoin-project/go-filecoin/pull/2115)

Part 1 / 3 was [here](https://github.com/filecoin-project/go-filecoin/pull/1795)

consensus/processor.go
```diff
@@ -504,26 +516,26 @@ type ApplyMessagesResponse struct {
...
 // Precondition: signatures of messages are checked by the caller.
-func (p *DefaultProcessor) ApplyMessagesAndPayRewards(ctx context.Context, st state.Tree, vms vm.StorageMap, messages []*types.SignedMessage, minerAddr address.Address, bh *types.BlockHeight, ancestors []types.TipSet) (ApplyMessagesResponse, error) {
+func (p *DefaultProcessor) ApplyMessagesAndPayRewards(ctx context.Context, st state.Tree, vms vm.StorageMap, messages []*types.SignedMessage, minerOwnerAddr address.Address, bh *types.BlockHeight, ancestors []types.TipSet) (ApplyMessagesResponse, error) {
 	var emptyRet ApplyMessagesResponse
 	var ret ApplyMessagesResponse
 
-	// transfer block reward to miner from network address.
-	if err := p.blockRewarder.BlockReward(ctx, st, minerAddr); err != nil {
+	// transfer block reward to miner's owner from network address.
+	if err := p.blockRewarder.BlockReward(ctx, st, minerOwnerAddr); err != nil {
```
phritz: i wish we had had a policy of only returning fault errors at the level of apply* so that it was easier to tell by inspection if fault errors are being returned. as it is we have to go look in each implementation we call and verify it. no action required, just musing.

consensus/processor_test.go
```diff
@@ -1074,3 +1099,11 @@ func mustSetup2Actors(t *testing.T, balance1 *types.AttoFIL, balance2 *types.Att
 	})
 	return addr1, act1, addr2, act2, st, mockSigner
 }
+
+func mustCreateMinerWithOwner(ctx context.Context, require *require.Assertions, st state.Tree, vms vm.StorageMap, minerAddr, minerOwner address.Address) (cid.Cid, *actor.Actor) {
```
phritz: name of this function implies to me that it creates the owner as well. looking at some of the tests that use it, seems like they might expect that it does

testhelpers/consensus.go
```diff
@@ -77,13 +79,10 @@ func (tv *TestPowerTableView) HasPower(ctx context.Context, st state.Tree, bstor
 
 // NewValidTestBlockFromTipSet creates a block for when proofs & power table don't need
 // to be correct
-func NewValidTestBlockFromTipSet(baseTipSet types.TipSet, height uint64, minerAddr address.Address) *types.Block {
+func NewValidTestBlockFromTipSet(baseTipSet types.TipSet, stateRoot cid.Cid, height uint64, minerAddr address.Address) *types.Block {
```
phritz: staterootcid

commands/payment_channel_daemon_test.go
```diff
@@ -68,7 +68,7 @@ func TestPaymentChannelLs(t *testing.T) {
 	t.Run("Works with default payer", func(t *testing.T) {
 		t.Parallel()
 
-		payer, err := address.NewFromString(fixtures.TestAddresses[0])
+		payer, err := address.NewFromString(fixtures.TestAddresses[2])
```
phritz: curious why the renumbering is required here and elsewhere?

chain/default_syncer_test.go
```diff
@@ -43,16 +44,22 @@ var (
 	genTS, link1, link2, link3, link4 types.TipSet
 
 	// utils
-	cidGetter    func() cid.Cid
-	minerAddress address.Address
+	cidGetter         func() cid.Cid
```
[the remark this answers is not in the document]
acruikshank: Agree. Already created #2116 to keep this PR shorter.

commands/payment_channel_daemon_test.go
ZenGround0: I have the same question, can't find any other place where [0] gets used

acruikshank: TestAddress[0] is the owner of TestMiner[0], so it now gets paid for mining. Most of these tests are concerned with tracking the payer account's balance. We could work around this in the tests, but it's easier just to remove it as a complication.

PR #2115, opened 2019-02-26
Feat/signed tickets part 2

shannonwells:
# Description
**PART 2 of 3 for #1054** 
Generates the ticket as a cryptographic signature of the hashed Proof. _Ticket validation will be temporarily disabled_ until Part 3, because changing this generation will break a bunch of other things anyway, although not the protocol.

### Edit: Wallet does not has CreateTicket.
signing logic remains in Wallet and CreateTicket stays in consensus/expected.

### Use an interface for a signer.  
[Edit] we have a consensus.TicketSigner interface, which MockSigner now implements.

### Increased *_test packages:
* wallet tests moved to wallet_test package.
* daemon tests moved to commands_test package, since daemon tests by definition are black box, and do not need to be part of the commands package.
* node test moved to node_test, except for a single test which tests a private function. It stays in node package, but it's in a file all by itself, called node_whitebox_test.go
* at this point it's questionable whether we even need a MockSigner; probably not for everything. So I have filed issue #2252.
* @acruikshank has concerns about how default_syncer_test overwrites FakeChildParams and wants a helper function instead.  This PR is already too long, so that is going into a follow up. See #2253
Please see inline comments for additional explanation.


node/node_test.go
```diff
@@ -198,102 +197,14 @@ func TestNodeStartMining(t *testing.T) {
 
 }
 
-// skipped anyway, now commented out.  With new mining we really need something here though.
```
shannonwells: killing this. It's been sitting here for months.

node/node_test.go
```diff
@@ -340,79 +251,50 @@ func TestOptionWithError(t *testing.T) {
...
 
 	opts = append(opts, errOpt)
 
-	_, err = New(ctx, opts...)
+	_, err = node.New(ctx, opts...)
 	assert.Error(err, scaryErr)
 
 }
 
-func TestMakePrivateKey(t *testing.T) {
```
shannonwells: moved to note_whitebox_test.go

consensus/expected.go
```diff
@@ -360,19 +361,6 @@ func CreateChallengeSeed(parents types.TipSet, nullBlkCount uint64) (proofs.PoSt
 	return h, nil
 }
 
-// CreateTicket computes a valid ticket using the supplied proof
```
shannonwells: moved to mining package

commands/commands_test.go
```diff
@@ -23,3 +25,13 @@ func requireSchemaConformance(t *testing.T, jsonBytes []byte, schemaName string)
 
 	require.Truef(t, result.Valid(), "Error schema validating: %s", string(jsonBytes))
 }
+
+func makeTestDaemonWithMinerAndStart(t *testing.T) *th.TestDaemon {
+	daemon := th.NewDaemon(
```
shannonwells: We were doing this in a bunch of places and most of the time it was fine to just use the first of everything (TestAddresses, TestMiners, KeyFilePaths), so I abstracted it.

types/testing.go
```diff
@@ -63,6 +63,14 @@ func NewMockSigner(kis []KeyInfo) MockSigner {
 	return ms
 }
 
+// NewMockSignersAndKeyInfo is a convenience function to generate a given number of (mock)
```
shannonwells: Was also seeing this setup all over the place, and it just made sense to me to abstract it.  Sometimes we use the resulting KeyInfos and sometimes we don't.

node/node_test.go
ZenGround0: 🎃 🎄 🎆

consensus/expected.go
```diff
@@ -265,18 +267,12 @@ func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ance
 // validateMining checks validity of the block ticket, proof, and miner address.
 //    Returns an error if:
 //    	* any tipset's block was mined by an invalid miner address.
-//      * the block ticket is incorrectly computed
```
ZenGround0: I think this condition is separate from the one on line 270.  I would maybe phrase it as "* the block ticket is not a valid signature of the proof under the miner's signing key"

consensus/expected_test.go
```diff
@@ -222,12 +233,6 @@ func TestIsWinningTicket(t *testing.T) {
...
 
-	minerAddress := address.NewForTestGetter()()
-	ctx := context.Background()
-	d := datastore.NewMapDatastore()
-	bs := blockstore.NewBlockstore(d)
-	var st state.Tree
-
 	t.Run("IsWinningTicket returns false + error when we fail to get total power", func(t *testing.T) {
 		ptv1 := NewFailingTestPowerTableView(testCase.myPower, testCase.totalPower)
 		ticket := [sha256.Size]byte{}
```
ZenGround0: I believe that this is a different size (32) than the signature size mentioned above (64). If this is correct is it weird that IsWinningTicket doesn't throw an error?

mining/worker.go
```diff
@@ -253,6 +255,22 @@ func createProof(challengeSeed proofs.PoStChallengeSeed, createPoST DoSomeWorkFu
...
+//    returns:  []byte -- the ticket.
+func CreateTicket(proof proofs.PoStProof, signerAddr address.Address, signer types.Signer) []byte {
+	buf := append(proof[:], signerAddr.Bytes()...)
+	h := blake2b.Sum256(buf)
+	//return h[:]
+
+	ticket, err := signer.SignBytes(h[:], signerAddr)
+	if err != nil {
+		errMsg := fmt.Sprintf("SignBytes error in CreateTicket: %s", err.Error())
+		panic(errMsg)
```
ZenGround0: Seems panic worthy to me.  Would be good to check with our local panic enforcer (@phritz) who has many times told me to stop panicing when at the time it was always similarly obvious to me.

commands/message_daemon_test.go
```diff
@@ -59,6 +63,7 @@ func TestMessageWait(t *testing.T) {
 	d := th.NewDaemon(
 		t,
 		th.WithMiner(fixtures.TestMiners[0]),
+		th.DefaultAddress(fixtures.TestAddresses[0]),
```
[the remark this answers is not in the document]
shannonwells: good catch, missed one!

shannonwells:
@ZenGround0 @acruikshank @phritz can you please take a second look?

acruikshank:
> @ZenGround0 @acruikshank @phritz can you please take a second look?

@shannonwells This PR is still marked as [WIP]. It's rather long, so it's not a good use of time to have 3 people review unfinished work. If you feel it's not WIP, I'll be happy to go through it in detail.

commands/miner_daemon_test.go
```diff
@@ -190,7 +190,7 @@ func TestMinerCreate(t *testing.T) {
 		var addr address.Address
 
 		tf := func(fromAddress address.Address, pid peer.ID) {
-			d1 := th.NewDaemon(t, th.WithMiner(fixtures.TestMiners[0]), th.KeyFile(fixtures.KeyFilePaths()[2])).Start()
+			d1 := makeTestDaemonWithMinerAndStart(t)
```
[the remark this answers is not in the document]
shannonwells: It seems so

shannonwells:
@acruikshank I just forgot to take the WIP off

config/config.go
```diff
@@ -101,7 +101,6 @@ func newDefaultBootstrapConfig() *BootstrapConfig {
 // MiningConfig holds all configuration options related to mining.
 type MiningConfig struct {
 	MinerAddress            address.Address `json:"minerAddress"`
```
shannonwells: It was decided that blockSignerAddress is not only redundant (use the miner owner public key instead which is already stored), but does not belong in config.

config/config_test.go
```diff
@@ -65,7 +65,6 @@ func TestWriteFile(t *testing.T) {
 	},
 	"mining": {
 		"minerAddress": "",
-		"blockSignerAddress": "",
```
shannonwells: kill blockSignerAddress (see above)

consensus/expected.go
shannonwells: block ticket checking is disabled for this PR anyway, unless you mean something else?

consensus/expected.go
```diff
@@ -265,18 +267,12 @@ func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ance
...
+//      * the block proof is invalid for the challenge
 //      * the block ticket fails the power check, i.e. is not a winning ticket
 //    Returns nil if all the above checks pass.
 // See https://github.com/filecoin-project/specs/blob/master/mining.md#chain-validation
 func (c *Expected) validateMining(ctx context.Context, st state.Tree, ts types.TipSet, parentTs types.TipSet) error {
 	for _, blk := range ts.ToSlice() {
-		computedTicket := CreateTicket(blk.Proof, blk.Miner)
```
shannonwells: validating mining will no longer compute a ticket since tickets have to be signed, and it can't (and shouldn't) be signed in verification.

consensus/expected_test.go
shannonwells: It doesn't, because it just copies the slice bytes into the Signature; however, it's kinda sloppy to make it a `[sha256.Size]byte{}`, so I'll at least declare the ticket as a types.Signature.

node/node_test.go
```diff
@@ -430,27 +312,11 @@ func TestNodeConfig(t *testing.T) {
 	}, cfg.Swarm)
 }
 
-func TestNode_getMinerOwnerPubKey(t *testing.T) {
-	seed := MakeChainSeed(t, TestGenCfg)
```
shannonwells: this function was delete

node/node.go
```diff
@@ -808,8 +807,17 @@ func (node *Node) StartMining(ctx context.Context) error {
 			return chain.GetRecentAncestors(ctx, ts, node.ChainReader, newBlockHeight, consensus.AncestorRoundsNeeded, consensus.LookBackParameter)
 		}
 		processor := consensus.NewDefaultProcessor()
+
+		queryer := msg.NewQueryer(node.Repo, node.Wallet, node.ChainReader, node.cborStore, node.Blockstore)
```
rosalinekarr: Replace with `porcelain.MinerGetKey`?

shannonwells: @rosalinekarr done, can you pls verify?

node/node.go
```diff
@@ -1010,14 +1018,16 @@ func (node *Node) CreateMiner(ctx context.Context, accountAddr address.Address,
 	defer func() {
 		log.FinishWithErr(ctx, err)
 	}()
-	pubKey, err := node.getMinerActorPubKey()
+
+	pubKey, err := node.Wallet.GetPubKeyForAddress(minerOwnerAddr)
 	if err != nil {
 		return nil, err
 	}
+	fmt.Printf("\n\ncreating a miner for %s with pubkey\n%v\n", minerOwnerAddr.String(), pubKey[:])
```
ZenGround0: Looks like this is leftover.

mining/worker.go
```diff
@@ -214,7 +219,12 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
 			return false
 		}
 		copy(proof[:], prChRead[:])
-		ticket = consensus.CreateTicket(proof, w.minerAddr)
+		signer := w.workerSigner
```
ZenGround0: why not just call `w.workerSignerCreateTicket`?

fixtures/constants.go
```diff
@@ -88,6 +88,7 @@ func init() {
 	for _, miner := range miners {
 		TestMiners = append(TestMiners, miner.Address)
 	}
+
```
ZenGround0: nit: extra newline

types/testing.go
```diff
@@ -65,6 +76,44 @@ func (ms MockSigner) SignBytes(data []byte, addr address.Address) (Signature, er
...
+			if err != nil {
+				return addr, errors.New("could not fetch address")
+			}
+			return addr, nil
+		}
+	}
+	return addr, errors.New("public key not found in wallet")
+}
+
+// CreateTicket is effectively a duplicate of Wallet CreateTicket for testing purposes.
```
ZenGround0: Since this code seems to exercise the actual signing calls why not use a wallet in tests?

wallet/wallet.go
```diff
@@ -106,6 +109,23 @@ func (w *Wallet) SignBytes(data []byte, addr address.Address) (types.Signature,
 	return backend.SignBytes(data, addr)
 }
 
+// AddressFromPubKey looks up a KeyInfo address associated with a given PublicKey
+func (w *Wallet) AddressFromPubKey(pk []byte) (address.Address, error) {
```
ZenGround0: nit: naming should be consistent across GetPubKeyForAddress and AddressFromPubKey.  Both should use or not use get, both should use from or for

wallet/wallet.go
```diff
@@ -152,6 +172,30 @@ func (w *Wallet) NewKeyInfo() (*types.KeyInfo, error) {
 	return w.keyInfoForAddr(newAddr)
 }
 
+// CreateTicket computes a valid ticket.
+// 	params:  proof  []byte, minerAddress address.Address.
+//  returns:  types.Signature ( []byte ), error
+func (w *Wallet) CreateTicket(proof proofs.PoStProof, signerPubKey []byte) (types.Signature, error) {
```
ZenGround0: It seems like this function belongs on an abstraction on top of wallet.  All of its operations depend on public methods of the wallet, and creating a ticket seems far enough removed from a wallet that another abstraction is sensible.

For the record I have a similar opinion about the `SignBytes` method which is well established so I might just be wrong here.

mining/worker.go
```diff
@@ -47,6 +47,11 @@ type Worker interface {
 	Mine(runCtx context.Context, base types.TipSet, nullBlkCount int, outCh chan<- Output) bool
 }
 
+// WorkerSigner is the interface used for signing blocks and tickets.
+type WorkerSigner interface {
+	CreateTicket(proof proofs.PoStProof, signerPubKey []byte) (types.Signature, error)
```
acruikshank: This should probably be SignBytes and FindAddressFromPublicKey or just SignBytesByPublicKey from wallet. See comment in wallet code.

acruikshank: Also the interface should be `TicketSigner` since that's what we're using it for.

types/testing.go
shannonwells: Because we are using a mockSigner in many tests where we don't always have access to, or want to go through the hassle of creating a wallet.  I tried just using Wallet and I wound up with massive issues, import conflicts being the least of it.

wallet/wallet.go
shannonwells: good catch, thanks

chain/default_syncer_test.go
```diff
@@ -137,14 +147,16 @@ func requireSetTestChain(require *require.Assertions, con consensus.Protocol, mo
...
+	fakeChildParams.NullBlockCount = uint64(2)
+	link4blk1 = chain.RequireMkFakeChildWithCon(require, fakeChildParams)
+	link4blk1.Proof, link4blk1.Ticket, err = chain.MakeProofAndWinningTicket(mockSignerPubKey, minerPower, totalPower, mockSigner)
 	require.NoError(err)
 
-	link4blk2 = chain.RequireMkFakeChildWithCon(require,
-		chain.FakeChildParams{Parent: link3, GenesisCid: genCid, StateRoot: link3State, Nonce: uint64(1), NullBlockCount: uint64(2), Consensus: con, MinerAddr: minerAddress})
-	link4blk2.Proof, link4blk2.Ticket, err = chain.MakeProofAndWinningTicket(minerAddress, minerPower, totalPower)
+	fakeChildParams.Nonce = uint64(2)
+	link4blk2 = chain.RequireMkFakeChildWithCon(require, fakeChildParams)
```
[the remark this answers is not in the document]
shannonwells: No, it was a typo.

mining/worker.go
shannonwells: herpty-derp

types/testing.go
shannonwells: @ZenGround0 it turns out with the requested refactor to put CreateTicket back into consensus, the interface implementation means we actually probably _can_ use a wallet.  But this PR is too long so I filed issue #2252 and will follow up with it.

wallet/wallet.go
shannonwells: @ZenGround0 check out the last commit.  A bunch of things were cleaned up so this should look better to you now.

chain/default_syncer_test.go
```diff
@@ -101,20 +109,20 @@ func requireSetTestChain(require *require.Assertions, con consensus.Protocol, mo
...
-		chain.FakeChildParams{Parent: link1, GenesisCid: genCid, StateRoot: link1State, Consensus: con, MinerAddr: minerAddress})
-	link2blk2.Proof, link2blk2.Ticket, err = chain.MakeProofAndWinningTicket(minerAddress, minerPower, totalPower)
+	link2blk2 = chain.RequireMkFakeChildWithCon(require, fakeChildParams)
+	link2blk2.Proof, link2blk2.Ticket, err = chain.MakeProofAndWinningTicket(mockSignerPubKey, minerPower, totalPower, mockSigner)
 	require.NoError(err)
 
-	link2blk3 = chain.RequireMkFakeChildWithCon(require,
-		chain.FakeChildParams{Parent: link1, GenesisCid: genCid, StateRoot: link1State, Nonce: uint64(1), Consensus: con, MinerAddr: minerAddress})
-	link2blk3.Proof, link2blk3.Ticket, err = chain.MakeProofAndWinningTicket(minerAddress, minerPower, totalPower)
+	fakeChildParams.Nonce = uint64(1)
```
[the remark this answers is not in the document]
shannonwells: This will be addressed in #2253

chain/default_syncer_test.go
```diff
@@ -456,28 +478,47 @@ func TestHeavierFork(t *testing.T) {
...
+		StateRoot:   genStateRoot,
+		MinerAddr:   minerAddress,
+		Signer:      signer,
+		MinerPubKey: mockSignerPubKey,
+		Nonce:       uint64(1),
+	}
+
+	forklink1blk1 := chain.RequireMkFakeChild(require, fakeChildParams)
+
+	fakeChildParams.Nonce = uint64(1)
```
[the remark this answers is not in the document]
shannonwells: See #2253

chain/default_syncer_test.go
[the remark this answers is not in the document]
shannonwells: Oversight will be fixed in #2253

PR #2023, opened 2019-02-19
Suppress repeated error log when hearbeat fails to connect.

anorth:
Logs once as a warning immediately on failure, then at error every 10 minutes.

Closes #1906

Issue #1906, opened 2019-02-13
Change heartbeat lost error log to warning and reduce frequency


anorth:
**Describe the bug**
Running a node after following the instructions in the wiki, the node sometimes logs errors sending heartbeats to the metrics server.

As the logs show, it eventually reconnected.

Observing https://stats.kittyhawk.wtf/storage/mining at the same time, it appeared that *all* miner peers lost connection at the same time. Then they came back.

This has been reported anecdotally many times, though many people think it's a permanent error, e.g. of misconfiguration.

**Log**
```21:01:40.835 WARNI    metrics: disconnecting from aggregator, failed to send heartbeat heartbeat.go:134
21:01:45.462 ERROR    metrics: failed to open stream, peerID: <peer.ID Qm*sseEjR>, targetAddr: /dns4/stats-infra.kittyhawk.wtf/tcp/8080 dial attempt failed: <peer.ID Qm*6QjZB2> --> <peer.ID Qm*sseEjR> dial attempt failed: connection refused heartbeat.go:216
21:01:46.279 ERROR    metrics: failed to open stream, peerID: <peer.ID Qm*sseEjR>, targetAddr: /dns4/stats-infra.kittyhawk.wtf/tcp/8080 dial backoff heartbeat.go:216
21:01:57.825 ERROR    metrics: failed to open stream, peerID: <peer.ID Qm*sseEjR>, targetAddr: /dns4/stats-infra.kittyhawk.wtf/tcp/8080 dial attempt failed: <peer.ID Qm*6QjZB2> --> <peer.ID Qm*sseEjR> dial attempt failed: connection refused heartbeat.go:216
...
21:03:46.275 ERROR    metrics: failed to open stream, peerID: <peer.ID Qm*sseEjR>, targetAddr: /dns4/stats-infra.kittyhawk.wtf/tcp/8080 dial backoff heartbeat.go:216
21:04:02.974 WARNI /fil/hello: failed to send hello handshake to peer <peer.ID Qm*sseEjR>: protocol not supported hello.go:150
21:04:05.035  INFO    metrics: successfully to open stream, peerID: <peer.ID Qm*sseEjR>, targetAddr: /dns4/stats-infra.kittyhawk.wtf/tcp/8080 heartbeat.go:219
```

**Desktop (please complete the following information):**
 - MacOS

phritz:
[...]

PR #2117, opened 2019-02-26
heartbeat conn failure is warning

frrist:


anorth:
The current behaviour is intentional. See https://github.com/filecoin-project/go-filecoin/issues/1906 and https://github.com/filecoin-project/go-filecoin/pull/2023

gmasgras:
We have 3 devnets running and only one (usernet) connects to the heartbeat service that feeds the dashboard at `stats.kittyhawk.wtf`. 
This results in the other 2 devnets spamming with non-actionable alerts every 10min.
Can we introduce an ENV that turns this into a warning when specified ?

anorth:
It sounds to me like we need a more general solution to ignoring, or at least siloing, warnings/errors from networks that aren't the devnet.

If the nightly net doesn't connect to `stats`, then it should probably be configured to *not connect*, and hence not fail, rather than downgrade this message.

frrist:
@gmasgras is there a way to avoid passing the environment variable `FIL_HEARTBEAT_ALERTS`  to the test and nightly devnets?

gmasgras:
> @gmasgras is there a way to avoid passing the environment variable `FIL_HEARTBEAT_ALERTS` to the test and nightly devnets?

I think we still want to pass that in, as we would stop getting alerts about nodes that are out of consensus or no blocks mined for $interval.

frrist:
After some discussion with @gmasgras we will close this in favor of a different approach

PR #2126, opened 2019-02-27
Make private key output consistent

### LINKED ISSUE #2004: [bug] wallet export gives different info with different encoding
**Describe the bug**

When `wallet export` is called without any encoding, it gives `Address`, `PrivateKey` (in some format), and `Curve`.
When called with `--enc=json`, it just gives `privateKey` (in a different format), and `curve`.

Feels unusual that the encoding changes what fields are available, and change the format of those fields.

**To Reproduce**

Initialize filecoin, then run:
```
export ADDR=$(go-filecoin wallet addrs ls)
go-filecoin wallet export $ADDR
go-filecoin wallet export $ADDR --enc=json
```
Compare the results.

**Expected behavior**
The two outputs having the same fields and the same values.

**Screenshots**
```
$ go-filecoin version
commit: 271a6812e23e3ccd80642d251e1c1f449459a8b7
```

rkowalick:
Resolves #2004 

The `wallet export` command with `--enc=json` will still only return the private key and the curve used, since this command is designed to write files suitable for consumption by the `wallet import` command, and isn't really designed for human use.

The `wallet export` command with no encoding specified will still return address, private key, and curve, but the private key will have the same output format as the one used for `--enc=json`.

Some examples:

```
$ ./go-filecoin wallet export fcqtu6574lys8nwsc39h4vpa4k0v7vnyj0xzzhsue
Address:	fcqtu6574lys8nwsc39h4vpa4k0v7vnyj0xzzhsue
PrivateKey:	ehfp0DiWZBqlYK4oBVUIBoAVIPJok4e/SVb5RPpL57E=
Curve:		secp256k1
```
and

```
$ ./go-filecoin wallet export fcqtu6574lys8nwsc39h4vpa4k0v7vnyj0xzzhsue --enc=json | json_pp
{
   "KeyInfo" : [
      {
         "privateKey" : "ehfp0DiWZBqlYK4oBVUIBoAVIPJok4e/SVb5RPpL57E=",
         "curve" : "secp256k1"
      }
   ]
}
```

### DIFF
--- a/commands/address.go
+++ b/commands/address.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"encoding/base64"
 	"fmt"
 	"io"
 
@@ -212,7 +213,8 @@ var walletExportCmd = &cmds.Command{
 				if err != nil {
 					return err
 				}
-				_, err = fmt.Fprintf(w, "Address:\t%s\nPrivateKey:\t%x\nCurve:\t\t%s\n\n", a.String(), k.PrivateKey, k.Curve)
+				privateKeyInBase64 := base64.StdEncoding.EncodeToString(k.PrivateKey)
+				_, err = fmt.Fprintf(w, "Address:\t%s\nPrivateKey:\t%s\nCurve:\t\t%s\n\n", a.String(), privateKeyInBase64, k.Curve)
 				if err != nil {
 					return err
 				}
--- a/commands/address_daemon_test.go
+++ b/commands/address_daemon_test.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


commands/address.go
```diff
@@ -212,7 +213,8 @@ var walletExportCmd = &cmds.Command{
 				if err != nil {
 					return err
 				}
-				_, err = fmt.Fprintf(w, "Address:\t%s\nPrivateKey:\t%x\nCurve:\t\t%s\n\n", a.String(), k.PrivateKey, k.Curve)
+				privateKeyInBase64 := base64.StdEncoding.EncodeToString(k.PrivateKey)
+				_, err = fmt.Fprintf(w, "Address:\t%s\nPrivateKey:\t%s\nCurve:\t\t%s\n\n", a.String(), privateKeyInBase64, k.Curve)
```
frrist: Lets add a test here to ensure it matches the json encoding otherwise LGTM

rkowalick: :+1:

PR #2130, opened 2019-02-27
add level and ls command to log

frrist:
This PR add commands:
```
  go-filecoin log level <subsystem> <level> - Change the logging level.
  go-filecoin log ls                        - List the logging subsystems.
```
Prior art here: https://github.com/ipfs/go-ipfs/blob/master/core/commands/log.go

### DIFF
--- a/commands/log.go
+++ b/commands/log.go
@@ -3,19 +3,15 @@ package commands
 import (
 	"fmt"
 	"io"
+	"strings"
 
 	"gx/ipfs/QmQtQrtNioesAWtrx8csBvfY37gTe94d6wQ3VikZUjxD39/go-ipfs-cmds"
 	logging "gx/ipfs/QmbkT7eMTyXfpeyB3ZMxxcxg7XH8t6uXp49jqzz4HB7BGF/go-log"
+	oldlogging "gx/ipfs/QmcaSwFc5RBg8yCq54QURwEU4nwjfCpjbpmaAm4VbdGLKv/go-logging"
 	"gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
 )
 
 var loglogger = logging.Logger("commands/log")
-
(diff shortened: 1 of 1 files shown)


commands/log.go
```diff
@@ -15,7 +27,9 @@ output of a running daemon.
 	},
 
 	Subcommands: map[string]*cmds.Command{
-		"tail": logTailCmd,
+		"tail":  logTailCmd,
+		"level": logLevelCmd,
```
phritz: alpha

commands/log.go
```diff
@@ -32,3 +46,69 @@ Outputs event log messages (not other log messages) as they are generated.
...
+	Helptext: cmdkit.HelpText{
+		Tagline: "Change the logging level.",
+		ShortDescription: `
+Change the verbosity of one or all subsystems log output. This does not affect
+the event log.
+`,
+	},
+
+	Arguments: []cmdkit.Argument{
+		// TODO use a different keyword for 'all' because all can theoretically
```
phritz: my suggestion is remove this todo. if it does clash then someone can fix it. or if you want give the level first and then an optional subsystem second on the command line. if they leave the subsystem off then treat that as all. we definitely definitely should not try to use * on the command line.

commands/log.go
```diff
@@ -32,3 +46,69 @@ Outputs event log messages (not other log messages) as they are generated.
...
+			One of: debug, info, warning, error, critical.
+		`),
+	},
+
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		args := req.Arguments
+		subsystem, level := args[0], args[1]
+
+		if subsystem == logAllKeyword {
+			subsystem = "*"
```
phritz: if per comment at top of file args 0 is an array of all files then how does this assignment work?

commands/log.go
frrist: I like the idea of having subsystem be optional, will go with that

commands/log.go
frrist: yup, this is gone if we make subsystem an option

Issue #2128, opened 2019-02-27
De-tangle syncer and store


ZenGround0:
### Description
From @whyrusleeping in #2067: "that object [the default store] is a tangle. It (and the syncer) appears to be doing the same things multiple times in several different overlapping ways. For example, the logic in the syncers 'maybe get blocks from network' is really weird... we first try to get things from the chain store, which pulls from our local disk. then we try to get from the 'offline' cbor store, which pulls from our local disk, then from an offline exchange, which pulls from the local disk (this should get fixed, theres no need for this offline thing, the blockservice should just learn to deal with a nil exchange), _then_ it fetches from the online cborstore, which checks the local disk, then asks (everyone) over bitswap.

### Acceptance criteria
Clean up this mess

anorth:
Another effect of the tangling is that it's hard to insert validation logic for blocks received over the network, without redundantly validating blocks already in the store.

ZenGround0:
Using this to track technical debt in syncer tests currently.  Setup is too laborious and needs rethinking.  The shared dependency of fetcher and store on global IPLD store seems like it can't be helped but is a headache for testing and at least needs to be done better than right now.  In the current state the test fetcher is not backed by a properly exposed or interfaced IPLD store so it can't be shared with the chain store.

anorth:
As part of this we should probably be able to get rid of the blockservice that's backed only by an "offline" exchange, replacing use of that code with a thin wrapper.

anorth:
We have cleaned up a few things:
- there's now just one IPLD store for chain data, instead of two
- tests are well on their way to being usable
- the chain/state store presents a more helpful interface

The tangling is still real and confusing though. The chain store interface does not expose blocks, only the tipsets for which the syncer has computed and provided the state root. When the syncer invokes the fetcher, the fetcher directly manipulates the IPLD store underlying the chain store, but completely opaque to the syncer: it can't even inspect it to check. The fetcher also returns tipsets in an arbitrarily-large slice, which the syncer uses to compute state but then discards. 

The syncer will never (and cannot) inspect the blocks the fetcher wrote to the store, but there's an assumption that they have been persisted such that `store.Load` will be able to rebuild the chain from whatever `setHead()` the syncer informs it of (the store doesn't check that it actually has those blocks).

[...]

Issue #2116, opened 2019-02-26
Remove global variables from default_syncer_test and default_store_test


acruikshank:
### Description

The default_syncer_test relies on several variables defined in package scope. These are statically initialized. This has the potential to cause erratic test behavior and is hard to read.

We should move all these variables into a struct that can be instantiated with an initializer. Tests that use these parameters will instantiate a struct instead. Should be a mechanical change.

### Acceptance criteria

There are no package scoped variables in default_syncer test or default_store_test.

Issue #2065, opened 2019-02-22
Implement ranged BlockSync for faster block fetching in chain sync


whyrusleeping:
### Description
The spec contains a "BlockSync" protocol for transferring large ranges of blockchain blocks from peers during chain sync. This is expected to reduce the time required for fetching large chains.

Should partially address #2050. 

### Acceptance criteria
- Nodes can fetch ranges of tipsets from peers when fetching/downloading the chain
- A comment on this issue quantifies the improvement in chain downloading time over the prior approach

### Risks + pitfalls
- Chain sync time appears to be dominated by validation rather than downloading, so don't do this work unless it will deliver a meaningful improvement to the whole
- This BlockSync is still an interim protocol, before a desired switch to GraphSync when it is ready for use. It may be lost effort in the long run (but might also be really worthwhile for its short-term benefits).

### Where to begin
Spec PR here: https://github.com/filecoin-project/specs/pull/151

anorth:
I have updated this issue title and description with a bit more info about the scope. This more efficient fetching is just one part of the whole chain sync story.

@ZenGround0 I don't believe you're actively working on it, so unassigning you now.

Issue #1883, opened 2019-02-12
[bug] balance changed to zero weirdly


whyrusleeping:
I got some funds from the faucet, then ran balance on my address. It showed 1000, and then ran it again and it showed zero. No idea what happened.

```
why@why-MS-7B46 ~/g/s/g/f/go-filecoin>
go-filecoin wallet balance fcq0egt5t0th7qc6yhh0aavlxz3u6vhflmn94efr
1000
why@why-MS-7B46 ~/g/s/g/f/go-filecoin>
go-filecoin wallet balance fcq0egt5t0th7qc6yhh0aavlxz3u6vhflmn94efr
0
```

rosalinekarr:
Renaming to "Add logging for chain reorg and extending" and updating labels.

@phritz, can you add some acceptance criteria?

phritz:
we got re org logging i think, either @rkowalick or @ZenGround0 implemented i think

rkowalick:
I don't think this ever landed. Will sync up with @ZenGround0 about it today.

ZenGround0:
We're logging which is something.  See #2127 for thoughts on how to make this better long term.

PR #2131, opened 2019-02-27
Feat/log reorgs

ZenGround0:
In response to #1883, leftover from launch. 

@rkowalick and I verified that this change causes reorg warnings in the log by manually triggering a reorg on a local node by disconnecting and reconnecting to the user net.  

FYI this change is quick and dirty, exposing a function on the defaultSyncer that probably belongs as a private method in order to make testing from outside the chain package possible.  As this method really should be private the tests effectively test a private abstraction which I understand is ugly and bad too.

My understanding is that this is good enough and that good alternatives to make this cleaner (like refactoring the chain syncer + tests OR figuring out a way to make assertions about log output, OR writing up boiler plate to allow test daemons to mine on forks and then asserting on logs that way) are lower priority than other things on my plate.  Furthermore [issues](https://github.com/filecoin-project/go-filecoin/issues/2065) [are](https://github.com/filecoin-project/go-filecoin/issues/2128) [open](https://github.com/filecoin-project/go-filecoin/issues/2116) that cover cleanup.

If this is unacceptably dirty I will slow down and focus efforts here but I think that's more time than this issue deserves.

### DIFF
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -66,7 +66,7 @@ type DefaultSyncer struct {
 var _ Syncer = (*DefaultSyncer)(nil)
 
 // NewDefaultSyncer constructs a DefaultSyncer ready for use.
-func NewDefaultSyncer(online, offline *hamt.CborIpldStore, c consensus.Protocol, s Store) Syncer {
+func NewDefaultSyncer(online, offline *hamt.CborIpldStore, c consensus.Protocol, s Store) *DefaultSyncer {
 	return &DefaultSyncer{
 		cstOnline:  online,
 		cstOffline: offline,
@@ -255,6 +255,16 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 	}
 
 	if heavier {
+		// Gather the entire new chain for reorg comparison.
+		// See Issue #2151 for making this scalable.
+		newChain, err := CollectTipSetsOfHeightAtLeast(ctx, syncer.chainStore.BlockHistory(ctx, parent), types.NewBlockHeight(uint64(0)))
+		if err != nil {
+			return err
+		}
+		newChain = append(newChain, next)
+		if IsReorg(syncer.chainStore.Head(), newChain) {
+			logSyncer.Infof("reorg occurring while switching from %s to %s", syncer.chainStore.Head().String(), next.String())
+		}
 		if err = syncer.chainStore.SetHead(ctx, next); err != nil {
 			return err
 		}
--- a/chain/default_syncer_test.go
+++ b/chain/default_syncer_test.go
@@ -157,7 +157,7 @@ func requireSetTestChain(require *require.Assertions, con consensus.Protocol, mo
 }
 
 // loadSyncerFromRepo creates a store and syncer from an existing repo.
-func loadSyncerFromRepo(require *require.Assertions, r repo.Repo) (chain.Syncer, *hamt.CborIpldStore) {
+func loadSyncerFromRepo(require *require.Assertions, r repo.Repo) (*chain.DefaultSyncer, *hamt.CborIpldStore) {
 	powerTable := &testhelpers.TestView{}
 	bs := bstore.NewBlockstore(r.Datastore())
 	cst := hamt.NewCborStore()
@@ -172,7 +172,7 @@ func loadSyncerFromRepo(require *require.Assertions, r repo.Repo) (chain.Syncer,
 
 // initSyncTestDefault creates and returns the datastructures (chain store, syncer, etc)
 // needed to run tests.  It also sets the global test variables appropriately.
-func initSyncTestDefault(require *require.Assertions) (chain.Syncer, chain.Store, *hamt.CborIpldStore, repo.Repo) {
+func initSyncTestDefault(require *require.Assertions) (*chain.DefaultSyncer, chain.Store, *hamt.CborIpldStore, repo.Repo) {
 	processor := testhelpers.NewTestProcessor()
 	powerTable := &testhelpers.TestView{}
 	r := repo.NewInMemoryRepo()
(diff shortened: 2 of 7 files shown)


chain/default_syncer.go
```diff
@@ -367,3 +370,21 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 	}
 	return nil
 }
+
+// IsReorg determines if a new tipset would cause a reorg if heavier than the
```
phritz: any tipset would cause a reorg if heavier than the syncers head. i think you mean if would cause a re org?

chain/default_syncer.go
```diff
@@ -367,3 +370,21 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
...
+// syncer's head.  The new tipset will cause a reorg if the syncer's head is
+// not one of its ancestors or a subset of the new tipset.
+func (syncer *DefaultSyncer) IsReorg(newAncestors []types.TipSet, new types.TipSet) bool {
```
phritz: this could just be a free function which would not pollute the syncer interface

chain/default_syncer.go
```diff
@@ -255,6 +255,9 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 	}
 
 	if heavier {
+		if syncer.IsReorg(ancestors, next) {
```
phritz: observation: this logic could be implemented as a thingy that consumes the new heaviest tipset channel, assuming it allows multiple consumers. look for ways to implement things outside of major abstractions.

phritz: this call assumes the ancestors list is sufficiently long to reach the current head

chain/default_syncer_test.go
```diff
@@ -922,3 +922,63 @@ func initGenesis(cst *hamt.CborIpldStore, bs bstore.Blockstore) (*types.Block, e
...
+
+func TestIsReorg(t *testing.T) {
+	ctx := context.Background()
+	require := require.New(t)
+	assert := assert.New(t)
+	t.Run("if chain is a fork of another chain, IsReorg is true", func(t *testing.T) {
+		syncer, chainStore, cst, _ := initSyncTestDefault(require)
+		requireGrowChain(ctx, require, cst, chainStore, 10)
+		fork1Head := chainStore.Head()
+		fork1Ancestors, err := chain.CollectTipSetsOfHeightAtLeast(ctx, chainStore.BlockHistory(ctx, chainStore.Head()), types.NewBlockHeight(uint64(0)))
```
phritz: oh, look! nice.

chain/default_syncer.go
phritz: the tests for this would be so much simpler if the caller passed in what the function needed, instead forcing the function to do it. i see this pattern a lot, it makes for lots of coupling and makes functions hard to understand. this thing is much simpler if it is a free function that is passed two things: 1 a slice of tipsets representing the new chain including the new head and 2 the current head. think about how much cleaner the unit tests can be, no syncer required at all, just tipsets.

this is the same pattern i saw with the lookback stuff: making functions hard to understand so the caller can be cleaner. it should be the other way around.

chain/default_syncer_test.go
```diff
@@ -922,3 +922,63 @@ func initGenesis(cst *hamt.CborIpldStore, bs bstore.Blockstore) (*types.Block, e
 		consensus.MinerActor(minerAddress, minerOwnerAddress, []byte{}, 1000, minerPeerID, types.ZeroAttoFIL),
 	)(cst, bs)
 }
+
+func TestIsReorg(t *testing.T) {
```
phritz: whenever you see unit tests this complicated for a function that checks to see if a thing is in a list, something is wrong. you can use that as a signal to look for other ways, eg make the function simpler with fewer dependencies.

chain/default_syncer.go
ZenGround0: This is not true if the new tipset is a child of the syncer's head.

ZenGround0:
@phritz mind giving this another look?  I know you gave me the green light but I want to see if any of your comments are unaddressed by this refactoring.  Thanks!

chain/default_syncer.go
```diff
@@ -255,6 +255,16 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
 	if heavier {
+		// Gather the entire new chain for reorg comparison.
+		// See Issue #2151 for making this scalable.
+		newChain, err := CollectTipSetsOfHeightAtLeast(ctx, syncer.chainStore.BlockHistory(ctx, parent), types.NewBlockHeight(uint64(0)))
+		if err != nil {
+			return err
+		}
+		newChain = append(newChain, next)
+		if IsReorg(syncer.chainStore.Head(), newChain) {
+			logSyncer.Warningf("reorg occurring while switching from %s to %s", syncer.chainStore.Head().String(), next.String())
```
phritz: maybe info or debug? re orgs are wai.

chain/reorg.go
```diff
@@ -0,0 +1,24 @@
+package chain
+
+import (
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// IsReorg determines if a choosing the end of the newChain as the new head
```
phritz: extra a in there

chain/reorg.go
```diff
@@ -0,0 +1,24 @@
...
+// would cause a "reorg" given the current head is at curHead.
+// syncer's head.  A reorg occurs when curHead is not a member of newChain
```
phritz: think this is an incomplete sentence

chain/reorg.go
```diff
@@ -0,0 +1,24 @@
...
+// AND curHead is not subset of newChain's head.
+func IsReorg(curHead types.TipSet, newChain []types.TipSet) bool {
```
phritz: omg so much nicer thank you!

PR #2135, opened 2019-02-27
Bug/dont connect heartbeat target when empty

gmasgras:
Based on discussion in https://github.com/filecoin-project/go-filecoin/pull/2117

Check that a non-empty HeartBeat.BeatTarget was set before starting a HeartbeatService


gmasgras:
> LGTM. Can you also update the README/Wiki stating that users will need to restart their nodes after setting the heartbeat target?

Updated https://github.com/filecoin-project/go-filecoin/wiki/Getting-Started#start-streaming-activity-from-your-node

Issue #1934, opened 2019-02-14
Messages should not send with insufficient funds


whyrusleeping:
Otherwise the message gets sent out, sits in my mempool forever, and never gets mined.

rosalinekarr:
@whyrusleeping, while I realize this is something the project needs to do, it seems more like an incomplete feature than a bug. I'm updating labels and title accordingly.

@acruikshank, can you add some story detail when you have the time?

acruikshank:
Yeah this will be part of the work around message nonces.

acruikshank:
## Description

There are many reasons why a message can fail with a permanent error. Permanent errors means the message never gets mined, and the miner never includes a message receipt for it in the block. 

The result for the sender is that the message sits in their message pool forever, and every subsequent message gets a higher nonce which prevents them from ever being processes (this behavior will be addressed in future work). 

Validating message before publishing them to the network and adding them to the message pool will greatly reduce the possibility of this bad situation.

## Acceptance Criteria

Messages with any one of any of the following criteria error out with an appropriate error message. They do not go on the network, and they do not land in the message pool.

1. Invalid signature.
2. Send to self.
3. Insufficient funds (insufficient gas)
4. Non-account actor
5. Attempt to transfer negative value
6. Gas limit above block limit

### where to start

* https://github.com/filecoin-project/go-filecoin/blob/e4bb60f8e75b7b61e30167ba01bd1d13430937bf/plumbing/msg/sender.go#L75
* https://github.com/filecoin-project/go-filecoin/blob/e4bb60f8e75b7b61e30167ba01bd1d13430937bf/consensus/processor.go#L558

anorth:
Some of the cases mentioned are "temporary" errors, where another message could change the state such that this message could then succeed. Also, some failures could be introduced only by prior, but unmined, messages.

I'm clarifying that it's out of scope for this issue to attempt to play forward the state transitions implied by other pooled messages to detect things that could succeed or fail only afterwards. In the case of failure that could succeed, the caller is instead going to have to wait for the earlier messages to be mined. In the case of success that turns to failure, they'll have to wait for the failure to time out ([design in progress](https://docs.google.com/document/d/1n224bJItvu44hd-WVd69gfNbzC-5XUu-HpOfUCcpgEc/edit#)). We may revisit this to optimize that UX later, but this simpler validation will catch most errors.

anorth:
Two remaining things to validate:
- Attempt to transfer negative value
[...]

PR #2136, opened 2019-02-28
Extract validator from processor.go to a new validation.go.

anorth:
I updated comments a bit, but there's no semantic change. This unfortunately exposes that there are no unit tests for this code directly (it's exercised
indirectly through processor tests).

This is prework for #1934.


shannonwells:
You don't want to add unit tests here?

anorth:
@shannonwells I'll get to it one day but trying to limit the time spent on collateral improvements while working towards my goal (message pre-validation).

PR #2138, opened 2019-02-28
Refactor next nonce calculation to decouple message sender from unrelated testing concerns.

anorth:
They key bits here are removing the empty message pool from MustGetNonce and inlining the
tree.GetActor() call, which then enables moving all the real nonce functionality to
sender, its only non-test call site.

This is prework for #1934, which will also inspect the actor's balance.


core/testing.go
```diff
@@ -24,8 +25,15 @@ import (
 
 // MustGetNonce returns the next nonce for an actor at the given address or panics.
```
phritz: this comment should probably be updated. as stated i would expect panic if the actor at the given address is not found

plumbing/msg/sender.go
```diff
@@ -56,7 +58,12 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
 	s.l.Lock()
 	defer s.l.Unlock()
 
-	nonce, err := nextNonce(ctx, s.chainReader, s.msgPool, from)
+	st, err := s.chainReader.LatestState(ctx)
+	if err != nil {
+		return cid.Undef, err
```
phritz: we wrap pretty liberally to give more context no errors. up to you.

core/testing.go
anorth: I changed it to panic instead 😄

PR #2139, opened 2019-02-28
 Plumb message validator to sender and validate outgoing messages before adding them to mpool.

anorth:
This is most of the work for #1934 (there are some cases the validator doesn't yet check).

I intend to rebase on #2136 and #2138 before merging, resolving some duplicated actor loading.

### DIFF
--- a/plumbing/msg/sender.go
+++ b/plumbing/msg/sender.go
@@ -8,11 +8,10 @@ import (
 	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/core"
-	"github.com/filecoin-project/go-filecoin/repo"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
-	"github.com/filecoin-project/go-filecoin/wallet"
 
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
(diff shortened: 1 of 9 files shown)


plumbing/msg/sender.go
```diff
@@ -56,22 +55,38 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
 	s.l.Lock()
 	defer s.l.Unlock()
 
-	nonce, err := nextNonce(ctx, s.chainReader, s.msgPool, from)
+	st, err := s.chainReader.LatestState(ctx)
+	if err != nil {
+		return cid.Undef, err
```
phritz: could be wrapped for more clarity, eg "couldnt get latest state: " here and below

plumbing/msg/sender.go
```diff
@@ -81,19 +96,13 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
...
 // nextNonce returns the next nonce for the given address. It checks
 // the actor's memory and also scans the message pool for any pending
 // messages.
-func nextNonce(ctx context.Context, chainReader chain.ReadStore, msgPool *core.MessagePool, address address.Address) (nonce uint64, err error) {
-	st, err := chainReader.LatestState(ctx)
-	if err != nil {
-		return 0, err
-	}
-
+func nextNonce(ctx context.Context, st state.Tree, msgPool *core.MessagePool, address address.Address) (nonce uint64, err error) {
```
frrist: nit: not a fan of named returns for a method like this.

anorth:
It seems I was premature putting this up to review. The test failure is non-trivial. I'll ping again for review if I fold the improved test support in here (but will try to do it in a different PR).

anorth:
@phritz please take another quick squiz before I land. The major new change is test setup in `message_propagate_test.go`. Note that there's also a flag on the validator now, opting out of the nonce-too-high check for outbound messages.

Issue #1997, opened 2019-02-16
DHT subcommand 


ZenGround0:
### Description
To make use of [upcoming DHT Improvements](https://github.com/libp2p/go-libp2p-kad-dht/pull/265#issuecomment-464148601) and enable DHT debugging, go-filecoin needs to expose a `dht` subcommand that makes calls into the node's DHT.  This issue tracks work adding in the basic commands and not doing more advanced things with improved APIs yet.

### Acceptance criteria
We need to have a `dht findprovs` command like ipfs does as this is directly helpful for debugging the autorelay system.  Move `findpeer` here, and ideally also add in `get`, `provide`, `put`, `query`.  

### Where to begin
Look at how ipfs does this and copy paste a bunch of that.

ZenGround0:
Note while this isn't necessarily a good first issue for someone totally new to this project, it is a good issue for a contributor familiar with IPFS commands who is interested in digging a little bit into the filecoin node's networking.

PR #2158, opened 2019-03-03
Feat/dht command

ZenGround0:
First of the subcommands referenced from #1997 

I'm adding in only one subcommand right now to make sure that other people are on board with some of the decisions I've made in this one that would apply to the others subcommands.

Check this out commit by commit and take a look at the initial comments for things that may need changing.

--edit--
One more note, the cid corresponding to the relay service is `zb2rhZ6FpTqFZyiAtpQFRKmybPMjq5A7oPHfmD5WeBko5kRAo` (the namespace tag "/libp2p/relay" is hashed with [this function](https://github.com/libp2p/go-libp2p-discovery/blob/master/routing.go#L64) )  When connecting to the test cluster with a daemon built on this branch I correctly see that the 5 bootstrap peers are providing the relay service using the `findprovs` command!

```
$ go-filecoin dht findprovs zb2rhZ6FpTqFZyiAtpQFRKmybPMjq5A7oPHfmD5WeBko5kRAo
QmXq6XEYeEmUzBFuuKbVEGgxEpVD4xbSkG2Rhek6zkFMp4
QmXhxqTKzBKHA5FcMuiKZv8YaMPwpbKGXHRVZcFB2DX9XY
QmZGDLdQLUTi7uYTNavKwCd7SBc5KMfxzWxAyvqRQvwuiV
QmZRnwmCjyNHgeNDiyT8mXRtGhP6uSzgHtrozc42crmVbg
Qmd6xrWYHsxivfakYRy6MszTpuAiEoFbgE1LWw4EvwBpp4
```

### DIFF
--- a/commands/dht.go
+++ b/commands/dht.go
@@ -1,0 +1,175 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	cmds "gx/ipfs/QmQtQrtNioesAWtrx8csBvfY37gTe94d6wQ3VikZUjxD39/go-ipfs-cmds"
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	pstore "gx/ipfs/QmRhFARzTHcFh8wUxwN5KvyTGq73FLC65EfFAhz8Ng7aGb/go-libp2p-peerstore"
+	notif "gx/ipfs/QmWaDSNoSdSXU9b6udyaq9T8y6LkzMwqWxECznFqvtcTsk/go-libp2p-routing/notifications"
+	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+)
+
+const (
+	dhtVerboseOptionName   = "verbose"
+	numProvidersOptionName = "num-providers"
+)
+
+// Note, most of this is copied directly from go-ipfs (https://github.com/ipfs/go-ipfs/blob/master/core/commands/dht.go).
+// A few simple modifications have been adapted for filecoin.
+var dhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Explore and manipulate the libp2p DHT.",
+		ShortDescription: ``,
+	},
+
+	Subcommands: map[string]*cmds.Command{
+		"findprovs": findProvidersDhtCmd,
+	},
+}
+
+var findProvidersDhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Find peers that can provide a given key's value.",
+		ShortDescription: "Outputs a list of newline-delimited provider Peer IDs for a given key.",
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("key", true, false, "The key whose provider Peer IDs are output.").EnableStdin(),
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption(dhtVerboseOptionName, "v", "Print extra information."),
+		cmdkit.IntOption(numProvidersOptionName, "n", "The max number of providers to find.").WithDefault(20),
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		numProviders, _ := req.Options[numProvidersOptionName].(int)
+		if numProviders < 1 {
+			return fmt.Errorf("number of providers must be greater than 0")
+		}
+
+		c, err := cid.Parse(req.Arguments[0])
+		if err != nil {
+			return err
+		}
+
+		ctx, cancel := context.WithTimeout(req.Context, time.Minute)
+		ctx, events := notif.RegisterForQueryEvents(ctx)
+
+		pchan := GetPorcelainAPI(env).NetworkFindProvidersAsync(ctx, c, numProviders)
+
+		go func() {
+			defer cancel()
+			for p := range pchan {
+				np := p
+				// Note that the peer IDs in these Provider
+				// events are the main output of this command.
+				// These results are piped back into the event
+				// system so that they can be read alongside
+				// other routing events which are output in
+				// verbose mode but otherwise filtered.
+				notif.PublishQueryEvent(ctx, &notif.QueryEvent{
+					Type:      notif.Provider,
+					Responses: []*pstore.PeerInfo{&np},
+				})
+			}
+		}()
+		for e := range events {
+			if err := res.Emit(e); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *notif.QueryEvent) error {
+			pfm := pfuncMap{
+				notif.FinalPeer: func(obj *notif.QueryEvent, out io.Writer, verbose bool) {
+					if verbose {
+						fmt.Fprintf(out, "* closest peer %s\n", obj.ID) // nolint: errcheck
+					}
+				},
+				notif.Provider: func(obj *notif.QueryEvent, out io.Writer, verbose bool) {
+					prov := obj.Responses[0]
+					if verbose {
+						fmt.Fprintf(out, "provider: ") // nolint: errcheck
+					}
+					fmt.Fprintf(out, "%s\n", prov.ID.Pretty()) // nolint: errcheck
+					if verbose {
+						for _, a := range prov.Addrs {
+							fmt.Fprintf(out, "\t%s\n", a) // nolint: errcheck
+						}
+					}
+				},
+			}
+
+			verbose, _ := req.Options[dhtVerboseOptionName].(bool)
+			printEvent(out, w, verbose, pfm)
+
+			return nil
+		}),
+	},
+	Type: notif.QueryEvent{},
+}
+
+type printFunc func(obj *notif.QueryEvent, out io.Writer, verbose bool)
+type pfuncMap map[notif.QueryEventType]printFunc
+
+// printEvent writes a libp2p event to a user friendly output on the out writer.
+// Note that this function is only needed to enable the output logging of
+// events that only show up during a "verbose" run. If we choose to eliminate
+// the verbose option this can be removed. However if we keep the verbose option
+// in findprovs and on any other dht subcommands we decide to copy over from
+// ipfs this function will stay needed.
+func printEvent(obj *notif.QueryEvent, out io.Writer, verbose bool, override pfuncMap) {
+	if verbose {
+		fmt.Fprintf(out, "%s: ", time.Now().Format("15:04:05.000")) // nolint: errcheck
+	}
+
+	if override != nil {
+		if pf, ok := override[obj.Type]; ok {
+			pf(obj, out, verbose)
+			return
+		}
+	}
+
+	switch obj.Type {
+	case notif.SendingQuery:
+		if verbose {
+			fmt.Fprintf(out, "* querying %s\n", obj.ID) // nolint: errcheck
+		}
+	case notif.Value:
+		if verbose {
+			fmt.Fprintf(out, "got value: '%s'\n", obj.Extra) // nolint: errcheck
+		} else {
+			fmt.Fprint(out, obj.Extra) // nolint: errcheck
+		}
+	case notif.PeerResponse:
+		if verbose {
+			fmt.Fprintf(out, "* %s says use ", obj.ID) // nolint: errcheck
+			for _, p := range obj.Responses {
+				fmt.Fprintf(out, "%s ", p.ID) // nolint: errcheck
+			}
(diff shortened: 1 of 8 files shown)


commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+
+	cmds "gx/ipfs/QmQtQrtNioesAWtrx8csBvfY37gTe94d6wQ3VikZUjxD39/go-ipfs-cmds"
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	pstore "gx/ipfs/QmRhFARzTHcFh8wUxwN5KvyTGq73FLC65EfFAhz8Ng7aGb/go-libp2p-peerstore"
+	notif "gx/ipfs/QmWaDSNoSdSXU9b6udyaq9T8y6LkzMwqWxECznFqvtcTsk/go-libp2p-routing/notifications"
+	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+)
+
+const (
+	dhtVerboseOptionName   = "verbose"
```
ZenGround0: I kept the same UI from ipfs, including this verbose mode.  It looks like it will be helpful for debugging (an explicit goal of this dht command) but it also departs from behavior in other commands.  Input welcome.

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+	numProvidersOptionName = "num-providers"
+)
+
+// Note, most of this is copied directly from go-ipfs (https://github.com/ipfs/go-ipfs/blob/master/core/commands/dht.go).
+// A few simple modifications have been adapted for filecoin.
+var dhtCmd = &cmds.Command{
```
ZenGround0: ipfs code has a comment about moving most subcommands (including this one) to a `routing` subcommand.  I kept this on dht because I don't have an opinion or know how canonical that TODO was.  Input welcome.

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *notif.QueryEvent) error {
+			pfm := pfuncMap{
+				notif.FinalPeer: func(obj *notif.QueryEvent, out io.Writer, verbose bool) {
+					if verbose {
+						fmt.Fprintf(out, "* closest peer %s\n", obj.ID) // nolint: errcheck
```
ZenGround0: The ipfs encoders don't do error checking when printing to the writer.  I put a bandaid on this with the "nolint" comments.  Happy to think through error checking if this seems important.

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil
+		}),
+	},
+	Type: notif.QueryEvent{},
+}
+
+type printFunc func(obj *notif.QueryEvent, out io.Writer, verbose bool)
+type pfuncMap map[notif.QueryEventType]printFunc
+
+func printEvent(obj *notif.QueryEvent, out io.Writer, verbose bool, override pfuncMap) {
```
ZenGround0: Note that this function is only needed to enable the output logging of events that only show up during a "verbose" run.  If we choose to eliminate the verbose option the code will get significantly shorter.  However if we keep the verbose option on the other (to be written) dht subcommands we can reuse this function and the overhead goes way down.

plumbing/api.go
```diff
@@ -181,6 +182,11 @@ func (api *API) NetworkGetPeerID() peer.ID {
 	return api.network.GetPeerID()
 }
 
+// RouterFindProvidersAsync issues a findProviders query to the router.
+func (api *API) RouterFindProvidersAsync(ctx context.Context, key cid.Cid, count int) <-chan pstore.PeerInfo {
```
ZenGround0: NetworkFindProvidersAsync seemed unnecessarily general but input welcome on the name.

plumbing/ntwk/ntwk.go
```diff
@@ -12,14 +13,16 @@ type Network struct {
 	host host.Host
 	*pubsub.Subscriber
 	*pubsub.Publisher
+	routing.IpfsRouting
```
ZenGround0: This is probably where I feel the least sure.  It seems weird to embed an interface here instead of a struct as the other embeddings are structs.  However embedding an interface nicely allows plumbing to be constructed the same way whether the node is in online or offline mode.

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+)
+
+const (
+	dhtVerboseOptionName   = "verbose"
+	numProvidersOptionName = "num-providers"
+)
+
+// Note, most of this is copied directly from go-ipfs (https://github.com/ipfs/go-ipfs/blob/master/core/commands/dht.go).
+// A few simple modifications have been adapted for filecoin.
```
ZenGround0: This copy-pasting is a lot of duplication across projects.  However my guess is that this is the right way to do it because we've done the same thing with other commands in the past and neither project likely wants to deal with abstractions that generalize across core APIs of the other project.  Please let me know if I should be abstracting this logic somewhere else to share amongst projects instead.

testhelpers/commands.go
```diff
@@ -849,11 +859,11 @@ func (td *TestDaemon) createNewProcess() {
 		td.test.Fatal(err)
 	}
 	// uncomment this and comment out the following 4 lines to output daemon stderr to os stderr
-	//td.process.Stderr = os.Stderr
```
ZenGround0: Temporary for WIP daemon test

commands/dht_daemon_test.go
```diff
@@ -0,0 +1,53 @@
...
+	require := require.New(t)
+	//	assert := assert.New(t)
+
+	dClient := th.NewDaemon(t, th.SwarmAddr("/ip4/0.0.0.0/tcp/6000")).Start()
+	defer dClient.ShutdownSuccess()
+
+	dProvider := th.NewDaemon(t, th.SwarmAddr("/ip4/0.0.0.0/tcp/6001"), th.IsRelay).Start()
+	defer dProvider.ShutdownSuccess()
+
+	time.Sleep(30 * time.Second)
```
ZenGround0: My biggest concern with this test is that it needs a long delay and is inherently flaky with this wait in here.  I think it's worth finally getting some relay test coverage but want to check in on this.

commands/dht_daemon_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	dClient := th.NewDaemon(t, th.SwarmAddr("/ip4/0.0.0.0/tcp/6000")).Start()
+	defer dClient.ShutdownSuccess()
+
+	dProvider := th.NewDaemon(t, th.SwarmAddr("/ip4/0.0.0.0/tcp/6001"), th.IsRelay).Start()
+	defer dProvider.ShutdownSuccess()
+
+	dClient.ConnectSuccess(dProvider)
+
+	time.Sleep(30 * time.Second)
```
ZenGround0: Any lower and this consistently fails.  It seems important to get some coverage on relay and dht command.  Do we think introducing this potential flake / test delay is worth it?

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+var dhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Issue commands directly through the DHT.",
```
anorth: Maybe "Explore and manipulate the libp2p DHT"?

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+	},
+
+	Subcommands: map[string]*cmds.Command{
+		"findprovs": findProvidersDhtCmd,
+	},
+}
+
+var findProvidersDhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Find peers that can provide a specific value, given a key.",
```
anorth: "Finds peers that can provide records for a key"?

commands/dht_daemon_test.go
anorth: I would say no, or at least not in the same class of tests as other daemon tests. Assuming it's not possible to fake out libp2p to exercise this faster 😢 , I would put this in a silo of long-running integration tests that are *not* part of normal developer workflows: maybe periodically on CI like the functional tests.

plumbing/api.go
anorth: `DHTFindProvidersAsync`? "Router" doesn't mean anything to me as a filecoin user.

plumbing/ntwk/ntwk.go
anorth: blocking: we have done work in pubsub to abstract libp2p away from the filecoin code in order to make testing and upgrading easy. Please add a wrapper in the same way as the local `pubsub` package referenced immediately above. See https://github.com/filecoin-project/go-filecoin/pull/2092

plumbing/api.go
ZenGround0: Hmm `Router` shows up as a dependency of the node and now the plumbing so it should mean something right?  The thing is the thingy that calls `FindProvidersAsync` isn't always a DHT, sometimes its an offline router that always returns nothing.

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+				return err
+			}
+		}
+
+		return nil
+	},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, out *notif.QueryEvent) error {
+			pfm := pfuncMap{
+				notif.FinalPeer: func(obj *notif.QueryEvent, out io.Writer, verbose bool) {
```
phritz: where do these come from? i only see provider events published above.

plumbing/ntwk/ntwk.go
phritz: that sounds like a good reason to me

commands/dht.go
```diff
@@ -0,0 +1,163 @@
...
+		ctx, cancel := context.WithCancel(req.Context)
+		ctx, events := notif.RegisterForQueryEvents(ctx)
+
+		pchan := GetPorcelainAPI(env).RouterFindProvidersAsync(ctx, c, numProviders)
+
+		go func() {
+			defer cancel()
+			for p := range pchan {
+				np := p
+				notif.PublishQueryEvent(ctx, &notif.QueryEvent{
```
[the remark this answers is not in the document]
ZenGround0: Yeah good question tracing this out and seeing what was going on was one of the trickier parts of copying over.  I'll add some comments.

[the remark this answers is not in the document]
ZenGround0: Short answer is that the peer ids we see are exactly what we care about outputting but the command is listening in only one place on the events channel which simplifies things in verbose mode.  In order to get results at this level of abstraction piped into the same notification system we make this call back into libp2p.

commands/dht.go
ZenGround0: They are generated internally by the DHT.  This is why verbose mode is good for debugging, we get to see libp2p system internals.

commands/dht.go
anorth: Move this comment into the code.

filnet/router.go
```diff
@@ -0,0 +1,25 @@
...
+// NewRouter builds a new router.
+func NewRouter(r routing.IpfsRouting) *Router {
+	return &Router{routing: r}
+}
+
+// FindProvidersAsync searches for and returns peers who are able to provide a
+// given key.
+func (r *Router) FindProvidersAsync(ctx context.Context, key cid.Cid, count int) <-chan pstore.PeerInfo {
+	return r.routing.FindProvidersAsync(ctx, key, count)
+}
```
anorth: Thanks. FYI I may centralise all the libp2p wrappers somewhere in the near future.

ZenGround0: Sweet, I vote that place is filnet!

commands/dht.go
phritz: good idea

commands/main.go
```diff
@@ -112,6 +112,7 @@ NETWORK COMMANDS
   go-filecoin id                     - Show info about the network peers
   go-filecoin ping <peer ID>...      - Send echo request packets to p2p network members
   go-filecoin swarm                  - Interact with the swarm
+  go-filecoin dht                    - Interact with the dht
```
phritz: alpha, like all the rest of the tables please

filnet/router.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	pstore "gx/ipfs/QmRhFARzTHcFh8wUxwN5KvyTGq73FLC65EfFAhz8Ng7aGb/go-libp2p-peerstore"
+	routing "gx/ipfs/QmWaDSNoSdSXU9b6udyaq9T8y6LkzMwqWxECznFqvtcTsk/go-libp2p-routing"
+)
+
+// Router is used to find information about who has what content.
```
phritz: could you make this more specific? what kind of a thing is 'who' and what kind of content are we talking about?

Issue #1863, opened 2019-02-11
IPTB localfilecoin plugin should make sure `go-filecoin` is using the provided binary through AttrFilecoinBinary.


travisperson:
### Description

FAST currently executions actions using whatever IPTB plugin is provided (through fast.Filecoin). The localfilecoin plugin can be configured to use a particular binary. However, it only uses that binary when it executes `go-filecoin` internally. This is used for `init`, `start`, and a few other things.

The plugins don't restrict the user to executing just `go-filecoin` commands (maybe it should, I remember there being a reason to not do this at some point).

This means that when we execute an action using `RunCmdWithStdin`, and specify something like `go-filecoin id`, it will actually just find whatever is in the path and use that. This means that the binary that is specified through options passed to `NewProcess`, needs to be the same in the PATH.

I think the correct way to handle this would be to copy the provided binary to a directory and add it to the path. This also has the added benefit that when we execute `Shell`, using `go-filecoin` will work nicely as well, because we can override the users PATH in the new shell.

### Acceptance criteria

- When `AttrFilecoinBinary` is provided to NewNode, it creates a directory under the provided dir, and copies the provided file.
-  Update the environment to prepend the directory to PATH.

### Where to begin

https://github.com/filecoin-project/go-filecoin/blob/master/tools/iptb-plugins/filecoin/local/localfilecoin.go#L77
https://github.com/filecoin-project/go-filecoin/blob/master/tools/iptb-plugins/filecoin/local/filecoinutil.go#L52

Issue #1810, opened 2019-02-05
Passing Init and Daemon flags to series


travisperson:
Running `init` and `daemon` in series is probably going to be pretty common because there are a lot of different basic rolls a node can run such as being the genesis node, a miner, or a client, and possibility others down the road.

Most of these configuration options are going to be the same regardless of which series. Some work was done in #1783 to allow defining the init / daemon flags at the environments `NewProcess` level.

I think there needs to be some more work here though, and a discussion around exactly how it will work.

Issue #1631, opened 2019-01-17
Implement FAT Series around FAT Actions


frrist:
### Description
Once #1629 gets moving along we can begin to implement yet an even high level set of wrapper methods herein referred to as **Series**. Series will be made up of multiple Actions, and will allow for complex operations to be performed against filecoin process(s). Some examples of Series include:
- `EnsureCreateMiner`: the method will get filecoin, create a miner, and ensure the create miner message is successfully mined.
- `MakeDealWithMiner`: the method will create an ask with a miner, propose a deal to the miner with a client, and ensure the deal is successfully mined.

### Acceptance criteria
- Enough Series are written such that the functional retrieval test can be composed of calls to story commands.
- [x] Genesis Series
- [ ] Retrieval Series
- [ ] Storage Series

anorth:
@travisperson is this about done now that FAST is usable? Could remaining work be broken out into a tighter scoped issue?

travisperson:
Ya, I think the larger concern of this issue has been address. A lot of this work was defined prior to FAST being implemented and we didn't really know what we needed. These series could be written, but I don't think they are strictly needed except to showcase as an example.

If we want these we can open additional issues, for now I think we can close this out.

PR #2168, opened 2019-03-05
FAST Updates

travisperson:
Includes fixes for #1810, #1631, #1863


mishmosh:
@travisperson can you tag your first-round reviewer(s) when ready?

travisperson:
> @travisperson can you tag your first-round reviewer(s) when ready?

Ya, I think I need #2193 to land as it is blocking CI from passing.

tools/fast/series/global_sleep_delay.go
```diff
@@ -0,0 +1,17 @@
...
+
+// GlobalSleepDelay is the time value that is used to wait for block processing
+// in series
+var GlobalSleepDelay = mining.DefaultBlockTime
+
+// SleepDelay is a helper method to make sure people
+// don't call `time.Sleep` themselves in series.
+func SleepDelay() {
+	time.Sleep(GlobalSleepDelay)
+}
```
travisperson: I expect this to possibility be controversial. One thought I had to help remove the global was to change series to be a package that you initiate, and maybe provide an environment too. That would also help with managing the passing of some redundant information in some places, and would be a good way to remove this global.

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -106,16 +106,52 @@ func init() {
...
+
+		dst := filepath.Join(dir, "bin", filepath.Base(binPath))
+		if err := copyFileContents(binPath, dst); err != nil {
+			return nil, err
+		}
+
+		if err := os.Chmod(dst, 0755); err != nil {
+			return nil, err
+		}
+
```
travisperson: Just so reviewers understand why this is here:

This makes it so we grab a copy of the binary that is used. This will either be whatever is found in the path (not shown in this change), or was is provided by the `AttrFilecoinBinary` attr (option, also not shown). This makes sure that when we drop into a shell we are using the same binary that the node was initiated with.

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -374,7 +410,25 @@ func (l *Localfilecoin) Shell(ctx context.Context, ns []testbedi.Core) error {
...
 	}
 
-	return syscall.Exec(shell, []string{shell}, nenvs)
+	pid, err := l.getPID()
+	if err != nil {
+		return err
+	}
+
+	nenvs = filecoin.UpdateOrAppendEnv(nenvs, "FC_PID", fmt.Sprintf("%d", pid))
+	nenvs = filecoin.UpdateOrAppendEnv(nenvs, "FC_BINARY", l.binPath)
```
travisperson: These are to enable an easy way to use `dlv` when in side of a shell; eg `dlv attach $FC_PID $FC_BINARY`

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,338 @@
...
+	if len(workdir) == 0 {
+		workdir, err = ioutil.TempDir("", "localnet")
+		if err != nil {
+			handleError(err)
+			os.Exit(1)
+		}
+	}
+
+	if ok, err := isEmpty(workdir); !ok {
+		handleError(err, "workdir is not empty;")
```
frrist: nit: can we use the errors package for wrapping errors like these?

tools/fast/series/series_createminerwithask.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/filecoin-project/go-filecoin/api"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+)
+
+// CreateMinerWithAsk setups a miner and sets an ask price. The created ask is
+// returned. The node will be mining as well.
+func CreateMinerWithAsk(ctx context.Context, miner *fast.Filecoin, pledge uint64, collateral *big.Int, price *big.Float, expiry *big.Int) (api.Ask, error) {
+
+	// Create miner
+	_, err := miner.MinerCreate(ctx, pledge, collateral, fast.AOPrice(big.NewFloat(1.0)), fast.AOLimit(300))
```
frrist: should we return the miner address we have created in addition to the ask?

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -374,7 +410,25 @@ func (l *Localfilecoin) Shell(ctx context.Context, ns []testbedi.Core) error {
 		nenvs = append(nenvs, fmt.Sprintf("NODE%d=%s", i, peerid))
 	}
 
-	return syscall.Exec(shell, []string{shell}, nenvs)
+	pid, err := l.getPID()
+	if err != nil {
+		return err
+	}
+
+	nenvs = filecoin.UpdateOrAppendEnv(nenvs, "FC_PID", fmt.Sprintf("%d", pid))
```
frrist: nit: stick with `FIL_*` pattern here

tools/fast/process.go
```diff
@@ -85,11 +85,23 @@ func NewFilecoinProcess(ctx context.Context, c IPTBCoreExt, eo EnvironmentOpts)
 
 // InitDaemon initializes the filecoin daemon process.
 func (f *Filecoin) InitDaemon(ctx context.Context, args ...string) (testbedi.Output, error) {
+	if len(args) == 0 {
```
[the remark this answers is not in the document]
travisperson: It means that you can override them, if anything I would make it an error if both are defined, but we should allow to pass args to `InitDaemon`.

tools/fast/series/series_createminerwithask.go
travisperson: The miner address is part of the `ask` structure, so it's available.

tools/fast/bin/localnet/main.go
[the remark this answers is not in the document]
travisperson: Ya, it's not really an error we are returning.

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// The genesis process is the filecoin node that loads the miner that is
+	// define with power in the genesis block, and the prefunnded wallet
+	genesis, err := env.NewProcess(ctx, lpfc.PluginName, options, fastenvOpts)
+	if err != nil {
+		exitcode = handleError(err, "failed to create genesis process;")
+		return
+	}
+
+	err = series.SetupGenesisNode(ctx, genesis, genesisMiner.Address, files.NewReaderFile(genesisMiner.Owner))
```
phritz: bootstrapMiner?

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+		miners = append(miners, miner)
+	}
+
+	// We will now go through the process of creating miners
+	// InitAndStart
+	// 1. Initialize node
+	// 2. Start daemon
+	//
+	// Connect
+	// 3. Connect to genesis
```
phritz: genesis is sufficiently overloaded we should be more specific, eg bootstrap node

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+		err = series.SendFilecoinDefaults(ctx, genesis, miner, fil)
+		if err != nil {
+			exitcode = handleError(err, "failed series.SendFilecoinDefaults;")
+			return
+		}
+
+		pledge := uint64(10)                    // sectors
+		collateral := big.NewInt(500)           // FIL
+		price := big.NewFloat(0.000000001)      // price per byte/block
+		expiry := big.NewInt(24 * 60 * 60 / 30) // ~24 hours
```
phritz: maybe should do it for longer so the miner continues to have power for more than a day?

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+			exitcode = handleError(err, "failed series.ImportAndStore;")
+			return
+		}
+
+		deals = append(deals, deal)
+
+	}
+
+	for _, deal := range deals {
+		err = series.WaitForDealState(ctx, genesis, deal, storagedeal.Posted)
```
phritz: seems like here and above we want to derive a new context that has a reasonable timeout. this will wait forever, or it looks like it would

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+		if err != nil {
+			exitcode = handleError(err, "failed series.WaitForDealState;")
+			return
+		}
+	}
+
+	if shell {
```
phritz: this model is a little different than what i expected, which was you run a thing to start the cluster up and then it tells you when it has. then you can run commands against it like shell or similar. i dont feel wed to the other model, just curious about the tradeoffs. my guess is this way we dont have to serialize any state, we can keep it all in process?

phritz: so this shell is with a new node connected into the network. my guess is that is easier than connecting to one of the miners?

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+		interval.Stop()
+		fmt.Println("===================================")
+		fmt.Println("===================================")
+		io.Copy(os.Stdout, interval) // nolint: errcheck
+		fmt.Println("===================================")
+		fmt.Println("===================================")
+	}
+
+	fmt.Println("Finished!")
+	fmt.Println("Ctrl-C to handleError")
```
phritz: unclear to me whether and why i would want to handlerror...

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+	if len(msg) != 0 {
+		fmt.Println(msg[0], err)
+	} else {
+		fmt.Println(err)
+	}
+
+	return 1
+}
+
+// https://stackoverflow.com/a/3070891://stackoverflow.com/a/30708914
```
phritz: eh? put down those crazy pills, they are affecting your judgement.

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+	defer f.Close() // nolint: errcheck
+
+	_, err = f.Readdirnames(1) // Or f.Readdir(1)
+	if err == io.EOF {
+		return true, nil
+	}
+	return false, err // Either not empty or error, suits both cases
+}
+
+func getSectorSize(smallSectors bool) (uint64, error) {
```
phritz: zomg this is such a painful way to do this. we should have an issue to surface sector size in plumbing as part of a params() call. fee free to just go do it, or file an issue on rose to do it after she is done removing her bits of the node api...

tools/fast/series/global_sleep_delay.go
phritz: i dont find it so onerous, but what you suggest sounds better, but more work

tools/fast/series/series_createminerwithask.go
```diff
@@ -0,0 +1,27 @@
+package series
```
phritz: strongly suggest we drop the stuttering and shorten filenames, eg series/createminer.go

tools/fast/series/series_genesisnode.go
```diff
@@ -13,9 +13,8 @@ import (
 // SetupGenesisNode will initialize, start, configure, and issue the "start mining" command to the filecoin process `node`.
 // Process `node` will use the genesisfile at `gcURI`, be configured with miner `minerAddress`, and import the address of the
```
phritz: comment should be updated

tools/fast/series/series_withwallet.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+)
+
+// ErrWithWalletRestoreFailed is returned if the original address could not be restored.
+var ErrWithWalletRestoreFailed = errors.New("failed to restore default wallet after WithWallet exited")
+
+// WithWallet can be used to temporarlly change the default wallet address of
+// the node to sessionWallet for all FAST actions exected inside of sessionFn.
```
phritz: say why one would use this

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+		exitcode = handleError(err, "failed to retrieve miner information from genesis;")
+		return
+	}
+
+	fastenvOpts := fast.EnvironmentOpts{
+		InitOpts:   []fast.ProcessInitOption{fast.POGenesisFile(genesisURI)},
+		DaemonOpts: []fast.ProcessDaemonOption{fast.POBlockTime(series.GlobalSleepDelay)},
+	}
+
+	// The genesis process is the filecoin node that loads the miner that is
```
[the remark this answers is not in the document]
frrist: Genesis node/process seems appropriate here, it's the first node created in the network. This means its the the only node with money, power, the capability to mine, and could eventually serve as the source of a faucet. I think it does a fair bit more than bootstrap.

tools/fast/process.go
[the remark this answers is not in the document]
frrist: This feels like a developer error if both are defined, dare I say panic?

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,338 @@
...
+	fmt.Println("Ctrl-C to handleError")
+
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, os.Interrupt, syscall.SIGTERM)
+
+	<-signals
+}
+
+func handleError(err error, msg ...string) {
+	if err == nil {
```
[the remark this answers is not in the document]
frrist: Can you add a comment on why this method is required?

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+	// 7. Generated some random data and import it to genesis
+	// 8. Genesis propposes a storage deal with miner
+	//
+	// WaitForDealState
+	// 9. Query deal till posted
+
+	var deals []*storagedeal.Response
+
+	for _, miner := range miners {
+		err = series.InitAndStart(ctx, miner)
```
frrist: Unsure if this is the correct place for this comment, but we will need to shut these daemons down at some point. As things currently stand the daemons will continue to run in the background until the user manually kills them.

tools/fast/process.go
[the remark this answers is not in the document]
travisperson: Possibility, but due to the fact that the panic could lead to leaked resources (not shutting things down) I'm going to opt for the error.

tools/fast/bin/localnet/main.go
```diff
@@ -0,0 +1,350 @@
...
+	lpfc "github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin/local"
+)
+
+var (
+	workdir      string
+	shell        bool
+	count        = 5
+	blocktime    = 5 * time.Second
+	err          error
+	fil          = 100000
```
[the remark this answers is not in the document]
travisperson: There is no good reason it's just a magic number :man_shrugging:

tools/fast/bin/localnet/main.go
travisperson: I've now made this a flag, along with the other values

tools/fast/bin/localnet/main.go
travisperson: find and replace error; `s/exit/handleError/`

tools/fast/bin/localnet/main.go
travisperson: Only if the user kills the process before the setup completes. This seems to be due to the fact that defer functions are not run after Ctrl-C.

tools/fast/bin/localnet/main.go
travisperson: The use case for this tool is that the user will be interfacing with it directly, so the user can arbitrarily choose when they want to cancel the context by calling Ctrl-C. I think forcing a timeout is less flexible for users.

tools/fast/bin/localnet/main.go
travisperson: > which was you run a thing to start the cluster up and then it tells you when it has. then you can run commands against it like shell or similar.

This is still possibly, just not yet built. I believe the model you are talking about is to have a localnet work setup, and then write additional FAST tests / programs that would create nodes to join the network. The nodes would then perform whatever actions they choose. This is the model that will be used for functional tests, so any functional test should be able to be executed against a localnet (it's a tad bit trickery as we need to communicate a bit more about the network, as localnet doesn't have a nice flag).

> so this shell is with a new node connected into the network. my guess is that is easier than connecting to one of the miners?

Yes, localnet is a tool to make it easier to setup a locally running network with miners. The shell flag, and this condition just make it easier to get access to a node that is connect to the network and ready to be used use. The shell could be opened against one of the miners, but starting with a fresh node is a bit cleaner and allows the user to go through the miner setup themselves if they choose. Think of the localnet tool as the hello world introduction program to Filecoin, it's a private sandbox.

The user can still initialize their own node, and connect to the localnet manually if they want too, or provide a repo flag to any node already running if they want to control the miner nodes directly.

localnet is just a tool to showcase FAST and provide utility to developers.

tools/fast/bin/localnet/main.go
travisperson: If this is added to the plumbing, would it still be possible to execute that plumbing in a stand alone way that has minimal requirement such as the way this function does? This needs to be executed prior to the initialization or starting of any filecoin process. We could move this implementation, but my guess is that it would still require the same process.

Another approach would be to provide a simpler way to get the value `GetMaxUserBytesPerStagedSector` without requiring the setup of `RustSectorBuilderConfig` that only requires providing the `sectorStoreType`. That method could then be exposed through plumbing as well.

@laser do you know if it would be possible to get this value through a more minimal process?

Issue #2155, opened 2019-03-02
Cannot sync chain on user devnet (regression)


jimpick:
I upgraded the two nodes I was running on the user devnet to master yesterday, and they both stopped syncing the chain.

I tried reseting the one I am running on Linux (delete ~/.filecoin, etc.), and even after an hour, it wouldn't sync the chain past a height of zero.

I reverted to code from Tuesday, rebuilt, and confirmed that the chain would sync with the older code.

Then I ran `git bisect` and tried a bunch of different builds, and determined that the chain stopped syncing once this commit was added:

https://github.com/filecoin-project/go-filecoin/commit/b2c75bd799e058f6a382c252d0dba3f425e35b5f

It takes a while to test, and it's possible I'm just seeing randomness, but it feels like it's fairly reproducible.

I'd recommend that somebody try to confirm these results - if true, that means that nobody has been able to successfully try out the user devnet since about Wednesday (the instructions on the wiki instruct building from source on master).

rkowalick:
@jimpick,

Looking into reproducing this right now.

phritz:
@rkowalick isn't thi wai? there is no expectation that you can interact with the user devent with code built from any other commit than the one the user devnet was cut from. in this case there is the breaking commit mentioned that will prevent interoperability.

jimpick:
I just did a Zoom screenshare with @rkowalick and we compared our setups ... his seems to sync OK, and mine definitely is acting strange.

I'm going to try to reproduce on OS X.

jimpick:
It's working on my iMac, but not on Ubuntu.

jimpick:
I discovered that `go-filecoin swarm peers` wasn't listing any peers outside of the bootstrap servers ... the wireless connection to my Ubuntu machine was very, very weak. I went out and bought an updated mesh router, and now the Linux machine is acting consistent with the OS X machine. So bad networking on my network was definitely part of the problem.

There appears to be an additional problem now ... once the chain syncs, the head doesn't move forward - it will continually loop with the following output:

```
23:18:16.527  INFO chain.sync: syncing the chain, currently at block height 34670 default_syncer.go:153
23:18:16.609  INFO chain.sync: syncing the chain, currently at block height 33576 default_syncer.go:153
23:18:16.670  INFO chain.sync: syncing the chain, currently at block height 32526 default_syncer.go:153
23:18:16.732  INFO chain.sync: syncing the chain, currently at block height 31440 default_syncer.go:153
[...]

PR #2175, opened 2019-03-05
Mark chain as bad if syncing fails

rkowalick:
We mark a chain as bad in the event that a syncOne call on a tipset
returns an error.

Part of resolution of #2155 .

### DIFF
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -371,6 +371,12 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 			}
 		}
 		if err = syncer.syncOne(ctx, parent, ts); err != nil {
+			// While `syncOne` can indeed fail for reasons other than consensus,
+			// adding to the badTipSets at this point is the simplest, since we
+			// have access to the chain. If syncOne fails for non-consensus reasons,
+			// there is no assumption that the running node's data is valid at all,
+			// so we don't really lose anything with this simplification.
+			syncer.badTipSets.AddChain(chain[i:])
 			return err
 		}
 		parent = ts
--- a/chain/bad_tipset_cache.go
+++ b/chain/bad_tipset_cache.go
@@ -7,7 +7,10 @@ import (
 )
 
 // badTipSetCache keeps track of bad tipsets that the syncer should not try to
-// download.  Readers and writers grab a lock.
+// download. Readers and writers grab a lock. The purpose of this cache is to
+// prevent a node from having to repeatedly invalidate a block (and its children)
+// in the event that the tipset does not conform to the rules of consensus. Note
+// that the cache is only in-memory, so it is reset whenever the node is restarted.
 // TODO: this needs to be limited.
 type badTipSetCache struct {
 	mu  sync.Mutex


chain/default_syncer.go
```diff
@@ -371,6 +371,7 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 			}
 		}
 		if err = syncer.syncOne(ctx, parent, ts); err != nil {
+			syncer.badTipSets.AddChain(chain[i:])
```
phritz: blocking: this doesn't differentiate between a bad tipset error and an error because something didn't work as intended. seems like it should be an important distinction, is it not?

phritz:
i'm headed ooo for 3 days, over to @acruikshank to approve when appropriate, see my blocking question above

chain/default_syncer.go
ZenGround0: @rkowalick and I talked a lot about this and we landed on this not being an important distinction, at least not right now.  

There is exactly one expected return error from `syncOne` during normal operation: https://github.com/filecoin-project/go-filecoin/blob/master/chain/default_syncer.go#L219.  If this error is returned it means this tipset does not conform to the consensus rules (invalid leader election proof, invalid state machine transition, etc).  In this case the tipset will always be bad and the chain based on this tipset should be marked as bad.

In addition to this one error there are many other errors that can be returned in the case that fundamental things are broken.  For example most of these will be thrown if the chain store is in a corrupt state.  These errors are like the "fault" errors explicitly denoted in the state machine processing code.  

Our reasoning for marking the tipset bad in this case is that we think the behavior of the node is undefined in the case there is a chain store fault, so it's not important to *not* add it to the cache in this case.  Additionally a situation like that is going to require a node restart (if not re-init) and that will clear the bad tipsets cache anyway.

This code can be modified to check the error and only add to the cache in that case.  I'm thinking the syncer code should also be more explicit about chain store faults and try to do something about them.  Probably a good start is to log them very loudly and explicitly.

anorth: Please summarise this comment in the code.

acruikshank: Thanks @ZenGround0. I buy the logic that's it's not a bad thing to mark a tip set as bad under fault conditions. 

What's still unclear to me as a relative newcomer to this code is why we're doing this in the first place. It appears the badTipSets cache prevents us from having to re-validate tip sets we've determined bad (either by being bad or by having a bad parent) in `collectChain`. It appears that we are expanding the definition of "bad" to include tip sets that fail to sync. Questions (cc @rkowalick):

1. How does this help resolve #2155? The issue seems to have been closed without this change.
2. Is this more than a performance optimization?
3. If a user has a long running node and encounters a hiccup syncing a valid block that creates a false positive here, how do they know to restart the node?

ZenGround0: 1. This behavior of seeing a chain syncing over and over in the logs came up as complaint which launched investigation of #2155.  @rkowalick figured out that the root cause was consensus state transition error and made this change to improve the UX of the failure.  I agree this should refer to a follow on and not #2155 itself.
2. No it does not change correctness.  It does makes logs less confusing.
3. Right now they wouldn't.  A power user might infer that behavior from seeing the `the chain store is in an unexpected state` in the logs.

rkowalick: I have added a comment in the `badTipSetCache` to clarify the purpose of the cache for the future. I am apprehensive about adding a comment at this call site, since, according to @ZenGround0, the change above certainly agrees with the original motivation behind this cache.

PR #2179, opened 2019-03-05
Feat/miner plum pork 1940

### LINKED ISSUE #1940: Remove No Longer Necessary Miner Node API Methods
### Description

Several methods in the `api/impl/miner.go` just call directly into plumbing. We should remove them and change their call sites to use plumbing instead.

rosalinekarr:
# Problem

We want to delete the `api` directory, but the miner API is part of that.

# Solution

Move miner api methods into plum pork

Resolves #1940

### DIFF
--- a/api/miner.go
+++ b/api/miner.go
@@ -3,7 +3,6 @@ package api
 import (
 	"context"
 
-	cid "gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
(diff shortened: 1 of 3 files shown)


api/miner.go
```diff
@@ -14,10 +12,4 @@ import (
 // Miner is the interface that defines methods to manage miner operations.
 type Miner interface {
 	Create(ctx context.Context, fromAddr address.Address, gasPrice types.AttoFIL, gasLimit types.GasUnits, pledge uint64, pid peer.ID, collateral *types.AttoFIL) (address.Address, error)
```
rkowalick: What makes removing this one hard?

PR #2183, opened 2019-03-05
integrate go-filecoin with new variable proof-producing generate_post and variable proof-consuming verify_post CGO calls

laser:
This PR fixes the go-filecoin side of [this rust-fil-proofs issue](https://github.com/filecoin-project/rust-fil-proofs/issues/505).

## What's in this PR?

- modify go-filecoin to conform to [the spec](https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost), which shows `submitPoSt` message receiving multiple PoSt proofs
- modify `generate_post` and `verify_post` CGO calls to return/accept variable number of PoSt proofs

### DIFF
--- a/abi/abi.go
+++ b/abi/abi.go
@@ -10,6 +10,7 @@ import (
 	cbor "gx/ipfs/QmcZLyosDwMKdB6NLRsiss9HXzDPhVhhRtPy67JFKTDQDX/go-ipld-cbor"
 
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -46,6 +47,8 @@ const (
 	SectorID
 	// CommitmentsMap is a map of stringified sector id (uint64) to commitments
 	CommitmentsMap
+	// PoStProofs is an array of proof-of-spacetime proofs
+	PoStProofs
 )
 
 func (t Type) String() string {
@@ -76,6 +79,8 @@ func (t Type) String() string {
 		return "uint64"
 	case CommitmentsMap:
 		return "map[string]Commitments"
+	case PoStProofs:
+		return "[]PoStProof"
 	default:
 		return "<unknown type>"
 	}
@@ -115,6 +120,8 @@ func (av *Value) String() string {
 		return fmt.Sprint(av.Val.(uint64))
 	case CommitmentsMap:
 		return fmt.Sprint(av.Val.(map[string]types.Commitments))
+	case PoStProofs:
+		return fmt.Sprint(av.Val.([]proofs.PoStProof))
 	default:
 		return "<unknown type>"
 	}
@@ -214,6 +221,13 @@ func (av *Value) Serialize() ([]byte, error) {
 		}
 
 		return cbor.DumpObject(m)
+	case PoStProofs:
+		m, ok := av.Val.([]proofs.PoStProof)
+		if !ok {
+			return nil, &typeError{[]proofs.PoStProof{}, av.Val}
+		}
+
+		return cbor.DumpObject(m)
 	default:
 		return nil, fmt.Errorf("unrecognized Type: %d", av.Type)
 	}
@@ -253,6 +267,8 @@ func ToValues(i []interface{}) ([]*Value, error) {
 			out = append(out, &Value{Type: SectorID, Val: v})
 		case map[string]types.Commitments:
 			out = append(out, &Value{Type: CommitmentsMap, Val: v})
+		case []proofs.PoStProof:
+			out = append(out, &Value{Type: PoStProofs, Val: v})
 		default:
 			return nil, fmt.Errorf("unsupported type: %T", v)
 		}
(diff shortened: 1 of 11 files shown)


proofs/sectorbuilder/interface.go
```diff
@@ -47,11 +47,11 @@ type SectorBuilder interface {
 	// which will fit into a newly-provisioned staged sector.
 	GetMaxUserBytesPerStagedSector() (uint64, error)
 
-	// GeneratePoST creates a proof-of-spacetime for the replicas managed by
+	// GeneratePoSt creates a proof-of-spacetime for the replicas managed by
 	// the SectorBuilder. Its output includes the proof-of-spacetime proof which
 	// is posted to the blockchain along with any faults. The proof can be
 	// verified by the VerifyPoSt method on the Verifier interface.
-	GeneratePoST(GeneratePoSTRequest) (GeneratePoSTResponse, error)
+	GeneratePoSt(GeneratePoStRequest) (GeneratePoStResponse, error)
```
phritz: new casing is less ELiTE

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -321,25 +321,40 @@ func (sb *RustSectorBuilder) GeneratePoST(req GeneratePoSTRequest) (GeneratePoST
...
 	}, nil
 }
 
+// goPoStProofs accepts a pointer to a C-allocated byte array and a size and
+// produces a Go-managed slice of PoStProof. Note that this function copies
+// values into the Go heap from C.
+func goPoStProofs(src *C.uint8_t, size C.size_t) []proofs.PoStProof {
+	chunkSize := int(proofs.PoStBytesLen)
+
+	out := make([]proofs.PoStProof, int(size)/chunkSize)
```
phritz: is it more safe to verify that size is an integral number of chunksizes?

phritz: do we care if src is nil but size is not zero?

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -321,25 +321,40 @@ func (sb *RustSectorBuilder) GeneratePoST(req GeneratePoSTRequest) (GeneratePoST
...
+	tmp := make([]byte, size)
+
+	if src != nil {
+		copy(tmp, (*(*[1 << 30]byte)(unsafe.Pointer(src)))[:size:size])
```
phritz: my head just exploded. can you comment this?

abi/abi.go
```diff
@@ -76,6 +79,8 @@ func (t Type) String() string {
 		return "uint64"
 	case CommitmentsMap:
 		return "map[string]Commitments"
+	case PoStProofs:
+		return "[]PoStProof"
```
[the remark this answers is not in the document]
laser: Hrm, I don't know why we do that. We certainly don't do it consistently (see `Commitments`, above). I'll change both `Commitments` and `PoStProof` such that they include a package qualifier.

proofs/sectorbuilder/interface.go
laser: Had to do what I had to do to [conform to the spec](https://github.com/filecoin-project/specs/blob/master/drafts/proof-of-spacetime.md#proof-of-spacetime).

proofs/sectorbuilder/rustsectorbuilder.go
laser: I don't, no. If `src` is `nil`, `goPoStProofs` will produce an empty slice.

laser: If size % PoSt proof size != 0 because of a bug in rust-fil-proofs, the miner will get slashed (and programmers will become very confused). I like your recommendation to enforce size invariant at runtime.

laser: What I just said was a lie; as written, when `src` is `nil` but `size` > `chunkSize`, the returned slice will be non-empty. That's a bug which I will fix.

proofs/sectorbuilder/rustsectorbuilder.go
laser: Just pushed comment. Check out [this link](https://github.com/golang/go/wiki/cgo#turning-c-arrays-into-go-slices) for more details.

abi/abi.go
```diff
@@ -357,6 +372,15 @@ func Deserialize(data []byte, t Type) (*Value, error) {
 			Type: t,
 			Val:  m,
 		}, nil
+	case PoStProofs:
+		var slice []proofs.PoStProof
+		if err := cbor.DecodeInto(data, &slice); err != nil {
```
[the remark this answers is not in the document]
laser: I don't understand what you're asking for. Are you asking that I propose a modification to `data-structures.md` in the specs repo to include the `proofs.PoStProof` type? 

I do not understand how the PR which you linked to interacts with the code in this PR.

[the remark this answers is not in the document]
phritz: that spec says how things in filecoin are to be cbor encoded. this type you are cbor encoding here, postproof, is not in the list of things that that spec says how to cbor encode. it seems like it should be to avoid ambiguity.

proofs/sectorbuilder/rustsectorbuilder.go
phritz: phritz: 1, laser: 0

abi/abi.go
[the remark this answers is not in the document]
whyrusleeping: @laser would be great to have this type in the data-structures spec doc :)

PR #2186, opened 2019-03-06
Feat/address from spec

frrist:
## Description
This PR implements addresses based on the [address spec](https://github.com/filecoin-project/specs/blob/master/address.md).

## Whats left before merging
- [x] Remove debug logs, they exists now to aid in debugging test failures during transition.
- [x]  Remove `allTheTests` script, exists now as go test output is a pita sometimes.
- [x] Address DoNotMerge Comments
- [x] Should address.String() panic or return an error


abi/abi.go
```diff
@@ -295,14 +295,9 @@ func FromValues(vals []*Value) []interface{} {
...
 	case Address:
-		addr, err := address.NewFromBytes(data)
-		if err != nil {
-			return nil, err
-		}
-
 		return &Value{
 			Type: t,
-			Val:  addr,
+			Val:  address.NewFromBytes(data),
```
whyrusleeping: NewFromBytes should probably still return an error. This is parsing user input, panicking is not a good thing

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-	if err != nil {
-		return err
-	}
-
-	copy(a[:], out[:])
-
-	return nil
+// NewFromBytes return the address represented by the bytes `addr`.
+func NewFromBytes(addr []byte) Address {
+	return Address{string(addr)}
```
whyrusleeping: definitely need some validation here, esp if this is coming from potentially untrusted sources.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+		cksm := Checksum(append([]byte{addr.Protocol()}, addr.Payload()...))
+		strAddr = ntwk + fmt.Sprintf("%d", addr.Protocol()) + AddressEncoding.WithPadding(-1).EncodeToString(append(addr.Payload(), cksm[:]...))
+	case ID:
+		strAddr = ntwk + fmt.Sprintf("%d", addr.Protocol()) + fmt.Sprintf("%d", leb128.ToUInt64(addr.Payload()))
+	default:
+		return "", errors.New("invalid protocol byte")
 	}
-
-	return hrp + string(encoded), nil
+	log.Debugf("encoded address: %s", strAddr)
```
whyrusleeping: spurious debug, remove

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-			return "", 0, nil, fmt.Errorf("invalid character")
-		}
+	if string(a[0]) != MainnetPrefix && string(a[0]) != TestnetPrefix {
+		return Undef, errors.New("invalid network prefix")
+	}
 
-		dataBytes = append(dataBytes, byte(Base32CharsetReverse[int(b)]))
+	var protocol Protocol
+	switch a[1] {
+	case 48:
```
whyrusleeping: to make this less confusing, i would do:

```go
case '0':
...
case '1':
...
```

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-// decode decodes a filecoin address and returns
-// the hrp(human-readable part), version (1byte) and data(32bit data array)
-// or an error.
-func decode(addr string) (string, byte, []byte, error) {
-	if len(addr) < 2 {
-		return "", 0, nil, fmt.Errorf("address too short")
-	}
-	if len(addr) > 41 {
-		return "", 0, nil, fmt.Errorf("too long: len=%d", len(addr))
+func decode(a string) (Address, error) {
```
whyrusleeping: should also have a check for max length.

address/address.go
```diff
@@ -1,41 +1,71 @@
...
-func MakeTestAddress(input string) Address {
-	hash := Hash([]byte(input))
+var addressAtlasEntry = atlas.BuildEntry(Address{}).Transform().
+	TransformMarshal(atlas.MakeMarshalTransformFunc(
+		func(a Address) ([]byte, error) {
+			return []byte(a.str), nil
+		})).
+	TransformUnmarshal(atlas.MakeUnmarshalTransformFunc(
+		func(x []byte) (Address, error) {
+			return Address{string(x)}, nil
```
whyrusleeping: This should be calling NewFromBytes, so that validation checks happen. Otherwise untrusted cbor input could give us potentially bad addresses.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+// String returns an address encoded as a string.
+func (a Address) String() string {
+	str, err := encode(Mainnet, a)
 	if err != nil {
-		return Address{}, err
-	}
-
-	if version != Version {
-		return Address{}, ErrUnknownVersion
+		panic(err)
```
whyrusleeping: This panic is acceptable as long as we can be confident that no address we ever create will trigger it. This panic being hit should be a result of programmer error, and never happen because of anything a user typed or was sent over the network

porcelain/payments.go
```diff
@@ -76,10 +76,10 @@ type CreatePaymentsReturn struct {
 // CreatePayments establishes a payment channel and create multiple payments against it
 func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePaymentsParams) (*CreatePaymentsReturn, error) {
 	// validate
-	if config.From.Empty() {
+	if config.From.Equal(address.Undef) {
```
whyrusleeping: We shouldnt need an `Equal` method here, the `==` operation should work fine

address/address_test.go
```diff
@@ -1,159 +1,104 @@
...
+	logging.SetDebugLogging()
+	rand.Seed(time.Now().Unix())
 }
 
-func TestEmptyAddress(t *testing.T) {
-	assert := assert.New(t)
-	assert.True((Address{}).Empty())
-	assert.False(NewMainnet(hashes[0]).Empty())
-}
+func TestIDAddress(t *testing.T) {
```
whyrusleeping: Let's get some test vectors in here. I want to see a few for ID addresses, like what is the address for ID 0, 1, 10, 150, 999999, and so on. I would expect them to be `f00`, `f01`, `f010`, etc, but let's make it very clear with tests.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-	if version > 31 {
-		return "", fmt.Errorf("version too large")
+	var ntwk string
+	switch network {
+	case Mainnet:
+		ntwk = MainnetPrefix
+	case Testnet:
+		ntwk = TestnetPrefix
+	default:
+		panic("invalid network byte")
```
acruikshank: Why not an error?

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-			return "", fmt.Errorf("invalid character human-readable part : hrp[%d]=%d", p, c)
-		}
+	log.Debugf("new address protocol: %x, payload: %v", protocol, payload)
+	return Address{string(buf)}
+}
+
+func encode(network Network, addr Address) (string, error) {
+	// TODO(frrist): not sure if this is the correct thing to do here
+	if addr == Undef {
+		return "", nil
```
acruikshank: Returning something like "Undef" or "[empty]" will make debug strings with empty actors easier to decode.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
 
-	if !verifyChecksum(hrp, dataBytes) {
-		return "", 0, nil, fmt.Errorf("invalid checksum")
+	raw := a[2:]
+	if protocol == ID {
+		id, err := strconv.ParseUint(raw, 10, 64)
+		if err != nil {
+			return Undef, errors.New("invalid ID payload")
+		}
+		return newAddress(protocol, leb128.FromUInt64(id)), nil
```
whyrusleeping: we probably shouldnt leb128 encode the bytes form for IDs. Its just taking up extra space (1 bit per byte) and doesnt actually get us anything. Addresses are always delimited in some way anyways. We should just use a standard little endian encoding: https://golang.org/pkg/encoding/binary/#pkg-variables

I realize I didnt catch this in the spec. I'll submit a PR to address this once we have some consensus here.

cc @dignifiedquire

dignifiedquire: I am not sure I agree, all other address formats have a fixed size, except the id version. Not having varints in there could make it tricky to nicely pack addresses in some places I would imagine.

dignifiedquire: For example, with the current impl I can give you a byte array of n addresses, without specifying anything else, and I can parse it in exactly one way (matching protocols, and then determining the length from that) removing the varints this ability gets lost

whyrusleeping: @dignifiedquire True, half of me says "we would never do that", the other half says "but we could".

whyrusleeping: Alright, I'm convinced. It makes sense to me to keep leb128 here

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-	case "fc":
-		return Mainnet, nil
-	case "tf":
-		return Testnet, nil
-	default:
-		return 0, ErrUnknownNetwork
-	}
-}
+// TestnetPrefix is the main network prefix.
+var TestnetPrefix = "t"
```
anorth: These should be consts.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-		return Address{}, ErrInvalidBytes
-	}
-
-	network := raw[0]
-	if network > Testnet {
-		return Address{}, ErrUnknownNetwork
-	}
+// Equal returns true if address `b` is equal to address.
+func (a Address) Equal(b Address) bool {
+	return bytes.Equal(a.Bytes(), b.Bytes())
```
anorth: Elsewhere @whyrusleeping suggested this is unnecessary and we can use `==` (which should be documented)

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+// NewSecp256k1Address returns an address using the SECP256K1 protocol.
+func NewSecp256k1Address(pubkey []byte) Address {
+	return newAddress(SECP256K1, addressHash(pubkey))
 }
 
-// Hash returns the hash part of the address.
-func (a Address) Hash() []byte {
-	return a[2:]
+// NewActorAddress returns an address using the Actor protocol.
+func NewActorAddress(data []byte) Address {
```
anorth: Are there any callers that don't start with a string here? Perhaps just define it that way.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+
+	buf[0] = protocol
+	if c := copy(buf[1:], payload); c != len(payload) {
+		panic("copy data length is inconsistent")
 	}
-	for p, c := range hrp {
-		if c < 33 || c > 126 {
-			return "", fmt.Errorf("invalid character human-readable part : hrp[%d]=%d", p, c)
-		}
+	log.Debugf("new address protocol: %x, payload: %v", protocol, payload)
```
anorth: This log is probably excessive.

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+	return Address{string(buf)}
+}
+
+func encode(network Network, addr Address) (string, error) {
+	// TODO(frrist): not sure if this is the correct thing to do here
```
anorth: Don't merge with this TODO still here. I suggest in general using a different tag like DONOTMERGE or similar to help you and reviewers address questions that you ask via the code.

commands/allTheTest.sh
```diff
@@ -0,0 +1,12 @@
...
+#ldirs=$(find . -name '*_test.go' | xargs dirname | sort | uniq)
+tests=$(find . -name '*_test.go' | xargs cat | grep '^func Test' | awk -F '[ (]' '{ print $2 }')
+
+for t in $tests; do
+  echo "=== $t"
+  #go clean -testcache
+  #go build -o go-filecoin main.go
+
+  go test -parallel=1  -run $t
+done
```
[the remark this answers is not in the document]
travisperson: Random file that was added by accident lol

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
-		// alphanumeric only
-		if !((b >= int32('0') && b <= int32('9')) || (b >= int32('A') && b <= int32('Z')) || (b >= int32('a') && b <= int32('z'))) {
-			return "", 0, nil, fmt.Errorf("non alphanumeric character")
-		}
 
-		// exclude [1,b,i,o]
-		if b == int32('1') || b == int32('b') || b == int32('i') || b == int32('o') {
-			return "", 0, nil, fmt.Errorf("invalid character")
-		}
+	if string(a[0]) != MainnetPrefix && string(a[0]) != TestnetPrefix {
```
acruikshank: I don't like the way we're half supporting networks. 

1. Testnet and Mainnet addresses decode to the same internal representation of an address.
2. We have no way of validating that an address encoded with a network can only be used in that network.
3. All addresses are encoded to Mainnet addresses.

This PR could: 

1. Add proper support for networks meaning that the node would be configured somehow with the right network (ideally in the genesis block somehow), use that prefix for encoding, and only permit that prefix in decoding. Also consider the implications of transmitting cbor encoded addresses that drop network information.
2. Table the above and require (and assume) all addresses to be Testnet addresses, since Testnet is our first target.
3. Table the concept of encoding the network at all. All addresses begin with 'f'. This makes it easier to distinguish our addresses from other cryptocurrencies. It provides no information otherwise.

@whyrusleeping ?

address/address_test.go
```diff
@@ -1,159 +1,104 @@
...
-			assert := assert.New(t)
-
-			a, err := NewFromBytes(tc.input.Bytes())
-			assert.NoError(err)
-			assert.Equal(a, tc.input)
-		})
-	}
 }
 
-func TestInvalidAddressCreation(t *testing.T) {
```
acruikshank: This test is still useful for secp256kAddresses. In general, test coverage has gone way down with this PR (more branches, fewer tests).

address/constants.go
```diff
@@ -1,62 +1,37 @@
...
-}
-
-// Base32 is a basex instance using the Base32Charset.
-var Base32 = basex.NewAlphabet(Base32Charset)
+	"gx/ipfs/QmZp3eKdYQHHAneECmeK6HhiMwTPufmjC8DuuaGKv3unvx/blake2b-simd"
+)
 
 var (
-	// TestAddress is an account with some initial funds in it
+	// TODO remove TestAddresses
```
acruikshank: TODO should explain why we want to remove TestAddresses. It should also explain why we're not doing it in this PR.

commands/failures
```diff
@@ -0,0 +1,3 @@
+TestMinerCreate
```
acruikshank: Are these notes? Probably shouldn't be here.

commands/allTheTest.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
acruikshank: This package isn't an appropriate place for scripts. If this is worth sharing, it should be documented and go in scripts. Otherwise it should stay out of the repo.

plumbing/msg/sender_test.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/testhelpers"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/wallet"
-	"github.com/pkg/errors"
+	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
```
acruikshank: Belongs with gx packages.

vm/context.go
```diff
@@ -194,16 +194,15 @@ func computeActorAddress(creator address.Address, nonce uint64) (address.Address
...
 	if err := binary.Write(buf, binary.BigEndian, nonce); err != nil {
-		return address.Address{}, err
+		return address.Undef, err
 	}
 
-	hash := address.Hash(buf.Bytes())
-
-	return address.NewMainnet(hash), nil
+	return address.NewActorAddress(buf.Bytes()), nil
+	//return address.NewFromBytes(buf.Bytes()), nil
```
acruikshank: rm

address/address.go
```diff
@@ -47,327 +77,236 @@ const (
...
+		panic("invalid protocol")
 	}
-	if len(data) != HashLength {
-		return "", fmt.Errorf("data is malformed: data length=%d", len(data))
+	explen := 1 + len(payload)
+	buf := make([]byte, explen)
+
+	buf[0] = protocol
+	if c := copy(buf[1:], payload); c != len(payload) {
+		panic("copy data length is inconsistent")
```
acruikshank: Checking c isn't necessary. It's only checking your math and is always true or always false.

address/address.go
frrist: agreed, will address.

mining/mqueue.go
```diff
@@ -99,7 +98,7 @@ func (pq queueHeap) Less(i, j int) bool {
 		return delta.GreaterThan(types.ZeroAttoFIL)
 	}
 	// Secondarily order by address to give a stable ordering.
-	return bytes.Compare(pq[i][0].From[:], pq[j][0].From[:]) < 0
+	return pq[i][0].From.Equal(pq[j][0].From)
```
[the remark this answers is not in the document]
frrist: ohh you are right, nice catch.

commands/failures
frrist: Yes sorry, these will be removed.

commands/allTheTest.sh
frrist: This was not supposed to be pushed, was helpful to debug some failure, will be removing

whyrusleeping: git add . is bad for you.

whyrusleeping: `git alias au="add -u"`. so much better

address/address.go
whyrusleeping: option 2 seems like the right approach if we want to keep the distinction. option 3 is what i would personally push for. I still don't think the difference between testnet and mainnet addressses makes that much of a difference.

frrist: I have opened an issue to track this question here: https://github.com/filecoin-project/go-filecoin/issues/2220. For now I will go with option 2 in this PR.

consensus/power_table_view.go
```diff
@@ -63,7 +63,7 @@ func (v *MarketView) Total(ctx context.Context, st state.Tree, bstore blockstore
 // should probably be increased for v1.
 func (v *MarketView) Miner(ctx context.Context, st state.Tree, bstore blockstore.Blockstore, mAddr address.Address) (uint64, error) {
 	vms := vm.NewStorageMap(bstore)
-	rets, ec, err := CallQueryMethod(ctx, st, vms, mAddr, "getPower", []byte{}, address.Address{}, nil)
+	rets, ec, err := CallQueryMethod(ctx, st, vms, mAddr, "getPower", []byte{}, address.Undef, nil)
```
frrist: Methods like this are a reason I think we should allow decoding a encoding and decoding a zero length address -- since we are using one here. Maybe I am misunderstanding the usage here?

address/address.go
```diff
@@ -47,327 +68,268 @@ const (
...
+	return newAddress(SECP256K1, addressHash(pubkey))
+}
 
-	out, err := NewFromString(string(in))
-	if err != nil {
-		return err
-	}
+// NewActorAddress returns an address using the Actor protocol.
+func NewActorAddress(data []byte) (Address, error) {
+	return newAddress(Actor, addressHash(data))
```
whyrusleeping: check length of input

address/address.go
```diff
@@ -47,327 +68,268 @@ const (
...
+}
 
-	copy(a[:], out[:])
+// NewBLSAddress returns an address using the BLS protocol.
+func NewBLSAddress(pubkey []byte) (Address, error) {
+	return newAddress(BLS, pubkey)
```
whyrusleeping: check length of input (more important here than on the other ones, but still good practice)

address/address.go
```diff
@@ -47,327 +68,268 @@ const (
...
+	default:
+		return "", ErrUnknownProtocol
 	}
+	return strAddr, nil
+}
 
-	if !verifyChecksum(hrp, dataBytes) {
-		return "", 0, nil, fmt.Errorf("invalid checksum")
+func decode(a string) (Address, error) {
+	// DONOTMERGE: should decoding a zero length address error?
```
whyrusleeping: donotmerge

consensus/power_table_view.go
whyrusleeping: the address used here shouldnt ever actually get serialized.

address/address.go
frrist: length check is done in `newAddress`: https://github.com/filecoin-project/go-filecoin/pull/2186/files#diff-2026d132a1aa6ec72a1af72e5763011cR218.

address/address.go
frrist: Length check is done in `newAddress`: https://github.com/filecoin-project/go-filecoin/pull/2186/files#diff-2026d132a1aa6ec72a1af72e5763011cR218

address/address.go
frrist: addressed here: https://github.com/filecoin-project/go-filecoin/pull/2186/files#diff-2026d132a1aa6ec72a1af72e5763011cR272

address/address.go
frrist: addressed

address/address.go
frrist: addressed

consensus/power_table_view.go
frrist: The message created by this call is serialized which causes the addresses to be serialized along with it.

address/address.go
frrist: I am fairly confident that this panic will only be caused by a programmer error. `encode()` will will return an error in 2 cases:
1. We pass it an unknown network prefix, this would be a programmer error
2. We pass it an address with an unknown protocol, `decode` will and `NewFromBytes` will prevent addresses from being created with invalid protocols.

It would be great to have fuzz tests on addresses to further guarantee this.

address/address.go
frrist: If addresses were fixed length arrays this would be true, but since they are variable length slices I believe `bytes.Equal` is required.

address/address.go
frrist: addressed

address/address.go
frrist: addressed

frrist:
@anorth @acruikshank @dignifiedquire @whyrusleeping this is ready for a second review at your convenience.

proofs/sectorbuilder/utils.go
```diff
@@ -9,7 +9,12 @@ import (
 
 // AddressToProverID creates a prover id by padding an address hash to 31 bytes
 func AddressToProverID(addr address.Address) [31]byte {
-	hash := addr.Hash()
+	// this code will no longer WAE when ID's or BLS pub keys are added
+	// as they will break the assumption of addresses all being the same length
+	if addr.Protocol() != address.Actor && addr.Protocol() != address.SECP256K1 {
+		panic("cannot create prover hash from new address protocol")
```
acruikshank: BLOCKING: This code does not assume addresses are all the same length (or even the desired length). This panic should be removed. If we do need that assurance, the check should be where we do make that assumption.

address/address.go
```diff
@@ -1,41 +1,62 @@
...
+|----------|---------|
+| protocol | payload |
+|----------|---------|
+|  1 byte  | n bytes |
+
+String:
+|------------|----------|---------|----------|
+|  network   | protocol | payload | checksum |
+|------------|----------|---------|----------|
+| 'f' or 't' |  1 byte  | n bytes | 4 bytes  |
```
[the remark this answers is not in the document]
frrist: I will remove this comment block and let readers refer to the spec.

address/address.go
```diff
@@ -47,327 +68,267 @@ const (
...
-	}
-
-	version := raw[1]
-	if version != Version {
-		return Address{}, ErrUnknownVersion
-	}
-
-	return New(network, raw[2:]), nil
+// Equal returns true if address `b` is equal to address.
+func (a Address) Equal(b Address) bool {
```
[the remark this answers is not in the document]
frrist: You are right, `==` is sufficient everywhere, removing.

address/address.go
```diff
@@ -47,327 +68,267 @@ const (
...
+		protocol = BLS
+	default:
+		return Undef, ErrUnknownProtocol
 	}
 
-	if !verifyChecksum(hrp, dataBytes) {
-		return "", 0, nil, fmt.Errorf("invalid checksum")
+	raw := a[2:]
+	if protocol == ID {
+		id, err := strconv.ParseUint(raw, 10, 64)
```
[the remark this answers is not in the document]
frrist: I will add a check specifically for `len(raw) > 20` (since 20 is length of math.MaxUint64 as a string), although I believe the `strconv.ParseUint` check handles this for us.

proofs/sectorbuilder/utils.go
frrist: I believe [this line will panic]( https://github.com/filecoin-project/go-filecoin/pull/2186/files/480b2a05b80a56b5e81399fb66aa64cdc622f7e0#diff-ab53d6373501c86e59a75d6590d62e7eR23 ) for Protocol 3 addresses since the length of a [BLS public key is 48 bytes](https://github.com/filecoin-project/go-filecoin/blob/master/bls-signatures/types.go#L10).

frrist: incase the 'this line will panic' link doesn't load for you (it doesn't for me) look at line 26 below.

frrist: More context, the current address code always returns a slice of [length 20](https://github.com/filecoin-project/go-filecoin/blob/master/address/constants.go#L8) for `addr.Hash()`, since `addr.Payload()` is replacing that call here, and since a BLS public key can be the value of payload, this breaks the assumptions made by this code.

address/address.go
frrist: This makes me sad, I really don't like returning an error from a `String()` method. I wrote a quick fuzz test for this, and it does indeed panic in our leb128 library, this means returning an error here won't help us. This is the first time I have tried fuzz testing, so I may be doing something wrong here -- the commit with the change is: 82820b656e437bd796967699cf167f4623c62434, failure should be in CI shortly.

phritz: you really really really should not be panicing on a codepath that could possibly be a function of user input. far better to deal with ugly return signatures or to return a known-bad string or something than to panic. seriously, don't do this, it stops the node from being able to do anything useless and causes the node to exit in an unknown state so you may not ever be able to start your node again. https://github.com/filecoin-project/go-filecoin/blob/master/CONTRIBUTING.md#error-handling

frrist: Alright, I will return an error. It can be removed later in a follow on if we don't like it (I don't like it, but I like it more than rebasing this branch)

address/constants.go
```diff
@@ -1,62 +1,87 @@
...
+	// ErrInvalidLength is returned when encountering an address of invalid length.
+	ErrInvalidLength = errors.New("invalid address length")
+	// ErrInvalidChecksum is returned when encountering an invalid address checksum.
+	ErrInvalidChecksum = errors.New("invalid address checksum")
+)
 
-	t = Hash([]byte("nakamoto"))
-	TestAddress2 = NewMainnet(t)
+// UndefAddressString is the string used to represent an empty address when encoded to a string.
+var UndefAddressString = "[UNDEFINED_ADDRESS]"
```
anorth: I suggest consistently using "empty" (preferred) or "undef" to refer to the non-address.

frrist:
Before I merge this I would like to point out that corrupt addresses using the ID protocol could panic, this will occur until https://github.com/filecoin-project/go-leb128/issues/9 is addressed and the calling code here is updated.

address/address.go
whyrusleeping: This is already what we do for Cids. More worrisome than panicking here is allowing invalid address objects to be created and propogated around the codebase, being sent to other nodes over the network, or even making it into the chain somehow. We should have strict confidence that anything parsing untrusted input catches all cases.

Let's merge this (aside from this discussion, i think we all agree it looks good) and open an issue to fuzz our inputs, for addresses, and for any other places we get random data from places.

anorth:
@frrist the commits in this PR were not squashed before merging, but should have been. I noticed because I attempted to revert it in debugging a test failure on master. From the commit messages, it seems doubtful that all commits have passing builds, so would also have made a bisect very difficult.

(The test failure turned out to be from a different PR).

PR #2193, opened 2019-03-07
Allow actors with undefined code to send messages.

### LINKED ISSUE #2184: [bug] Failed to deploy nightly devnet
**Describe the bug**

The command which exited to fail the deployment was
```
++ docker exec filecoin-1 go-filecoin --repodir=/var/local/filecoin miner create --price 0.000000001 --limit 300 10 10
Error: Could not create miner. Please consult the documentation to setup your wallet and genesis block correctly: invalid message: message from non-account actor
```

The only suspicious thing I've been able to find so far is a warning from badger

```
+ docker run -v /home/ubuntu/car:/var/filecoin/car -v /mnt/storage/filecoins:/var/local/filecoins --entrypoint=/usr/local/bin/go-filecoin 657871693752.dkr.ecr.us-east-1.amazonaws.com/filecoin:latest init --genesisfile=/var/filecoin/car/genesis.car --repodir=/var/local/filecoins/1 --peerkeyfile=/var/filecoin/car/keyFile1 --auto-seal-interval-seconds 0
initializing filecoin node at /var/local/filecoins/1
badger 2019/03/06 06:21:11 INFO: All 0 tables opened in 0s
badger 2019/03/06 06:21:11 INFO: All 0 tables opened in 0s
badger 2019/03/06 06:21:11 INFO: All 0 tables opened in 0s
badger 2019/03/06 06:21:11 INFO: All 0 tables opened in 0s
badger 2019/03/06 06:21:11 INFO: Storing value log head: {Fid:0 Len:43 Offset:3815}
badger 2019/03/06 06:21:11 INFO: Force compaction on level 0 done
badger 2019/03/06 06:21:11 INFO: Storing value log head: {Fid:0 Len:42 Offset:134}
badger 2019/03/06 06:21:11 INFO: Force compaction on level 0 done
badger 2019/03/06 06:21:11 INFO: Storing value log head: {Fid:0 Len:42 Offset:803}
badger 2019/03/06 06:21:11 INFO: Force compaction on level 0 done
badger 2019/03/06 06:21:11 WARNING: While forcing compaction on level 0: Unable to fill tables
```

@rkowalick also saw these same warnings when he upgrade to golang 1.12. We build the docker container images using [`circleci/golang:latest`](https://github.com/filecoin-project/go-filecoin/blob/master/.circleci/config.yml#L322), so I'm wondering if this is the same issue.

https://gist.github.com/travisperson/a77d38cb7958b610beeef9544b65bd51

anorth:
An empty actor may be initialised by sending FIL to an address. The actor is not recognized as
an account actor until a message is sent from it's address, at which point it is transparently
upgraded in processing. This change updates the validation logic to reflect that.

I re-ordered the validation and upgrade blocks in the processor to exercise the code. This PR is unfortunately not accompanied by a regression test, which is a bit difficult to construct. I intend to follow up with some plumbing such that an in-process integration test (c.f. `message_propagate_test`) can mine a block and we can test the sequence of sending FIL to a new address, then a message from that address.

Fixes #2184.

### DIFF
--- a/consensus/processor.go
+++ b/consensus/processor.go
@@ -439,20 +439,20 @@ func (p *DefaultProcessor) attemptApplyMessage(ctx context.Context, st *state.Ca
 		return nil, errors.FaultErrorWrapf(err, "failed to get From actor %s", msg.From)
 	}
 
-	// processing an external message from an empty actor upgrades it to an account actor.
-	if !fromActor.Code.Defined() {
-		err := account.UpgradeActor(fromActor)
-		if err != nil {
-			return nil, errors.FaultErrorWrap(err, "failed to upgrade empty actor")
-		}
-	}
-
 	err = p.signedMessageValidator.Validate(ctx, msg, fromActor)
 	if err != nil {
 		return &types.MessageReceipt{
 			ExitCode:   errors.CodeError(err),
 			GasAttoFIL: types.ZeroAttoFIL,
 		}, err
+	}
+
+	// Processing an external message from an empty actor upgrades it to an account actor.
+	if !fromActor.Code.Defined() {
+		err := account.UpgradeActor(fromActor)
+		if err != nil {
+			return nil, errors.FaultErrorWrap(err, "failed to upgrade empty actor")
+		}
 	}
 
 	toActor, err := st.GetOrCreateActor(ctx, msg.To, func() (*actor.Actor, error) {
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -44,12 +44,13 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
 		return errSelfSend
 	}
 
-	// sender must be an account actor.
-	if !fromActor.Code.Equals(types.AccountActorCodeCid) {
+	// Sender must be an account actor, or a undefined actor which will be upgraded to an account actor
+	// when the message is processed.
+	if fromActor.Code.Defined() && !fromActor.Code.Equals(types.AccountActorCodeCid) {
 		return errNonAccountActor
 	}
 
-	// avoid processing messages for actors that cannot pay.
+	// Avoid processing messages for actors that cannot pay.
 	if !canCoverGasLimit(msg, fromActor) {
 		log.Info("Insufficient funds to cover gas limit: ", fromActor, msg)
 		return errInsufficientGas
--- a/actor/actor.go
+++ b/actor/actor.go
@@ -31,7 +31,8 @@ func init() {
 //
 // Not safe for concurrent access.
 type Actor struct {
-	// Code is a CID of the VM code for this actor's implementation (or a constant for actors implemented in Go code)
+	// Code is a CID of the VM code for this actor's implementation (or a constant for actors implemented in Go code).
+	// Code may be nil for an uninitialized actor (which exists because it has received a balance).
 	Code cid.Cid `refmt:",omitempty"`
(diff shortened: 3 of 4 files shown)


consensus/processor.go
```diff
@@ -455,6 +447,14 @@ func (p *DefaultProcessor) attemptApplyMessage(ctx context.Context, st *state.Ca
...
 	}
 
+	// Processing an external message from an empty actor upgrades it to an account actor.
+	if !fromActor.Code.Defined() {
+		err := account.UpgradeActor(fromActor)
+		if err != nil {
+			return nil, errors.FaultErrorWrap(err, "failed to upgrade empty actor")
+		}
+	}
+
```
travisperson: What is the reasoning for moving this code?

whyrusleeping:
Note that once the new address code lands, we don’t have to worry about any of this. The whole undefined actor upgrade thing goes away completely as we no longer have ambiguity around the key/actor account distinction.

whyrusleeping:
(that is to say, don’t spend too much time worrying about testing this)

consensus/processor.go
```diff
@@ -455,6 +447,14 @@ func (p *DefaultProcessor) attemptApplyMessage(ctx context.Context, st *state.Ca
 		}, err
 	}
 
+	// Processing an external message from an empty actor upgrades it to an account actor.
+	if !fromActor.Code.Defined() {
```
acruikshank: I should have done this when I originally wrote this code, but it would be great to have an isEmpty() method on actor (or a free function) rather than relying on comments to define the concept of an empty actor.

consensus/validation.go
```diff
@@ -44,12 +44,13 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
 		return errSelfSend
 	}
 
-	// sender must be an account actor.
-	if !fromActor.Code.Equals(types.AccountActorCodeCid) {
+	// Sender must be an account actor, or a undefined actor which will be upgraded to an account actor
```
acruikshank: It's an "empty" actor, not "undefined".

travisperson:
> In the old code, an undefined Cid would force an upgrade before validation, so validation wouldn't fail. In the new code, validation ignores undefined Cid's and always upgrades if validation succeeds. In either case, a validation failure rolls back changes, so no upgrade will occur.

As I understand it, the issue was that we added a validation check (in #2139) on the sender side, before the message is sent to the network. The upgrade only occurs on the receiver side. This meant that a node could not send a message from an account actor, as it would never validate, and therefor could not be upgrade on the receivers end. The goal of the fix is to allow for empty actors to sent, and rely on the validation of the receiver side.

I think this is fine, because we can't control what is sent to the network anyways.

Now that I look at this again though, it appears that this validate is the same for both sides. I think we really want to only have this skip on the sender side, and require non-empty actors on the receiving side. This is why I'm questioning the move of the upgrade code below validation.

@acruikshank

acruikshank:
Oh, I see now. When we validate on the sending side, we now accept empty actors. None of the sending code was touched by this PR except the shared validation logic. This makes sense and I fully approve.

anorth:
Thanks. Sorry that the code move in `processor.go` wasn't sufficiently explained. It doesn't change functionality, but it means that both sender and receiver are validating against the same state (with an empty actor), so it exercises the validation code more thoroughly (in lieu of a solid test). The goal is to have the validation be as "same" as possible between the two cases, rather than skip some things in some case.

Issue #2177, opened 2019-03-05
Add DHT query subcommand (Copy from IPFS)


ZenGround0:
### Description
It will be useful for debugging purposes (for example to make sure that DHT bootstrapping is wai) to include this subcommand.  Copy over [this](https://github.com/ipfs/go-ipfs/blob/master/core/commands/dht.go#L50) ipfs command to filecoin.  For reference on how to do this see PR #2158.  This should be even simpler because the porcelain api is already set up with a router and all the command helpers are already in place.
 
Good first issue for a go-ipfs contributor.  Not a bad first issue for an experienced dev looking at filecoin for the first time.

### Where to begin
1. copy code from IPFS
2. modify code to call into router from go-filecoin porcelain API instead of ipfs node
3. add a `GetClosestPeers` call to filnet/router, inside that method cast the wrapped interface and error if routing isn't a kad dht.  Then pipe the call through
4. add a `NetworkGetCLosesPeers` call to plumbing/api.go calling `GetClosestPeers` on the ntwk struct.

firmianavan:
#2158 has not been merged yet

PR #2200, opened 2019-03-07
Add DHT query subcommand to get closest peers

firmianavan:
Simply wrap the same subcommand in IPFS as mentioned in #2177
Please check if there is anything that may need changing.

### DIFF
--- a/commands/dht.go
+++ b/commands/dht.go
@@ -30,7 +30,71 @@ var dhtCmd = &cmds.Command{
 	Subcommands: map[string]*cmds.Command{
 		"findprovs": findProvidersDhtCmd,
 		"findpeer":  findPeerDhtCmd,
+		"query":     queryDhtCmd,
 	},
+}
+
+var queryDhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Find the closest Peer IDs to a given Peer ID by querying the DHT.",
+		ShortDescription: "Outputs a list of newline-delimited Peer IDs.",
+	},
+
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("peerID", true, false, "The peerID to run the query against."),
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption(dhtVerboseOptionName, "v", "Print extra information."),
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+
+		id, err := peer.IDB58Decode(req.Arguments[0])
+		if err != nil {
+			return cmds.ClientError("invalid peer ID")
+		}
+
+		ctx, cancel := context.WithCancel(req.Context)
+		ctx, events := notif.RegisterForQueryEvents(ctx)
+
+		closestPeers, err := GetPorcelainAPI(env).NetworkGetClosestPeers(ctx, string(id))
+		if err != nil {
+			cancel()
+			return err
+		}
+
+		go func() {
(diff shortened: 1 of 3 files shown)


commands/dht.go
```diff
@@ -27,9 +28,72 @@ var dhtCmd = &cmds.Command{
...
 	},
 }
+var queryDhtCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Find the closest Peer IDs to a given Peer ID by querying the DHT.",
+		ShortDescription: "Outputs a list of newline-delimited Peer IDs.",
+	},
+
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("peerID", true, true, "The peerID to run the query against."),
```
ZenGround0: Please set the third argument `false`.  go-ipfs defaults all of these to variadic even though the command body only every uses the first request argument which seems inconsistent.

filnet/router.go
```diff
@@ -39,3 +42,12 @@ func NewRouter(r routing.IpfsRouting) *Router {
 func (r *Router) FindProvidersAsync(ctx context.Context, key cid.Cid, count int) <-chan pstore.PeerInfo {
 	return r.routing.FindProvidersAsync(ctx, key, count)
 }
+
+// GetClosestPeers returns peers of the K closest peers to the given key
```
ZenGround0: To help make it clear what K is please modify to something like "GetClosestPeers returns a channel of the K closest peers to the given key where K is the 'K Bucket' parameter of the Kademlia DHT protocol.

Also please be sure to put a fullstop at the end of a comment like this that is a complete sentence as some people here are a little OCD about that 😀.

commands/dht.go
```diff
@@ -27,9 +28,72 @@ var dhtCmd = &cmds.Command{
 	},
 
 	Subcommands: map[string]*cmds.Command{
+		"query":     queryDhtCmd,
```
phritz: we generally order by alpha

plumbing/api.go
```diff
@@ -187,6 +187,11 @@ func (api *API) NetworkFindProvidersAsync(ctx context.Context, key cid.Cid, coun
 	return api.network.FindProvidersAsync(ctx, key, count)
 }
 
+// NetworkGetClosestPeers issues a getClosestPeers query to the filecoin network.
```
phritz: @rosalinekarr @anorth once we kill the node api we might want to make a sweep of plumbing and move anything that is not of general utility out to a direct dependency of the command that uses it, eg this and some of the other network calls.

firmianavan:
> Thanks! my only question is whether it is possible to write a test for this...? generally if we have some functionality, esp with a nontrivial implementation, we want it tested. since this is more for debugging it is less critical, esp since it is copy pasted from ipfs, but still worth seeing if we could write one easily. if so, we should.

Sorry for late response, I am working on the dht test

anorth:
@ZenGround0 could you please shepherd this PR to completion?

ZenGround0:
Hey @firmianavan this is looking good.  Have you started writing a test for this?  If not I will go ahead and merge this and create a follow on issue for testing.

firmianavan:
> Hey @firmianavan this is looking good. Have you started writing a test for this? If not I will go ahead and merge this and create a follow on issue for testing.

I have not finish yet, please do merge and create a follow issue

ZenGround0:
@firmianavan would you mind rebasing on master and then pinging me?

firmianavan:
@ZenGround0 It was done now

PR #2204, opened 2019-03-07
Feat/actor-ls-plum-pork-1798

### LINKED ISSUE #1798: move actor ls command to rely on plumbing
### Description

background #1469, depends on #1797 which adds actorget to plumbing. do not start this story unless 1797 is closed.

actor ls command:
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/commands/actor.go#L22
implementation:
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/impl/actor.go#L32

currently actor ls command relies on the node api. we want to move it to rely on plumbing and to remove the actor interface from the node api:
https://github.com/filecoin-project/go-filecoin/blob/ce708768d3dd06e3597a4af9bd6879e57fd35a39/api/actor.go#L35

in #1797 an actorget interface was added to plumbing. we should use it to implement the actor ls command. actor ls enumerates the actors in the tree and returns a view on them appropriate for printing. lets introduce a addressls plumbing call that returns a list of all addresses in the state tree. dont worry about memory efficiency here, just return a huge list for now. then we can have a porcelain call that iterates these addresses, calls actorget on each, and creates the actorview for it which is printed. 

### Acceptance criteria

- there are zero files named actor in api/ and in api/impl, and zero actor methods in the api interface
- actor ls command uses plumbing/porcelain

rosalinekarr:
# Problem

We want to get rid of the `api` directory, but Actor Ls is part of that.

# Solution

Move Actor Ls to plumbing.

Resolves #1798


state/tree.go
```diff
@@ -243,42 +243,47 @@ func (t *tree) debugPointer(ps []*hamt.Pointer) {
...
+// GetAllActorsResult is the struct returned via a channel by the GetAllActors
+// method. This struct contains only an address string and the actor itself.
+type GetAllActorsResult struct {
+	Address string
+	Actor   *actor.Actor
+}
+
 // GetAllActors returns a slice of all actors in the StateTree, t.
-func GetAllActors(t Tree) ([]string, []*actor.Actor) {
+func GetAllActors(t Tree) <-chan GetAllActorsResult {
```
anorth: I can't tell from the code why this has been changed to a channel. Is this because it's slow and so as to provide incremental results to the UI. Please document that if so (the comment above is now wrong in any case). 

(In future, it would be easier to review and discuss if the functional change were in a separate PR).

state/tree.go
```diff
@@ -243,42 +243,47 @@ func (t *tree) debugPointer(ps []*hamt.Pointer) {
...
+func GetAllActorsFromStore(ctx context.Context, store *hamt.CborIpldStore, stateRoot cid.Cid) (<-chan GetAllActorsResult, error) {
 	st, err := LoadStateTree(ctx, store, stateRoot, nil)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-
-	addrs, actors := GetAllActors(st)
-	return addrs, actors, nil
+	return GetAllActors(st), nil
```
phritz: blocking: seems like getallactors needs to respect the context else it cannot be canceled

state/tree.go
```diff
@@ -243,42 +243,47 @@ func (t *tree) debugPointer(ps []*hamt.Pointer) {
...
 }
 
 // NOTE: This extracts actors from pointers recursively. Maybe we shouldn't recurse here.
-func (t *tree) getActorsFromPointers(ps []*hamt.Pointer) (addresses []string, actors []*actor.Actor) {
+func (t *tree) getActorsFromPointers(out chan<- GetAllActorsResult, ps []*hamt.Pointer) {
```
acruikshank: BLOCKING: This needs to take a context parameter and select from ctx.Done to permit cancellation. The ctx parameter needs to be added on up the chain.

state/tree.go
rosalinekarr: This change was suggested by @acruikshank because GetAllActors could potentially return a very large number of actors. I've updated the comment.

state/tree.go
```diff
@@ -243,42 +243,54 @@ func (t *tree) debugPointer(ps []*hamt.Pointer) {
...
-			addresses = append(addresses, kv.Key)
-			actors = append(actors, &a)
+			result := GetAllActorsResult{
+				Address: kv.Key,
+				Actor:   &a,
+			}
+
+			select {
+			case <-ctx.Done():
+				return
```
phritz: it is more idiomatic to surface ctx.err in this case to the caller, as it is they can't distinguish between timing out and finishing getting all the actors

state/tree.go
```diff
@@ -243,42 +243,54 @@ func (t *tree) debugPointer(ps []*hamt.Pointer) {
...
 
-			addresses = append(addresses, kv.Key)
-			actors = append(actors, &a)
+			result := GetAllActorsResult{
+				Address: kv.Key,
+				Actor:   &a,
+			}
+
+			select {
+			case <-ctx.Done():
```
phritz: maybe add a note that select is non deterministic so there is some chance that if receiver is waiting we will send them a result even when the context has been canceled. to fix this we could put an if ctx.err != nil return above this select, and keep the select as is, but dunno if that is worth it or just confusing... cc @anorth

Issue #2035, opened 2019-02-20
Check connection to miner before proposing storage deal


anorth:
### Description
Clients require a direct connection to a miner in order to propose a deal. Many asks may be on chain from miners that are no longer active, or just temporarily inaccessible. At present, clients go to all the work of setting up payment channels (on-chain) before discovering that the target miner is inaccessible. `propose-storage-deal` fails with `error sending proposal: context deadline exceeded`.

Clients should test connection to the miner, with a relatively short timeout, before making on-chain operations. Maybe this can be done in parallel with ingesting the file and other sanity checks like it's size.

### Acceptance criteria
- Attempting to propose a deal to an offline miner fails in only a few seconds, with a clear error message
- If a miner goes offline subsequent to the ping, the error message when failing to propose the deal clearly articulates the problem

### Risks + pitfalls
None

### Where to begin
protocol/client/miner.go

magik6k:
https://github.com/filecoin-project/go-filecoin/pull/2216 should address part of this

anorth:
@rkowalick could you evaluate whether this is sufficiently fixed, or maybe needs some error message improvements to bring it home?

rkowalick:
Looking now

rkowalick:
For both Acceptance Criteria cases, I get the following error message:

```
Error: error sending proposal: failed to establish connection with the peer: dial attempt failed:
<peer.ID Qm*osca5L> --> <peer.ID Qm*5jwVEW> dial attempt failed: dial tcp4
127.0.0.1:50593: connect: connection refused
```

I'm not sure if this error message is satisfactory or not. Let me know if want me to change it @anorth

anorth:
Looks good enough to me, thanks!

PR #2216, opened 2019-03-08
client: Check connection to miner before proposing storage deal

magik6k:
See https://github.com/filecoin-project/go-filecoin/issues/2035


whyrusleeping:
Thanks @magik6k :)

protocol/storage/client.go
```diff
@@ -178,6 +197,26 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
...
+	}
+
+	select {
+	case _, ok := <-res:
+		if !ok {
+			return errors.New("couldn't establish connection to miner: ping channel closed")
+		}
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("couldn't establish connection to miner: %s", ctx.Err())
```
anorth: add "timed out after %s" to this message

protocol/storage/client.go
```diff
@@ -178,6 +197,26 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 	return &response, nil
 }
 
+func (smc *Client) pingMiner(ctx context.Context, pid peer.ID) error {
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Second) //TODO: better place to (not) hard-code this?
```
anorth: For now, please make the timeout a parameter to this function, then move this hardcoded value next to the other hardcoded value at line 90. Remove the TODO before merging.

PR #2221, opened 2019-03-09
[wip] Migrate to go modules

### LINKED ISSUE #2208: Use Go modules
### Description
So long gx 👋 we are migrating dependency management to go modules

* See the [go modules wiki](https://github.com/ipfs/go-ipfs/issues/5848)

* See the [historical issue](https://github.com/ipfs/go-ipfs/issues/5850) beginning the discussion in go-ipfs

* See some [prior art](https://github.com/ipfs/go-ipfs/pull/5849) [in go-ipfs](https://github.com/ipfs/go-ipfs/issues/5848)

* A [script](https://gist.github.com/raulk/708debe71b9abd395fc7935ee1215ddd) from @raulk for moving to go-mod 

### Acceptance criteria
Given the above resources, figure out how to move from gx to go-mod and do this.

### Where to begin
A close reading of the above resources, or talk to @Kubuxu.  Seems like @anacrolix and @raulk are also good people to consult on this.

hsanjuan:
Fixes #2208


hsanjuan:
I have only been able to get this far:

* go-metalinter is deprecated and is going to be killed. It does not work with Go modules. I have replaced it with golangci-lint. Running the linter throws a bunch of errors for unchecked return types etc. Probably the previous linter was not checking so strictly.
* A bug in go modules requires upgrade of go version to 1.11.5
* There is no more `smartdeps` mode to build deps. Gomod pretty much hits the network all the time. Gomod deps are stored in `$GOPATH/pkg`, but this is only worth caching once the whole thing is built (or even after running tests), so I have altered the order of the cache creation in circle.
* While the binaries build, tests don't pass for iptb*/filecoin/docker. Probably the go-moded version of iptb is not compatible with filecoin in its current state.

anorth:
Great progress, thanks. Could you please pull out the linter changes to a separate PR that we can discuss and land first? Ditto for any other prerequisite changes that can stand alone.

hsanjuan:
#2307 overseeds this

Issue #2146, opened 2019-03-01
Implement an outbound message queue, distinct from the message pool


anorth:
### Description
We wish to separate the message pool from the outbound message queue, including enabling non-mining nodes to have no mpool, only an outbound queue.

The outbound queue is described in this [design doc](https://docs.google.com/document/d/1MO6MloDs1plJsBb8pLKYILJQPbmO9Hy8hdrGCCno6UI/edit#)

PR #2227, opened 2019-03-11
Initial implementation of outbound message queue, including expiry mechanism

anorth:
This is the core of #2146.

This is not yet integrated into the message sender plumbing, which will provide values for chain height,
or chain syncing, which will remove/expire messages.

### DIFF
--- a/core/message_queue.go
+++ b/core/message_queue.go
@@ -1,0 +1,132 @@
+package core
+
+import (
+	"sync"
+
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/pkg/errors"
+)
+
+// MessageQueue stores an ordered list of messages (per actor) and enforces that their nonces form a contiguous sequence.
+// Each message is associated with a "stamp" (an opaque integer), and the queue supports expiring any list
+// of messages where the first message has a stamp below some threshold. The relative order of stamps in a queue is
+// not enforced.
+// A message queue is intended to record outbound messages that have been transmitted but not yet appeared in a block,
+// where the stamp could be block height.
+// MessageQueue is safe for concurrent access.
+type MessageQueue struct {
+	lk sync.RWMutex
+	// Message queues keyed by sending actor address, in nonce order
+	queues map[address.Address][]*queuedMessage
+}
+
+type queuedMessage struct {
+	msg   *types.SignedMessage
+	stamp uint64
+}
+
+// NewMessageQueue constructs a new, empty queue.
+func NewMessageQueue() *MessageQueue {
+	return &MessageQueue{
+		queues: make(map[address.Address][]*queuedMessage),
+	}
+}
+
+// Enqueue appends a new message for an address. If the queue already contains any messages for
+// from same address, the new message's nonce must be exactly one greater than the largest nonce
+// present.
+func (mq *MessageQueue) Enqueue(msg *types.SignedMessage, stamp uint64) error {
+	mq.lk.Lock()
+	defer mq.lk.Unlock()
+
+	q := mq.queues[msg.From]
+	if len(q) > 0 {
+		nextNonce := q[len(q)-1].msg.Nonce + 1
+		if msg.Nonce != nextNonce {
+			return errors.Errorf("Invalid nonce %d, expected %d", msg.Nonce, nextNonce)
+		}
+	}
+	mq.queues[msg.From] = append(q, &queuedMessage{msg, stamp})
+	return nil
+}
+
+// RemoveNext removes and returns a single message from the queue, if it bears the expected nonce value, with found = true.
+// Returns found = false if the queue is empty or the expected nonce is less than any in the queue for that address
+// (indicating the message had already been removed).
+// Returns an error if the expected nonce is greater than the smallest in the queue.
+// The caller may wish to check that the returned message is equal to that expected (not just in nonce value).
+func (mq *MessageQueue) RemoveNext(sender address.Address, expectedNonce uint64) (msg *types.SignedMessage, found bool, err error) {
+	mq.lk.Lock()
+	defer mq.lk.Unlock()
+
+	q := mq.queues[sender]
+	if len(q) > 0 {
+		head := q[0]
+		if expectedNonce == uint64(head.msg.Nonce) {
+			mq.queues[sender] = q[1:] // pop the head
+			msg = head.msg
+			found = true
+		} else if expectedNonce > uint64(head.msg.Nonce) {
+			err = errors.Errorf("Next message for %s has nonce %d, expected %d", sender, head.msg.Nonce, expectedNonce)
+		}
+		// else expected nonce was before the head of the queue, already removed
+	}
+	return
+}
+
+// Clear removes all messages for a single sender address.
+func (mq *MessageQueue) Clear(sender address.Address) bool {
+	mq.lk.Lock()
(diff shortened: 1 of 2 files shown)


core/message_queue.go
```diff
@@ -0,0 +1,131 @@
...
+		head := q[0]
+		if expectedNonce < uint64(head.msg.Nonce) {
+			return nil, nil
+		} else if expectedNonce == uint64(head.msg.Nonce) {
+			mq.queues[sender] = q[1:] // pop the head
+			return head.msg, nil
+		}
+		return nil, errors.Errorf("Next message for %s has nonce %d, expected %d", sender, head.msg.Nonce, expectedNonce)
+	}
+	return nil, nil
```
acruikshank: I'd rather this function return special not found error or add an `ok` return value. Not a big deal, and all the options are a little ugly, but signalling by returning nils is not idiomatic in Go.

core/message_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	mq.queues[msg.From] = append(q, &queuedMessage{msg, stamp})
+	return nil
+}
+
+// RemoveNext removes and returns a single message from the queue, if it bears the expected nonce value.
+// Returns the message removed, or nil if the queue is empty or the expected nonce was less than any
+// in the queue for that address (indicating the message had already been removed).
+// Returns an error if the expected nonce is greater than the smallest in the queue.
+// The caller may wish to check that the returned message is equal to that expected (not just in nonce value).
+func (mq *MessageQueue) RemoveNext(sender address.Address, expectedNonce uint64) (*types.SignedMessage, error) {
```
acruikshank: It's not clear how the caller is going to know the expected nonce.
In general, there's a lot of logic around nonce and block height tracking that I was expecting to see inside this queue. Will this data structure be wrapped in another, smarter message queue looking thing that handles that logic? It seems a bit much for a plumbing struct.

core/message_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	q := mq.queues[sender]
+	if len(q) > 0 {
+		return uint64(q[len(q)-1].msg.Nonce), true
+	}
+	return 0, false
+}
+
+// NextStamp returns the stamp for the next message in the queue for an address, or zero if the
+// queue is empty.
+func (mq *MessageQueue) NextStamp(sender address.Address) uint64 {
```
acruikshank: Why is this exposed?

core/message_queue.go
```diff
@@ -0,0 +1,131 @@
...
+// MessageQueue is safe for concurrent access.
+type MessageQueue struct {
+	lk sync.RWMutex
+	// Message queues keyed by sending actor address, in nonce order
+	queues map[address.Address][]*queuedMessage
+}
+
+type queuedMessage struct {
+	msg   *types.SignedMessage
+	stamp uint64
```
acruikshank: Why not just call this `blockHeight`? That's the intent, and calling it `stamp` when it's actually a block height will be confusing.

core/message_queue.go
```diff
@@ -0,0 +1,131 @@
...
+func (mq *MessageQueue) Enqueue(msg *types.SignedMessage, stamp uint64) error {
+	mq.lk.Lock()
+	defer mq.lk.Unlock()
+
+	q := mq.queues[msg.From]
+	if len(q) > 0 {
+		nextNonce := q[len(q)-1].msg.Nonce + 1
+		if msg.Nonce != nextNonce {
+			return errors.Errorf("Invalid nonce %d, expected %d", msg.Nonce, nextNonce)
+		}
```
shannonwells: thank you for error messages useful for diagnostics.

core/message_queue_test.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+var keys = types.MustGenerateKeyInfo(3, types.GenerateKeyInfoSeed())
+var addresses = make([]address.Address, len(keys))
+var signer = types.NewMockSigner(keys)
+var msgSeq = 0
+var gasPrice = *types.ZeroAttoFIL
+var gasUnits = types.NewGasUnits(0)
+
```
shannonwells: Can we avoid having global variables in tests? Other testing functions will share this space. I'd prefer setting this stuff inside a test suite, in helper functions from within tests.  It has caused me headaches in places like default_syncer_test.go. It is a lot like RSpec before(:all) which also can have unexpected side effects, but there  it's at least confined to the file itself.

core/message_queue_test.go
```diff
@@ -0,0 +1,291 @@
...
+		nonce,
+		types.NewAttoFILFromFIL(0),
+		"method"+fmt.Sprintf("%d", seq),
+		[]byte("params"))
+	signed, err := types.NewSignedMessage(*msg, signer, gasPrice, gasUnits)
+	require.NoError(t, err)
+	return signed
+}
+
+func mustEnqueue(t *testing.T, q *core.MessageQueue, msg *types.SignedMessage, stamp uint64) {
```
shannonwells: We should standardize on naming for these things that call require.NoError and friends.  I don't have any preference, but I've seen both mustDoAThing and requireDoAThing.  I'll even go back and change my own stuff if people prefer mustDoAThing.

anorth:
@acruikshank I can see we have a difference in philosophy here between this and #2237 (I reviewed that before reading here). I have intentionally decoupled the data structure from the hookup/strategy code that will use it (not yet written) and put it up for review to ensure that it stands alone. The abstraction over the definition of block height, etc, is intentional. There is some chance the interface will need changing as I integrate.

I don't yet know what the integration code will be called, but certainly it will be more than just plumbing. A `MessageQueueExpirer` perhaps. I hope it will be testable in ~isolation, though.

I prefer small PRs that don't do any more than necessary for a self-contained change. I want feedback on the internals here before I do all the hookup work, in case I'm wildly off-target. One alternative approach I could follow (next time) is to do the hookup work but to a stub implementation in one or more PRs, and then fill out the implementation (i.e. top-down). Another alternative is to do all the work in secret and then put up 3-4 PRs, based off each other, for review in parallel. That would give you a lot more context and make it more likely I get the interface exactly right, but I fear the constant chained rebasing involved in such an approach.

core/message_queue.go
anorth: The caller is going to see a message in a block and read its nonce from that. Nonces should be contiguous, so a gap will error here. But the sequence may also restart from an earlier point after a reorg.

core/message_queue.go
anorth: I hear you and it's worth talking about it. I intend this data structure to be self contained, in that it implemented the interface and semantics described in its comments. While it does have an intended use case, it's correctness need not depend on that. 

It wasn't an obvious answer to use block height for the timeout. If we change our mind, nothing in this structure should need to change.

core/message_queue_test.go
anorth: Thank you, I was looking for a pattern here and couldn't find a clear one. I will do that.

The minor reason I chose top-level vars was so I could define helper functions below the test code. If I move them inside the test method I think I have to declare the helpers before the test code, as `func` vars.

core/message_queue_test.go
anorth: I think I saw `must` pattern in godocs somewhere, but can't find it now.

anorth: I'll resolve in a follow-up if I've ended up doing the wrong thing (seems likely)

PR #2229, opened 2019-03-11
Serialize Uint64 to string in JSON

### LINKED ISSUE #1924: integers should be json encoded to something readable
see nonce, gaslimit below. a decimal string would be better

```
temporary ApplyMessage failure, [SignedMessage cid=[zDPWYqFCxYhhKMiB44R9cUUS3wukJ8Xm3YJUSjY21Yw5g6cuMNXB]: {
  "meteredMessage": {
    "message": {
      "to": "fcqgu23wf2arsfhck08gvkeq07ke7treq074ccppr",
      "from": "fcqmehdkmrtpj7utgzvwh5aqyrjmyfc9gkafg38fw",
      "nonce": "Ag==",
      "value": null,
      "method": "commitSector",
      "params": "hUEBWCDFZ6qHdDXSL4fuq3p30grnR7TnGpHwYuejcj8WiMShLlggq0wvChl7Eo1mPROvgCd8AF0MQENbb/AKhVPFBnJT7xpYIGFN9apxv8Zf+HN9cF0ic+B0+vWcesqw9EoE2/3PRogpWQGAlkX+ngdcRWW/mnp+BdWohPUyuTmO5yd4U8Vp49+lPcL7f5PgXWvK2JLQFBWeNt84hvTQuxjdDnj1aCxvZVIjq1bNnwu0toTZPRYLlnb5kxkITIfb5+akO60g783JBmHXGe7S/mXZKD7eGGjgQZS2AqWjLdEhPWQDhDL1v3jl3V817jabFA4WDl5RSUeaP1o/khdUlCN21/JbLqGbG/33bjCo69rG7l8lVem/dZz+KLq+6t/NGFtKYMP7nM4rb3XThv9TNNNtAZOl4yAjdQtSsXxdmr8UVP7rlosHjiVGqrLqZklGJ2fL3NnZafZuUTSWiEtlvcrv6KcHei5EujCx/hUeLSATN7hfQt1XywEKbeaEQDRI5W+Um+S4gAVQot7RC8dt2qHuWBeWHDRoXOLWs9ru4NkPimKZV9bD7JbAxqEmFv2OOh9v6jiWDWifbR42oJ69g1ZcRFQRo+etHvpdWqT40SORChIbEnZVjyBIcOHt0AZt2v7uw5sXbzJGu/YP"
    },
    "gasPrice": "0",
    "gasLimit": "gK3iBA=="
  },
  "signature": "fjCOhcNNGwRncKrCHHtGgeJ6FKMxEtG4kANXLDa66CoEUuYQFFuwKnSj3B+xxIkMDygp+XvVxyhrF/RZLZdeXAE="
}] (apply message failed: message gas limit too high for current block)
```

mslipper:
Closes #1924.


types/uint64.go
```diff
@@ -1,12 +1,11 @@
...
 
 import (
-	"encoding/base64"
-	"strings"
-
 	"gx/ipfs/QmSKyB5faguXT4NqbrXpnRXqaVj5DhSm7x9BtzFydBY1UK/go-leb128"
 	cbor "gx/ipfs/QmcZLyosDwMKdB6NLRsiss9HXzDPhVhhRtPy67JFKTDQDX/go-ipld-cbor"
 	"gx/ipfs/QmdBzoMxsBpojBfN1cv5GnKtB7sfYBMoLH7p9qSyEVYXcu/refmt/obj/atlas"
+	"strconv"
+	"strings"
```
anorth: Please revert this reordering. Our conventions are documented in https://github.com/filecoin-project/go-filecoin/blob/master/CONTRIBUTING.md#conventions-and-style

phritz:
the point of Uint64 was to have a custom type that we could encode efficiently using leb128, as opposed to say a fixed size integer encoding. however that whole exercise turns out to have been a waste of time because i recently learned we are using cbor's native variable length encoding https://github.com/filecoin-project/specs/pull/131#discussion_r259549966. i cannot think of a good reason to keep Uint64 off the top of my head except for custom json encoding and maybe something to do with the go-ipld-cbor library which uses a global (!) type registry eg https://github.com/filecoin-project/go-filecoin/blob/e1697c1c33b9024f042dc546c715cea9713337d3/types/atto_fil.go#L19. if you have a type you want encoded two different ways you have to create a type alias for it, so it could be the case that there is a custom registry or default behavior for uint64 that we might want to circumvent for some reason. @warpfork has canonical bits for how refmt works, but will be lacking any context from go-filecoin.

the fact that this change doesn't break any schema tests https://github.com/filecoin-project/go-filecoin/search?q=requireschemaconformance&unscoped_q=requireschemaconformance i guess is not surprising because the field still has string type. but we definitely want a test that verifies that this change has the behavior intended. i think we need a test in commands that asks for json output and verifies that the height for example we get out is an integer string, as opposed to an base64 encoding of leb128-encoded bytes. i seem to want to recall some hidden gotcha here, but can't off the top of my head. 

would any other context be helpful?

ZenGround0:
@anorth @phritz are we ok writing a followup issue to decide the fate of uint64 and merging this?

anorth:
I think we need the tests @phritz requested, but otherwise yes I'm very happy for a follow-up to address Uint64.

mslipper:
Cool, I'll add the tests tonight.

warpfork:
> which uses a global (!) type registry

fwiw, that's somewhat in "known issue" town and will be fixed again when we get to go-ipld-prime, which is intended to supplant the current go-ipld-cbor and address design issues like this.  (It's a little odd that go-ipld-cbor ended up with a global type registry, since refmt was explicitly designed to avoid this, but, that ship sailed a while ago now.)

---

I'll refrain from any other strong opinions about the semantics being discussed here, since I'm not super synced up with the overall context, but as a general note, numerics right around the max bit size limits tend to terrify me.  I always wonder to myself -- if the 64th bit really mattered, is there some reason that one was more important than the 63rd, and is there a stable reason the 65th never will?  In other words, is it certain this shouldn't bite the bullet and do a BigInt?  If it's literally a 64-bit bitfield for some algorithmic reason, that's fine, but any other application is quite scary; and if it's a bitfield rather than a number, don't bytes as a type make more sense anyway?  /2c.

travisperson:
> I think the whole point of the Uint64 type is to achieve JSON encoding, so without that we might be better off nuking it.

> i cannot think of a good reason to keep Uint64 off the top of my head except for custom json encoding

I think that is a big reason for keeping the custom type around. We shouldn't encode it to a JSON number because JSON doesn't define the semantics of what a number is besides the sequence of valid characters. This makes it much harder to work with in different languages or with different JSON libraries.

We could use JSON numbers, but make it much harder for others to support as well. Regardless we would need the custom type to handle the encoding if we wanted to have more than 53 bits (or 52 I can never remember) of precision as golang IEEE-754 for float64 (which is the internal type used when handling JSON numbers).

It seems much easier to use a string, and then allow application developers to custom convert to a numerical type they want after whatever library / built-in parses the JSON tokens.

mishmosh:
The uint64 discussion is parked in #2267.

Tagging @phritz for maintainer review now.

phritz:
>  In other words, is it certain this shouldn't bite the bullet and do a BigInt? 

fair point. i think native types are just easier to work with, so we prefer them where they are going to last a long time, but i can see both sides.

anorth:
I have reverted this because the test failed on master. I'm not quite sure what went wrong, but please try again!

mslipper:
`tar: go/src/github.com/gogo/protobuf/proto/table_merge.go: Cannot write: Cannot allocate memory` - seems transient. Will rebase and rebuild.

PR #2233, opened 2019-03-11
Message validator checks value is non-negative and gas limit is below block limit

### LINKED ISSUE #1934: Messages should not send with insufficient funds

anorth:
Also adds tests for the validator 🎉 

Closes #1934.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -51,6 +51,16 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
 		return errNonAccountActor
 	}
 
+	if msg.Value.IsNegative() {
+		log.Info("Cannot transfer negative value", fromActor, msg.Value)
+		return errNegativeValue
+	}
+
+	if msg.GasLimit > types.BlockGasLimit {
+		log.Info("Message gas limit above block limit", fromActor, msg, types.BlockGasLimit)
+		return errGasAboveBlockLimit
+	}
+
 	// Avoid processing messages for actors that cannot pay.
 	if !canCoverGasLimit(msg, fromActor) {
 		log.Info("Insufficient funds to cover gas limit: ", fromActor, msg)
--- a/commands/message_daemon_test.go
+++ b/commands/message_daemon_test.go
@@ -112,18 +112,11 @@ func TestMessageSendBlockGasLimit(t *testing.T) {
 	result := struct{ Messages []interface{} }{}
 
 	t.Run("when the gas limit is above the block limit, the message fails", func(t *testing.T) {
-		d.RunSuccess(
+		d.RunFail("block gas limit",
 			"message", "send",
 			"--price", "0", "--limit", doubleTheBlockGasLimit,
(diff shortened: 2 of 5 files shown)


consensus/validation.go
```diff
@@ -51,6 +51,16 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
 		return errNonAccountActor
 	}
 
+	if msg.Value.LessThan(types.ZeroAttoFIL) {
```
rkowalick: This check is already present here: https://github.com/filecoin-project/go-filecoin/blob/414464d5f086d35fcdcdfb1a7a46e50fbe0e28c6/vm/vm.go#L69

I'm not sure if the check is needed in both places, but I don't think we should be duplicating these consensus rules.

consensus/validation.go
```diff
@@ -51,6 +51,16 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
...
+		log.Info("Cannot transfer negative value", fromActor, msg.Value)
+		return errNegativeValue
+	}
+
+	if msg.GasLimit > types.BlockGasLimit {
```
rkowalick: Similar to the above comment, this check is also performed here: https://github.com/filecoin-project/go-filecoin/blob/25b666da7e07f0a8b9f739ad2af2eec5b149ec56/consensus/processor.go#L605

acruikshank: @rkowalick This code is used both before sending and prior to processing. We don't want users to send messages that are going to fail because they are over the block limit. Also, messages with this problem should fail faster on the mining side, because we can't charge gas for these messages. The check in the gas tracker isn't hurting anything, but messages really should not have gotten that far.

consensus/validation.go
acruikshank: @rkowalick Messages should fail faster for these types of errors, so this is appropriate (it will fail the message before it's sent).

@anorth You should use the `IsNegative` method, though.

consensus/validation.go
rkowalick: Since validate is called both in `Sender.send` and in `attemptApplyMessage`, the old "gas block limit" check is redundant and should be removed.

consensus/validation.go
rkowalick: As below, this check happens in both `Sender.send` and `attemptApplyMessage`, so the older version should be removed, since they are redundant.

anorth:
@rkowalick I feel pretty uncomfortable about removing either of the lower-level checks. The VM as a module should be able to implement whatever invariants it wants, and we *want* to replicate them close to the "user" for fast feedback and to reduce confusion. C.f. client and server-side validation in a web app. I don't think the fact that the check is made three times in total (once in sender, twice in miner) is a big problem.

I think the block gas limit check is similar.

If there were a call to remove, I think it would be the call to Validate in `attemptApplyMessage`, but we'd then want to be very confident that everything subsequent enforces everything that it must. I think the risk tradeoff is too one-sided at the moment.

PR #2234, opened 2019-03-11
Remove redundant error return from KeyInfo.Address()

anorth:
The error value was removed in 619b0eb1f946a18f87b232834c8c6df1f6fc9b32.

This change is mechanical, I did nothing except pull the thread.

### DIFF
--- a/api/address.go
+++ b/api/address.go
@@ -19,7 +19,7 @@ type Address interface {
 
 // Addrs is the interface that defines method to interact with addresses.
 type Addrs interface {
-	New(ctx context.Context) (address.Address, error)
+	New(ctx context.Context) address.Address
 	Ls(ctx context.Context) ([]address.Address, error)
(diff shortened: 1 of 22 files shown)


travisperson:
Before merging this I would check to see if this change make sense when we land the new address spec.

PR #2186
Spec https://github.com/filecoin-project/specs/pull/135

anorth:
@shannonwells already reflected.

@travisperson thank you for a great catch that makes me really sad. @frrist your PR invokes `NewSecp256k1Address` which looks like it returns an error, but I can't find that method definition anywhere? Should I abandon this change?

frrist:
`NewSecp256k1Address` is defined [here](https://github.com/filecoin-project/go-filecoin/pull/2186/files#diff-2026d132a1aa6ec72a1af72e5763011cR169) -- added in the new Address spec PR. Sadly I think we may need to abandon PR. Nice catch @travisperson and sorry @anorth :disappointed:.

PR #2237, opened 2019-03-11
Timeout messages from message pool

### LINKED ISSUE #2133: Timeout messages in message pool
### Description

Messages that fail processing with a permanent error never exit the message pool, because we never receive receipts to cause us to remove them. To avoid this:

1. Track the block height when a message enters the message pool.
2. Establish a timeout interval.
3. When updating to remove messages with receipts from the pool, also remove any message that has exceeded the timeout.

### Acceptance criteria

Messages that have exceeded the timeout interval are no longer in the message pool.

acruikshank:
closes #2133

### Problem

The message pool currently has no invalidation mechanism. Temporary failures and other issues can cause a message to remain in the message pool indefinitely. This also means there's no clean recovery mechanism when several messages are sent and the first message fails.

### Solution

Add  a timeout mechanism to the message pool. The mechanism involves the following steps:

1. Store `timedMessages` instead of just messages. `timedMessages` are a message plus a block height.
2. On `UpdateMessagePool`, search `MessageTimeOut` tipsets back in the chain to get the timeout block height.
3. Remove every message in the `pending` map that has a block height below the timeout.

It's important that the timeout block height be based on the number of tipsets seen, rather than a specific block height. It's possible for the chain to have upwards of 10 null tipsets, so the block time when a message is added might be lower than expected. 

For example: If a message is added to the pool ~5 minutes after the node received tip set 100, and then 10 null tip sets occur, when tip set 110 arrives the message appears to be 10 blocks old, even though it hasn't really had a chance to have been mined in a block. Walking back tip sets bases the timeout on the number of opportunities the block should have had to have been mined.

### DIFF
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -373,6 +373,13 @@ func (store *DefaultStore) Head() types.TipSet {
 	return store.head
 }
 
+// BlockHeight returns the chain height of the head tipset.
+// Strictly speaking, the block height is the number of tip sets that appear on chain plus
+// the number of "null blocks" that occur when a mining round fails to produce a block.
+func (store *DefaultStore) BlockHeight() (uint64, error) {
+	return store.Head().Height()
+}
+
 // LatestState returns the state associated with the latest chain head.
 func (store *DefaultStore) LatestState(ctx context.Context) (state.Tree, error) {
 	h := store.Head()
--- a/core/message_pool.go
+++ b/core/message_pool.go
@@ -12,6 +12,19 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// MessageTimeOut is the number of tipsets we should receive before timing out messages
+const MessageTimeOut = 6
+
+type timedmessage struct {
+	message *types.SignedMessage
+	addedAt uint64
+}
+
+// BlockTimer defines a interface to a struct that can give the current block height.
+type BlockTimer interface {
+	BlockHeight() (uint64, error)
+}
+
 // MessagePool keeps an unordered, de-duplicated set of Messages and supports removal by CID.
 // By 'de-duplicated' we mean that insertion of a message by cid that already
 // exists is a nop. We use a MessagePool to store all messages received by this node
@@ -22,26 +35,37 @@ import (
 type MessagePool struct {
 	lk sync.RWMutex
 
-	pending map[cid.Cid]*types.SignedMessage // all pending messages
+	timer   BlockTimer
+	pending map[cid.Cid]*timedmessage // all pending messages
 }
 
 // Add adds a message to the pool.
 func (pool *MessagePool) Add(msg *types.SignedMessage) (cid.Cid, error) {
+	blockTime, err := pool.timer.BlockHeight()
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	return pool.addTimedMessage(&timedmessage{message: msg, addedAt: blockTime})
+}
+
+func (pool *MessagePool) addTimedMessage(msg *timedmessage) (cid.Cid, error) {
 	pool.lk.Lock()
 	defer pool.lk.Unlock()
 
-	c, err := msg.Cid()
+	c, err := msg.message.Cid()
 	if err != nil {
 		return cid.Undef, errors.Wrap(err, "failed to create CID")
 	}
 
 	// Reject messages with invalid signatires
-	if !msg.VerifySignature() {
+	if !msg.message.VerifySignature() {
 		return cid.Undef, errors.Errorf("failed to add message %s to pool: sig invalid", c.String())
 	}
 
 	pool.pending[c] = msg
 	return c, nil
+
 }
 
 // Pending returns all pending messages.
@@ -50,21 +74,23 @@ func (pool *MessagePool) Pending() []*types.SignedMessage {
 	defer pool.lk.Unlock()
 	out := make([]*types.SignedMessage, 0, len(pool.pending))
 	for _, msg := range pool.pending {
-		out = append(out, msg)
+		out = append(out, msg.message)
 	}
 
 	return out
 }
 
 // Get retrieves a message from the pool by CID.
-func (pool *MessagePool) Get(c cid.Cid) (value *types.SignedMessage, ok bool) {
+func (pool *MessagePool) Get(c cid.Cid) (*types.SignedMessage, bool) {
 	pool.lk.Lock()
 	defer pool.lk.Unlock()
-	value, ok = pool.pending[c]
-	if ok && value == nil {
+	value, ok := pool.pending[c]
+	if !ok {
+		return nil, ok
+	} else if value == nil {
 		panic("Found nil message for CID " + c.String())
 	}
-	return
+	return value.message, ok
 }
 
 // Remove removes the message by CID from the pending pool.
@@ -76,9 +102,10 @@ func (pool *MessagePool) Remove(c cid.Cid) {
 }
 
 // NewMessagePool constructs a new MessagePool.
-func NewMessagePool() *MessagePool {
+func NewMessagePool(timer BlockTimer) *MessagePool {
 	return &MessagePool{
-		pending: make(map[cid.Cid]*types.SignedMessage),
+		timer:   timer,
+		pending: make(map[cid.Cid]*timedmessage),
 	}
 }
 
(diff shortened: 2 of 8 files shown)


node/node.go
```diff
@@ -389,11 +389,7 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 
 	// only the syncer gets the storage which is online connected
 	chainSyncer := chain.NewDefaultSyncer(&cstOnline, &cstOffline, nodeConsensus, chainStore)
-	chainReader, ok := chainStore.(chain.ReadStore)
```
acruikshank: This cast is redundant. The compiler will complain if `chainStore` isn't a `chain.ReadStore`.

chain/default_store.go
```diff
@@ -373,6 +373,11 @@ func (store *DefaultStore) Head() types.TipSet {
 	return store.head
 }
 
+// BlockHeight returns the height of the head
```
anorth: Please clarify whether "height" counts non-null blocks/tipsets, or rounds (i.e. including "null blocks"). I assume the former, based on conversations with others, with "round number" or "ticket number" being the latter. But I fear it is in fact the latter.

core/message_pool.go
```diff
@@ -12,6 +12,19 @@ import (
...
+const MessageTimeOut = 6
+
+type timedmessage struct {
+	message *types.SignedMessage
+	addedAt uint64
+}
+
+// BlockTimer defines a interface to a struct that can give the current block height.
+type BlockTimer interface {
+	BlockHeight() (uint64, error)
```
anorth: FYI In my WIP message queue I use the term "stamp" to refer to the time-like mark associated with each message. It's opaque to the queue/pool and should make no difference if the stamps and age limit were converted to seconds.

So this could then become `Stamper` with `CurrentStamp()` method, no reference to "blocks" or height etc. The wrapper struct could be `StampedMessage. The interpretation as block height is done by the user/constructor that hooks things up.

core/message_pool.go
```diff
@@ -12,6 +12,19 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// MessageTimeOut is the number of tipsets we should receive before timing out messages
+const MessageTimeOut = 6
```
anorth: Please make this a construction parameter, or even better, a parameter to the timeout/expire method.

core/message_pool.go
```diff
@@ -149,7 +178,7 @@ func collectChainsMessagesToHeight(ctx context.Context, store *hamt.CborIpldStor
 // TODO there is considerable functionality missing here: don't add
 //      messages that have expired, respect nonce, do this efficiently,
 //      etc.
-func UpdateMessagePool(ctx context.Context, pool *MessagePool, store *hamt.CborIpldStore, old, new types.TipSet) error {
+func (pool *MessagePool) UpdateMessagePool(ctx context.Context, store *hamt.CborIpldStore, old, new types.TipSet) error {
```
anorth: Blocking: I do not think this functionality should be tightly tied to the message pool. I have it as a low-pri cleanup to move the integration with the chain store out of this package, so that message_pool is a simple data structure, and these algorithms use its public interface. Please revert the receiver change so that the move is easier later.

See my top-level comment about moving the timestamp to a parameter of `Add`.

anorth:
@acruikshank #2227, which I see you've seen since doing this work.

core/message_pool.go
```diff
@@ -230,12 +263,49 @@ func UpdateMessagePool(ctx context.Context, pool *MessagePool, store *hamt.CborI
 		pool.Remove(cid)
 	}
 
+	// prune all messages that have been in the pool too long
+	return pool.timeoutMessages(ctx, store, new)
+}
+
+// timeoutMessages removes all messages from the pool that arrived more than MessageTimeout tip sets ago.
+// Note that we measure the timeout in the number of tip sets we have received rather than a fixed block
+// height. This prevents us from prematurely timing messages that arrive during long chains of null blocks.
```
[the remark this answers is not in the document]
anorth: @acruikshank and I discussed this earlier and I'm in favour. But we want to align this with outbound queue policy, so it would be helpful if the block counting logic were broken out to a `FindAncestor(n int)` function.

core/message_pool.go
acruikshank: My reasoning for not having a public timeout/expire method are explained in my comments about the role of `MessagePool`.

If I make this a construction parameter, I'll just have to define this constant somewhere else so that I can pass it in. Right now, it's `Node` that would need to know about this constant and pass it in. This seems a bad trade.

We do need to figure what to do with all these constants we have lying around the code. They need to be accounted for in our model for change. Right now I don't know what that looks like, so, in the mean time, this is the simplest option.

core/message_pool.go
acruikshank: The fact that this is a BlockHeight is central to invalidation logic. It can't be abstracted away. See comment on the role of `MessagePool`.

core/message_pool.go
```diff
@@ -22,26 +35,37 @@ import (
...
 	lk sync.RWMutex
 
-	pending map[cid.Cid]*types.SignedMessage // all pending messages
+	timer   BlockTimer
+	pending map[cid.Cid]*timedmessage // all pending messages
 }
 
 // Add adds a message to the pool.
 func (pool *MessagePool) Add(msg *types.SignedMessage) (cid.Cid, error) {
+	blockTime, err := pool.timer.BlockHeight()
```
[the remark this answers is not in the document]
acruikshank: This comment confuses me. The block height should potentially change between calls to Add. This is the point of tracking block height with messages.

core/message_pool.go
acruikshank: See comment about the role of `MessagePool`

core/message_pool.go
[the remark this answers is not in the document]
acruikshank: @phritz I think it is. Fixing the expiry to a block height gives us an ugly tradeoff between making the user wait an unreasonable amount of time for invalidation and the possibility of invalidating invalid messages due to a long chain of null blocks. As @anorth points out, the inbound and outbound queues will be in sync on this. The outbound queue will block new messages until old message clear, so there's no mental math.

acruikshank:
@anorth @phritz 
It seems like there's some disagreement about the __role of `MessagePool`__ here. I get that a data structure could exist that:

1. Can add messages with an associated generic timestamp.
2. Remove messages by cid (or message)
3. Return a slice all the messages.
4. Remove all the messages less that a timestamp.

The problem is that the above describes some code that only supports our concept of an inbound message pool. We'd be left with a lot of business logic about removing messages from the pool that needs to live somewhere. This logic is something that consumers of the outbound message pool shouldn't need to deal with at all.

This means the MessagePool data structure described above would be a dependency of this larger business logic-y thing, and the business logic-y thing will have roughly the same interface as the `MessagePool` does now. The fact that this fairly major change in expiration logic requires no alteration to calling code attests to the fact that MessagePool's current interface is correct abstraction. 

The business logic-y thing and the data structure will be 1-to-1 (both singleton in the system). Note that the inbound and outbound queues are structured differently because the outbound queue cares about nonces. 

If you accept the above, converting message pool into a generic data structure looks more like extracting code internal to the existing message pool into a standalone data structure and maybe renaming `MessagePool` to something else.

I'm happy to make this internal refactor in this PR as long as we can agree that something like `MessagePool` needs to exist with the essentially the same interface. If you disagree, please provide some details about where the business logic should live, how it integrates with the rest of code, and why the changes are in scope for this PR.

anorth:
Thanks for your responses @acruikshank. I agree that our disagreement is primarily about the role and scope of responsibilities of the message pool object.

I am still of the opinion that the implementation and config/policy can be profitably separated. However, the most pragmatic course here might be to merge this, and I'll try to demonstrate an alternative approach in the follow-ups for the outbound queue. If we attempt to alight too tightly now we will be in continual merge conflicts with each other on the shared code. 

I'm approving, but perhaps you could revert the receiver change and expand on in-code comments about block height.

Small responses:
- I don't expect the business logic to have a similar interface to the pool/queue, but be more like `HandleNewChainHead(...)`
- I don't think the fact that new functionality was added without the owner of the pool having an opportunity to configure it is necessarily good

core/message_pool.go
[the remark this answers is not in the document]
phritz: yeah but if the caller has a bunch of messages they want to expire at some point it is out of their control with the stamp on this side of the interface. not a big deal i dont think.

core/message_pool_test.go
```diff
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"fmt"
+	"github.com/filecoin-project/go-filecoin/testhelpers"
```
phritz: oops

core/message_pool.go
```diff
@@ -230,12 +263,49 @@ func UpdateMessagePool(ctx context.Context, pool *MessagePool, store *hamt.CborI
 		pool.Remove(cid)
 	}
 
+	// prune all messages that have been in the pool too long
+	return pool.timeoutMessages(ctx, store, new)
```
anorth: @acruikshank I think you missed the fact that this method mutates `new`, so it's not passed through as expected to this method. I think the tests you added break when this is fixed (e.g. by #2272).

PR #2238, opened 2019-03-11
Moved Ping from node API to porcelain and plumbing

### LINKED ISSUE #1980: Move Ping into Plumbing
### Description

Ping is largely a wrapper around libp2p, so it should be moved into the network plumbing, as that's where we're moving libp2p utilities.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Ping` is part of it.

# Solution

Move Ping to the plumbing porcelain.

Resolves #1980


porcelain/network.go
```diff
@@ -0,0 +1,63 @@
...
+	ch := make(chan *PingResult)
+
+	go func() {
+		defer close(ch)
+		defer cancel()
+
+		ch <- &PingResult{Text: fmt.Sprintf("PING %s", pid)}
+		for i := 0; i < int(count); i++ {
+			select {
+			case dur := <-times:
```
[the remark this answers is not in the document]
rosalinekarr: No longer applicable

porcelain/network.go
```diff
@@ -0,0 +1,63 @@
...
+// pingTimeout is the maximum timeout that is waited for when pinging another peer.
+const pingTimeout = time.Second * 10
+
+type pPlumbing interface {
+	NetworkGetPeerID() peer.ID
+	NetworkPing(ctx context.Context, pid peer.ID) (<-chan time.Duration, error)
+}
+
+// NetworkPingWithCount pings a peer repeatedly returning delay information each time
+func NetworkPingWithCount(ctx context.Context, plumbing pPlumbing, pid peer.ID, count uint, delay time.Duration) (<-chan *PingResult, error) {
```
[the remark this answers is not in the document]
rosalinekarr: Removed it and rewrote logic in commands instead.

rosalinekarr:
@phritz @anorth @travisperson I removed the `NetworkPingWithCount` porcelain at @phritz's suggestion, and rewrote some of the loop logic in commands. I also changed the logic a bit since I noticed a lot of the error cases we handled are impossible or irrelevant, and @whyrusleeping [suggested some changes](https://filecoinproject.slack.com/archives/CEHHJNJS3/p1552406762354700?thread_ts=1552404552.350200&cid=CEHHJNJS3) to the behavior of the UX command.

Now, `go-filecoin ping TARGET_PEER_ID` will continue pinging the target until Ctrl-C'd, and `go-filecoin ping TARGET_PEER_ID --count X` will ping for X times. I also discarded the timeout logic since closed channels are now handled in the loop and very long pings can be exited by the user manually.

commands/ping.go
```diff
@@ -35,29 +45,33 @@ trip latency information.
...
+			pong, pingChOpen := <-pingCh
+			result := &PingResult{
+				Count: i,
+				Time:  pong,
+			}
+			if err := re.Emit(result); err != nil {
 				return err
 			}
+			if !pingChOpen {
+				return errors.New("Ping channel closed by receiver")
```
phritz: by sender

commands/ping.go
```diff
@@ -35,29 +45,33 @@ trip latency information.
...
-			if len(p.Text) > 0 {
-				fmt.Fprintln(w, p.Text) // nolint: errcheck
-			} else if p.Success {
-				fmt.Fprintf(w, "Pong received: time=%.2f ms\n", p.Time.Seconds()*1000) // nolint: errcheck
-			} else {
-				fmt.Fprintf(w, "Pong failed\n") // nolint: errcheck
-			}
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, result *PingResult) error {
+			milliseconds := result.Time.Seconds() * 1000
+			fmt.Fprintf(w, "Pong received: seq=%d time=%.2f ms\n", result.Count, milliseconds) // nolint: errcheck
```
phritz: might be easier to understand if we just print result.Time.Round(time.Millisecond).String()? at the very least fractions of ms dont matter in this context, ms is sufficient

protocol/storage/client.go
```diff
@@ -308,16 +307,20 @@ type ClientNodeImpl struct {
 	dserv     ipld.DAGService
 	host      host.Host
 	blockTime time.Duration
-	*ping.PingService
+	plumbing  clientNodeImplPlumbing
+}
+
+type clientNodeImplPlumbing interface {
```
phritz: blocking: sorry i'm confused about why we need to do this. i expected to see ping removed from clientNode above, and added to clientPorcelainAPI above, and the call on line 203 to smc.node.Ping changed to smc.porcelainapi.ping. that eliminates node as an intermediary for this dependency. cc @anorth

rosalinekarr: `clientNodeImplPlumbing`'s dependency on Ping was introduced in a rebase. Originally, I had removed Ping from the node entirely as it wasn't needed anywhere porcelain wasn't available. After the dependency was added, I did this as an attempt to avoid re-adding porcelain to the node.

Now, I've put Ping back in the node and removed porcelain from `clientNodeImplPlumbing`.

net/network.go
```diff
@@ -49,3 +56,8 @@ func (network *Network) GetPeerID() peer.ID {
 func (network *Network) GetBandwidthStats() metrics.Stats {
 	return network.Reporter.GetBandwidthTotals()
 }
+
+// Ping pings a peer and returns a channel with the amount of time for a response
+func (network *Network) Ping(ctx context.Context, pid peer.ID) (<-chan time.Duration, error) {
```
phritz: you could avoid having this method if you wanted by directly embedding the ping service eg
```
type Network struct {
  ...
  *ping.PingService
}
```

PR #2240, opened 2019-03-11
WIP Add genesis block to binary

### LINKED ISSUE #2182: Prevent Users From Joining User Devnet with Wrong Code
### Description

We would like to prevent users from joining the user devnet unless they have the same exact code as what is deployed to the user devnet.

Some possible implementations include:
- Ensure the genesis block is unique (and easily reproducible) for each code snapshot used for the user devnet
- Have the software raise an error if it sees an attempt to connect to the user devnet with the wrong code
- Use libp2p protocol version incrementation for the user devnet.

### Acceptance criteria

As a user with any code not exactly equal to what is on user devnet, I should not be able to connect to devnet.

### Risks + pitfalls

This plan must ensure that both past _and_ future versions of the code are flagged as incompatible.

### Protocol Changes

None, as our protocol model for change has not yet been solidified

### Where to begin

The above suggestions are a great place to start.

rkowalick:
Add genesis block to binary

Ideally, resolves #2182


acruikshank:
@rkowalick can we close this in favor of  #2262?

rkowalick:
@acruikshank Yes we can.

PR #2241, opened 2019-03-11
Feat/id plum pork 1962

### LINKED ISSUE #1962: Move ID from the Node API into Plumbing Network
### Description

The ID method wraps libp2p, so it should be moved into the ntwk plumbing.

rosalinekarr:
# Problem

We want to delete the `api` directory, but ID details are part of it.

# Solution

Move ID details into commands and plumbing.

Resolves #1962


commands/id.go
```diff
@@ -22,15 +31,28 @@ var idCmd = &cmds.Command{
...
-		details, err := GetAPI(env).ID().Details()
-		if err != nil {
-			return err
+		addrs := GetPorcelainAPI(env).NetworkGetPeerAddresses()
+		hostID := GetPorcelainAPI(env).NetworkGetPeerID()
+
+		details := IDDetails{
+			Addresses: make([]ma.Multiaddr, len(addrs)),
+			ID:        hostID,
+			// TODO: fill out the other details
```
phritz: fwiw i think we should not have speculative fields. it is so misleading. i would prefer we didn't have an agentversion or protocol version or pubkey if they are not used.

ZenGround0: I know this is not your bug but could you file an issue describing the fields you are taking out so that someone can go through and add them in the future?

commands/id.go
```diff
@@ -37,7 +37,6 @@ var idCmd = &cmds.Command{
 		details := IDDetails{
 			Addresses: make([]ma.Multiaddr, len(addrs)),
 			ID:        hostID,
-			// TODO: fill out the other details
```
ZenGround0: Is there an issue corresponding to this?  I'm guessing that its being written but just flagging in case.

commands/id.go
rosalinekarr: Are we ever going to add other fields here? Unless we have specific ones in mind, I'm not sure we should write an issue for it.

PR #2243, opened 2019-03-11
Feat/addrs plum pork

### LINKED ISSUE #1865: Remove Address New from Node API
### Description

Currently, node has several methods for new addresses under the addresses module, but this functionality already exists on wallet and most of the address module code is unnecessary.

Assuming, #911 is merged, it should be possible to replace any call to `Address().Addrs().New()` with `plubmingAPI.WalletNewAddress()`.

### LINKED ISSUE #1866: Remove Address Ls from Node API
### Description

This method just calls ListAddresses on wallet. It should be removed and a plumbing API method for wallet should be added to replace it.

### LINKED ISSUE #1867: Move Address Lookup from Node API to Plumbing
### Description

ChainLookupService is already essentially extracted from the node itself. We just need to move the remaining method calls for Address Lookup from the node into the plumbing api. Likely, it can just be attached to the plumbing api object in the same way chain reader and wallet are attached.

rosalinekarr:
# Problem

We want to remove the `api` directory, but Address Addrs is part of it.

# Solution

Just remove Address Addrs. It's all redundant.

Resolves #1865 
Resolves #1866 
Resolves #1867


api/impl/retrieval_client.go
```diff
@@ -5,20 +5,31 @@ import (
...
+	"gx/ipfs/QmTu65MVbemtUxJEWgsTtzv9Zv9P8rvmqNA4eG9TrTRGYc/go-libp2p-peer"
 
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/types"
 )
 
 type nodeRetrievalClient struct {
-	api *nodeAPI
+	api      *nodeAPI
+	plumbing retrievalClientPorcelainAPI
```
shannonwells: I'm confused; it's a PorcelainAPI but it's called plumbing?

rosalinekarr: Sorry, this is just bad naming on my part

Issue #2049, opened 2019-02-20
Provide porcelain and cli command to retrieve default wallet address


travisperson:
### Description

Currently the only way to retrieve the default wallet address is by using the config. However, this may not be set when it is needed.

https://github.com/filecoin-project/go-filecoin/pull/1864#discussion_r258696860

### Acceptance criteria

- Users should be able to retrieve their default wallet address from the CLI.

jaybatra26:
Is this taken. If not I would like to contribute.

GeetikaBatra:
Can I take this one?

waynewyang:
@travisperson I think I can add  "DefaultWalletAddress" method in [go-filecoin/porcelain/wallet.go](https://github.com/filecoin-project/go-filecoin/blob/master/porcelain/wallet.go), references to  "GetAndMaybeSetDefaultSenderAddress" in [go-filecoin/porcelain/message.go](https://github.com/filecoin-project/go-filecoin/blob/master/porcelain/message.go),
and add the CLI "go-filecoin wallet addrs default", is it feasible?

travisperson:
That seems like a reasonable approach.

waynewyang:
@travisperson ok, thanks,I'll make a PR later.

PR #2250, opened 2019-03-12
Provide porcelain and cli command to retrieve default wallet address

waynewyang:
Provide porcelain and cli command to retrieve default wallet address
1 Add the method DefaultWalletAddress
2 Add the cli command "go-filecoin address default"

referenced #2049

thanks


anorth:
@rosalinekarr could you please shepherd this contribution, and add a maintainer for review when ready?

porcelain/wallet.go
```diff
@@ -27,3 +32,14 @@ func WalletBalance(ctx context.Context, plumbing walletPlumbing, addr address.Ad
...
+
+// DefaultWalletAddress returns a default wallet address from the config.
+func DefaultWalletAddress(plumbing walletPlumbing) (address.Address, error) {
+	ret, err := plumbing.ConfigGet("wallet.defaultAddress")
+	addr := ret.(address.Address)
+	if err != nil || addr != (address.Address{}) {
+		return addr, err
+	}
+
+	return address.Address{}, ErrNoDefaultWalletAddress
```
rosalinekarr: If no address is present in the config, we should attempt to set one using the first available address in the wallet store. Check out the `GetAndMaybeSetDefaultSenderAddress` method in `porcelain/message.go` around line 60. I believe this method does what you're looking for here.

This method has a better name, so you could move that method's logic here, delete the old one, and replace any invocations of `GetAndMaybeSetDefaultSenderAddress` with `DefaultWalletAddress `.

waynewyang: Ok, thank you, I'll fix it tomorrow.

rosalinekarr: Thanks!

waynewyang: I've fix it just now, please let me know if there is any problems, thanks!

anorth:
Thank you! It looks like some files have changed under you, so please rebase and @rosalinekarr can merge after maintainer review.

porcelain/api.go
```diff
@@ -144,18 +144,17 @@ func (a *API) MinerPreviewSetPrice(
...
-}
-
 // WalletBalance returns the current balance of the given wallet address.
 func (a *API) WalletBalance(ctx context.Context, address address.Address) (*types.AttoFIL, error) {
 	return WalletBalance(ctx, a, address)
 }
 
+// DefaultWalletAddress returns a default wallet address from the config.
+// If none is set it picks the first address in the wallet and sets it as the default in the config.
+func (a *API) DefaultWalletAddress() (address.Address, error) {
```
phritz: WalletDefaultAddress would be more idomatic and cause this method to sort with its friends. Would you mind making the update?

porcelain/wallet.go
```diff
@@ -2,15 +2,22 @@ package porcelain
...
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// ErrNoDefaultFromAddress is returned when a default wallet address couldn't be determined (eg, there are zero addresses in the wallet).
+var ErrNoDefaultFromAddress = errors.New("unable to determine a default wallet address")
+
 type walletPlumbing interface {
 	ActorGet(ctx context.Context, addr address.Address) (*actor.Actor, error)
+	ConfigGet(dottedPath string) (interface{}, error)
```
phritz: This mixes the dependencies for these two calls together. They should be separate so we can what what each depends on. Would you please break this into two interfaces, wbPlumbing which has actorget for walletbalance and wdaPlumbing which has configget/set/walletaddresses for walletdefaultadddress? then ensure that walletdefaultaddress takes wdaPlumbing.

porcelain/wallet.go
```diff
@@ -27,3 +34,27 @@ func WalletBalance(ctx context.Context, plumbing walletPlumbing, addr address.Ad
...
+		addr := plumbing.WalletAddresses()[0]
+		err := plumbing.ConfigSet("wallet.defaultAddress", addr.String())
+		if err != nil {
+			return address.Address{}, err
+		}
+
+		return addr, nil
+	}
+
+	return address.Address{}, ErrNoDefaultFromAddress
```
phritz: this is so new you could never have know about this, but we now prefer to return address.Undef instead of the empty struct.

porcelain/api.go
waynewyang: thanks, agree with you.

porcelain/wallet.go
waynewyang: thanks,I've done it

porcelain/wallet.go
waynewyang: I've fix it yesterday, thanks.

porcelain/wallet_test.go
```diff
@@ -39,3 +73,54 @@ func TestWalletBalance(t *testing.T) {
...
+
+		wdatp := newWdaTestPlumbing(require)
+
+		addrA, err := wdatp.WalletNewAddress()
+		require.NoError(err)
+		wdatp.ConfigSet("wallet.defaultAddress", addrA.String())
+
+		addrB, err := porcelain.WalletDefaultAddress(wdatp)
+		require.NoError(err)
+		assert.Equal(addrA.String(), addrB.String())
```
phritz: i think you shouldn't need to stringify them to compare

protocol/storage/client.go
```diff
@@ -63,7 +63,7 @@ type clientPorcelainAPI interface {
 	DealGet(cid.Cid) *storagedeal.Deal
 	DealPut(*storagedeal.Deal) error
 	DealsLs() ([]*storagedeal.Deal, error)
-	GetAndMaybeSetDefaultSenderAddress() (address.Address, error)
+	WalletDefaultAddress() (address.Address, error)
```
phritz: please sort this lexicographically more towards the bottom

porcelain/miner.go
```diff
@@ -24,7 +24,7 @@ import (
 type mcAPI interface {
 	ConfigGet(dottedPath string) (interface{}, error)
 	ConfigSet(dottedPath string, paramJSON string) error
-	GetAndMaybeSetDefaultSenderAddress() (address.Address, error)
+	WalletDefaultAddress() (address.Address, error)
```
phritz: please sort this lexicographically more towards the bottom of this list. this placement made sense when it started with 'g', but now it starts with 'w'.

waynewyang:
@phritz @rosalinekarr I've Addressed it(sort the methods lexicographically) ,thanks

rosalinekarr:
@waynewyang Looks great! I'll merge it

rosalinekarr:
@waynewyang Thanks for the contribution!

PR #2262, opened 2019-03-13
Simple way to guard user devnet

### LINKED ISSUE #2182: Prevent Users From Joining User Devnet with Wrong Code

rkowalick:
A way less crazy way to resolve #2182 

## Problem

Users can easily join the user devnet with the wrong version of the filecoin codebase

## Solution

Require a specific code version for user devnet in the hello protocol. This version specifier is capable of being overwritten during build time if a developer needs to test something with a different code version.
    
    
Additionally:
    - Clean up some redundant or repetitive code in daemon.Init
    - Bolster checks for using multiple `devnet` flags at once

### DIFF
--- a/api/impl/daemon.go
+++ b/api/impl/daemon.go
@@ -16,7 +16,6 @@ import (
 	car "gx/ipfs/QmUGpiTCKct5s1F7jaAnY9KJmoo7Qm1R2uhSjq5iHDSUMn/go-car"
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
 
-	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/api"
 	"github.com/filecoin-project/go-filecoin/config"
 	"github.com/filecoin-project/go-filecoin/consensus"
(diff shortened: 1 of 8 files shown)


protocol/hello/hello.go
```diff
@@ -96,10 +106,16 @@ func (h *Handler) handleNewStream(s net.Stream) {
 // ErrBadGenesis is the error returned when a missmatch in genesis blocks happens.
 var ErrBadGenesis = fmt.Errorf("bad genesis block")
 
+// ErrWrongVersion is the error returned when a missmatch in the code version happens.
+var ErrWrongVersion = fmt.Errorf("bad genesis block")
```
travisperson: copy paste bug

api/impl/daemon.go
```diff
@@ -91,57 +90,42 @@ func (nd *nodeDaemon) Init(ctx context.Context, opts ...api.DaemonInitOpt) error
 
 	initopts = append(initopts, node.AutoSealIntervalSecondsOpt(cfg.AutoSealIntervalSeconds))
 
-	if cfg.WithMiner != (address.Address{}) {
-		newConfig := rep.Config()
-		newConfig.Mining.MinerAddress = cfg.WithMiner
-		if err := rep.ReplaceConfig(newConfig); err != nil {
-			return err
-		}
+	if cfg.DevnetTest && cfg.DevnetNightly || cfg.DevnetTest && cfg.DevnetUser || cfg.DevnetNightly && cfg.DevnetUser {
```
acruikshank: The associativity here isn't clear (I forget the rule of && and ||s). Please parenthesize.

rkowalick: done and done

PR #2263, opened 2019-03-13
Fix units for value in message send

### LINKED ISSUE #1819: [bug] `message send` sends in FIL, not AttoFIL as it describes in usage
**Describe the bug**

From the `message send` help, the `--value` option is suppose to be AttoFIL. However, it is implemented as FIL.

>   --value int    - Value to send with message, in AttoFIL.

https://github.com/filecoin-project/go-filecoin/blob/726e6705860ddfc8ca4e55bc3610ad2230a95c0c/commands/message.go#L77

rkowalick:
Resolves #1819


rkowalick:
@travisperson I have just filed #2265

Issue #2257, opened 2019-03-13
Instrument Metrics on ProcessBlock time and Number of ProcessBlock calls


frrist:
### Description
Add instrumentation to process block. Metrics should include:
- Process Block duration
- Process Block call count

### Acceptance Criteria
- Process Block duration metric is accessible to infrastructure via prometheus
- Process Block count metric is accessible to infrastructure via prometheus

PR #2266, opened 2019-03-13
Feat/process block metric

frrist:
### What
This PR addresses https://github.com/filecoin-project/go-filecoin/issues/2257, to facilitate this we have:
- Added a metrics config, this is used to configure how and where the filecoin node exposes prometheus metrics
- Added a prometheus metric via opencensus to expose how long `ProcessBlock` and `ApplyMessage` take.

These metrics will give us the average duration of each methods execution, the number of times the method is called and a distribution of durations [defined here](https://github.com/filecoin-project/go-filecoin/pull/2266/files#diff-6bdbd3e56444b9285aa20f6442e879bdR24).

### DIFF
--- a/build/main.go
+++ b/build/main.go
@@ -124,8 +124,6 @@ func deps() {
 		cmd("go get -u github.com/docker/docker/pkg/stdcopy"),
 		cmd("go get -u github.com/ipsn/go-secp256k1"),
 		cmd("go get -u github.com/json-iterator/go"),
-		cmd("go get -u github.com/prometheus/client_golang/prometheus"),
-		cmd("go get -u github.com/prometheus/client_golang/prometheus/promhttp"),
 		cmd("go get -u github.com/jstemmer/go-junit-report"),
 		cmd("go get -u github.com/pmezard/go-difflib/difflib"),
 		cmd("./scripts/install-rust-fil-proofs.sh"),
--- a/config/config.go
+++ b/config/config.go
@@ -25,6 +25,7 @@ type Config struct {
 	Wallet    *WalletConfig    `json:"wallet"`
 	Heartbeat *HeartbeatConfig `json:"heartbeat"`
 	Net       string           `json:"net"`
+	Metrics   *MetricsConfig   `json:"metrics"`
 }
 
 // APIConfig holds all configuration options related to the api.
@@ -148,6 +149,24 @@ func newDefaultHeartbeatConfig() *HeartbeatConfig {
 	}
 }
 
+// MetricsConfig holds all configuration options related to node metrics.
+type MetricsConfig struct {
+	// Enabled will enable prometheus metrics when true.
+	PrometheusEnabled bool `json:"prometheusEnabled"`
+	// ReportInterval represents how frequently filecoin will update its prometheus metrics.
+	ReportInterval string `json:"reportInterval"`
+	// PrometheusEndpoint represents the address filecoin will expose prometheus metrics at.
+	PrometheusEndpoint string `json:"prometheusEndpoint"`
+}
+
+func newDefaultMetricsConfig() *MetricsConfig {
+	return &MetricsConfig{
+		PrometheusEnabled:  false,
+		ReportInterval:     "5s",
+		PrometheusEndpoint: "/ip4/0.0.0.0/tcp/9400",
+	}
+}
+
 // NewDefaultConfig returns a config object with all the fields filled out to
 // their default values
 func NewDefaultConfig() *Config {
@@ -160,6 +179,7 @@ func NewDefaultConfig() *Config {
 		Wallet:    newDefaultWalletConfig(),
 		Heartbeat: newDefaultHeartbeatConfig(),
 		Net:       "",
+		Metrics:   newDefaultMetricsConfig(),
 	}
 }
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -79,7 +79,12 @@ func TestWriteFile(t *testing.T) {
 		"reconnectPeriod": "10s",
 		"nickname": ""
 	},
-	"net": ""
+	"net": "",
+	"metrics": {
+		"prometheusEnabled": false,
+		"reportInterval": "5s",
+		"prometheusEndpoint": "/ip4/0.0.0.0/tcp/9400"
+	}
 }`,
 		string(content),
 	)
--- a/consensus/processor.go
+++ b/consensus/processor.go
@@ -3,16 +3,24 @@ package consensus
 import (
 	"context"
 	"math/big"
-	"time"
 
 	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/account"
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/metrics"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/vm"
 	"github.com/filecoin-project/go-filecoin/vm/errors"
 )
+
+var pbTimer *metrics.Float64Timer
+var amTimer *metrics.Float64Timer
+
+func init() {
+	amTimer = metrics.NewTimer("consensus/apply_message", "Duration of message application in milliseconds")
+	pbTimer = metrics.NewTimer("consensus/process_block", "Duration of block processing in milliseconds")
+}
 
 // BlockRewarder applies all rewards due to the miner's owner for processing a block including block reward and gas
 type BlockRewarder interface {
@@ -95,9 +103,10 @@ func NewConfiguredProcessor(validator SignedMessageValidator, rewarder BlockRewa
 func (p *DefaultProcessor) ProcessBlock(ctx context.Context, st state.Tree, vms vm.StorageMap, blk *types.Block, ancestors []types.TipSet) ([]*ApplicationResult, error) {
 	var emptyResults []*ApplicationResult
 
-	processBlkTimer := time.Now()
+	pbsw := pbTimer.Start(ctx)
 	defer func() {
-		log.Infof("[TIMER] DefaultProcessor.ProcessBlock BlkCID: %s - elapsed time: %s", blk.Cid(), time.Since(processBlkTimer).Round(time.Millisecond))
+		dur := pbsw.Stop(ctx)
+		log.Infof("[TIMER] DefaultProcessor.ProcessBlock BlkCID: %s - elapsed time: %s", blk.Cid(), dur)
 	}()
 
 	// find miner's owner address
@@ -267,9 +276,10 @@ func (p *DefaultProcessor) ApplyMessage(ctx context.Context, st state.Tree, vms 
 		return nil, errors.FaultErrorWrap(err, "could not get message cid")
 	}
 
-	applyMsgTimer := time.Now()
+	amsw := amTimer.Start(ctx)
 	defer func() {
-		log.Infof("[TIMER] DefaultProcessor.ApplyMessage CID: %s - elapsed time: %s", msgCid.String(), time.Since(applyMsgTimer).Round(time.Millisecond))
+		dur := amsw.Stop(ctx)
+		log.Infof("[TIMER] DefaultProcessor.ApplyMessage CID: %s - elapsed time: %s", msgCid.String(), dur)
 	}()
 
 	cachedStateTree := state.NewCachedStateTree(st)
--- a/metrics/heartbeat_test.go
+++ b/metrics/heartbeat_test.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/config"
-	"github.com/filecoin-project/go-filecoin/fixtures"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -137,7 +136,7 @@ func TestHeartbeatRunSuccess(t *testing.T) {
 	expHeight := types.Uint64(444)
 	expTs := mustMakeTipset(t, expHeight)
 
-	addr, err := address.NewFromString(fixtures.TestAddresses[0])
+	addr, err := address.NewActorAddress([]byte("miner address"))
 	require.NoError(err)
 
 	// The handle method will run the assertions for the test
--- a/metrics/timers.go
+++ b/metrics/timers.go
@@ -1,0 +1,66 @@
+package metrics
+
+import (
+	"context"
+	"time"
+
+	"gx/ipfs/QmNVpHFt7QmabuVQyguf8AbkLDZoFh7ifBYztqijYT1Sd2/go.opencensus.io/stats"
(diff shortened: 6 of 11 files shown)


consensus/processor.go
```diff
@@ -5,9 +5,13 @@ import (
 	"math/big"
 	"time"
 
+	"go.opencensus.io/stats"
```
acruikshank: Hopefully we'll be on go modules soon, but until then, these need to be gx-ified.

config/config_test.go
```diff
@@ -77,7 +77,12 @@ func TestWriteFile(t *testing.T) {
 		"reconnectPeriod": "10s",
 		"nickname": ""
 	},
-	"net": ""
+	"net": "",
+	"metrics": {
+		"enabled": false,
+		"reportInterval": "3s",
+		"prometheusEndpoint": "/ip4/0.0.0.0/tcp/8888"
```
acruikshank: What would it look like to support another (or multiple) monitoring system? Opencensus seems to support many exporters. If we were to, say, support Graphite, would we add a `graphiteEndpoint`? If so, we would either need to set the endpoint to nil to indicate it's enabled (and no enable flag) or have separate enable flags for each endpoint.

My preference would be the latter:
``` go
"metrics": {
		"reportInterval": "3s",
		"prometheusEnabled": false,
		"prometheusEndpoint": "/ip4/0.0.0.0/tcp/8888"
		"graphiteEnabled": true,
		"graphiteEndpoint": "/ip4/0.0.0.0/tcp/2003"
}
```

If you agree, maybe we should rename `enabled` to `prometheusEnabled` now.

metrics/setup.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+	"time"
+
+	ma "gx/ipfs/QmNTCey11oxhb1AxDnQBRHtdhap6Ctud872NjAYPYYXPuc/go-multiaddr"
+	manet "gx/ipfs/QmZcLBXKaFe8ND5YHPkJRAwmhJGrVsi1JqDZNyJ4nRK5Mj/go-multiaddr-net"
+
+	prom "github.com/prometheus/client_golang/prometheus"
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/zpages"
```
acruikshank: These also need to be gxed.

build/main.go
```diff
@@ -149,6 +154,10 @@ func smartdeps() {
 	cmds := []command{
 		cmd("gx install"),
 		cmd("gx-go rewrite"),
+		// because we call go install on pkgs and that will fail
+		cmd("go get -u go.opencensus.io"),
+		// because go getting opencensus doesn't actually get everything :(
+		cmd("go get -u google.golang.org/grpc"),
```
acruikshank: BLOCKING We should not be implementing ad hoc package management for production code. The opencensus code needs to come in through gx like everything else (unless the go modules PR lands soon).

consensus/processor.go
frrist: I can do that -- think they already exist somewhere as gx packages.

config/config_test.go
frrist: This is definitely worth considering down the road, but I think it may be outside the scope of this PR, we don't use graphite anywhere so I would prefer not to add configuration options for it here. Is that alright with you?

metrics/setup.go
frrist: I will gx all opencensus and prometheus imports

build/main.go
frrist: agreed, I did this to get the thing working, will remove these on gx add.

config/config_test.go
acruikshank: Non-blocking: the suggestion is just to rename the enabled flag to prometheusEnabled to future proof the config. It's not a big deal.

package.json
```diff
@@ -194,6 +194,18 @@
       "hash": "QmceTjrpdhYXocRzx9hMEwztLyWUEvyDdRqrkSjLQeq6pE",
       "name": "go-libp2p-autonat-svc",
       "version": "1.0.10"
+    },
+    {
+      "author": "hsanjuan",
+      "hash": "QmNVpHFt7QmabuVQyguf8AbkLDZoFh7ifBYztqijYT1Sd2",
+      "name": "go.opencensus.io",
```
acruikshank: 👍

build/main.go
```diff
@@ -124,8 +124,6 @@ func deps() {
 		cmd("go get -u github.com/docker/docker/pkg/stdcopy"),
 		cmd("go get -u github.com/ipsn/go-secp256k1"),
 		cmd("go get -u github.com/json-iterator/go"),
-		cmd("go get -u github.com/prometheus/client_golang/prometheus"),
```
frrist: have been gx'd

config/config.go
```diff
@@ -60,6 +61,7 @@ type DatastoreConfig struct {
 // arguments, and must return either an error or nil depending on whether or not
 // the given key and value are valid. Validators will only be run if a property
 // being set matches the name given in this map.
+// TODO add validators for BeatTarget and PrometheusEndpoint asserting they are valid addresses
```
anorth: Please remove this in favour of an issue, or fix it now.

consensus/processor.go
```diff
@@ -95,9 +99,16 @@ func NewConfiguredProcessor(validator SignedMessageValidator, rewarder BlockRewa
 func (p *DefaultProcessor) ProcessBlock(ctx context.Context, st state.Tree, vms vm.StorageMap, blk *types.Block, ancestors []types.TipSet) ([]*ApplicationResult, error) {
 	var emptyResults []*ApplicationResult
 
+	ctx, err := tag.New(ctx, tag.Insert(metrics.KeyMethod, "ProcessBlock"))
```
anorth: Blocking: Errors here are very unfortunate, but we should not wedge the node due to a metrics failure. Log an ERROR instead (or better, have the error happen much sooner like at node initialisation). I suggest factoring the 4-line block out to a simple helper function in `metrics` to reduce noise and document what it's doing.

Non-blocking, but could you confirm whether this is the best practise way of defining the metrics? I would have expected it to happen at construction, in a factory method, or possibly statically (or DI) rather than at point of use.

Edit: in light of metrics/constants.go I'm not sure what this is doing at all, actually.

metrics/constants.go
```diff
@@ -0,0 +1,39 @@
...
+	applyMessageView = &view.View{
+		Name:        "consensus/apply_message",
+		Measure:     MApplyMessageMs,
+		Description: "The distribution of the durations",
+
+		// Latency in buckets:
+		// [>=0ms, >=25ms, >=50ms, >=75ms, >=100ms, >=200ms, >=400ms, >=600ms, >=800ms, >=1s, >=2s, >=4s, >=8s]
+		Aggregation: view.Distribution(25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 8000),
+		TagKeys:     []tag.Key{KeyMethod}}
+)
```
anorth: Blocking: this seems to be doing multiple things, it's not sufficiently documented what they are, and I think too centralised/static.

I infer that the `MProcessBlockMs` and `MApplyMessageMs` are definitions of the things that actually track metrics, "measures", and are referenced from `processor.go`. We should not have an explicit global list of all of these measures that's imported by everything in the code in this way. Better would be for the metrics to be defined by the modules exporting them. Even better would be for them to be less static so that they can be faked, tested and so the other stuff here doesn't need to import the world either (NB actually doing any of that faking and testing is way outside scope here).

I infer that the "View" stuff is about how values are exported to Prometheus. Is that right? It's not so bad for that to be centralised/static, but I do think we could do a lot better than all this boilerplate. The name and description belong with the metric definition in the implementing module. I would suggest for now that we can put the aggregation stuff there for now too and defer more complex aggregations to later.

So my summary comes down to moving almost all of this into processor.go. I recommend a helper function to remove the boilerplate and repetition of Measure/View construction, so a caller can do something like:

```
// In processor.go and other modules
// Under the hood this registers a view for exporting with prometheus.
pbLatency := metrics.Registry().NewTimer("consensus/process_block", "The duration in milliseconds of ProcessBlock", stats.UnitMilliseconds) // default aggregations and tags
```

An ideal outcome is that almost none of the go-filecoin code imports `opencensus` directly, but I can't quite tell whether that's reasonable yet.

metrics/setup.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"gx/ipfs/QmNVpHFt7QmabuVQyguf8AbkLDZoFh7ifBYztqijYT1Sd2/go.opencensus.io/exporter/prometheus"
+	"gx/ipfs/QmNVpHFt7QmabuVQyguf8AbkLDZoFh7ifBYztqijYT1Sd2/go.opencensus.io/stats/view"
+	"gx/ipfs/QmNVpHFt7QmabuVQyguf8AbkLDZoFh7ifBYztqijYT1Sd2/go.opencensus.io/zpages"
+	prom "gx/ipfs/QmaQtvgBNGwD4os5VLWtBLR6HM6TY6ApX6xFqSnfjDF2aW/client_golang/prometheus"
+
+	"github.com/filecoin-project/go-filecoin/config"
+)
+
+// SetupMetrics registers and serves prometheus metrics
```
anorth: Rename to prometheus.go (or export.go) and specialise it to just that functionality. See comments elsewhere.

consensus/processor.go
frrist: Tags are used in a few of the opencensus [examples](https://github.com/census-instrumentation/opencensus-go/tree/master/examples), and according to the [opencensus spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md) tags are:
 >used to label anything that is associated with a specific operation, such as an HTTP request.

I think adding tags is is more reduant than helpful since we already get the method name via the view -- I will remove them.

metrics/constants.go
frrist: I will prototype something like this and push it up for discussion

consensus/processor.go
```diff
@@ -8,12 +8,23 @@ import (
...
 	"github.com/filecoin-project/go-filecoin/vm"
 	"github.com/filecoin-project/go-filecoin/vm/errors"
 )
 
+var consensusRegistry *metrics.Registry
+var pbTimer *metrics.Float64Timer
+var amTimer *metrics.Float64Timer
+
+func init() {
+	consensusRegistry = metrics.NewRegistry()
```
frrist: I'm not sure where a good place for this is yet, I don't think it belongs here (dare I say we put it on node?). However since it is only referenced here I think we can leave it for a follow on.

metrics/heartbeat_test.go
```diff
@@ -19,7 +19,6 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/config"
-	"github.com/filecoin-project/go-filecoin/fixtures"
```
frrist: This was done to remove a circular dep with `metrics/` and `config/`

metrics/timers.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Float64Timer contains a opencensus measurement and view
+type Float64Timer struct {
+	measure *stats.Float64Measure
+	view    *view.View
+}
+
+// Begin starts a timer and returns a Stopwatch.
+func (t *Float64Timer) Begin(ctx context.Context) *Stopwatch {
```
frrist: Thinking of a better name, open to suggestions -- maybe `Time`?

consensus/processor.go
```diff
@@ -8,12 +8,23 @@ import (
...
+	amTimer = consensusRegistry.NewTimer("consensus/apply_message", "Duration of message application")
+	pbTimer = consensusRegistry.NewTimer("consensus/process_block", "Duration of block processing")
```
anorth: Nit: how about camelCase rather than snake_case? Less typing, and idiomatic in many languages except Python.

consensus/processor.go
```diff
@@ -95,9 +106,12 @@ func NewConfiguredProcessor(validator SignedMessageValidator, rewarder BlockRewa
 func (p *DefaultProcessor) ProcessBlock(ctx context.Context, st state.Tree, vms vm.StorageMap, blk *types.Block, ancestors []types.TipSet) ([]*ApplicationResult, error) {
 	var emptyResults []*ApplicationResult
 
+	pbsw := pbTimer.Begin(ctx)
 	processBlkTimer := time.Now()
 	defer func() {
-		log.Infof("[TIMER] DefaultProcessor.ProcessBlock BlkCID: %s - elapsed time: %s", blk.Cid(), time.Since(processBlkTimer).Round(time.Millisecond))
+		dur := time.Since(processBlkTimer).Round(time.Millisecond)
+		log.Infof("[TIMER] DefaultProcessor.ProcessBlock BlkCID: %s - elapsed time: %s", blk.Cid(), dur)
+		pbsw.Stop(ctx)
```
anorth: Great. Maybe we can return the duration from `Stop()` and avoid the need for a separate time measurement just for logging.

metrics/timers.go
```diff
@@ -0,0 +1,87 @@
...
+
+// NewTimer creates a Float64Timer that wraps an opencensus flat64 measurement.
+// The time defaults to milliseconds.
+func (r *Registry) NewTimer(name, desc string) *Float64Timer {
+	r.msurMu.Lock()
+	defer r.msurMu.Unlock()
+	timer, ok := r.measures[name]
+	if ok {
+		return timer
+	}
```
anorth: Consider failing hard here - it's probably an inadvertent error and two semantically different measurements will be conflated.

metrics/timers.go
```diff
@@ -0,0 +1,87 @@
...
+		measure: fMeasure,
+		view:    fView,
+	}
+	r.measures[name] = timer
+	return timer
+}
+
+// Float64Timer contains a opencensus measurement and view
+type Float64Timer struct {
+	measure *stats.Float64Measure
```
anorth: Document the units (milliseconds?). Even better, name it `measureMs`.

metrics/timers.go
anorth: `Start` would be symmetric with `Stop` and natural for a stopwatch.

metrics/timers_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert := assert.New(t)
+
+	ctx := context.Background()
+	reg := NewRegistry()
+	assert.Equal(len(reg.measures), 0)
+
+	testTimer := reg.NewTimer("testName", "testDesc")
+	// some view state is kept around after tests exit, doing this to clean that up.
+	// e.g. view will remain registered after a test exits.
+	defer view.Unregister(testTimer.view)
```
anorth: Won't the registry be garbage collected and take everything with it? Why is there no unregistration in the other tests?

consensus/processor.go
frrist: Prometheus prefers sneak case: https://prometheus.io/docs/practices/naming/

metrics/timers.go
frrist: This comment has made me realize the registry isn't needed, removing it.

metrics/timers_test.go
frrist: I did not observe the garbage collector running between the tests in these files, without this line the code on line 41 panics.

PR #2268, opened 2019-03-13
Pull block mining commands out of node api.

shannonwells:
# Problem
Protocol things need to be in their own area with their own API(s), starting with consensus block mining. 

# Quick and Dirty Solution
Yank block mining out of nodeAPI.  MiningOnce, StartMining and StopMining are now part of the `protocolapi` package. For `StartMining`/`StopMining`, for now just give the new `protocolapi.API` a handle to call `node.StartMining` and `node.StopMining`.  Since node may need to create a SectorBuilder when mining starts, and there are other intricacies like mutex locks that happen when we set mining state, this is done for the sake of speed.

# Questions
I'm not sure how to package this.  When I put it into the protocol directory it had problems recognizing where the api functions were, so I left them in their own directory, like porcelain does. Suggestions welcome.

P.S.  🎼Nodebody nodes the trouble I've seen...🎶


api/api.go
```diff
@@ -9,7 +9,6 @@ type API interface {
 	Daemon() Daemon
 	Dag() Dag
 	Miner() Miner
-	Mining() Mining
```
shannonwells: Get out.

commands/mining.go
```diff
@@ -50,9 +50,7 @@ var miningStartCmd = &cmds.Command{
 
 var miningStopCmd = &cmds.Command{
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		if err := GetAPI(env).Mining().Stop(req.Context); err != nil {
```
shannonwells: Mining().Stop() already was never returning anything but nil error, same as node.StopMining(), so no point in checking for a fake error.

protocolapi/block_mining_api.go
```diff
@@ -0,0 +1,160 @@
...
+		syncer:            syncer,
+		ticketSigner:      signer,
+	}
+
+	return API{apiDeps: newDeps}
+}
+
+// MineOnce mines a single block in the given context, and returns the new block.
+func (a *API) MineOnce(ctx context.Context) (*types.Block, error) {
+
```
shannonwells: this is more or less identical to the now-deleted Once function in node API

commands/daemon.go
```diff
@@ -122,8 +123,23 @@ func getRepo(req *cmds.Request) (repo.Repo, error) {
 	return repo.OpenFSRepo(getRepoDir(req))
 }
 
-func runAPIAndWait(ctx context.Context, node *node.Node, config *config.Config, req *cmds.Request) error {
-	api := impl.New(node)
+func runAPIAndWait(ctx context.Context, nd *node.Node, config *config.Config, req *cmds.Request) error {
+	api := impl.New(nd)
+	blockTime, mineDelay := nd.MiningTimes()
+	protocolAPI := bapi.New(
```
phritz: strong suggestion, though i guess not strictly blocking: i think it makes sense to keep all the construction logic together in one place until we have a chance to refactor it. suggest that this protocol api be instantiated in config.build somewhere around https://github.com/filecoin-project/go-filecoin/blob/124cd65964c2d1a55c0816fd0d5700087d1a8099/node/node.go#L403, set on node, and then we can pull it out of node here like we do the porcelainapi.

down the road node just becomes a constructor and container for the various layers: core apis (repo, wallet, cborstore, host, etc), plumbing, porcelain, protocols, and protocol apis.

commands/daemon.go
```diff
@@ -122,8 +123,23 @@ func getRepo(req *cmds.Request) (repo.Repo, error) {
...
+		nd.AddNewBlock,
+		nd.Blockstore,
+		nd.CborStore(), nd.OnlineStore,
```
phritz: for some reason this one and minedely are two to a line

commands/daemon.go
```diff
@@ -133,7 +149,8 @@ func runAPIAndWait(ctx context.Context, node *node.Node, config *config.Config,
 		// TODO: should this be the passed in context?
 		ctx:          context.Background(),
 		api:          api,
-		porcelainAPI: node.PorcelainAPI,
+		porcelainAPI: nd.PorcelainAPI,
+		protocolAPI:  &protocolAPI,
```
phritz: it might make sense to shoot for separate api containers for each protocol to make it easier to pull out or use individual protocols, eg if you wanted to embed just the retreival protocol in your binary. no action required i dont think.

protocolapi/block/api.go
```diff
@@ -0,0 +1,160 @@
+package block
```
phritz: i think it would be simplest to have the api live in the same package as the protocol. however it seems like something is preventing that ("When I put it into the protocol directory it had problems recognizing where the api functions were"), but i'm having trouble understanding what the problem is. can you help me understand?

protocolapi/block/api.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// API provides an interface to protocols.
+type API struct {
+	apiDeps
+}
+
+type apiDeps struct {
+	addNewBlockFunc   func(context.Context, *types.Block) (err error)
```
phritz: ugh, half these deps are for the default worker. it would be really nice to introduce a deps struct to the worker, then we could pass in its deps here in that struct and be able to see the actual deps of this dude. would you mind filing an issue that we can use to track protocol cleanups that we want to do after we kill the node? this one should go in that list, and there is one below.

protocolapi/block/api.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if err := a.addNewBlockFunc(ctx, res.NewBlock); err != nil {
+		return nil, err
+	}
+
+	return res.NewBlock, nil
+}
+
+// MiningStart calls the node's StartMining function
+func (a *API) MiningStart(ctx context.Context) error {
```
phritz: i think we should pass the mining node's mining worker to the api and have the api call directly into it. probably all the bookkeeping stuff strewn throughout the node that keeps track of mining state should move here as well. can you add this to the list of cleanups?

phritz:
also cc @anorth as the direction that protocol api is evolving. we are doing the expedient thing here.

commands/daemon.go
shannonwells: it was just to follow the declaration format. I can put in the newlines.

commands/daemon.go
shannonwells: it's self-contained right now anyway so I could just rename it.

protocolapi/block/api.go
shannonwells: > would you mind filing an issue that we can use to track protocol cleanups that we want to do after we kill the node? this one should go in that list, and there is one below.

Ok: #2276

protocolapi/block/api.go
shannonwells: #2277

.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 
-parameters:
+params:
```
phritz: ?

commands/daemon.go
```diff
@@ -122,8 +123,23 @@ func getRepo(req *cmds.Request) (repo.Repo, error) {
 	return repo.OpenFSRepo(getRepoDir(req))
 }
 
-func runAPIAndWait(ctx context.Context, node *node.Node, config *config.Config, req *cmds.Request) error {
-	api := impl.New(node)
+func runAPIAndWait(ctx context.Context, nd *node.Node, config *config.Config, req *cmds.Request) error {
+	api := impl.New(nd)
+	blockTime, mineDelay := nd.MiningTimes()
+	blockpAPI := bapi.New(
```
phritz: boo on not moving this into https://github.com/filecoin-project/go-filecoin/blob/cdf18316b9f65b9d474b56520f944fef4081fcba/node/node.go#L314 but ok

.circleci/config.yml
shannonwells: that's the CI fix; it was just merged into master. the diff should have gone away

PR #2269, opened 2019-03-13
Implemented chain tipset and block iterators.

anorth:
My primary goal here is to factor chain traversal out of the MessageQueue so I can re-use
logic for collecting new/old messages in the case of a re-org and expire messages in the
outbound queue. I'm informed the iterators may be useful elsewhere too.

### DIFF
--- a/chain/traversal.go
+++ b/chain/traversal.go
@@ -1,0 +1,64 @@
+package chain
+
+import (
+	"context"
+
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// BlockProvider provides blocks. This is a subset of the ReadStore interface.
+type BlockProvider interface {
+	GetBlock(ctx context.Context, cid cid.Cid) (*types.Block, error)
+}
+
+// GetParentTipSet returns the parent tipset of a tipset.
+// The result is empty if the tipset has no parents (including if it is empty itself)
+func GetParentTipSet(ctx context.Context, store BlockProvider, ts types.TipSet) (types.TipSet, error) {
+	newTipSet := types.TipSet{}
+	parents, err := ts.Parents()
+	if err != nil {
+		return nil, err
+	}
+	for it := parents.Iter(); !it.Complete() && ctx.Err() == nil; it.Next() {
+		newBlk, err := store.GetBlock(ctx, it.Value())
+		if err != nil {
+			return nil, err
+		}
+		if err := newTipSet.AddBlock(newBlk); err != nil {
+			return nil, err
+		}
+	}
+	return newTipSet, nil
+}
+
+// IterAncestors returns an iterator over tipset ancestors, yielding first the start tipset and
+// then its parent tipsets until (and including) the genesis tipset.
+func IterAncestors(ctx context.Context, store BlockProvider, start types.TipSet) *TipsetIterator {
+	return &TipsetIterator{ctx, store, start, nil}
+}
+
+// TipsetIterator is an iterator over tipsets.
+type TipsetIterator struct {
+	ctx   context.Context
+	store BlockProvider
+	value types.TipSet
+	err   error
+}
+
+// Value returns the iterator's current value, if not Complete().
+func (it *TipsetIterator) Value() types.TipSet {
+	return it.value
+}
+
+// Complete tests whether the iterator is exhausted.
+func (it *TipsetIterator) Complete() bool {
+	return len(it.value) == 0
+}
+
+// Next advances the iterator to the next value.
+func (it *TipsetIterator) Next() error {
+	it.value, it.err = GetParentTipSet(it.ctx, it.store, it.value)
+	return it.err
+}
--- a/chain/traversal_test.go
+++ b/chain/traversal_test.go
@@ -1,0 +1,120 @@
+package chain_test
+
+import (
+	"context"
+	"testing"
+
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/assert"
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/require"
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
+
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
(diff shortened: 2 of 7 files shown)


chain/traversal.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// BlockProvider provides blocks. This is a subset of the ReadStore interface.
+type BlockProvider interface {
+	GetBlock(ctx context.Context, cid cid.Cid) (*types.Block, error)
+}
+
+// GetParentTipSet returns the parent tipset of a tipset.
+// The result is empty if the tipset has no parents (including if it is empty itself)
+func GetParentTipSet(ctx context.Context, store BlockProvider, ts types.TipSet) (types.TipSet, error) {
```
anorth: This is moved from `message_queue.go`, the only change being the store parameter changing to this narrow interface in place of the concrete `hamt.CborIpldStore`.

wallet/wallet_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/wallet"
-	"github.com/stretchr/testify/require"
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/require"
```
laser: NON-BLOCKING: Was this change intentional?

chain/traversal.go
```diff
@@ -0,0 +1,100 @@
...
+// the genesis.
+// Blocks from each tipset are yielded in an unspecified but stable order.
+// NOTE: it would be great to align this order with the canonical ordering and define it as such.
+func IterAncestorBlocks(ctx context.Context, store BlockProvider, start types.TipSet) *BlockIterator {
+	tsIter := start.ToSortedCidSet().Iter()
+	return &BlockIterator{TipsetIterator{ctx, store, start, nil}, tsIter}
+}
+
+// BlockIterator is an iterator over blocks in tipsets.
+type BlockIterator struct {
```
laser: NON-BLOCKING: Cool idea!

chain/traversal_test.go
```diff
@@ -0,0 +1,154 @@
...
+		b.Height = parents[0].Height + 1
+		b.StateRoot = parents[0].StateRoot
+		for _, p := range parents {
+			b.Parents.Add(p.Cid())
+		}
+	}
+
+	bs.blocks[b.Cid()] = b
+	return b
+
```
laser: NON-BLOCKING: Extra newline most likely unintentional.

core/message_queue.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/types"
-	"github.com/pkg/errors"
+	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
```
laser: NON-BLOCKING: Was this change intentional?

wallet/wallet_test.go
phritz: looks like an improvement to me

chain/traversal_test.go
```diff
@@ -0,0 +1,154 @@
+package chain_test
+
+import (
+	"context"
+	"testing"
+
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/assert"
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
```
phritz: imports here and elsewhere are off.

chain/traversal_test.go
```diff
@@ -0,0 +1,154 @@
...
+		ts := requireTipset(t, root)
+		parent, e := chain.GetParentTipSet(ctx, store, ts)
+		require.NoError(e)
+		assert.Empty(parent)
+	})
+	t.Run("plural tipset", func(t *testing.T) {
+		ts := requireTipset(t, b11, b12)
+		parent, e := chain.GetParentTipSet(ctx, store, ts)
+		require.NoError(e)
+		assert.Equal(requireTipset(t, root), parent)
```
phritz: there is a tipset.equals which is presumably better than default object comparison(?)

chain/traversal.go
```diff
@@ -0,0 +1,100 @@
...
+	return it.TipsetIterator.Complete() && it.cidIterator.Complete()
+}
+
+// Next advances the iterator to the next value.
+func (it *BlockIterator) Next() error {
+	it.cidIterator.Next()
+	if it.cidIterator.Complete() {
+		err := it.TipsetIterator.Next()
+		if err == nil {
+			it.cidIterator = it.TipsetIterator.Value().ToSortedCidSet().Iter()
```
ZenGround0: It is unclear why blocks are not returned in the canonical order if we are iterating over the tipset in sorted cid order.

core/message_pool.go
```diff
@@ -109,32 +109,12 @@ func NewMessagePool(timer BlockTimer) *MessagePool {
 	}
 }
 
-// getParentTips returns the parent tipset of the provided tipset
-// TODO msgPool should have access to a chain store that can just look this up...
-func getParentTipSet(ctx context.Context, store *hamt.CborIpldStore, ts types.TipSet) (types.TipSet, error) {
```
ZenGround0: 🎉

ZenGround0:
At the very least I want to see [walk chain](https://github.com/filecoin-project/go-filecoin/blob/master/chain/default_store.go#L430) (the shared traversal function for BlockHistory and Load) rewritten using this someday soon.  It probably makes sense to replace some or all calls to BlockHistory with this too.  Good deal.

In the near future I want to understand and talk about the tradeoffs between putting these GetIterator functions on the chain package vs. on some object, either the store, the tipset, or some other thing.

chain/traversal.go
anorth: I couldn't find a bit of code that clearly identified this as the canonical order. Is this the canonical order?

ZenGround0: This is a good point.  In retrospect we aren't being good about enforcing one canonical order.  Tipsets are printed and serialized to slices via `ToSlice` in this order but more fundamentally their messages are serialized in the order of increasing ticket bytes.  I think it probably makes sense to enforce the ticket ordering everywhere.

anorth: Agreed. I'll leave this and the comment about it here for now.

chain/traversal.go
```diff
@@ -0,0 +1,100 @@
...
+		}
```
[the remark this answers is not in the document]
anorth: Thanks, fixed.

anorth:
@ZenGround0 it could go on the store. A free utility function is at least not a wrong place to put it, and easy to move later. It depends on our philosophy toward minimal interfaces.

@acruikshank fair point about the BlockIterator. I wrote it because UpdateMessagePool iterates messages in blocks, and it's easy to stop at a tipset boundary by checking block height. However, I ended up having to iterate tipsets directly. I will retract and stash the BlockIterator for potential use some day.

PR #2271, opened 2019-03-13
[WIP] remove bootstrap miner

laser:


actor/builtin/miner/miner.go
```diff
@@ -653,8 +646,15 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, postProofs []proofs.PoStProof) (
...
 		var commRs []proofs.CommR
-		for _, v := range state.SectorCommitments {
-			commRs = append(commRs, v.CommR)
+		for k, v := range state.SectorCommitments {
+			n, err := strconv.ParseUint(k, 10, 64)
+			if err != nil {
+				return nil, errors.FaultErrorWrap(err, "failed to parse commitment sector id to uint64")
+			}
+
+			if n >= 10000 {
```
anorth: Use the symbolic constant

laser: Ugh. Nice catch.

actor/builtin/miner/miner.go
```diff
@@ -441,7 +434,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		return 1, errors.NewRevertError("invalid sized commRStar")
 	}
 
-	if !ma.Bootstrap {
+	if sectorID >= types.FirstNonGenGenSectorID {
```
acruikshank: BLOCKING: sectorID is a parameter to CommitSector. This PR will prevent a node from sending a message with sectorID < FirstNonGenGenSectorID, but nothing prevents a miner from crafting such a message. I'm currently unclear whether this PR would allow miners to gain storage power without proving it.

laser:
> This PR allows non-bootstrap miners to commit sectors that are not subject to verification. I don't think this is acceptable even if temporary.

No `commitSector` message will see its seal proof verified if its sector id is less than 10000.

A PoSt cannot be generated for any of these bogus sectors (whose id is less than 10000), because there does not exist any sealed sector on the miner's disk corresponding to the CommR in the `commitSector` message.

So, in usernet (before [this work is done](
https://github.com/filecoin-project/go-filecoin/issues/227)), a naughty miner could create a bunch of these messages to inflate their power - but they're going to slashed out of all their pledge collateral.

PR #2272, opened 2019-03-14
Factor out message pool traversals to extract re-usable logic for the outbound queue

anorth:
Part of #2146.

This code is now testable in isolation, but only exercised via mpool tests. I'd love to add more tests soon, but on a deadline for re-using this code in the outbound queue.

### DIFF
--- a/core/chains.go
+++ b/core/chains.go
@@ -1,0 +1,75 @@
+package core
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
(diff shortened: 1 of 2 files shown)


anorth:
I understand your concern about isolated blocks. I'm going to roll forward, but let's discuss and I'll come back to improve it.

In case it's not obvious, I exposed blocks to get at the block height for creation of `timedMessage` (and an analogue in the outbound queue). So maybe we want to collect all messages in a tipset into a single structure associated with height. Those messages should be in the canonical processing order, which I believe is ticket bytes. @ZenGround0 can you point me to that logic in the code?

PR #2278, opened 2019-03-14
Bug/message pool timeouts from wrong head

acruikshank:
### Problem

We now timeout message after a constant number of tipsets from the head. This calculation was being made off the `new` head after an algorithm that mutated that value down. A +/- issue mask the bug in the tests.

### Solution

Keep a reference to the head from before it mutates. Fix the tests and test the consequences of adding a chain with multiple tip sets.


anorth:
@acruikshank forgive me for merging this for you so I can work from it

acruikshank:
@anorth you are so forgiven.

PR #2280, opened 2019-03-14
Feat/address plum pork 1869

### LINKED ISSUE #1869: Move Address Import from Node API to Wallet
### Description

Address Import only depends on wallet and fits well with the responsibilities of wallet, so we should move it into the wallet and access it through plumbing in commands.

### Acceptance criteria

Address Import should no longer be a method on the Node API and should live in wallet instead.

### Risks + pitfalls

Address Import currently uses wallet which exists at the plumbing level, so this will have to be porcelain.

### LINKED ISSUE #1937: Move Address Export from Node API into Wallet
### Description

Address Export only depends on wallet and fits well with the responsibilities of wallet, so we should move it into the wallet and access it through plumbing in commands.

### Acceptance criteria

Address Export should no longer be a method on the Node API and should live in wallet instead.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Address` is part of it

# Solution

Move import and export into the wallet core module and use plumbing instead of the node api

Resolves #1869 
Resolves #1937


wallet/wallet.go
```diff
@@ -182,3 +182,49 @@ func (w *Wallet) keyInfoForAddr(addr address.Address) (*types.KeyInfo, error) {
 	}
 	return info, nil
 }
+
+// Import adds the given keyinfos to the wallet
```
phritz: seems like we should have a unit test or at least a daemon test for the round trip. if not i dont want to get in the way of more progress on killing nodeapi but probably good to start a list of things to circle back on once the api is killed...

rkowalick: There is `TestWalletExportImportRoundTrip`

PR #2281, opened 2019-03-14
Moved DAG from node api to plumbing

### LINKED ISSUE #1959: Add the DAG to plumbing
### Description

The merkledag is used by several other methods which will need to become porcelain, so we'll need to add the DAG to plumbing, likely with a thin wrapper around go-merkledag.

rosalinekarr:
# Problem

We want to remove the `api` directory, but DAG is part of it.

# Solution

Move DAG to plumbing

Resolves #1959


node/node.go
```diff
@@ -147,6 +149,9 @@ type Node struct {
 	// Blockservice is a higher level interface for fetching data
 	blockservice bserv.BlockService
 
+	// DAGService is the merkledag service for accessing the merkledag
+	DAGService dagFormat.DAGService
```
anorth: Un-export this name

plumbing/dag/dag.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"context"
+
+	ipld "gx/ipfs/QmRL22E4paat7ky7vx9MLpR97JHHbFPrg3ytFQw6qp1y1s/go-ipld-format"
+	"gx/ipfs/QmVUojkFtcsrVBa8kYZiM6LhxpYXaKDTxE4aF1NFj4RfBv/go-path"
+	"gx/ipfs/QmVUojkFtcsrVBa8kYZiM6LhxpYXaKDTxE4aF1NFj4RfBv/go-path/resolver"
+)
+
+// DAG is the
```
anorth: Incomplete comment

node/node.go
```diff
@@ -1128,3 +1137,8 @@ func (node *Node) CborStore() *hamt.CborIpldStore {
 func (node *Node) ChainReadStore() chain.ReadStore {
 	return node.ChainReader
 }
+
+// DAG is the interface that defines methods to interact with IPLD DAG objects.
+func (node *Node) DAG() dagFormat.DAGService {
```
acruikshank: BLOCKING: Avoiding exporting this type of functionality from node is one of the goals of this refactor. This function should be inlined.

rosalinekarr: Removed from the node entirely.

PR #2286, opened 2019-03-14
Update TOC in CODEWALK.md

anorth:


frrist:
Failure is due to a config change in CI, if you rebase on latest master it'll pass.

PR #2289, opened 2019-03-14
replace usage of address.Address{} with address.Undef

frrist:


frrist:
note that this doesn't remove _all_ usages of `address.Address{}`, its still used when declaring an array or using reflection.

Issue #2171, opened 2019-03-05
Include rust-fil-proofs groth parameters in go-filecoin github release artifacts


ognots:
### Description

Include the rust-fil-proofs groth parameters as an archive in `go-filecoin` [binary releases](https://github.com/filecoin-project/go-filecoin/releases)

Ensure the parameters match the latests `rust-fil-proofs` (parameters.json)[https://github.com/filecoin-project/rust-fil-proofs/blob/master/parameters.json] 

Add or edit README.md and any other documentation to reflect how to install the groth parameters

### Acceptance criteria

* As a consumer of go-filecoin binary releases, I want the rust-fil-proofs groth parameters associated with a binary release to be available for download within the go-filecoin release
* As a consumer of go-filecoin binary releases, I want clear instructions how to install the rust-fil-proofs groth parameters

### Risks + pitfalls

minimal

### Where to begin

groth parameters are already present in CI workspace, so this feature can be built in to existing CI functionality

ognots:
One is now able to fetch groth parameters via `paramfetch`, included in [0.1.4 release](https://github.com/filecoin-project/go-filecoin/releases/tag/0.1.4).

This approach was favored over original proposal in this issue, as groth parameter size will become prohibitively large to bundle in a `go-filecoin` release.

See the [Getting Started Wiki "Install Filecoin and its Dependencies" section](https://github.com/filecoin-project/go-filecoin/wiki/Getting-Started#install-filecoin-and-its-dependencies) for detailed instructions.

PR #2299, opened 2019-03-15
include paramfetch and parameters.json in GH release bundle

gmasgras:
Implements https://github.com/filecoin-project/go-filecoin/issues/2171

- Include `paramfetch` binary and `parameters.json` in the GH bundle
- Update WIKI to reflect changes: https://github.com/filecoin-project/go-filecoin/wiki/Getting-Started#installing-from-binary

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,14 +179,14 @@ jobs:
           name: Create macos bundle
           command: ./scripts/build-bundle.sh
       - store_artifacts:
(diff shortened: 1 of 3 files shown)


gmasgras:
> We'll want to parameterize the release tag so we can do nightly releases.
> 
> Up to you if you want to do that now or I can do that in a separate PR with additional nightly release CI step.
> 
> Otherwise LGTM

Will track that work in a separate issue: https://github.com/filecoin-project/go-filecoin/issues/2318

PR #2300, opened 2019-03-15
Moved Client ListAsks from node api to porcelain

### LINKED ISSUE #1802: move asks ls to use plumbing/porcelain
### Description
 
this builds on #1798 and #1797 , do them first

the list-asks command:
https://github.com/filecoin-project/go-filecoin/blob/bf50d67e1d92150042075a909b923668ec14eb01/commands/client.go#L193
uses the node api implementation:
https://github.com/filecoin-project/go-filecoin/blob/bf50d67e1d92150042075a909b923668ec14eb01/api/impl/client.go#L64

we should switch it from using the node api to using plumbing/porcelain. in particular we can move the implementation to porcelain and use the addressls command above to iterate addresses. 

note there is no unit test but there is a nice integration test in https://github.com/filecoin-project/go-filecoin/blob/bf50d67e1d92150042075a909b923668ec14eb01/commands/client_daemon_test.go#L13. we should probably have both.

also looks like the implementation might be more complicated than it needs to be -- why are we using channels? the addressls and chainls etc calls would be good places for iterators, but we dont have them yet. suspect that we can simplify this impl and when it comes time for iterators use that as a solution to stream results.

### Acceptance criteria

- no list asks impl in the client interface of the node api, and code removed from impl above
- list-asks uses porcelain

### Where to begin
see above

rosalinekarr:
# Problem

We want to remove the `api` directory, but Client ListAsks is part of it

# Solution

Moved Client ListAsks to porcelain

Resolves #1802

### DIFF
--- a/porcelain/client_test.go
+++ b/porcelain/client_test.go
@@ -1,0 +1,139 @@
+package porcelain_test
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/filecoin-project/go-filecoin/actor"
+	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/porcelain"
+	"github.com/filecoin-project/go-filecoin/state"
+	"github.com/filecoin-project/go-filecoin/types"
+
+	"gx/ipfs/QmPVkJMTeRC6iBByPWdrRkD3BE5UXsj5HPzb4kPqL186mS/testify/assert"
+	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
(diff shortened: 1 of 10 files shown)


porcelain/client_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+type claPlubming struct {
+	actorFail   bool
+	actorChFail bool
+	messageFail bool
+
+	MinerAddress address.Address
+}
+
+func (cla *claPlubming) ActorLs(ctx context.Context) (<-chan state.GetAllActorsResult, error) {
```
shannonwells: typo, `claPlu_bm_ing` ,  puns will go unspecified.

state/tree.go
```diff
@@ -256,8 +256,8 @@ func GetAllActors(ctx context.Context, t Tree) <-chan GetAllActorsResult {
 	st := t.(*tree)
 	out := make(chan GetAllActorsResult)
 	go func() {
+		defer close(out)
```
shannonwells: 👍

Issue #2030, opened 2019-02-19
move protocol calls from node api into protocol apis


phritz:
We should have protocol apis of some sort that sit on top of protocols on which commands can depend.  Mining start/stop is the canonical example of this  -- it should move out of the node api, and not into plumbing/porcelain,  but into its own;  a new api that is higher level. Anything that interacts with storage miner / cilent and retrieval miner / client, things which themselves rely on plumbing/porcelain, should move out of node api into this / these new apis.

See [Architecture Overview](https://github.com/filecoin-project/go-filecoin/blob/master/CODEWALK.md#architecture-overview) 

cc @rosalinekarr who is working on a plan to move the non protocol stuff out of node api in #1858

shannonwells:
All the actual protocol stuff is moved now as of #2325

PR #2303, opened 2019-03-15
Destroy node/api retrieval.  Retrieval API in protocol

shannonwells:
# In support of 
#2030 , moving things into a Protocol API

# Change Summary
* Rename bapi import and use block.  
* Rename ProtocolAPI to BlockAPI, because that's what it is.  
* Create and use protocol/RetrievalApi.
* I changed my mind. API struct has the deps directly inside, and not in a deps struct underneath.


api/api.go
```diff
@@ -6,6 +6,5 @@ type API interface {
 	Client() Client
 	Daemon() Daemon
 	Ping() Ping
-	RetrievalClient() RetrievalClient
```
phritz: omg only four left!!!!

node/node.go
```diff
@@ -509,7 +492,7 @@ func (node *Node) Start(ctx context.Context) error {
 		return errors.Wrap(err, "Could not make new storage client")
 	}
 
-	node.RetrievalClient = retrieval.NewClient(node)
+	node.setupProtocols()
```
phritz: personally i like having all the construction inline so i can see what gets what, but to each her own

protocol/retrieval/api.go
```diff
@@ -0,0 +1,36 @@
...
+	//types.Signer
+}
+
+// NewAPI creates a new API for a retrieval client.
+func NewAPI(rc *Client, minerPorc retrievalClientPorcelainAPI) API {
+	return API{rc: rc, plumbing: minerPorc}
+}
+
+// RetrievePiece retrieves bytes referenced by CID pieceCID
+func (a *API) RetrievePiece(ctx context.Context, pieceCID cid.Cid, minerAddr address.Address) (io.ReadCloser, error) {
```
phritz: it strikes me that if the caller gets the peerid, which is already in porcelain then we could just delete this api abstraction and surface the retrieval client directly. or i guess since we know more stuff is going to go into retrieval we could keep it, but simplify it by eliminating its dep on porcelain

protocol/retrieval/api.go
```diff
@@ -0,0 +1,36 @@
...
+	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
+	"gx/ipfs/QmTu65MVbemtUxJEWgsTtzv9Zv9P8rvmqNA4eG9TrTRGYc/go-libp2p-peer"
+
+	"github.com/filecoin-project/go-filecoin/address"
+)
+
+// API here is the API for a retrieval client.
+type API struct {
+	rc       *Client
+	plumbing retrievalClientPorcelainAPI
```
phritz: maybe call it porcelain?

shannonwells: 🤷‍♀️ it's called plumbing elsewhere, I guess because it's not meant to be used externally?

node/node.go
shannonwells: Start() is hella long though. I want to abstract more stuff.

protocol/retrieval/api.go
shannonwells: let's leave it for cleanup

node/node.go
phritz: it definitely needs abstracting. partitioning helps i guess but we need a builder thingy to really help

protocol/retrieval/api.go
phritz: probably a mistake, we originally depended o plumbing everywhere before we had porcelain, and there was probably a rename that missed unexported struct members

commands/daemon.go
```diff
@@ -133,7 +133,8 @@ func runAPIAndWait(ctx context.Context, nd *node.Node, config *config.Config, re
 		ctx:          context.Background(),
 		api:          api,
 		porcelainAPI: nd.PorcelainAPI,
-		blockpAPI:    nd.BlockAPI,
+		blockAPI:     nd.BlockAPI,
```
ZenGround0: I know this is after the fact but imo blockAPI should be blockMiningAPI.

node/node.go
```diff
@@ -57,7 +57,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/porcelain"
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
-	bapi "github.com/filecoin-project/go-filecoin/protocol/block"
+	"github.com/filecoin-project/go-filecoin/protocol/block"
```
ZenGround0: Please keep this as `bapi` here. Block is so overloaded and in the context of the node package readers will assume it means many other things.

node/node.go
ZenGround0: Chiming in: I prefer your way @shannonwells

protocol/retrieval/api.go
ZenGround0: As non-trivial versions of retrieval will be using porcelain functionality (for example to setup payment channels) I think it makes since to keep wrapping abstractions this way, if only to get us ready for the future.

commands/daemon.go
phritz: Agree

phritz:
Agree re wrapping too

commands/daemon.go
shannonwells: Naming is hard.  I struggled between FooMining, FooProtocol and between keeping names short a la go convention.  I'll put it as a todo in the cleanup issue I just filed.

PR #2297, opened 2019-03-15
 Fix golangci-lint warnings 

magik6k:
Required for https://github.com/filecoin-project/go-filecoin/issues/2207.

There seem to be some issues with daemon tests, but I'm not sure what causes that (also, circle seems dead - https://status.circleci.com/)

Issue #2207, opened 2019-03-08
Use something besides gometalinter as it is no longer being maintained (golangci-lint?)


shannonwells:
### Description
The [gometalinter](https://github.com/alecthomas/gometalinter) maintainers have marked this package as deprecated and direct people to [golangci-lint](https://github.com/golangci/golangci-lint) instead.  It claims it is "5x faster than gometalinter".

### Acceptance criteria
Equivalent linting that works for CircleCI and command line.

### Risks + pitfalls
Might have to update the filtering of colored output for non-ttys

### Where to begin
build/main.go `func lint(packages... string)`

PR #2305, opened 2019-03-16
test: add a "genesis" block

Stebalien:
Otherwise, if the initial tipset is empty, the first block has a height of 1.

Note: As far as I can tell, the genesis block _has_ to be empty because `UpdateMessagePool` skips the 0th (genesis) block.

(not sure if this is the right fix but the chain needs to start at 0)


Stebalien:
This fixes #2297 which is required for #2207.

anorth:
Thanks for tracing this down @Stebalien!

As it happens, I have in parallel removed the skipping of the first block in https://github.com/filecoin-project/go-filecoin/pull/2272. It seemed poorly justified and didn't cause any tests to fail. Will that alone fix the issue? 

AFAIK a real genesis block *will* have messages, needed to bootstrap any actors to be able to send any further messages. If change is necessary, I think it would be better to force the caller to provide a non-empty root tipset than magically insert one.

Stebalien:
Nope. In that case, we just need to make sure the block height starts at the right place. I'll open a new PR and close this one.

PR #2306, opened 2019-03-16
build: switch to golangci-lint

### LINKED ISSUE #2207: Use something besides gometalinter as it is no longer being maintained (golangci-lint?)

Stebalien:
This also switches to a single lint stage because the linters run in ~10s (on my machine, at least).

fixes #2207 

(currently based off #2305)

### DIFF
--- a/tools/iptb-plugins/filecoin/docker/iptb_metrics.go
+++ b/tools/iptb-plugins/filecoin/docker/iptb_metrics.go
@@ -14,8 +14,10 @@ func (l *Dockerfilecoin) Events() (io.ReadCloser, error) {
 
 // StderrReader provides an io.ReadCloser to the running daemons stderr
 func (l *Dockerfilecoin) StderrReader() (io.ReadCloser, error) {
-	cli, _ := l.GetClient()
-	ctx, _ := context.WithCancel(context.Background()) // nolint: vet
+	cli, err := l.GetClient()
+	if err != nil {
+		return nil, err
+	}
 
 	options := types.ContainerLogsOptions{
 		ShowStdout: false,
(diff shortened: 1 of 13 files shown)


anorth:
@frrist could you please shepherd this PR?

tools/iptb-plugins/filecoin/docker/iptb_metrics.go
```diff
@@ -14,30 +14,34 @@ func (l *Dockerfilecoin) Events() (io.ReadCloser, error) {
...
+	}
 
 	options := types.ContainerLogsOptions{
 		ShowStdout: false,
 		ShowStderr: true,
 		Follow:     true,
 	}
 
-	return cli.ContainerLogs(ctx, l.ID, options)
+	return cli.ContainerLogs(context.Background(), l.ID, options)
```
acruikshank: We are no longer making the context cancel-able here. Is this a conscious change? Do we know why it was cancel-able in the first place?

tools/iptb-plugins/filecoin/docker/iptb_metrics.go
```diff
@@ -14,30 +14,34 @@ func (l *Dockerfilecoin) Events() (io.ReadCloser, error) {
...
+	}
 
 	options := types.ContainerLogsOptions{
 		ShowStdout: true,
 		ShowStderr: false,
 		Follow:     true,
 	}
 
-	return cli.ContainerLogs(ctx, l.ID, options)
+	return cli.ContainerLogs(context.Background(), l.ID, options)
```
acruikshank: Same question here.

tools/iptb-plugins/filecoin/docker/iptb_metrics.go
frrist: Since we were ignoring the `cancel()` function returned by `context.WithCancel()` and since it didn't have a parent context I don't think we were ever canceling it. I believe this change is alright.

travisperson: I think it just slipped through code review when this was originally merged. It started out as a cancel-able context to implement the Closer interface for the io.ReadCloser, turns out that ContainerLogs already returns a io.ReadCloser so it was no longer needed to shutdown the stream.

PR #2313, opened 2019-03-18
Hook up message sender to put messages in outbound queue

### LINKED ISSUE #2146: Implement an outbound message queue, distinct from the message pool

anorth:
The sender now uses the outbound queue for nonce calculation. Inbox/outbox terminology suggested by @rosalinekarr, and I think we could push it further. Closes #2146.

- [x] Based on #2272
- [x] Merge #2311 first (which expires old messages from the queue)

@acruikshank this about finishes the hookup of the outbound queue (still to add a status command and msgwait equivalent). Not for this PR, but we can discuss more cohesion with the message pool.
- I think the `Sender` is doing too much work for a bit of plumbing and could be factored out next to the message pool/queue implementations
- I would also extract UpdateMessagePool next to the the outbound queue expiry policy thingy

### DIFF
--- a/plumbing/msg/sender.go
+++ b/plumbing/msg/sender.go
@@ -10,7 +10,6 @@ import (
 	"github.com/filecoin-project/go-filecoin/chain"
 	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/core"
-	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
@@ -66,7 +65,7 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
 		return cid.Undef, errors.Wrapf(err, "no actor at address %s", from)
 	}
 
-	nonce, err := nextNonce(ctx, st, s.msgPool, from)
+	nonce, err := nextNonce(fromActor, s.msgPool, from)
 	if err != nil {
 		return cid.Undef, errors.Wrapf(err, "failed calculating nonce for actor %s", from)
 	}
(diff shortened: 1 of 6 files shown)


plumbing/msg/sender.go
```diff
@@ -20,17 +20,27 @@ import (
...
 	// Signs messages.
 	signer types.Signer
-	// Provides actor state (we could reduce this dependency to only LatestState()).
-	chainReader chain.ReadStore
-	// Pool of existing message, and receiver of the sent message.
-	msgPool *core.MessagePool
+	// Provides actor state
+	chainState chainState
+	// Provides the current block height
+	blockTimer core.BlockTimer
```
anorth: @acruikshank in a follow-up I suggest removing the blocktimer from message pool in favour of this one.

While the chain store ends up providing both `chainState` and `core.blockTimer` it's very useful in testing that they're separate interfaces here.

plumbing/msg/sender.go
```diff
@@ -87,9 +107,17 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
...
-	// Add to the local message pool at the last possible moment before broadcasting to network.
-	if _, err := s.msgPool.Add(smsg); err != nil {
-		return cid.Undef, errors.Wrap(err, "failed to add message to the message pool")
+	height, err := s.blockTimer.BlockHeight()
+	if err != nil {
+		return cid.Undef, errors.Wrap(err, "failed to get block height")
+	}
+
+	// Add to the local message queue/pool at the last possible moment before broadcasting to network.
+	if err := s.outbox.Enqueue(smsg, height); err != nil {
```
acruikshank: Non-Blocking: Still think block height should be the message queue's responsibility, not the callers. Calling Enqueue with anything other than the current block height will cause the timeout mechanism to misbehave.

plumbing/msg/sender.go
```diff
@@ -56,7 +76,7 @@ func (s *Sender) Send(ctx context.Context, from, to address.Address, value *type
 	s.l.Lock()
 	defer s.l.Unlock()
 
-	st, err := s.chainReader.LatestState(ctx)
+	st, err := s.chainState.LatestState(ctx)
```
shannonwells: from the Naming Is Hard Department: I hope that we intend to converge on one convention for a chain.ReadStore -- I've seen chainreader, chain, chainstore, and now chainstate.  This isn't a request for changes here; maybe it's something we can discuss next week.

anorth: This one I renamed because I inserted a scoped interface with this name (that happens to be satisfied by the chain store).

anorth:
Whoops, I can't land this until implementing equality for signed messages #2332.

(The code using the equality is in the expiry policy).

Issue #1873, opened 2019-02-12
[bug] Flaky test TestPaymentChannelReclaimSuccess and TestStorageDealsAfterRestart


ZenGround0:
**Describe the bug**
Flaky test: TestPaymentChannelReclaimSuccess

**To Reproduce**
Run CI

**Expected behavior**
Test should never fail.

**Screenshots**
https://circleci.com/gh/filecoin-project/go-filecoin/10794?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

**Desktop (please complete the following information):**
 Travis CI

**Additional context**
Haven't seen this before and couldn't find another issue reporting

shannonwells:
This one is back again; failed in my dev branch and at least once on master.

anorth:
Another one: https://circleci.com/gh/filecoin-project/go-filecoin/12314

`TestStorageDealsAfterRestart` failed too.

rkowalick:
I don't know if fixing the payment channel test failures make sense given that @travisperson is working on converting this stuff to FAST in https://github.com/filecoin-project/go-filecoin/pull/2319. I have a fix for TestStorageDealsAfterRestart, so I will say that "resolves" this issue.

We can open a new one or reopen this if the new FAST payment channels end up being flaky too.

phritz:
still flaky i think: https://circleci.com/gh/filecoin-project/go-filecoin/13258?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link 

@travisperson can you tell us if we should ignore these flakes because this test is being replaced, and if so when you think it will be? probably worth an announcement to the team if we should be ignoring

travisperson:
Yes, these can probably be ignored unless they are really consistent and causing a lot of pain. Payment channel FAST tests are currently WIP in PR #2319. I hope to have those merged later this week or early next week.

travisperson:
Merged #2528, hopefully this resolves these issues.

Issue #1811, opened 2019-02-05
Passing `price` and `limit` to series


travisperson:
### Description

FAT series are standalone functions which which accept a context, one or more nodes, and additional arguments. Any action that occurs in a series that submits a message is going to have to provide a way to set a `price` and a `limit`. There might be more than one action as well that will need to use these options.

I want to start a discussion to determine the best way to manage these values.

travisperson:
I think a solution to this is to follow a pattern I've started to use for managing sleeping in series (#2514) and allowing for synchronous mining (#2515).

We would set a price and limit value on the context which would be extracted from the context inside of the series and passed into the actions. This will make it easy to set a price / limit for all actions, and the value can be changed when needed.

For example, we may want to set a price of 0 during the setup, and then later update the price to another value for the rest of a test which can be done by replacing the ctx.

travisperson:
Does this seem like an appropriate use for context @whyrusleeping?

whyrusleeping:
Yeah, that seems appropriate to me. 
You could have a 'params' object stored in the context that has all these variables.

travisperson:
@frrist would you want to work on this?

frrist:
In light of the recent [comment](https://github.com/filecoin-project/go-filecoin/pull/2515#discussion_r273763967) from anorth I think you are best suited to address this.

PR #2319, opened 2019-03-18
FAST payment channel tests

travisperson:
This is still a bit of a work in progress. There are some patterns that I think need to be worked out.

Also, currently tests written with FAST can flake due to the added local go-filecoin iptb plugin feature to copy the target binary (to ensure binary capability for all `go-filecoin` commands). This appears to be related to a bug being tracked in golang (https://github.com/golang/go/issues/22315) due to an issue around file descriptors being held onto longer than needed inside of threads.

A possible work around might be to use exec to run `cp` to ensure there are no open file descriptors being held by go threads.

### DIFF
--- a/commands/payment_channel_daemon_test.go
+++ b/commands/payment_channel_daemon_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	cbor "github.com/ipfs/go-ipld-cbor"
+	logging "github.com/ipfs/go-log"
 	"github.com/multiformats/go-multibase"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -18,6 +19,11 @@ import (
 	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
 )
+
+func init() {
+	logging.SetDebugLogging()
+	series.GlobalSleepDelay = 5 * time.Second
+}
 
 func VoucherFromString(data string) (paymentbroker.PaymentVoucher, error) {
 	_, cborVoucher, err := multibase.Decode(data)
@@ -38,7 +44,7 @@ func TestPaymentChannelCreateSuccess(t *testing.T) {
 	require := require.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -74,8 +80,6 @@ func WithPaymentChannel(t *testing.T, ctx context.Context, p *fast.Filecoin, fro
 	mcid, err := p.PaychCreate(ctx, target, amt, eol, fast.AOFromAddr(from), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 	require.NoError(err)
 
-	series.MiningOnceFromCtx(ctx)
-
 	response, err := p.MessageWait(ctx, mcid)
 	require.NoError(err)
 
@@ -92,7 +96,7 @@ func TestPaymentChannelLs(t *testing.T) {
 		assert := assert.New(t)
 
 		// This test should run in 20 block times
-		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 		defer cancel()
 
 		// Get basic testing environment
@@ -136,7 +140,7 @@ func TestPaymentChannelLs(t *testing.T) {
 		assert := assert.New(t)
 
 		// This test should run in 20 block times
-		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 		defer cancel()
 
 		// Get basic testing environment
@@ -180,7 +184,7 @@ func TestPaymentChannelLs(t *testing.T) {
 		assert := assert.New(t)
 
 		// This test should run in 20 block times
-		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 		defer cancel()
 
 		// Get basic testing environment
@@ -219,7 +223,7 @@ func TestPaymentChannelVoucherSuccess(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -262,7 +266,7 @@ func TestPaymentChannelRedeemSuccess(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -296,8 +300,6 @@ func TestPaymentChannelRedeemSuccess(t *testing.T) {
 		mcid, err := targetDaemon.PaychRedeem(ctx, voucherStr, fast.AOFromAddr(targetAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
 
-		series.MiningOnceFromCtx(ctx)
-
 		_, err = targetDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
 
@@ -316,7 +318,7 @@ func TestPaymentChannelRedeemTooEarlyFails(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -350,8 +352,6 @@ func TestPaymentChannelRedeemTooEarlyFails(t *testing.T) {
 		mcid, err := targetDaemon.PaychRedeem(ctx, voucherStr, fast.AOFromAddr(targetAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
 
-		series.MiningOnceFromCtx(ctx)
-
 		_, err = targetDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
 
@@ -370,7 +370,7 @@ func TestPaymentChannelReclaimSuccess(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(40*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -394,11 +394,7 @@ func TestPaymentChannelReclaimSuccess(t *testing.T) {
 	bh, err := series.GetHeadBlockHeight(ctx, p)
 	require.NoError(err)
 
-	// Expiry is current height, plus 3
-	// - Setting up the payment channel
-	// - Redeeming one voucher
-	// - Expires on third block
-	channelExpiry := types.NewBlockHeight(3).Add(bh)
+	channelExpiry := types.NewBlockHeight(10).Add(bh)
 	channelAmount := types.NewAttoFILFromFIL(1000)
 
 	balanceBefore, err := payerDaemon.WalletBalance(ctx, payerAddr)
@@ -414,8 +410,6 @@ func TestPaymentChannelReclaimSuccess(t *testing.T) {
 		mcid, err := targetDaemon.PaychRedeem(ctx, voucherStr, fast.AOFromAddr(targetAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
 
-		series.MiningOnceFromCtx(ctx)
-
 		_, err = targetDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
 
@@ -426,12 +420,10 @@ func TestPaymentChannelReclaimSuccess(t *testing.T) {
 		assert.Equal(channelAmount, channel.Amount)
 		assert.Equal(voucherAmount, channel.AmountRedeemed)
 
-		series.MiningOnceFromCtx(ctx)
+		series.WaitForBlockHeight(ctx, payerDaemon, channel.Eol)
 
 		mcid, err = payerDaemon.PaychReclaim(ctx, chanid, fast.AOFromAddr(payerAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
-
-		series.MiningOnceFromCtx(ctx)
 
 		_, err = payerDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
@@ -453,7 +445,7 @@ func TestPaymentChannelCloseSuccess(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -493,8 +485,6 @@ func TestPaymentChannelCloseSuccess(t *testing.T) {
 		mcid, err := targetDaemon.PaychClose(ctx, voucherStr, fast.AOFromAddr(targetAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
 
-		series.MiningOnceFromCtx(ctx)
-
 		_, err = targetDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
 
@@ -518,7 +508,7 @@ func TestPaymentChannelExtendSuccess(t *testing.T) {
 	assert := assert.New(t)
 
 	// This test should run in 20 block times
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
 	defer cancel()
 
 	// Get basic testing environment
@@ -559,8 +549,6 @@ func TestPaymentChannelExtendSuccess(t *testing.T) {
 
 		mcid, err := payerDaemon.PaychExtend(ctx, chanid, extendAmount, extendExpiry, fast.AOFromAddr(payerAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
 		require.NoError(err)
-
-		series.MiningOnceFromCtx(ctx)
 
 		_, err = payerDaemon.MessageWait(ctx, mcid)
 		require.NoError(err)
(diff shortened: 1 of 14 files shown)


commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,620 @@
...
-	assert := assert.New(t)
-
-	d := makeTestDaemonWithMinerAndStart(t)
-	defer d.ShutdownSuccess()
-
-	args := []string{"paych", "create"}
-	args = append(args, "--from", fixtures.TestAddresses[0], "--gas-price", "0", "--gas-limit", "300")
-	args = append(args, fixtures.TestAddresses[1], "10000", "20")
+func init() {
+	_, _ = os.Getwd()
```
travisperson: delete: was used to disable annoying unused import.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,620 @@
...
-			"--return",
-			"--message=false",
-			"--receipt=false",
-			messageCid.String(),
-		)
-		_, ok := types.NewChannelIDFromString(strings.Trim(wait.ReadStdout(), "\n"), 10)
-		assert.True(ok)
-		wg.Done()
-	}()
+func VoucherFromString(data string) (paymentbroker.PaymentVoucher, error) {
```
travisperson: This should probably be a function in `types`? However, `PaymentVoucher` is not defined there. Should we move it there?

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,620 @@
...
-
-		daemonTestWithPaymentChannel(t, &payer, &target, amt, eol, func(d *th.TestDaemon, channelID *types.ChannelID) {
-			ls := listChannelsAsStrs(d, &payer)[0]
-
-			assert.Equal(fmt.Sprintf("%s: target: %s, amt: 10000, amt redeemed: 0, eol: 20", channelID, target.String()), ls)
-		})
-	})
+	// Create a directory for the test using the test name (mostly for FAST)
+	// TODO(tperson) use a different TempDir that uses MkdirAll
+	dir, err := ioutil.TempDir("", strings.Replace(t.Name(), "/", ".", -1))
```
travisperson: `TempDir` only expects that there is a single directory level used in the provided name, and that it's not a path. However, nested tests are separated by `/`, which makes this fail.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,620 @@
...
+
+	WithPaymentChannel(t, ctx, payerDaemon, payerAddr, targetAddr, channelAmount, channelExpiry, func(chanid *types.ChannelID) {
+		require.NotNil(chanid)
 	})
 }
 
-func TestPaymentChannelRedeemSuccess(t *testing.T) {
-	t.Parallel()
+// WithPaymentChannel takes a filecoin daemon and constructs a payment channel to the `target` for the `amt` for `eol`, and calls fn with the channel id
+func WithPaymentChannel(t *testing.T, ctx context.Context, p *fast.Filecoin, from, target address.Address, amt *types.AttoFIL, eol *types.BlockHeight, fn func(chanid *types.ChannelID)) {
```
travisperson: This doesn't really need a callback function, it would just return the `ChannelID`. This is because we don't have the goroutine anymore. I kept this as a callback for explicit in updating tests and I wasn't sure of the exact pattern I wanted to use. Really, either way I don't think it's needed as we can always mine, and then wait.

I will say the callback does help break up the tests and I think makes it easier to read.

travisperson:
> Tests now run in ~30-35 seconds, prior tests were in the range of 12s, so this is a significant increase in time. I think this is largely due to the background mining.

I swapped background mining out for explicit mining.

With a block time configured to 1s (same as TestDaemon), the FAST payment channel tests run in ~14 seconds. This is pretty good, and the additional time is from a bit more overhead in setting up nodes.

I also tested with a millisecond block time (which I believe is used for the delay function), and have achieved test runs in the ~8 second range without issue.

tools/fast/series/mining_once_from_ctx.go
```diff
@@ -0,0 +1,19 @@
...
+// MiningOnceFunc is the type for the value stored under the CKMiningOnce key
+type MiningOnceFunc func()
+
+// MiningOnceFromCtx extracts the MiningOnceFunc through key CKMiningOnce
+func MiningOnceFromCtx(ctx context.Context) {
+	MiningOnce, ok := ctx.Value(CKMiningOnce).(MiningOnceFunc)
+	if ok {
+		MiningOnce()
+	}
+}
```
travisperson: This is a solution to enabling the ability to run `MiningOnce` in integration tests.

Integration tests in general do not setup with async mining (`mining start`). The issue with using async mining is the lack of control over block height, and the increased test time that can come from it. One of the goals of FAST is to enable reuse of code. A simple example of this is sending tokens from one node to another. Ideally, in a single function call, the sending of the message, and the waiting would be performed to ensure that when the function returns, the message has been mined and the targeted wallet has the funds. This works well when there is async mining because the function can wait around for the message. However, when working with tests that need to have more control over the block height, eg. payment channel expiry, using async miners requires more padding around expiry values and waiting around for blocks to be mined. This increases the test time substantially. 

The idea is that this can be used in FAST series where mining may be required, such as sending tokens. When running in a functional testing environment (devnets with async mining) the operation is a noop. When explicit mining control is required, a MiningOnceFunc can be added to the context enabling series to manage explicit mining without having to pass in a dedicated miner node, or requiring that all nodes be miners (would suck).

tools/fast/series/wait_for_block_height.go
```diff
@@ -0,0 +1,28 @@
...
+		hh, err := GetHeadBlockHeight(ctx, client)
+		if err != nil {
+			return err
+		}
+
+		if hh.GreaterEqual(bh) {
+			break
+		}
+
+		SleepDelay()
```
travisperson: Should call `MiningOnceFromCtx`

tools/fast/series/setup_genesis_node.go
```diff
@@ -41,5 +41,5 @@ func SetupGenesisNode(ctx context.Context, node *fast.Filecoin, minerAddress add
 		return err
 	}
 
-	return node.MiningStart(ctx)
+	return nil
```
travisperson: Remove iferr and just return `err`.

tools/fast/series/mining_once_from_ctx.go
```diff
@@ -0,0 +1,19 @@
+package series
+
+import (
+	"context"
+)
+
+// CKMiningOnce is the key used to store the MiningOnceFunc in a context
+var CKMiningOnce = struct{}{}
```
travisperson: Name suggestions? Probably can be private and export a function to set the value on an accepted context.

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -129,27 +129,15 @@ func init() {
...
-	}
-	defer func() {
-		cerr := out.Close()
-		if err == nil {
-			err = cerr
-		}
-	}()
-	if _, err = io.Copy(out, in); err != nil {
-		return
+	cmd := exec.Command("cp", src, dst)
```
acruikshank: This won't work in windows. At the very least we should at a compiler flag to note that.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
+
+	addrs, err := targetDaemon.AddressLs(ctx)
+	require.NoError(err)
+	targetAddr := addrs[0]
+
+	addrs, err = payerDaemon.AddressLs(ctx)
+	require.NoError(err)
+	payerAddr := addrs[0]
+
+	channelExpiry := types.NewBlockHeight(20)
```
frrist: Do I understand correctly that the `20` here is related to this: https://github.com/filecoin-project/go-filecoin/pull/2319/files#diff-d61a1f499164476e93615aeea69423c2R47
If so, consider making it a variable here and elsewhere.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
 	})
 
-	t.Run("Notifies when channels not found", func(t *testing.T) {
+	t.Run("No results when listing with different from address", func(t *testing.T) {
 		t.Parallel()
+		require := require.New(t)
+		assert := assert.New(t)
+
+		// This test should run in 20 block times
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
```
frrist: Seeing this pattern is used in every payment channel test, should we consider making it a private helper method?
```go
func fastSetupWithBlockDeadline(t *testing.T, deadline time.Duration) (context.Context, fast.Environment, func() *fast.Filecoin) 
```

tools/iptb-plugins/filecoin/local/localfilecoin.go
travisperson: The original bug probably doesn't exist in windows, so I can probably move this out to a new file and flag it.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
-	daemonTestWithPaymentChannel(t, &payer, &target, amt, eol, func(d *th.TestDaemon, channelID *types.ChannelID) {
-		assert := assert.New(t)
+	WithPaymentChannel(ctx, t, payerDaemon, payerAddr, targetAddr, channelAmount, channelExpiry, func(chanid *types.ChannelID) {
+		voucherAmount := types.NewAttoFILFromFIL(10)
+		voucherValidAt := types.NewBlockHeight(0)
+		voucherStr, err := payerDaemon.PaychVoucher(ctx, chanid, voucherAmount, fast.AOFromAddr(payerAddr), fast.AOValidAt(voucherValidAt))
+		require.NoError(err)
 
-		d.ConnectSuccess(targetDaemon)
+		// TODO(tperson) allow setting Price / Limit on process?
```
frrist: What does this mean? A change to FAST or the test?

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,75 @@
...
+	fastenvOpts = fast.EnvironmentOpts{
+		InitOpts:   append([]fast.ProcessInitOption{fast.POGenesisFile(genesisURI)}, fastenvOpts.InitOpts...),
+		DaemonOpts: append([]fast.ProcessDaemonOption{fast.POBlockTime(series.GlobalSleepDelay)}, fastenvOpts.DaemonOpts...),
+	}
+
+	// Create a node for the test
+	genesis, err := env.NewProcess(ctx, localplugin.PluginName, options, fastenvOpts)
+	require.NoError(err)
+
+	err = series.SetupGenesisNode(ctx, genesis, genesisMiner.Address, files.NewReaderFile(genesisMiner.Owner))
```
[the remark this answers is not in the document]
frrist: I believe we called it this because it's the first node created, capable of mining, and distributing filecoin to other nodes. It is also the only node that gets power/funds from the genesis block. How about `GenesisMiner` instead?

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
-		// payer creates a voucher to be redeemed by target (off-chain)
-		voucher := createVoucherStr(t, d, channelID, types.NewAttoFILFromFIL(10), &payer, uint64(0))
-
-		// target redeems the voucher (on-chain)
-		mustRedeemVoucher(t, targetDaemon, voucher, &target)
-
-		lsStr := listChannelsAsStrs(targetDaemon, &payer)[0]
-		assert.Equal(fmt.Sprintf("%v: target: %s, amt: 1000, amt redeemed: 10, eol: %s", channelID, target.String(), eol.String()), lsStr)
-
-		d.RunSuccess("mining once")
```
frrist: BLOCKING: I am concerned about removing these. This pattern allowed for deterministic testing (in theory at least), now that mining is a long running process I fear that we may reintroduce test flakiness.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,567 @@
...
+	WithPaymentChannel(t, ctx, payerDaemon, payerAddr, targetAddr, channelAmount, channelExpiry, func(chanid *types.ChannelID) {
+		voucherAmount := types.NewAttoFILFromFIL(10)
+		voucherValidAt := types.NewBlockHeight(10)
+		voucherStr, err := payerDaemon.PaychVoucher(ctx, chanid, voucherAmount, fast.AOFromAddr(payerAddr), fast.AOValidAt(voucherValidAt))
+		require.NoError(err)
 
-		// target redeems the voucher (on-chain) and simultaneously closes the channel
-		mustCloseChannel(t, targetDaemon, voucher, target)
+		// TODO(tperson) allow setting Price / Limit on process?
+		mcid, err := targetDaemon.PaychRedeem(ctx, voucherStr, fast.AOFromAddr(targetAddr), fast.AOPrice(big.NewFloat(0)), fast.AOLimit(300))
```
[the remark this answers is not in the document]
travisperson: > checking for an error response here

Can you clarify this? There is a `NoError` check below.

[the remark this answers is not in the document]
travisperson: Do you mean to check the response of `targetDaemon.MessageWait` below for errors?

commands/payment_channel_daemon_test.go
travisperson: Yes, they do appear to be the same exact code

commands/payment_channel_daemon_test.go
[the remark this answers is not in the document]
frrist: The test is called `TestPaymentChannelRedeemTooEarlyFails`, at first glance this method should fail, right?

commands/payment_channel_daemon_test.go
travisperson: Not related, just happen to be the same.

commands/payment_channel_daemon_test.go
travisperson: FAST it's a place holder comment for myself to address https://github.com/filecoin-project/go-filecoin/issues/1811

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,567 @@
...
-	daemonTestWithPaymentChannel(t, &payer, &target, amt, eol, func(d *th.TestDaemon, channelID *types.ChannelID) {
-		assert := assert.New(t)
+	WithPaymentChannel(t, ctx, payerDaemon, payerAddr, targetAddr, channelAmount, channelExpiry, func(chanid *types.ChannelID) {
+		voucherAmount := types.NewAttoFILFromFIL(10)
+		voucherValidAt := types.NewBlockHeight(0)
+		voucherStr, err := payerDaemon.PaychVoucher(ctx, chanid, voucherAmount, fast.AOFromAddr(payerAddr), fast.AOValidAt(voucherValidAt))
+		require.NoError(err)
 
-		d.ConnectSuccess(targetDaemon)
+		// TODO(tperson) allow setting Price / Limit on process?
```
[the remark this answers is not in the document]
travisperson: It's a place holder comment for myself when I wrote this to address https://github.com/filecoin-project/go-filecoin/issues/1811

commands/payment_channel_daemon_test.go
travisperson: See the commit prior to this one which uses the synchronous testing, as well as the top level discussion. More context in these slack discussions.

https://filecoinproject.slack.com/archives/CEHHJNJS3/p1554145557069100
https://filecoinproject.slack.com/archives/CEHHJNJS3/p1553814863008600

travisperson: https://github.com/filecoin-project/go-filecoin/issues/1873

tools/fast/action_chain.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/json"
 
-	"github.com/ipfs/go-cid"
+	cid "github.com/ipfs/go-cid"
```
frrist: redundant rename

commands/payment_channel_daemon_test.go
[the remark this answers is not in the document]
acruikshank: Yes. The message wait command doesn't error out, but the message response it returns should contain an error.

commands/payment_channel_daemon_test.go
frrist: The commit prior to this (dcf5b35782b4dba4892588b8528a70b6b85c149a) is called (and uses) async mining. But based on this comment: https://github.com/filecoin-project/go-filecoin/pull/2319#discussion_r270228398 it sounds like we are going the synchronous route via the MiningFromCtx method, this sounds good to me (nice idea!). 

However this: https://github.com/filecoin-project/go-filecoin/pull/2319/files#r270234693 leaves me with the impression that we have not yet completed the changes required for synchronous mining. This comment is nonblocking, just trying to understand whats happening here.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
-	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fasting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+func init() {
+	logging.SetDebugLogging()
+	series.GlobalSleepDelay = 5 * time.Second
```
anorth: Does this change persist after this test? I'm guessing so. We need a better pattern for this. Can it go into the FAST environment, or some other piece of code that can be scoped to a function lifetime?

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
-			"--return",
-			"--message=false",
-			"--receipt=false",
-			messageCid.String(),
-		)
-		_, ok := types.NewChannelIDFromString(strings.Trim(wait.ReadStdout(), "\n"), 10)
-		assert.True(ok)
-		wg.Done()
+	// This test should run in 20 block times
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(20*series.GlobalSleepDelay))
```
anorth: Factor this out so you can `ctx, cancel := contextWithBlockTimes(20)`. Maybe after moving the global sleep delay to be less global, it becomes a method on the environment.

commands/payment_channel_daemon_test.go
anorth: Please document that, or give the two constants different named `consts`, and/or make the values different to clarify that they are not coupled.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,598 @@
...
+	channelAmount := types.NewAttoFILFromFIL(1000)
 
-	wg.Wait()
+	WithPaymentChannel(ctx, t, payerDaemon, payerAddr, targetAddr, channelAmount, channelExpiry, func(chanid *types.ChannelID) {
+		require.NotNil(chanid)
+	})
+}
+
+// WithPaymentChannel takes a filecoin daemon and constructs a payment channel to the `target` for the `amt` for `eol`, and calls fn with the channel id
+func WithPaymentChannel(ctx context.Context, t *testing.T, p *fast.Filecoin, from, target address.Address, amt *types.AttoFIL, eol *types.BlockHeight, fn func(chanid *types.ChannelID)) {
```
anorth: Nit: move all the helper methods below all the test methods so that a top-down reading of the file describes what is being tested, then delves into progressive detail.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,567 @@
...
-			"--receipt=false",
-			messageCid.String(),
-		)
-		_, ok := types.NewChannelIDFromString(strings.Trim(wait.ReadStdout(), "\n"), 10)
-		assert.True(ok)
-		wg.Done()
-	}()
+	addrs, err = payerDaemon.AddressLs(ctx)
+	require.NoError(err)
+	payerAddr := addrs[0]
```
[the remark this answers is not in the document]
anorth: +1. If the number of return values gets unwieldy, return a struct (to which you can also add methods for repeated patterns).

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
+	localplugin "github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin/local"
+)
+
+// BasicFastSetup creates a environment and a single node, and environment options
+func BasicFastSetup(ctx context.Context, t *testing.T, fastenvOpts fast.EnvironmentOpts) (fast.Environment, *fast.Filecoin, func() *fast.Filecoin, context.Context) {
+	require := require.New(t)
+
+	// Create a directory for the test using the test name (mostly for FAST)
+	// TODO(tperson) use a different TempDir that uses MkdirAll
```
anorth: Don't merge with TODOs. Either do them, or leave a comment describing an alternative that wasn't valuable enough to be worth doing now, or an issue, or no comment at all.

anorth:
Tests now run in ~30-35 seconds, prior tests were in the range of 12s, so this is a significant increase in time. I think this is largely due to the background mining.

I swapped background mining out for explicit mining.

With a block time configured to 1s (same as TestDaemon), the FAST payment channel tests run in ~14 seconds. This is pretty good, and the additional time is from a bit more overhead in setting up nodes.

commands/payment_channel_daemon_test.go
travisperson: Opened an issue to move payment vouchers: https://github.com/filecoin-project/go-filecoin/issues/2527

travisperson:
PR replace by #2528

PR #2320, opened 2019-03-18
Flaky Payment Channel Reclaim Test

### LINKED ISSUE #1873: [bug] Flaky test TestPaymentChannelReclaimSuccess and TestStorageDealsAfterRestart

rkowalick:
Was supposed to help resolve #1873 

CLOSED due to FAST conversion, will reopen if needed.

### DIFF
--- a/build/main.go
+++ b/build/main.go
@@ -335,7 +335,7 @@ func install() {
 func test(args ...string) {
 	log.Println("Testing...")
 
-	runCmd(cmd(fmt.Sprintf("go test -timeout 30m -parallel 8 ./... %s", strings.Join(args, " "))))
+	runCmd(cmd(fmt.Sprintf("go test -timeout 30m -parallel 8 --count=20 ./commands/payment_channel_daemon_test.go %s", strings.Join(args, " "))))
 }
 
 func main() {
--- a/commands/payment_channel_daemon_test.go
+++ b/commands/payment_channel_daemon_test.go
@@ -2,7 +2,6 @@ package commands_test
 
 import (
 	"fmt"
-	"strings"
 	"sync"
 	"testing"
 
@@ -33,7 +32,7 @@ func TestPaymentChannelCreateSuccess(t *testing.T) {
 	args = append(args, fixtures.TestAddresses[1], "10000", "20")
 
 	paymentChannelCmd := d.RunSuccess(args...)
-	messageCid, err := cid.Parse(strings.Trim(paymentChannelCmd.ReadStdout(), "\n"))
+	messageCid, err := cid.Parse(paymentChannelCmd.ReadStdout())
 	require.NoError(t, err)
 
 	var wg sync.WaitGroup
@@ -46,7 +45,7 @@ func TestPaymentChannelCreateSuccess(t *testing.T) {
 			"--receipt=false",
 			messageCid.String(),
 		)
-		_, ok := types.NewChannelIDFromString(strings.Trim(wait.ReadStdout(), "\n"), 10)
+		_, ok := types.NewChannelIDFromString(wait.ReadStdout(), 10)
 		assert.True(ok)
 		wg.Done()
 	}()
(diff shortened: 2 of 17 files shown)


build/main.go
```diff
@@ -335,7 +335,7 @@ func install() {
 func test(args ...string) {
 	log.Println("Testing...")
 
-	runCmd(cmd(fmt.Sprintf("go test -timeout 30m -parallel 8 ./... %s", strings.Join(args, " "))))
+	runCmd(cmd(fmt.Sprintf("go test -timeout 30m -run TestPaymentChannelReclaimSuccess --count=50 -parallel 8 ./... %s", strings.Join(args, " "))))
```
rkowalick: This _MUST_ be removed before merge

commands/payment_channel_daemon_test.go
```diff
@@ -205,7 +204,7 @@ func TestPaymentChannelRedeemTooEarlyFails(t *testing.T) {
 		voucher := createVoucherStr(t, d, channelID, types.NewAttoFILFromFIL(111), &payer, uint64(8))
 
 		// Wait for the voucher message to be processed.
-		mustRedeemVoucher(t, targetDaemon, voucher, &target)
+		mustRedeemVoucher(t, targetDaemon, voucher, &target, d)
```
acruikshank: `listChannelsAsStrs` should also run from d. Also might be worth renaming `d` and `targetDaemon` to `sender` and `miner` at this point to clarify roles.

commands/payment_channel_daemon_test.go
```diff
@@ -448,14 +447,14 @@ func mustExtendChannel(t *testing.T, d *th.TestDaemon, channelID *types.ChannelI
 	wg.Wait()
 }
 
-func mustRedeemVoucher(t *testing.T, d *th.TestDaemon, voucher string, targetAddress *address.Address) {
+func mustRedeemVoucher(t *testing.T, d *th.TestDaemon, voucher string, targetAddress *address.Address, watcher *th.TestDaemon) {
```
acruikshank: d -> miner, watcher -> sender. Should be the first two parameters after `t`

commands/payment_channel_daemon_test.go
rkowalick: I feel like payer and payee might be more clear? There isn't any actual storage mining going on here.

rkowalick: actually, maybe renaming them to sender and miner and only have one of them mine would improve the situation a lot.

rkowalick: Eh, scratch that. It'd definitely mean a lot more changes in the tests.

acruikshank: It would be good to establish roles for the the daemons. Possibly just in in this test and the other you changed. The problem is that the transaction is between a payer and a target, but the test requires a role for the sender to send the message and a block miner to mine it. The node sending the message needs to have a key for the sender or target account depending on the message, but otherwise the sender and miner nodes are independent of the other roles.

We started with a single daemon that handled all the roles. We moved to two daemons to resolve some timing issues around sending and mining. For this reason, if we persist with multiple daemons, we should focus on the sender and miner relationship. All messages (payer and target) will be sent from one daemon. All messages will be mined from the other. All messages will be waited on by the sender, and all assertions will be made against sender state.

This is pretty close to what you have here. We should use naming to establish the pattern.

rkowalick:
I think I'm going to close this for now, since @travisperson is converting these tests to FAST in #2319 .
We can reevaluate if they end up being flaky.

PR #2321, opened 2019-03-18
Moved Client Cat from node api to plum pork

### LINKED ISSUE #1955: Move Client Cat from Node API into Porcelain
### Description

The Node API's Client Cat method should live in porcelain. This method makes heavy use of the merkledag which will need to be moved into plumbing with a #1959 first.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Client().Cat()` is part of it.

# Solution

Move `Client().Cat()` to DAG plumbing, where it makes sense.

Resolves #1955

### DIFF
--- a/plumbing/dag/dag.go
+++ b/plumbing/dag/dag.go
@@ -7,6 +7,7 @@ import (
 	"gx/ipfs/QmNRAuGmvnVw8urHkUZQirhu42VTiZjVWASa2aTznEMmpP/go-merkledag"
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
 	"gx/ipfs/QmRDWTzVdbHXdtat7tVJ7YC7kRaW7rTZTEF79yykcLYa49/go-unixfs"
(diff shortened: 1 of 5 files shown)


plumbing/dag/dag.go
```diff
@@ -70,3 +71,17 @@ func (dag *DAG) GetFileSize(ctx context.Context, c cid.Cid) (uint64, error) {
 		return 0, fmt.Errorf("unrecognized node type: %T", fnode)
 	}
 }
+
+// Cat returns an iostream with a piece of data stored on the merkeldag with
+// the given cid.
+func (dag *DAG) Cat(ctx context.Context, c cid.Cid) (io.DagReader, error) {
+	// TODO: this goes back to 'how is data stored and referenced'
```
ZenGround0: Not your TODO but is this reflected in an issue in go-filecoin?  I'm guessing yes but would be good to confirm and link the issue here, and if not write the issue and then link here.

I don't know exactly what the TODO is talking about. If you're in the same boat it's worth git blaming and badgering the author for their meaning / any issues they know about.

phritz: yeah this is almost certainly why, and you should ask him whether it is ok to remove this todo. if we need a todo then we should have an associated issue that it can point to.

PR #2324, opened 2019-03-18
Moved Client ImportData from node api to plum pork

### LINKED ISSUE #1956: Move Client Import Data from Node API into Porcelain
### Description

The Node API's Client Import Data method should live in porcelain. This method makes heavy use of the merkledag which will need to be moved into plumbing with #1959.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Client().ImportData()` is a part of it.

# Solution

Move `Client().ImportData()` to DAG plumbing, where it makes sense.

Resolves #1956

### DIFF
--- a/plumbing/api.go
+++ b/plumbing/api.go
@@ -294,8 +294,9 @@ func (api *API) DAGCat(ctx context.Context, c cid.Cid) (uio.DagReader, error) {
 	return api.dag.Cat(ctx, c)
 }
 
-// DAGImportData adds data from an io stream to the merkledag and returns the
-// Cid of the given data
+// DAGImportData adds data from an io reader to the merkledag and returns the
+// Cid of the given data. Once the data is in the DAG, it can fetched from the
+// node via Bitswap and a copy will be kept in the blockstore.
 func (api *API) DAGImportData(ctx context.Context, data io.Reader) (ipld.Node, error) {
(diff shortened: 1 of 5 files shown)


plumbing/api.go
```diff
@@ -291,3 +293,9 @@ func (api *API) DAGGetFileSize(ctx context.Context, c cid.Cid) (uint64, error) {
 func (api *API) DAGCat(ctx context.Context, c cid.Cid) (io.DagReader, error) {
 	return api.dag.Cat(ctx, c)
 }
+
+// DAGImportData adds data from an io stream to the merkledag and returns the
```
phritz: fyi io stream is more of a c++ thing, just calling it a reader is fine i think

plumbing/api.go
```diff
@@ -291,3 +293,9 @@ func (api *API) DAGGetFileSize(ctx context.Context, c cid.Cid) (uint64, error) {
...
+// Cid of the given data
+func (api *API) DAGImportData(ctx context.Context, data io.Reader) (ipld.Node, error) {
```
phritz: can you please comment on the side effect of this, which i think that it makes the piece available to fetch from the node via bitswap and keeps a copy in the blockstore though @ZenGround0 knows canonically

PR #2332, opened 2019-03-19
Implement Equals() method for Message, MeteredMessage and SignedMessage

anorth:
I need this for runtime checks that messages mined into blocks are identical to the message sent by a node.

I could alternatively use `CID()` for the inner Message, but it looks like that would be slower with adding extra complexity to cache the CID etc.

CC @whyrusleeping

### DIFF
--- a/types/message.go
+++ b/types/message.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -37,6 +38,19 @@ type Message struct {
 
 	Method string `json:"method"`
 	Params []byte `json:"params"`
+	// Pay attention to Equals() if updating this struct.
+}
+
+// NewMessage creates a new message.
+func NewMessage(from, to address.Address, nonce uint64, value *AttoFIL, method string, params []byte) *Message {
+	return &Message{
+		From:   from,
+		To:     to,
+		Nonce:  Uint64(nonce),
+		Value:  value,
+		Method: method,
+		Params: params,
+	}
 }
 
 // Unmarshal a message from the given bytes.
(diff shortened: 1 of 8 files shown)


types/message.go
```diff
@@ -84,14 +98,12 @@ func (msg *Message) String() string {
...
-func NewMessage(from, to address.Address, nonce uint64, value *AttoFIL, method string, params []byte) *Message {
-	return &Message{
-		From:   from,
-		To:     to,
-		Nonce:  Uint64(nonce),
-		Value:  value,
-		Method: method,
-		Params: params,
-	}
+// Equals tests whether two metered messages are equal
```
phritz: two messages are equal

types/metered_message.go
```diff
@@ -21,16 +21,7 @@ type MeteredMessage struct {
...
-
-// Unmarshal a message from the given bytes.
-func (msg *MeteredMessage) Unmarshal(b []byte) error {
-	return cbor.DecodeInto(b, msg)
-}
-
-// Marshal the message into bytes.
-func (msg *MeteredMessage) Marshal() ([]byte, error) {
-	return cbor.DumpObject(msg)
+	// Pay attention to Equals() if updating this struct.
```
phritz: we could enforce this with a unit test that uses reflection to assert the fields in message are what equal tests against

types/message.go
```diff
@@ -84,14 +98,12 @@ func (msg *Message) String() string {
...
+func (msg *Message) Equals(other *Message) bool {
+	return msg.To == other.To &&
```
phritz: trading off this way instead of using cid does make me nervous but i'm willing to try it

types/metered_message.go
ZenGround0: +1 this has saved me a few times with block: https://github.com/filecoin-project/go-filecoin/blob/master/types/block_test.go#L107

types/signed_message.go
```diff
@@ -28,6 +29,28 @@ func init() {
...
 	Signature      Signature `json:"signature"`
+	// Pay attention to Equals() if updating this struct.
+}
+
+// NewSignedMessage accepts a message `msg` and a signer `s`. NewSignedMessage returns a `SignedMessage` containing
+// a signature derived from the serialized `msg` and `msg.From`
+func NewSignedMessage(msg Message, s Signer, gasPrice AttoFIL, gasLimit GasUnits) (*SignedMessage, error) {
+	meteredMsg := NewMeteredMessage(msg, gasPrice, gasLimit)
+
+	bmsg, err := meteredMsg.Marshal()
```
ZenGround0: Total nit && not your bug: `bmsg` seems like an out of date name unless I'm missing something.

types/metered_message.go
anorth: Great suggestion, thanks

PR #2334, opened 2019-03-19
Update networking.md

anacrolix:
Add mention of BEP 5 for additional clarification (it's much more approachable for newcomers to DHTs, even if it is quite implementation heavy, skimming it is the best way to get started). I've also removed mention of k-buckets as it isn't helpful.

### DIFF
--- a/networking.md
+++ b/networking.md
@@ -24,17 +24,14 @@ Here we'll explain some of the mechanics that drive the Filecoin network layer.
 
 ## Peer discovery
 
-We use the libp2p Kademlia Distributed Hashtable (Kad DHT) service to:
+We use the [`go-libp2p-kad-dht`](https://github.com/libp2p/go-libp2p-kad-dht) service to:
 
-1. discover Filecoin peers to interface with upon boot.
-2. lookup the network addresses of a specific peer we need to interact with, starting from a peer ID, for the purposes of storage and retrieval.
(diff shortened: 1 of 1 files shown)


mgoelzer:
+1 on adding BEP 5 as an easier getting started read than the Kademlia paper.

networking.md
```diff
@@ -24,17 +24,14 @@ Here we'll explain some of the mechanics that drive the Filecoin network layer.
 
 ## Peer discovery
 
-We use the libp2p Kademlia Distributed Hashtable (Kad DHT) service to:
+We use the [`go-libp2p-kad-dht`](https://github.com/libp2p/go-libp2p-kad-dht) service to:
 
-1. discover Filecoin peers to interface with upon boot.
-2. lookup the network addresses of a specific peer we need to interact with, starting from a peer ID, for the purposes of storage and retrieval.
+1. Discover Filecoin peers at start up.
+2. Lookup the network addresses of a specific peer ID, for the purposes of Filecoin-related storage and retrieval.
```
ZenGround0: Should we also include autorelay provider record search / provide here?  We currently only mention this dht use in the autorelay section.  No strong opinions just pointing out.  cc @anorth

anorth: @ZenGround0 could you make an edit to that effect in another PR? The libp2p people don't necessarily know how we use it.

PR #2335, opened 2019-03-19
generate a wallet key file when run the 'go-filecoin wallet export' c…

yangjian102621:
It make user puzzled when run the `go-filecoin wallet import` command we should pass the wallet file path,
but user count not get the wallet file when run the `go-filecoin wallet export` command.

Generating a wallet key file when run the 'go-filecoin wallet export'  command, so that user can import it directly.

@frrist @travisperson @phritz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 go-filecoin
+localnet
 .task
 *.coverprofile
 *.out
--- a/commands/address.go
+++ b/commands/address.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/types"
+	"os"
 )
 
 var walletCmd = &cmds.Command{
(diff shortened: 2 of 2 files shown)


travisperson:
I think having a built in way to export the key to a file is a good idea, instead of having to rely on io redirection.

We should handle this with an `[-o | --output]` flag however. Not sure on a clean and easy way to allow exporting to the address though.

yangjian102621:
> I think having a built in way to export the key to a file is a good idea, instead of having to rely on io redirection.
> 
> We should handle this with an `[-o | --output]` flag however. Not sure on a clean and easy way to allow exporting to the address though.

Yes, handle this with an `[-o | --output]` flag is a good idea, and rely on io redirection is not a good choice.
I want to deal with this. If `--output` flag is given, I will not print the wallet to the console and generate the wallet file directly. Otherwise, I will directly output the wallet information to the console.

Do you think this is appropriate or we should provide another method to export wallet file?

yangjian102621:
> I think having a built in way to export the key to a file is a good idea, instead of having to rely on io redirection.
> 
> We should handle this with an `[-o | --output]` flag however. Not sure on a clean and easy way to allow exporting to the address though.

Well, I just found out that the feature that to export wallet keystore file is already can be achieved.

By run the follow command:

```bash
./go-filecoin wallet export --enc=json {address} > ./wallet.key
```
and import command

```bash
./go-filecoin wallet inport ./wallet.key
``` 

Thanks.

PR #2338, opened 2019-03-19
deps: fallback to paramcache if paramfetch fails

### LINKED ISSUE #2219: [bug] Installing deps doesn't work if ipfs.io is unreachable.
**Describe the bug**

This is in relation to #2213, where ipfs.io is not reachable for the user. This issue is for tracking the failure here in go-filecoin. However, the fix is a change to `paramfetch`, which is filed in rust-fil-proof https://github.com/filecoin-project/rust-fil-proofs/issues/535

**To Reproduce**
1. Add `127.0.0.1 ipfs.io` to our host file
2. Run `go run build/*.go deps`

**Expected behavior**
paramfetch fails, but deps continues and parameters are generated with `paramcache`.

travisperson:
paramfetch exits with a non-zero status code if any parameters were unable to be fetched.

Closes #2219

### DIFF
--- a/scripts/install-filecoin-parameters.sh
+++ b/scripts/install-filecoin-parameters.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+fetch_params() {
+  ./proofs/bin/paramfetch --all --verbose --json=./proofs/misc/parameters.json
+}
+
+generate_params() {
+  ./proofs/bin/paramcache
+}
+
+if ! fetch_params; then
+  generate_params
+fi
--- a/scripts/install-filecoin-parameters.sh
+++ b/scripts/install-filecoin-parameters.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
+set -e
+
 fetch_params() {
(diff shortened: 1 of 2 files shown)


scripts/install-filecoin-parameters.sh
```diff
@@ -0,0 +1,13 @@
...
+
+fetch_params() {
+  ./proofs/bin/paramfetch --all --verbose --json=./proofs/misc/parameters.json
+}
+
+generate_params() {
+  ./proofs/bin/paramcache
+}
+
+if ! fetch_params; then
```
laser: BLOCKING: We should always run `paramcache`, regardless of whether or not `./proofs/bin/paramfetch` produces a non-zero exit code. 

We should do this because it's possible that a rust-fil-proofs developer forgot to update the `parameters.json` file to reflect the state of the Rust code. Running `paramcache` will _always_ check/download the correct parameters - regardless of what's in the `parameters.json` file. This ensures that we won't exercise proofs code w/out the cache being hydrated ahead of time.

PR #2341, opened 2019-03-19
Feat/daemon plum pork 2340

### LINKED ISSUE #2340: Move Daemon out of API
### Description

We want to get rid of the `api` directory, but `Daemon()` is part of it. We need to find a way to remove it.

### Acceptance criteria

`Daemon()` is no longer part of `api`

### Risks + pitfalls

We don't know how to do this yet.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Daemon()` is part of it.

# Solution

Remove `Daemon` from the api directory, moving it into `commands` and `node` as appropriate.

Resolves #2340


rosalinekarr:
@shannonwells and I decided to do the bare minimum here for deleting `api` because we intend to come back for more refactoring and cleanup later when that work will result in fewer conflicts with other work.

commands/init.go
```diff
@@ -57,18 +92,110 @@ var initCmd = &cmds.Command{
...
-			api.PeerKeyFile(peerKeyFile),
-			api.WithMiner(withMiner),
-			api.DevnetTest(devnetTest),
-			api.DevnetNightly(devnetNightly),
-			api.DevnetUser(devnetUser),
-			api.AutoSealIntervalSeconds(autoSealIntervalSeconds),
-			api.DefaultAddress(defaultAddress),
-		)
+		// load configuration options
+		cfg := &DaemonInitConfig{
```
acruikshank: BLOCKING: This is too much logic to live in a command. It is entirely concerned with populating the repo. recommend moving this to repo/setup.go or repo/init and then write tests for it.

commands/init.go
```diff
@@ -27,70 +41,174 @@ var initCmd = &cmds.Command{
...
-			}
+		defer func() {
+			if closeErr := rep.Close(); closeErr != nil {
+				if err == nil {
+					err = closeErr
+				} else {
+					err = errors.Wrap(err, closeErr.Error())
+				}
+			} // else err may be set and returned as normal
+		}()
```
travisperson: I don't think this error is going to be handled correctly. I believe the function needs to have a named return value.

repo/fsrepo.go
```diff
@@ -67,6 +67,26 @@ type FSRepo struct {
...
+func FSRepoPath() string {
+	envRepoDir := os.Getenv("FIL_PATH")
+	if envRepoDir != "" {
+		return envRepoDir
+	}
+	return "~/.filecoin"
+}
+
+// CreateRepo provides a quick shorthand for initializing and opening a repo
+func CreateRepo(repoDir string, cfg *config.Config) (*FSRepo, error) {
```
phritz: repodir could be easier to understand if it required a non empty repodir, and then callers used a new getrepodir() function to get the repodir from option or fsrepopath.

commands/daemon.go
```diff
@@ -119,19 +118,22 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment)
 }
 
 func getRepo(req *cmds.Request) (repo.Repo, error) {
-	return repo.OpenFSRepo(getRepoDir(req))
+	repoDir, _ := req.Options[OptionRepoDir].(string)
+	if repoDir == "" {
```
phritz: getrepodir(string) could replace several repetitions of this logic in this pr

commands/init.go
```diff
@@ -27,70 +41,174 @@ var initCmd = &cmds.Command{
 		cmdkit.BoolOption(DevnetUser, "when set, populates config bootstrap addrs with the dns multiaddrs of the user devnet and other user devnet specific bootstrap parameters"),
 	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		repoDir := getRepoDir(req)
+		newConfig, err := getConfigFromOptions(req.Options)
+		if err != nil {
+			return err
+		}
+
+		repoDir, _ := req.Options[OptionRepoDir].(string)
```
phritz: this is a place where we could call getrepodir with the option string and pass it into createrepo

commands/init.go
```diff
@@ -27,70 +41,174 @@ var initCmd = &cmds.Command{
...
-			api.GenesisFile(genesisFile),
-			api.PeerKeyFile(peerKeyFile),
-			api.WithMiner(withMiner),
-			api.DevnetTest(devnetTest),
-			api.DevnetNightly(devnetNightly),
-			api.DevnetUser(devnetUser),
-			api.AutoSealIntervalSeconds(autoSealIntervalSeconds),
-			api.DefaultAddress(defaultAddress),
-		)
+		// TODO: don't create the repo if this fails
```
phritz: maybe kill this todo? or turn it into a note: if this fails we're not cleaning up repo

commands/init.go
```diff
@@ -27,70 +41,174 @@ var initCmd = &cmds.Command{
...
+		return consensus.DefaultGenesis, nil
+	}
+
+	sourceURL, err := url.Parse(sourceName)
+	if err != nil {
+		return nil, fmt.Errorf("invalid filepath or URL for genesis file: %s", sourceURL)
+	}
+
+	var source io.ReadCloser
+	if sourceURL.Scheme == "http" || sourceURL.Scheme == "https" {
```
phritz: when this merges could you please file an issue to remove http support, labeling it i-security. does not have to be prioritized, can just sit in the backlog

commands/main.go
```diff
@@ -265,7 +264,11 @@ func getAPIAddress(req *cmds.Request) (string, error) {
 
 	// we will read the api file if no other option is given.
 	if len(rawAddr) == 0 {
-		rawPath := filepath.Join(filepath.Clean(getRepoDir(req)), repo.APIFile)
+		repoDir, _ := req.Options[OptionRepoDir].(string)
+		if repoDir == "" {
```
phritz: another repetition of this logic

repo/fsrepo_test.go
```diff
@@ -342,6 +342,77 @@ func TestRepoAPIFile(t *testing.T) {
 	})
 }
 
+func TestCreateRepo(t *testing.T) {
+	cfg := config.NewDefaultConfig()
+
+	t.Run("folder exists", func(t *testing.T) {
```
phritz: i expected this to be an error case, based on the name. maybe: successfully creates if dir exists

repo/fsrepo_test.go
```diff
@@ -342,6 +342,77 @@ func TestRepoAPIFile(t *testing.T) {
...
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		_, err = CreateRepo(dir, cfg)
+		assert.NoError(err)
+
+		assert.True(ConfigExists(dir))
+	})
+
+	t.Run("folder does not exist", func(t *testing.T) {
```
phritz: successfully creates if dir does not exist

repo/fsrepo_test.go
```diff
@@ -342,6 +342,77 @@ func TestRepoAPIFile(t *testing.T) {
...
+
+		dir = filepath.Join(dir, "nested")
+
+		_, err = CreateRepo(dir, cfg)
+		assert.NoError(err)
+
+		assert.True(ConfigExists(dir))
+	})
+
+	t.Run("folder is not writable", func(t *testing.T) {
```
phritz: fails with error if dir is not writeable

repo/fsrepo_test.go
```diff
@@ -342,6 +342,77 @@ func TestRepoAPIFile(t *testing.T) {
...
+		err = os.Mkdir(dir, 0444)
+		assert.NoError(err)
+
+		_, err = CreateRepo(dir, cfg)
+		assert.Contains(err.Error(), "permission denied")
+
+		assert.False(ConfigExists(dir))
+	})
+
+	t.Run("config file already exists", func(t *testing.T) {
```
phritz: fails with error if config already exists

repo/fsrepo_test.go
```diff
@@ -342,6 +342,77 @@ func TestRepoAPIFile(t *testing.T) {
...
+		dir, err := ioutil.TempDir("", "init")
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		ioutil.WriteFile(filepath.Join(dir, "config.json"), []byte("hello"), 0644)
+
+		_, err = CreateRepo(dir, cfg)
+		assert.Contains(err.Error(), "repo already initialized")
+
+		assert.True(ConfigExists(dir))
```
phritz: seems like this assertion should move up above the createrepo call?

commands/init.go
rosalinekarr: This was copy pasted from the previous implementation. It turns out we don't need the error or a go func at all. The only condition under which `Close()` throws an error is if the file was already closed. I'm replacing this with a simple defer.

PR #2343, opened 2019-03-19
Feat/swarm plum pork

### LINKED ISSUE #1984: Move Swarm Peers from Node API to Porcelain
### Description

Move the swarm peers method from the Node API into Porcelain. This will require moving several other aspect of libp2p into network plumbing.

### LINKED ISSUE #1985: Move Swarm Connect from Node API into Porcelain
### Description

Move the swarm connect method from the Node API into Porcelain. This will require moving several other aspect of libp2p into network plumbing.

### LINKED ISSUE #1986: Move Swarm Find Peers from Node API into Porcelain
### Description

Move the swarm find peers method from the Node API into Porcelain. This will require moving several other aspect of libp2p into network plumbing.

rosalinekarr:
# Problem

We want to remove the `api` directory, but `Swarm()` is part of it.

# Solution

Move `Swarm` to `net` and `plumbing` as necessary.

Resolves #1984 
Resolves #1985 
Resolves #1986


commands/swarm.go
```diff
@@ -131,7 +131,7 @@ var findPeerDhtCmd = &cmds.Command{
 			return err
 		}
 
-		out, err := GetAPI(env).Swarm().FindPeer(req.Context, peerID)
```
ZenGround0: Blocking: please move this command to `go-filecoin dht` in the dht.go file.

net/network.go
```diff
@@ -1,22 +1,72 @@
...
+}
+
 // Network is a unified interface for dealing with libp2p
 type Network struct {
 	host host.Host
 	*pubsub.Subscriber
 	*pubsub.Publisher
 	metrics.Reporter
-	*Router
+	Router routing.IpfsRouting
```
ZenGround0: Blocking: hey @anorth, this works for me but you blocked this when I first did it so wanted to run it by you to make sure you are on board.

net/network.go
```diff
@@ -53,3 +103,77 @@ func (network *Network) GetPeerID() peer.ID {
 func (network *Network) GetBandwidthStats() metrics.Stats {
 	return network.Reporter.GetBandwidthTotals()
 }
+
+// Connect connects to peers at the given addresses
+func (network *Network) Connect(ctx context.Context, addrs []string) ([]peer.ID, error) {
```
ZenGround0: Nice catch removing the unused `Success` field

net/network.go
```diff
@@ -53,3 +103,77 @@ func (network *Network) GetPeerID() peer.ID {
 func (network *Network) GetBandwidthStats() metrics.Stats {
 	return network.Reporter.GetBandwidthTotals()
 }
+
+// Connect connects to peers at the given addresses
```
phritz: should say that it stops trying once it hits its first error. probably it should return results like the original did, but set the success field to false on connection error *and then keep trying the others*. i think maybe follow up with this change?

net/network.go
```diff
@@ -53,3 +103,77 @@ func (network *Network) GetPeerID() peer.ID {
...
+	for _, c := range conns {
+		pid := c.RemotePeer()
+		addr := c.RemoteMultiaddr()
+
+		ci := SwarmConnInfo{
+			Addr: addr.String(),
+			Peer: pid.Pretty(),
+		}
+
+		/* FIXME(steb)
```
phritz: can we turn this into an issue or just delete it?

net/network.go
```diff
@@ -1,22 +1,72 @@
...
+	return ci.Peers[i].Addr < ci.Peers[j].Addr
+}
+
+func (ci SwarmConnInfos) Len() int {
+	return len(ci.Peers)
+}
+
+func (ci SwarmConnInfos) Swap(i, j int) {
+	ci.Peers[i], ci.Peers[j] = ci.Peers[j], ci.Peers[i]
+}
```
anorth: I think this could go in a separate `net/swarm.go` file, but happy for that to come later.

net/network.go
anorth: Yes, please indirect through the net.Router. The reason for this type is to abstract out libp2p/ipfs from most of the code so that we can more easy fake it for testing and/or upgrade the dependee code with minimal disruption elsewhere. I realise we have yet to clearly demonstrate how and the benefits of doing so in the rest of the codebase (but you can refer to pubsub for an example).

Thanks Zen.

shannonwells:
> Approved conditioned on two blocking comments being addressed
Will do.

net/network.go
shannonwells: Deleted.

commands/dht_daemon_test.go
```diff
@@ -28,3 +31,56 @@ func TestDhtFindPeer(t *testing.T) {
...
+
+	op1 := d1.RunSuccess("chain", "ls")
+	results := strings.Split(op1.ReadStdout(), "\n")
+	assert.True(len(results) > 1)
+	genesisKey := results[0]
+
+	t.Run("basic command succeeds", func(t *testing.T) {
+		op1 = d1.Run("dht", "findprovs", genesisKey)
+		results = strings.Split(op1.ReadStdoutTrimNewlines(), "\n")
+		// all miners should have the genesis block key
```
ZenGround0: BLOCKING: Whaa?  Since when are we using the dht for content routing?  I don't understand how this is passing.  If it is passing it seems like it should not be (cc @whyrusleeping).

Could be that I don't understand pubsub + dht interactions well enough but lets sync before merging this.

ZenGround0: FYI I had a test that I loved with all my heart in the initial `findprovs` PR that exercised both this command and the auto relay daemon flags.  However it was too flaky to make it in because I could not figure out how to guarantee a discovery time.  I am concerned that even if for some reason this should be working this test will suffer from similar flakiness.

ZenGround0: To be more clear: the only thing that I think this command should resolve is the cid representing the autorelay service.

ZenGround0: Ok I filed an issue here #2357.  This was a good bug to find.  IMO the right way to move forward with this PR is to delete all `findprovs` tests.  If you can figure out a way to test `findprovs` on the autorelay records that is not flaky then I would welcome that test but I believe that is too much to ask / out of scope for this PR.

shannonwells: @ZenGround0 these tests haven't flaked on me locally; they're pretty reliable so far.

Issue #2315, opened 2019-03-18
DefaultSyncer#HandleNewBlocks prints debug information to stdout


laser:
### Description

[This line](https://github.com/filecoin-project/go-filecoin/blob/master/chain/default_syncer.go#L329), introduced on March 15, writes debugging information to stdout. It should instead use our go-filecoin log.

### Acceptance criteria

Debug information is written to log instead of stdout.

PR #2356, opened 2019-03-20
replace printf with logger as required by #2315

contrun:


laser:
Hello @contrun. Thanks for your PR! I implemented and merged this change [in a separate PR](https://github.com/filecoin-project/go-filecoin/pull/2358) just a few moments ago.

PR #2367, opened 2019-03-20
localnet: Binary flag and README

travisperson:
Add support to specify the `go-filecoin` binary that will be used instead of requiring the user to build. However, for easy of use for developers, localnet will default first to the binary in the project directory, then fallback to searching the $PATH.

The binary that will be used will be printed as the default.

_binary built in the go-filecoin project_
```
Usage of ./localnet:
  -binpath go-filecoin
    	set the binary used when executing go-filecoin commands (default "/home/travis/src/github.com/filecoin-project/go-filecoin/go-filecoin")
```

_binary **not** built in the go-filecoin project, but found in path_
```
Usage of ./localnet:
  -binpath go-filecoin
    	set the binary used when executing go-filecoin commands (default "/home/travis/bin/go-filecoin")
```

_no binary found anywhere. Running without setting the `-binpath` flag will result in an error_
```
Usage of ./localnet:
  -binpath go-filecoin
    	set the binary used when executing go-filecoin commands
```

### DIFF
--- a/tools/fast/bin/localnet/README.md
+++ b/tools/fast/bin/localnet/README.md
@@ -1,0 +1,83 @@
+# localnet
+
+localnet is a FAST binary tool that quickly and easily, sets up a local network
+on the users computer. The network will stay standing till the program is closed.
+
+### Example
+
+```
+localnet -shell
+```
+
+### Building
+
+The localnet tool expects that you can already build `go-filecoin`. Please refer
+to the README in the root of this project for details.
+
+```
+go build -o localnet main.go
+```
+
+### Usage
+
+```
+Usage of ./localnet:
+  -binpath go-filecoin
+    	set the binary used when executing go-filecoin commands
+  -blocktime duration
+    	duration for blocktime (default 5s)
+  -miner-collateral string
+    	amount of fil each miner will use for collateral (default "500")
+  -miner-count int
+    	number of miners (default 5)
+  -miner-expiry string
+    	expiry value used when creating ask for miners (default "86400")
+  -miner-pledge uint
+    	number of sectors to pledge for each miner (default 10)
+  -miner-price string
+    	price value used when creating ask for miners (default "0.0000000010")
+  -shell
+    	setup a filecoin client node and enter into a shell ready to use
+  -small-sectors
+    	enables small sectors (default true)
+  -workdir string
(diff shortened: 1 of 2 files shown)


tools/fast/bin/localnet/README.md
```diff
@@ -0,0 +1,83 @@
+# localnet
```
phritz: are we linking to this file from anywhere?

tools/fast/bin/localnet/main.go
```diff
@@ -66,8 +71,27 @@ func init() {
 		minerExpiryArg     = minerExpiry.Text(10)
 	)
 
+	// We default to the binary built in the project directory, fallback
+	// to searching path.
+	binpath, err = getFilecoinBinary()
```
phritz: personally id distinguish between an unexpected error (error) and not finding the file (which you could signal with a bool return value). in the first case this should bail out. in the second, continue into the conditional below.

tools/fast/bin/localnet/main.go
```diff
@@ -66,8 +71,27 @@ func init() {
...
+	if err != nil {
+		// Look for `go-filecoin` in the path to set `binpath` default
+		// If the binary is not found, an error will be returned. If the
+		// error is ErrNotFound we ignore it.
+		// Error is handled after flag parsing so help can be shown without
+		// erroring first
+		binpath, err = exec.LookPath("go-filecoin")
+		if err != nil {
+			xerr, ok := err.(*exec.Error)
+			if ok && xerr.Err == exec.ErrNotFound {
```
phritz: if not ok bail?

tools/fast/bin/localnet/README.md
travisperson: No, not yet. It may get linked to by a will page at some point I'm sure.

tools/fast/bin/localnet/main.go
travisperson: No, we only want to check the second half of the expression if the cast was successful.

tools/fast/bin/localnet/main.go
travisperson: The issue is I don't want to bail out before parsing the flags, as I want to still handle `-help` even if this errors, and if the user passes in a value on the flag, the error is mute.

Issue #2276, opened 2019-03-14
Make a worker deps struct so it can be used by node and protocolapi


shannonwells:
### Description
For the new protocolapi, almost all the deps passed in for mining are for the DefaultWorker.  Let's make a deps struct for it and make node and protocolapi use it (once #2268 is in).

### Acceptance criteria
Node uses the worker deps struct to instantiate a worker
ProtocolAPI takes a worker deps struct with many of the current params inside it instead of in params.

### Risks + pitfalls
None really

### Where to begin
worker.go, DefaultWorker

PR #2370, opened 2019-03-20
Some refactoring to make a cleaner block mining api

### LINKED ISSUE #2277: Pass in mining node's mining worker to protocolapi
### Description
Cleanup after "quick n dirty" yanking of stuff out of node api.  From @phritz :
> pass the mining node's mining worker to the api and have the api call directly into it. probably all the bookkeeping stuff strewn throughout the node that keeps track of mining state should move here as well. 

### Detail
The block API in protocol/block/api.go requires a lot of dependencies that likely could be reduced if it were passed the DefaultWorker as a dependency.  So do that.   If there are opportunities to pull the so-called "bookkeeping stuff that keeps track of mining state" out of node, do that too.

### Where to start
* protocol/block/api.go
* node.go:
    `setupProtocols`, possibly also `StartMining`, `StopMining`

shannonwells:
# Problem
Closes #2277 
Too many dependencies for the block api.

# Solution
1. Rename block.API to block.MiningAPI (as requested)
1. Some refactoring to reduce dependencies in protocol/block/mining_api.
1. Also pulls out some inline functions and sets them as dependencies on node:
    * because it's cleaner
    * to assist with peeling off these deps/responsibilities later.


node/node.go
```diff
@@ -122,6 +123,9 @@ type Node struct {
 	miningDoneWg       *sync.WaitGroup
 	AddNewlyMinedBlock newBlockFunc
 	blockTime          time.Duration
+	GetAncestorsFunc   mining.GetAncestors
+	GetStateTreeFunc   mining.GetStateTree
+	GetWeightFunc      mining.GetWeight
```
shannonwells: These 3 above can now be set for testing.

node/node.go
```diff
@@ -450,6 +454,11 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 		Router:       router,
 	}
 
+	// set up mining worker funcs
+	nd.GetAncestorsFunc = nd.getAncestors
+	nd.GetStateTreeFunc = nd.getStateTree
+	nd.GetWeightFunc = nd.getWeight
+
```
shannonwells: these are now the default settings.

protocol/block/mining_api_test.go
```diff
@@ -47,30 +47,23 @@ func TestAPI_StartStopMining(t *testing.T) {
...
-		bt, md,
-		nd.MsgPool,
-		nd.PorcelainAPI,
-		nd.PowerTable,
+		bt,
 		nd.StartMining,
 		nd.StopMining,
-		nd.Syncer,
-		nd.Wallet), nd
+		nd.CreateMiningWorker), nd
```
shannonwells: Here we see the other reason for abstracting creation of the mining worker.

protocol/block/mining_api.go
```diff
@@ -0,0 +1,73 @@
+package block
```
shannonwells: rename.

protocol/block/mining_api.go
```diff
@@ -0,0 +1,73 @@
...
+// MiningOnce mines a single block in the given context, and returns the new block.
+func (a *MiningAPI) MiningOnce(ctx context.Context) (*types.Block, error) {
+	ts := a.chainReader.Head()
+
+	miningWorker, err := a.createWorkerFunc(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := mining.MineOnce(ctx, miningWorker, a.mineDelay, ts)
```
shannonwells: MineOnce (formerly known as Once) was always creating its own worker; this at least puts such code in a single spot so we know it's done the same way here as in node.StartMining.

node/node.go
```diff
@@ -770,48 +779,15 @@ func (node *Node) StartMining(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to get mining owner address for miner %s", minerAddr)
 	}
 
-	blockTime, mineDelay := node.MiningTimes()
-
-	if node.MiningScheduler == nil {
-		getStateFromKey := func(ctx context.Context, tsKey string) (state.Tree, error) {
```
ZenGround0: Really happy to see these closures going away

node/node.go
```diff
@@ -1077,6 +1039,69 @@ func (node *Node) setupProtocols() error {
...
+		return nil, errors.Wrap(err, "could not get key from miner actor")
+	}
+
+	minerOwnerAddr, err := node.miningOwnerAddress(ctx, minerAddr)
+	if err != nil {
+		log.Errorf("could not getKey from miner actor")
+		return nil, err
+	}
+	return mining.NewDefaultWorker(
+		node.MsgPool, node.getStateTree, node.getWeight, node.getAncestors, processor, node.PowerTable,
```
ZenGround0: --Disclaimer-- Not at all suggesting this change in this PR.

The miner worker is clearly a consumer of a lot of core pieces.  If we raise its logic up into the block mining protocol then we can probably hit all of these dependencies through the porcelain api.  This might open up a way to get rid of the need to use these somewhat clunky functions for initializing the worker.  Instead the blockMiner (previously some combo of scheduler and worker), as defined in `protocol/block_mining.go`  could be modularly tested by using the carefully designed mocking features of the porcelain.

ZenGround0: In retrospect this is probably exactly what you are already thinking as moving the mining api to `protocol/block` seems to imply a blockMiner by symmetry with `protocol/storage`'s storageMiner and `protocol/retrieval`'s retrievalMiner.

node/node.go
```diff
@@ -1077,6 +1039,69 @@ func (node *Node) setupProtocols() error {
...
+		node.Blockstore, node.CborStore(), minerAddr, minerOwnerAddr, minerPubKey,
+		node.Wallet, node.blockTime), nil
```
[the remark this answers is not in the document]
shannonwells: that's planned - I filed this last week: https://github.com/filecoin-project/go-filecoin/issues/2276

PR #2376, opened 2019-03-21
Add message status command unifying outbox, pool and chain.

### LINKED ISSUE #2165: Provide message status command
### Description
Provide a command that shows status for a message in the message pool and/or outbound message queue and/or chain. Include its age (calculated locally) and nonce (and possibly the actor's nonce for reference) and whether it has been mined.

A motivating use case is tracking the status of an outbound message that appears first in the outbound queue, the message pool (if the node is mining), but eventually either appears on chain or expires. Neither appearing on chain or expiring should cause this status command to fail (at least for a while or until the node restarts).

Note existing command for `mpool ls|show` and a likely new command for listing the outbound queue, which will share some functionality.

### Acceptance criteria
- A new command prints a single message's status and age, regardless where it appears in the queue, pool or chain (within reason).
    - It's ok if rebooting the node causes status to be lost. This story is scoped only to making status accessible, not keeping track of the information internally.

### Risks + pitfalls
- A message may be in either or both the message pool and outbound queue
- Finding messages on chain may be expensive or require a new index

### Where to begin
commands/message.go

anorth:
Closes #2165.

### DIFF
--- a/commands/message.go
+++ b/commands/message.go
@@ -8,24 +8,26 @@ import (
 
 	"gx/ipfs/QmR8BauakNcBa3RbE4nbQu76PDiJgoQgz8AJdhJuiU4TAw/go-cid"
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
-	cmdkit "gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
-	cmds "gx/ipfs/Qmf46mr235gtyxizkKUkTH5fo62Thza2zwXR4DWC7rkoqF/go-ipfs-cmds"
+	"gx/ipfs/Qmde5VP1qUkyQXKCfmEUA7bP64V2HAptbJ7phuPp7jXWwg/go-ipfs-cmdkit"
+	"gx/ipfs/Qmf46mr235gtyxizkKUkTH5fo62Thza2zwXR4DWC7rkoqF/go-ipfs-cmds"
 
 	"github.com/filecoin-project/go-filecoin/abi"
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/core"
 	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/plumbing/msg"
 	"github.com/filecoin-project/go-filecoin/plumbing/mthdsig"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 var msgCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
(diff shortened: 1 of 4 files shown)


commands/message.go
```diff
@@ -221,6 +223,82 @@ var msgWaitCmd = &cmds.Command{
...
+		// Look on chain
+		result.ChainMsg, result.OnChain, err = api.MessageFind(req.Context, msgCid)
+		if err != nil {
+			return err
+		}
+		return re.Emit(&result)
+	},
+	Type: &MessageStatusResult{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, res *MessageStatusResult) error {
```
phritz: i would have expected printing the message out below as well

plumbing/msg/waiter.go
```diff
@@ -79,39 +96,46 @@ func (w *Waiter) Wait(ctx context.Context, msgCid cid.Cid, cb func(*types.Block,
...
 	}()
 
+	chainMsg, found, err := w.waitForMessage(ctx, ch, msgCid)
+	if found {
+		return cb(chainMsg.Block, chainMsg.Message, chainMsg.Receipt)
+	}
+	return err
+}
+
+func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgCid cid.Cid) (*ChainMessage, bool, error) {
```
phritz: please comment that the second return value means found

commands/message.go
```diff
@@ -221,6 +223,82 @@ var msgWaitCmd = &cmds.Command{
 	},
 }
 
+// MessageStatusResult is the status of a message on chain or in the message queue/pool
+type MessageStatusResult struct {
+	InPool    bool
+	PoolMsg   *types.SignedMessage
+	InOutbox  bool
```
shannonwells: this is a confusing name, but I can't think of anything better. A comment would help since I had to read all the way down to know what it means.

commands/message_daemon_test.go
```diff
@@ -138,3 +138,38 @@ func TestMessageSendBlockGasLimit(t *testing.T) {
...
+		d.RunSuccess("mining once")
+
+		status = d.RunSuccess("message", "status", msgcid).ReadStdout()
+
+		assert.NotContains(status, "In outbox")
+		assert.NotContains(status, "In mpool")
+		assert.Contains(status, "On chain")
+		assert.Contains(status, "1234") // the "value"
+	})
+}
```
shannonwells: blocking-ish -- we can has sad path tests?  e.g. invalid CID, or MessageFind fails

commands/message.go
anorth: It does at line 294

commands/message_daemon_test.go
anorth: The MessageFind not finding the message is already exercised. I added a test for a CID not found at all.

Issue #2363, opened 2019-03-20
`testing` package included by packages


travisperson:
**Describe the bug**

The `testing` package injects `flag` options when it's included which makes the flags show up in help text when using the standard `flag` singleton. We currently include the `testing` package in a few packages.

-  actor/builtin/testing.go
-  core/testing.go
-  net/testing.go
-  node/testing.go
-  types/testing_messages.go

This makes it hard to use the standard `flag` package when building binaries and consuming `go-filecoin` packages.

The `testing` package should only be included into packages explicitly used for testing (`testhelpers`) or files which have the `_test` suffix.

I put together a PR (#2362) to move files (basically call `testing.go` files) to use the proper suffix. However, these packages are consumed by `testhelpers`, and result in build errors.

There is a bit of cleanup that needs to occur to properly handle this and something we need to do so that `go-filecoin` packages can be consumed by others without polluting the standard `flag` package.

The work around at the moment is to create a new flag set and not use the global for any binary that consumes go-filecoin packages.
```go
import (
    flg "flag"
    "os"
)

var flag = flg.NewFlagSet(os.Args[0], flg.ExitOnError)
....
flag.Parse(os.Args[1:])
```

phritz:
possibly related: https://github.com/filecoin-project/go-filecoin/issues/329

anorth:
Just to clarify, I think @travisperson is referring to the Go SDK package named `testing`, not something of ours.

I have also experimented with moving the contents of `testing.go` files into `*_test` packages but have determined that it's probably not the right approach. Test support code from one module often needs to be used by test support code in a dependent module. Instead, I propose we form `package/testing/foo.go` subpackages provide test-support code for `package`. These packages can depend on other `anotherpackage/testing` subpackages. We could probably enforce the desired constraints with `grep`, though I don't think it's particularly high value if we have clear precedents.

This could happen as part of #2148 (but need not wait for it, either).

PR #2383, opened 2019-03-21
gengen: Add README

travisperson:
- Adds README to gengen
- Fixes testing flags in gengen help (#2363)

### DIFF
--- a/gengen/README.md
+++ b/gengen/README.md
@@ -41,8 +41,8 @@ Usage of ./gengen:
 The gengen configuration file defines the artifacts produced.
 
 - `keys` defines the number of keys which will be produced
-- `preAlloc` is an array defining the amount of FIL for the key
-- `miners` is an array defining miners, the `owner` is the key index, and `powe` is the amount of power the miner will have in the genesis block.
+- `preAlloc` is an array defining the amount of FIL for each key
+- `miners` is an array defining miners, the `owner` is the key index, and `power` is the amount of power the miner will have in the genesis block.
 
 Example
 
--- a/gengen/README.md
+++ b/gengen/README.md
@@ -1,0 +1,37 @@
+# gengen
+
+gengen is a filecoin tool that builds the genesis.car, which is used to seed the
+genesis block.
+
+### Example
+
+```
+go-filecon $ ./gengen/gengen --keypath fixtures --out-car fixtures/genesis.car --out-json fixtures/gen.json --config fixtures/setup.json
+```
+
+### Building
+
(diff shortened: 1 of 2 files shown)


gengen/README.md
```diff
@@ -0,0 +1,37 @@
+# gengen
+
+gengen is a filecoin tool that builds the genesis.car, which is used to seed the
+genesis block.
+
+### Example
```
frrist: non-blocking: in addition to this example it would be nice if we provided examples of:
- How to initialize a filecoin node with the `genesis.car` file
- How to use the `*.key` artifacts of the gengen call
- What the `setup.json` file does -- e.g. it allocates power and funds to an address that is in the genesis block.

gengen/README.md
```diff
@@ -0,0 +1,64 @@
...
+    "1000000000000",
+    "1000000000000",
+    "1000000000000"
+  ],
+  "miners": [{
+    "owner": 0,
+    "power": 1
+  }]
+}
+```
```
travisperson: I'm not too familiar with the configuration (and it honestly needs to be updated as it's a bit overly complex to describe). Suggestions are welcome, but you'll need to write out exact wording.

gengen/README.md
```diff
@@ -0,0 +1,64 @@
...
+    	provides the seed for randomization, defaults to current unix epoch (default 1553189402)
+```
+
+#### Configuration File
+
+The gengen configuration file defines the artifacts produced.
+
+- `keys` defines the number of keys which will be produced
+- `preAlloc` is an array defining the amount of FIL for the key
+- `miners` is an array defining miners, the `owner` is the key index, and `powe` is the amount of power the miner will have in the genesis block.
```
anorth: Typo in `powe`

gengen/README.md
```diff
@@ -0,0 +1,64 @@
...
+  -seed int
+    	provides the seed for randomization, defaults to current unix epoch (default 1553189402)
+```
+
+#### Configuration File
+
+The gengen configuration file defines the artifacts produced.
+
+- `keys` defines the number of keys which will be produced
+- `preAlloc` is an array defining the amount of FIL for the key
```
anorth: "the key" -> "each key"

gengen/README.md
anorth: I think "owner" is an index into the keys, so you could clarify that above.

travisperson: It's already stated. I'm happy to update, but you'll need to provide exact wording as I said above, the configuration is overly complex to describe. I'd have to write in paragraph form to describe how it actually works.

> the `owner` is the key index

PR #2384, opened 2019-03-21
Feat/pork-swarm-follow-up-2382

### LINKED ISSUE #2382: Swarm Connect should not exit upon the first failure
### Description

Swarm Connect accepts a slice of addresses to connect to, and exits at the first failed connection. As suggested by @phritz in [this comment](https://github.com/filecoin-project/go-filecoin/pull/2343#discussion_r267086964), Swarm Connect should instead continue trying to connect to other addresses.

One solution to this would be to return a channel for each connection with a results struct containing the resulting peer ids and any errors, rather than a slice and one error as it does now.

### Acceptance criteria

Swarm Connect no longer stops after a single failure.

### Where to begin

rosalinekarr:
# Problem

`go-filecoin swarm connect` stops after the first address to fail rather than trying other addresses.

# Solution

Change swarm connect to return a channel with errors and peer ids for each attempted connection.

Resolves #2382

### DIFF
--- a/net/network.go
+++ b/net/network.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 
 	ma "gx/ipfs/QmNTCey11oxhb1AxDnQBRHtdhap6Ctud872NjAYPYYXPuc/go-multiaddr"
+	"gx/ipfs/QmRhFARzTHcFh8wUxwN5KvyTGq73FLC65EfFAhz8Ng7aGb/go-libp2p-peerstore"
 	"gx/ipfs/QmTu65MVbemtUxJEWgsTtzv9Zv9P8rvmqNA4eG9TrTRGYc/go-libp2p-peer"
 	"gx/ipfs/QmU7iTrsNaJfu1Rf5DrvaJLH9wJtQwmP4Dj8oPduprAU68/go-libp2p-swarm"
 	"gx/ipfs/QmVmDhyTTUcQXFD1rRQ64fGLMSAoaQvNH3hwuaCFAPq2hy/errors"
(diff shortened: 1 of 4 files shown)


net/network.go
```diff
@@ -116,18 +125,20 @@ func (network *Network) Connect(ctx context.Context, addrs []string) ([]peer.ID,
...
-
-		output[i] = pi.ID
-
-		if err := network.host.Connect(ctx, pi); err != nil {
-			return nil, errors.Wrapf(err, "peer: %s", output[i].Pretty())
+	go func() {
+		for _, pi := range pis {
+			swrm.Backoff().Clear(pi.ID)
+			err := network.host.Connect(ctx, pi)
+			out <- ConnectionResult{
```
phritz: this is going to block until the receiver receives the value, might be better to send the result to the channel on a goroutine

net/network.go
```diff
@@ -103,9 +105,18 @@ func (network *Network) GetBandwidthStats() metrics.Stats {
 	return network.Reporter.GetBandwidthTotals()
 }
 
+// ConnectionResult represents the result of an attempted connection from the
+// Connect method
```
ZenGround0: nit channeling inner @phritz: no period at the end of the sentence

net/network.go
```diff
@@ -103,9 +105,18 @@ func (network *Network) GetBandwidthStats() metrics.Stats {
...
+type ConnectionResult struct {
+	PeerID peer.ID
+	Err    error
+}
+
 // Connect connects to peers at the given addresses. Does not retry, and does not
 // try to connect to any more peers if any connection fails.
```
ZenGround0: BLOCKING: this comment is no longer true

PR #2390, opened 2019-03-22
Eliminate some direct dependencies on node (not via api)

### LINKED ISSUE #2371: eliminate some direct dependencies on node (not via api)
1. https://github.com/filecoin-project/go-filecoin/blob/6043562c26a2af6c90bfe1bb6fc06323c5af6013/protocol/retrieval/client.go#L23 should be eliminated in favor of the client taking the network or host directly. 

2. https://github.com/filecoin-project/go-filecoin/blob/6043562c26a2af6c90bfe1bb6fc06323c5af6013/protocol/storage/client.go#L51 should be eliminated as well. blocktime should be plumbing and should be part of a protocolparams struct that is returned; we'll end up surfacing a lot of parameters this way. as for makeprotocolrequest we should eliminate this abstraction https://github.com/filecoin-project/go-filecoin/blob/6043562c26a2af6c90bfe1bb6fc06323c5af6013/protocol/storage/client.go#L306 and just have a thing that depends on the host, which is passed indirect (or maybe network). that thing could be a free function or a member of the client, up to you.

shannonwells:
# Problem
closes #2371 

# Solution
* Get rid of dependency on node, pull dependencies out of there and use those in storage and retrieval APIs

### Also
* For your consideration:  following the advice of “make all dependencies explicit” in this [go best practices article](https://peter.bourgon.org/go-best-practices-2016/#top-tip-9),  the log variable is a part of the client struct and not set up at the package level.  I think this is a good practice and would like to see us start using this pattern with our logs in particular.
* Kill xargs file that was mistakenly checked in


protocol/retrieval/client.go
```diff
@@ -19,31 +22,27 @@ import (
...
-// TODO: better name
-type clientNode interface {
-	Host() host.Host
-}
-
 // Client is a client interface to the retrieval market protocols.
 type Client struct {
-	node clientNode
+	host host.Host
+	log  logging.EventLogger
```
shannonwells: make all dependencies explicit

protocol/storage/client.go
```diff
@@ -70,23 +66,29 @@ type clientPorcelainAPI interface {
...
 	smc := &Client{
-		node: nd,
-		api:  api,
+		api:       api,
+		blockTime: blockTime,
+		host:      host,
+		log:       logging.Logger("storage/client"),
 	}
-	return smc, nil
+	smc.ProtocolRequestFunc = MakeProtocolRequest
```
phritz: curious why set this outside the above?

shannonwells: good catch, it was originally not a free function and I forgot to move it back in.

PR #2399, opened 2019-03-22
move net/testing to testhelpers/net

shannonwells:
# Problem
This problem is also blocking #2390 due to import cycles issues, uncovered (or perhaps caused) by the move of chain/testing.go to testhelpers/chain.go

Supports #2363 as well.

# Solution
As the title says.
Also: unite RequireRandomPeerID calls into one and Don't Panic (since it was decided to follow the custom of Must --> panic on fail, Require --> require.NoError )


net/bootstrap_test.go
```diff
@@ -6,19 +6,20 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
phritz: blocking: why do this? why not just import assert through gx as assert and eliminate the githubby (another domain for you) path? my guess is your editor added the githubby path because you go got it at some point? in any case do we need both?

shannonwells: yeah that was an auto-add.

PR #2402, opened 2019-03-25
Combine publish and deploy workflow; add test-devnet target

ognots:
combining publish and deploy workflow reduces duplicated build_linux jobs running, increasing execution time.

allow targetting test-devnet OR user-devnet to add permanent test-devnet Continuous Deployment capabilities

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,11 @@ params:
       description: "job is being invoked from user devnet workflow"
       type: boolean
       default: false
+  test_devnet_param: &test_devnet_param
+    test_devnet:
+      description: "job is being invoked from test devnet workflow"
+      type: boolean
+      default: false
 
 commands:
   rust_fil_proofs_checksum:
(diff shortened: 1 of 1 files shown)


ognots:
Tested in https://circleci.com/workflow-run/831c0aef-01be-486c-87fd-ee6a6c2ba625

ognots:
Screen Shot 2019-03-25 at 12 19 03 PM

This screenshot demonstrates concurrency in this workflow

ognots:
https://circleci.com/gh/filecoin-project/go-filecoin/13245 is a test deploying user devnet that passes

ognots:
PR should be merged alongside https://github.com/filecoin-project/go-filecoin-infra/pull/205

.circleci/config.yml
```diff
@@ -682,8 +746,18 @@ workflows:
...
+            - build_macos
+          filters:
+            branches:
+              ignore:
+                - /.*/
+            tags:
+              only:
                 - /^(test\-devnet\-)*\d+\.\d+\.\d+$/
 
       - build_docker_img:
```
travisperson: Below this line, `user_devnet: true` needs to be updated.

travisperson:
@ognots should we (for clarity purposes) make tags for the user devnet match the test devnet by requiring that tags be `user-devnet-x.y.z`?

/cc @mishmosh @phritz

.circleci/config.yml
ognots: nice catch, thanks @travisperson !

ognots:
@travisperson the only slightly complication with that, is the release name and docker tag will be `user-devnet-x.y.z` instead of `x.y.z`

currently I would consider the User Devnet the closest thing to a "stable" environment, so having it deploy our stable release `x.y.z` is more fitting than `user-devnet-x.y.z` IMHO

curious to hear other opinions.

anorth:
My opinion is that the explicitness in the longer name is better. PL holds self-describing structures as a key design principle; the label that describes what it's labelling does this. It's too early to guess at labels for future networks and whether or not the current user devnet will become one of them, or remain distinct.

Issue #2210, opened 2019-03-08
Propose-storage-deal can fail with "context deadline exceeded"


pengxiankaikai:
**Describe the bug**
execute:
go-filecoin client propose-storage-deal --allow-duplicates fcqdzp28x8vnph7aedzw3pwlpwrnxjh77zayx75c5 $CID 3 200
Error: error creating payment: context deadline exceeded

return Error, but balance change:
go-filecoin wallet balance fcq2j4j2hu7mc0fm0xyxk7euucezhzwg9mfpkp7fq
801

go-filecoin wallet balance fcq2j4j2hu7mc0fm0xyxk7euucezhzwg9mfpkp7fq
735

go-filecoin wallet balance fcq2j4j2hu7mc0fm0xyxk7euucezhzwg9mfpkp7fq
668.6124205

The miner I specified also did not store changes in file space.

minerAddress:
fcqdzp28x8vnph7aedzw3pwlpwrnxjh77zayx75c5   
Proven Storage： 0MB

thinks!

pengxiankaikai:
supplement：
Transaction charge record found in the block：
{
                "meteredMessage":{
                    "message":{
                        "to":"fcqp606qfk5gwmq6ac24g4mhv3cr8zzf67vqkpulh",
                        "from":"fcq2j4j2hu7mc0fm0xyxk7euucezhzwg9mfpkp7fq",
                        "nonce":"Bg==",
                        "value":"66",
                        "method":"createChannel",
                        "params":"glYAALqFPeGzjsI5kjYR2PWRdmGxgiy7Q/3CAg=="
                    },
                    "gasPrice":"0",
                    "gasLimit":"rAI="
                },
                "signature":"cz5ndFgRlNh4v+Bl1a13cWfEADms/vvBCwjAHMLazV97zKjo9OMb8cYUMXHEQxFv2dGPLAYzwafXYryykvD5AAA="
            }

rkowalick:
@pengxiankaikai Seeing `context deadline exceeded` when running `client propose-storage-deal` is definitely a bug.

The fact that your payment channel was created and your balanced was deducted is a known problem, as we have not yet implemented verification of storage before the client pays the miner.



Currently, the payment vouchers are sent out without any confirmation that the data was stored by the storage miner, so I think this might be related.

pengxiankaikai:
[...]

PR #2410, opened 2019-03-25
remove context deadline from file transfer

ZenGround0:
Right now the context deadline covers file transfer which prevents large files from getting transferred.

See this issue for more details https://github.com/filecoin-project/go-filecoin/issues/2411 and proposed follow up improvements.

Many bug reports have mentioned this, #2210 for example.


protocol/storage/client.go
```diff
@@ -178,6 +178,9 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 
 	// send proposal
 	var response storagedeal.Response
+	// We reset the context to not timeout to allow large file transfers
+	// to complete.
+	ctx := context.Background()
```
phritz: should be the context passed in

PR #2422, opened 2019-03-26
Feat/chain api improvements 1847

### LINKED ISSUE #1847: Chain API improvements
### Description
Discussing with @phritz `BlockHistory` seems a little off.  A better solution (that could be used to implement block history) is to have a `Head` and `GetNext` method allowing tipsets to traverse themselves.

### Acceptance criteria
BlockHistory is removed from the chain interface.  It is reimplemented elsewhere with more irreduceable and ergonomic api methods.  

### Risks + pitfalls
Depending on how we go about it this work has the potential to blow up in scope.  If we want tipsets to be able to traverse themselves maybe that means the `GetNext` method is defined on tipset which opens up a lot of complexity.  This work could also start bleeding into questions of whether state or state cids should be stored on tipsets.  This could start a chain reaction where we start untangling lots of technical debt in the chain store interface.

rosalinekarr:
# Problem

The Chain API is inconsistent and poorly typed. Specifically, the BlockHistory method should be implemented elsewhere.

# Solution

Refactor the Chain API

Resolves #1847


plumbing/api.go
```diff
@@ -204,7 +196,7 @@ func (api *API) MessagePreview(ctx context.Context, from, to address.Address, me
 // MessageQuery calls an actor's method using the most recent chain state. It is read-only,
 // it does not change any state. It is use to interrogate actor state. The from address
 // is optional; if not provided, an address will be chosen from the node's wallet.
-func (api *API) MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, *exec.FunctionSignature, error) {
+func (api *API) MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
```
rosalinekarr: Ok, this is a pretty big change, so I want to explain it a bit. Apologies for the wall of text.

At the time of this commit, we're only using the method signature returned from `MessageQuery` in three places:

- `getLastUsedSectorID` method in `node/node.go`
- `isBootstrapMinerActor` method in `protocol/storage/miner.go`
- `getActorSectorCommitments` method in `protocol/storage/miner.go`

We also use the `MessageQuery` method and discard the method signature in over 36 different places. (I'm not going to list them all.) The method signature of a message query seems to be largely irrelevant on most invocations, since the calling code typically knows what to expect back in terms of an encoded response.

There may be some argument that by using a method signature and decoding with it we can avoid brittleness in the call sites when changing the method signatures themselves, but I'm inclined to believe that said brittleness can never be completely avoided because, at some point, we will almost always have to work with the typed object returned in some meaningful way, so if we change function signatures, we'll end up breaking these method invocations anyway. Using the method signature like this at best will only ever hide these problems, not solve them.

Additionally, because `MessageQuery` returns the method signature, `plumbing/msg` has a hard dependency on `plumbing/mthdsig` which contradicts our expressed goal of not creating dependencies between plumbing packages. Further, the *only* places the `mthdsig` package is used are here and in the `ActorGetSignature` plumbing method, which essentially does the exact same thing as `mthdsig`. This makes me think `mthdsig` should just be part of `actor` plumbing instead and `MessageQuery` shouldn't do two things.

The reason I made this change in this PR rather than another is because `mthdsig` plumbing is dependent on the chain core API's `Head` method, which was removed as part of the goal of this issue in favor of the chain plumbing's `ChainHead` method, so leaving method signatures in the `MessageQuery` return would mean either making `msg` plumbing dependent on `chain` plumbing or moving `MessageQuery` into porcelain, both of which seem significantly more wrong than just dropping this mostly unused return value.

phritz: > mthdsig should just be part of actor plumbing instead and MessageQuery shouldn't do two things.

hear! hear! fully agree that messagequery was doing too much and should not return the signature. if someone needs it, they can make a call to get it. 

i have more thoughts re plumbing/porcelain organization that will follow with, maybe here or maybe in an issue. i think some of us have different assumptions about what should depend on what and my hypothesis -- sit down for it -- is that this is because we never really talked about it or wrote it down. so lets do that.

porcelain/payments.go
```diff
@@ -11,16 +11,16 @@ import (
...
 )
 
 // cpPlumbing is the subset of the plumbing.API that CreatePayments uses.
 type cpPlumbing interface {
 	MessageSend(ctx context.Context, from, to address.Address, value *types.AttoFIL, gasPrice types.AttoFIL, gasLimit types.GasUnits, method string, params ...interface{}) (cid.Cid, error)
-	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, *exec.FunctionSignature, error)
+	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
 	MessageWait(ctx context.Context, msgCid cid.Cid, cb func(*types.Block, *types.SignedMessage, *types.MessageReceipt) error) error
-	ChainLs(ctx context.Context) <-chan interface{}
+	ChainLs(ctx context.Context) <-chan *chn.ChainLsResult
```
phritz: alpha (not your bug)

plumbing/msg/sender.go
```diff
@@ -54,12 +58,13 @@ type Sender struct {
 
 // NewSender returns a new Sender. There should be exactly one of these per node because
 // sending locks to reduce nonce collisions.
-func NewSender(signer types.Signer, chainReader chain.ReadStore, blockTimer core.BlockTimer,
+func NewSender(signer types.Signer, chainReader chain.ReadStore, cst *hamt.CborIpldStore, blockTimer core.BlockTimer,
```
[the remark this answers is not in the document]
ZenGround0: An alternate proposal for a cleanup followon: tipsets are given a pointer to a state tree upon creation and `ts.State()` returns the tree.  This should be evaluated in more depth before going down this road

chain/default_store.go
```diff
@@ -359,94 +358,19 @@ func (store *DefaultStore) writeTipSetAndState(tsas *TipSetAndState) error {
 	return store.ds.Put(key, val)
 }
 
-// Head returns the current head.
-func (store *DefaultStore) Head() types.TipSet {
+// GetHead returns the current head.
+func (store *DefaultStore) GetHead() types.SortedCidSet {
```
anorth: What motivates this change? It seems inconvenient and led to lots of callers to change. Is it to hide the head tipset to prevent external modification?

Please document why it's structured this ways so that a future developer who finds it inconvenient doesn't mistakenly change it back without context.

ZenGround0: @anorth I asked the same question and @phritz left an in depth answer in slack.  I'll tag you there so you get access to it.

chain/default_syncer_test.go
```diff
@@ -952,7 +956,9 @@ func TestTipSetWeightDeep(t *testing.T) {
 	verifier = proofs.NewFakeVerifier(true, nil)
 	con = consensus.NewExpected(cst, bs, th.NewTestProcessor(), &consensus.MarketView{}, calcGenBlk.Cid(), verifier)
 	syncer := chain.NewDefaultSyncer(cst, con, chainStore, blockSource)
-	baseTS := chainStore.Head() // this is the last block of the bootstrapping chain creating miners
+	headTipSetAndState, err := chainStore.GetTipSetAndState(ctx, chainStore.GetHead())
```
ZenGround0: nit: a helper function `requireGetHeadTipSet` (or whichever better name you find) that runs `GetHead` and then `GetTipSetAndState` would help simplify tests for readers.

chain/get_ancestors.go
```diff
@@ -85,57 +88,35 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
...
+			return
 		}
 	}
 }
 
 // CollectAtMostNTipSets collect N tipsets from the input channel.  If there
 // are fewer than n tipsets in the channel it returns all of them.
-func CollectAtMostNTipSets(ctx context.Context, ch <-chan interface{}, n uint) ([]types.TipSet, error) {
-	var ret []types.TipSet
+func CollectAtMostNTipSets(ctx context.Context, chainReader ReadStore, ts *types.TipSet, n uint) (ret []types.TipSet, err error) {
```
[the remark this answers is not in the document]
ZenGround0: my b fam.  Y'all can assign me the testing follow on issue.

chain/get_ancestors.go
```diff
@@ -70,12 +74,11 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
 	// This is needed because CollectTipSetsOfHeightAtLeast necessarily reads out
```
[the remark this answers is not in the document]
ZenGround0: Comment is still valid if we move to an iterator as suggested below (we can't use the same iterator as it will read past by one).  If we keep it like this the comment should be dropped but the code fine.

chain/get_ancestors.go
```diff
@@ -85,57 +88,35 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
 // CollectTipSetsOfHeightAtLeast collects all tipsets with a height greater
 // than or equal to minHeight from the input channel.  Precondition, the input
```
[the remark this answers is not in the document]
ZenGround0: +1 to giving both of these functions iterators of some kind as input.

chain/get_ancestors.go
```diff
@@ -85,57 +88,35 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
...
 // channel contains interfaces which may be tipsets or errors.
-func CollectTipSetsOfHeightAtLeast(ctx context.Context, ch <-chan interface{}, minHeight *types.BlockHeight) ([]types.TipSet, error) {
-	var ret []types.TipSet
+func CollectTipSetsOfHeightAtLeast(ctx context.Context, chainReader ReadStore, ts *types.TipSet, minHeight *types.BlockHeight) (ret []types.TipSet, err error) {
```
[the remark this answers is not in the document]
ZenGround0: question from curiosity: why is delegating all ctx error passing to `ts.GetNext` (the only method waiting on the context) a bad thing?  Is it wrong to assume that this call will handle context errors correctly?  It seems like you are advocating for assuming that other parts of the system are doing things wrong which makes each individual function more robust to programmer error elsewhere in the system but potentially adds in unnecessary redundancy.

chain/get_ancestors.go
```diff
@@ -85,57 +88,35 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
...
-				}
-				ret = append(ret, raw)
-			}
+		ret = append(ret, *ts)
+		ts, err = ts.GetNext(ctx, chainReader)
+		if ts == nil || err != nil {
+			return
+		}
+		// Add tipset to ancestors.
+		h, err = ts.Height()
```
ZenGround0: blocking: I believe we should return a nil slice in the case any error is thrown.  Because of the way you are doing named return variables a bunch of tipsets are being returned.

This belief comes from a memory of being told the same thing long ago on this project.  @phritz feel free to override this block if this isn't actually a big deal.

ZenGround0: If I'm right to block this then it also means the concise if statement on 96 will need to be broken into two or otherwise expanded a bit.

ZenGround0: same for 118 in `CollectAtMostNTipSets` below

chain/get_ancestors_test.go
```diff
@@ -21,7 +21,10 @@ func setupGetAncestorTests(require *require.Assertions) (context.Context, *th.Te
 // requireGrowChain grows the given store numBlocks single block tipsets from
 // its head.
 func requireGrowChain(ctx context.Context, require *require.Assertions, blockSource *th.TestFetcher, chainStore chain.Store, numBlocks int) {
-	link := chainStore.Head()
+	head := chainStore.GetHead()
```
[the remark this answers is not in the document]
ZenGround0: Same helper can be used as suggested in the review of the default syncer!

chain/default_store.go
```diff
@@ -359,94 +358,19 @@ func (store *DefaultStore) writeTipSetAndState(tsas *TipSetAndState) error {
...
-			p, err := store.GetBlock(ctx, pid)
-			if err != nil {
-				return errors.Wrap(err, "error retrieving block from store")
-			}
-			tips = append(tips, p)
-		}
-	}
-
-	return nil
+	return store.head.Height()
```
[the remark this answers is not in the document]
rosalinekarr: wow, so nothing is atomic in go, not even pointer assignment

that's warps my C-minded brain

chain/default_syncer.go
```diff
@@ -225,37 +225,41 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 
 	// TipSet is validated and added to store, now check if it is the heaviest.
 	// If it is the heaviest update the chainStore.
-	nextParentSt, err := syncer.tipSetState(ctx, parent.String()) // call again to get a copy
+	nextParentSt, err := syncer.tipSetState(ctx, parent.ToSortedCidSet()) // call again to get a copy
```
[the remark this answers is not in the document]
rosalinekarr: Can you elaborate? I'm not sure how an iterator would work here. I don't see us iterating over anything here.

chain/get_ancestors.go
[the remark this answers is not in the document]
rosalinekarr: Since this is going to use the new TipSet iterator in `chain/traversal.go`, I'll just change said iterator to respect `ctx.Err()` on `Next()`. That should cover this case, the next comment, and any future uses of `IterAncestors`.

chain/get_ancestors.go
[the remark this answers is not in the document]
rosalinekarr: Addressed here: https://github.com/filecoin-project/go-filecoin/pull/2422#discussion_r274073642

node/node.go
```diff
@@ -981,11 +996,8 @@ func (node *Node) miningOwnerAddress(ctx context.Context, miningAddr address.Add
 
 // BlockHeight returns the current block height of the chain.
 func (node *Node) BlockHeight() (*types.BlockHeight, error) {
```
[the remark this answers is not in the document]
rosalinekarr: @phritz, after looking into it, the only places we use `BlockHeight()` are protocols where `PorcelainAPI` should be available anyway, so I'm just going to remove this method and replace all those invocations with direct calls to `ChainHead()`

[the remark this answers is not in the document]
rosalinekarr: also, @ZenGround0, after digging into `chain` a bit more, it looks like we don't ever actually use the context. Don't know how we missed that, but I'm going to remove it as a function argument.

plumbing/chn/chain.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Ls returns a channel of tipsets from head to genesis
+func (chn *Chain) Ls(ctx context.Context) <-chan *ChainLsResult {
+	out := make(chan *ChainLsResult)
+	go func() {
+		defer close(out)
+		tsas, err := chn.reader.GetTipSetAndState(ctx, chn.reader.GetHead())
+		ts := &tsas.TipSet
+		for ts != nil {
```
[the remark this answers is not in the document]
rosalinekarr: Addressed here: https://github.com/filecoin-project/go-filecoin/pull/2422#discussion_r274073642

chain/traversal.go
```diff
@@ -59,6 +58,12 @@ func (it *TipsetIterator) Complete() bool {
...
-	it.value, it.err = GetParentTipSet(it.ctx, it.store, it.value)
-	return it.err
+	var err error
+	select {
+	case <-it.ctx.Done():
+		return it.ctx.Err()
+	default:
+		it.value, err = GetParentTipSet(it.ctx, it.store, it.value)
+	}
+	return err
```
anorth: @rosalinekarr please propose and land this independent change in a separate PR and rebase onto it. The iterator has tests and they should be extended to exercise this branch.

Landing independent changes ahead of a larger PR is a good pattern for keeping reviews manageable. I call it a "prefactor".

rosalinekarr: Done: https://github.com/filecoin-project/go-filecoin/pull/2544

plumbing/chn/chain.go
```diff
@@ -0,0 +1,71 @@
...
+	return chain.GetRecentAncestorsOfHeaviestChain(ctx, chn.reader, descendantBlockHeight)
+}
+
+type ChainLsResult struct {
+	TipSet *types.TipSet
+	Error  error
+}
+
+// Ls returns a channel of tipsets from head to genesis
+func (chn *Chain) Ls(ctx context.Context) <-chan *ChainLsResult {
```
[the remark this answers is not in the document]
rosalinekarr: Changed to an iterator

plumbing/api.go
```diff
@@ -118,41 +128,23 @@ func (api *API) ConfigGet(dottedPath string) (interface{}, error) {
...
+func (api *API) ChainGetBlock(ctx context.Context, id cid.Cid) (*types.Block, error) {
+	return api.chain.GetBlock(ctx, id)
 }
 
-// BlockGet gets a block by CID
-func (api *API) BlockGet(ctx context.Context, id cid.Cid) (*types.Block, error) {
-	return api.chain.GetBlock(ctx, id)
+// GetRecentAncestorsOfHeaviestChain returns the recent ancestors of the
+// `TipSet` with height `descendantBlockHeight` in the heaviest chain.
+func (api *API) GetRecentAncestorsOfHeaviestChain(ctx context.Context, descendantBlockHeight *types.BlockHeight) ([]types.TipSet, error) {
```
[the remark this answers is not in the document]
rosalinekarr: This plumbing was only necessary because of the `SampleChainRandomness` method in porcelain. I moved said method into plumbing and removed `GetRecentAncestorsOfHeaviestChain`.

chain/default_store.go
```diff
@@ -359,94 +358,19 @@ func (store *DefaultStore) writeTipSetAndState(tsas *TipSetAndState) error {
...
 	defer store.mu.RUnlock()
 
-	return store.head
+	return store.head.ToSortedCidSet()
 }
 
 // BlockHeight returns the chain height of the head tipset.
 // Strictly speaking, the block height is the number of tip sets that appear on chain plus
 // the number of "null blocks" that occur when a mining round fails to produce a block.
 func (store *DefaultStore) BlockHeight() (uint64, error) {
```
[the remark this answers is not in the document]
rosalinekarr: This method is used by the vm. See line 74 of `exec/exec.go`

plumbing/msg/sender.go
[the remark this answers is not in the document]
rosalinekarr: I'll add an issue to clean this up after this PR is merged.

Fwiw, I would prefer to add a `GetLatestState` method or function somewhere, rather than adding more state to the TipSet struct. I think linking low level objects to their containers is a bad pattern as it introduces a lot of undefined behavior when dealing with those objects in isolation.

For example, we may want to temporarily create isolated TipSets for tests or certain chain operations, and leaving the store pointer nil could introduce mistakes if we ever call a method that utilizes said pointer.

chain/get_ancestors.go
```diff
@@ -66,76 +71,50 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
...
-				return nil, raw
-			case types.TipSet:
-				// Add tipset to ancestors.
-				ret = append(ret, raw)
-			}
+	var err error
+	for i := uint(0); i < n && !iterator.Complete(); i++ {
+		ret = append(ret, iterator.Value())
+		if err = iterator.Next(); err != nil {
+			return nil, err
```
anorth: Thanks, this diff is a compelling argument for iterators as a pattern in preference to channels.

chain/default_store.go
acruikshank: I'm also confused about this change. It seems unnatural and unnecessary. Could @phritz, @ZenGround0 or @rosalinekarr respond in this PR?

anorth: Copy-paste from [slack](https://filecoinproject.slack.com/archives/CEHHJNJS3/p1554234370122200?thread_ts=1554213262.096100&cid=CEHHJNJS3):

> there are a lot of considerations for how to define good apis but here are a few generalizations it is hard to go wrong with: 1. push complexity onto the caller side of the api; generally dont provide convenience functions if your callers could. 2. keep your apis as small as possible 3. keep your calls as orthogonal as possible. All three of these principles are in evidence in the new api, which is not perfect. It 1 eliminates convenience methods where a caller can simply make two calls and 2 reduces the number of calls in the api and 3 makes the calls that remain more orthogoal, though not completely so. having apis that follow these principles make the api easier to optimize, easier to understand, and easier to satisfy with a variety of implementations. they limit the blast radius of changes, should you have to make them. martin fowler, one of the gang of four, identifies this philosophy of api design as the minimal interface and articulates some of its benefits here https://martinfowler.com/bliki/MinimalInterface.html.  this is in contrast with sprawling expansive convenience interfaces he calls humane interfaces https://martinfowler.com/bliki/HumaneInterface.html. the systems and distributed systems worlds are strongly oriented around minimal interfaces these days for the reasons articulated above. many web frameworks eg rails are oriented around the latter in the name of developer productivity. I find the maintenance and understandability of narrow interfaces to be far more compelling than saving a few lines when writing a consumer of the interface.

rosalinekarr:
In addition to this work, I've added the following issues for follow up refactors:

https://github.com/filecoin-project/go-filecoin/issues/2551
https://github.com/filecoin-project/go-filecoin/issues/2552
https://github.com/filecoin-project/go-filecoin/issues/2553
https://github.com/filecoin-project/go-filecoin/issues/2554

chain/default_store.go
```diff
@@ -118,32 +118,30 @@ func (store *DefaultStore) Load(ctx context.Context) error {
...
 	var genesii types.TipSet
-	err = store.walkChain(ctx, headTs.ToSlice(), func(tips []*types.Block) (cont bool, err error) {
-		if logStatusEvery != 0 && (tips[0].Height%logStatusEvery) == 0 {
-			logStore.Infof("load tipset: %s, height: %v", tips[0].Cid().String(), tips[0].Height)
+	for iterator := IterAncestors(ctx, store, headTs); !iterator.Complete(); err = iterator.Next() {
+		if err != nil {
+			return err
 		}
-		ts, err := types.NewTipSet(tips...)
+		genesii = iterator.Value()
```
phritz: i find it confusing to call these intermediate tipsets genesii. i prefer the pre-change way of calling them simply ts, which is more accurate. anything but genesii, except for the actual genesis ts.

chain/default_syncer.go
```diff
@@ -179,16 +179,16 @@ func (syncer *DefaultSyncer) tipSetState(ctx context.Context, tsKey string) (sta
 // Precondition: the caller of syncOne must hold the syncer's lock (syncer.mu) to
 // ensure head is not modified by another goroutine during run.
 func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.TipSet) error {
-	head := syncer.chainStore.Head()
+	head := syncer.chainStore.GetHead()
```
phritz: i'm a little iffy on how this works, but it must because locking behavior hasn't changed but the caller has an exclusive lock, how are we getting a reader lock in gethead?

chain/get_ancestors.go
```diff
@@ -66,76 +71,50 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, chainReader Read
...
-
-				// Check for termination.
-				if types.NewBlockHeight(h).LessThan(minHeight) {
-					return ret, nil
-				}
-				ret = append(ret, raw)
-			}
+	var err error
+	var h uint64
+	for ; !iterator.Complete(); err = iterator.Next() {
```
phritz: nice, i find this much nicer with iterators than channels personally

commands/chain.go
```diff
@@ -54,19 +58,19 @@ var chainLsCmd = &cmds.Command{
...
+		iter, err := GetPorcelainAPI(env).ChainLs(req.Context)
+		if err != nil {
+			return err
+		}
+		for ; !iter.Complete(); err = iter.Next() {
+			if err != nil {
+				return err
+			}
+			if len(iter.Value()) == 0 {
+				panic("tipsets from this channel should have at least one member")
```
phritz: from this iterator

commands/message.go
```diff
@@ -164,9 +163,9 @@ var msgWaitCmd = &cmds.Command{
 		found := false
 		err = GetPorcelainAPI(env).MessageWait(req.Context, msgCid, func(blk *types.Block, msg *types.SignedMessage, receipt *types.MessageReceipt) error {
 			found = true
-			sig, err2 := GetPorcelainAPI(env).ActorGetSignature(req.Context, msg.To, msg.Method)
-			if err2 != nil && err2 != mthdsig.ErrNoMethod && err2 != mthdsig.ErrNoActorImpl {
-				return errors.Wrap(err2, "Couldn't get signature for message")
+			sig, err := GetPorcelainAPI(env).ActorGetSignature(req.Context, msg.To, msg.Method)
+			if err != nil && err.Error() != "no method" && err.Error() != "failed to load actor code" {
```
phritz: a little iffy matching on the string, can we not define a type some place that doesn't make bad dependencies?

commands/message.go
```diff
@@ -164,9 +163,9 @@ var msgWaitCmd = &cmds.Command{
 		found := false
 		err = GetPorcelainAPI(env).MessageWait(req.Context, msgCid, func(blk *types.Block, msg *types.SignedMessage, receipt *types.MessageReceipt) error {
 			found = true
-			sig, err2 := GetPorcelainAPI(env).ActorGetSignature(req.Context, msg.To, msg.Method)
-			if err2 != nil && err2 != mthdsig.ErrNoMethod && err2 != mthdsig.ErrNoActorImpl {
-				return errors.Wrap(err2, "Couldn't get signature for message")
+			sig, err := GetPorcelainAPI(env).ActorGetSignature(req.Context, msg.To, msg.Method)
```
phritz: dont we need to not mask err here (hence err2 in original)?

plumbing/api.go
```diff
@@ -121,43 +128,26 @@ func (api *API) ConfigGet(dottedPath string) (interface{}, error) {
...
-// GetRecentAncestorsOfHeaviestChain returns the recent ancestors of the
-// `TipSet` with height `descendantBlockHeight` in the heaviest chain.
-func (api *API) GetRecentAncestorsOfHeaviestChain(ctx context.Context, descendantBlockHeight *types.BlockHeight) ([]types.TipSet, error) {
-	return chain.GetRecentAncestorsOfHeaviestChain(ctx, api.chain, descendantBlockHeight)
+// ChainHead returns the head tipset
+func (api *API) ChainHead() (*types.TipSet, error) {
+	return api.chain.Head()
 }
 
 // ChainLs returns a channel of tipsets from head to genesis
```
phritz: iterator

plumbing/bcf/blockchain_facade.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/sampling"
+	"github.com/filecoin-project/go-filecoin/state"
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+	hamt "github.com/ipfs/go-hamt-ipld"
+	"github.com/pkg/errors"
+)
+
+// BlockChainFacade is a simple facade for the chain core api
```
phritz: say why this thing needs to exist, or what purpose it serves

chain/default_syncer.go
rosalinekarr: a lock was added on lines 366 & 367

commands/message.go
rosalinekarr: `err` is only masked in the context of the anonymous callback function where the original `err` is never used, so it's not an issue. If we need to access the original `err` again, we can change it back.

PR #2425, opened 2019-03-26
extend block retrieval timeout

### LINKED ISSUE #2424: Chain fails to sync due to short block retrieval timeout
### Description

After the deploy, we're seeing a lot of:
```
processBlock(): processing block from network: failed to fetch all requested blocks: context deadline exceeded
```

This appears to be due to a 1 second timeout when trying to retrieve blocks. This is short for any sort of network call, so this should be lengthened.

### Acceptance criteria

1. Better timeout is determined and implemented.

acruikshank:
closes #2424

### Problem

We currently only wait for 1 second before timing out when fetching a block for chain syncing. This is a network call and should be increased on principle. We have observed many of these timeouts when bringing up the new user net, so we now know it should be longer.

### Solution

Increase the timeout to 30 seconds. We don't know of any reason why this should not be 30 times longer and it should greatly reduce these timeouts.

### DIFF
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -317,17 +317,8 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
 // attempt to validate and caches invalid blocks it has encountered to
 // help prevent DOS.
 func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.Cid) error {
-	// ********** WARNING **********
-	//
-	// This concurrency model is flawed.  The mutex is held during a possibly
-	// long call to the network.  TODO: re-evaluate / re-design the concurrency
-	// model to allow for collectChain to be called outside the lock.
-	// FYI the two biggest concurrency concerns at present would be addressed
-	// by locking after collectChain completes, so my hunch is this can be
-	// fixed by simply moving the lock call below collectChain.
 	logSyncer.Debugf("trying to sync %v\n", blkCids)
-	syncer.mu.Lock()
-	defer syncer.mu.Unlock()
+
 	// If the store already has all these blocks the syncer is finished.
 	if syncer.chainStore.HasAllBlocks(ctx, blkCids) {
 		return nil
@@ -349,6 +340,11 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 		return err
 	}
 	parent := parentTsas.TipSet
+
+	// Start locking now, so simultaneous calls to this method don't confuse each other about
+	// the state of the chain head.
+	syncer.mu.Lock()
+	defer syncer.mu.Unlock()
 
 	// Try adding the tipsets of the chain to the store, checking for new
 	// heaviest tipsets.
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -19,7 +19,7 @@ import (
 
 // The amount of time the syncer will wait while fetching the blocks of a
 // tipset over the network.
-var blkWaitTime = time.Second // TODO set this parameter in an informed way too
+var blkWaitTime = 30 * time.Second
 var (
 	// ErrChainHasBadTipSet is returned when the syncer traverses a chain with a cached bad tipset.
 	ErrChainHasBadTipSet = errors.New("input chain contains a cached bad tipset")
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -49,12 +49,12 @@ type syncFetcher interface {
 // tipset in the incoming chain, and assumptions regarding the existence of
 // grandparent state in the store.
 type DefaultSyncer struct {
-	// This mutex ensures at most one call to HandleNewBlocks executes at
+	// This mutex ensures at most one call to HandleNewTipset executes at
 	// any time.  This is important because at least two sections of the
 	// code otherwise have races:
 	// 1. syncOne assumes that chainStore.Head() does not change when
 	// comparing tipset weights and updating the store
-	// 2. HandleNewBlocks assumes that calls to widen and then syncOne
+	// 2. HandleNewTipset assumes that calls to widen and then syncOne
 	// are not run concurrently with other calls to widen to ensure
 	// that the syncer always finds the heaviest existing tipset.
 	mu sync.Mutex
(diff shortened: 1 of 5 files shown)


chain/default_syncer.go
```diff
@@ -350,6 +348,11 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 	}
 	parent := parentTsas.TipSet
 
+	// Start locking now, so simultaneous calls to this method don't confuse each other about
+	// the state of the chain head.
+	syncer.mu.Lock()
+	defer syncer.mu.Unlock()
```
acruikshank: We've moved this lock down so chain syncing won't deadlock with the longer timeout. We've analyzed the code and believe this is safe.

chain/default_syncer.go
```diff
@@ -183,6 +183,13 @@ func (syncer *DefaultSyncer) tipSetState(ctx context.Context, tsKey string) (sta
 // Precondition: the caller of syncOne must hold the syncer's lock (syncer.mu) to
 // ensure head is not modified by another goroutine during run.
 func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.TipSet) error {
+	head := syncer.chainStore.Head()
+
+	// if tipset is already head, we've been here before. do nothing.
+	if head.Equals(next) {
```
acruikshank: Added this check to avoid errors when processing the same tipset twice. This is more likely after moving the lock after the chain fetch.

chain/default_syncer.go
```diff
@@ -350,6 +348,11 @@ func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.
 	}
 	parent := parentTsas.TipSet
 
+	// Start locking now, so simultaneous calls to this method don't confuse each other about
+	// the state of the chain head.
+	syncer.mu.Lock()
+	defer syncer.mu.Unlock()
+
```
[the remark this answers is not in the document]
ZenGround0: Good question. Nope.  There are three places where the head can be set at the end of HandleNewBlocks: 
1. the old head
2. the head of the incoming chain
3. the widened tipset

In general we need to compare 2 and 3 and can't just assume 2 is heavier.

chain/default_syncer.go
ZenGround0: Already talked to @acruikshank about this but acking that duplicate network traffic and duplicate block processing will happen as is and that this should change for efficiency.

ZenGround0:
@phritz 
> Is it weird that getBlksMaybeFromNet attempts to get all the blocks from the network even if some of the blocks are available locally in the chain store (but not say the whole tipset)? seems like we could do an awful lot of extra fetching. (i guess the bitswap service might have them, but only might right?)

The fetcher wraps the blockservice that wraps the blockstore that holds onto all blocks ever fetched from the network.  All blocks in the chainstore were fetched through this path and so today there will be not extra fetching afaik.

> fyi i dont think i see discussion of widen in the ec spec, though it makes sense.

Yes, right now "widen" is an implementation detail needed to be up to spec with EC but it is not explicitly mentioned in the spec.  It can be inferred from the heaviest chain head selection rule, the tipset rule and the weighting function.  It is unclear to me if widen and things like it have a place in the spec but I am open to that.

chain/default_syncer.go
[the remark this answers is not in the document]
phritz: i think just a comment saying that the fall through is intentional and why. makes sense now that you have explained.

chain/default_syncer.go
phritz: this looks like a really bad tradeoff to me, but ok. ac, i guess i ask that you explicitly call out the two consequences above of lock placement in a comment. also still curious about 1.

phritz:
> Yes, right now "widen" is an implementation detail needed to be up to spec with EC but it is not explicitly mentioned in the spec. It can be inferred from the heaviest chain head selection rule, the tipset rule and the weighting function. It is unclear to me if widen and things like it have a place in the spec but I am open to that.

i would strongly suggest that we not rely on people to infer things from the spec that are important.  it might be an easy inference to make for someone who really understands ec but if you're a developer who just wants to implement a thing you are unlikely to properly derive much from whatever first principles are presented.

chain/default_syncer.go
```diff
@@ -19,7 +19,7 @@ import (
 
 // The amount of time the syncer will wait while fetching the blocks of a
 // tipset over the network.
-var blkWaitTime = time.Second // TODO set this parameter in an informed way too
+var blkWaitTime = 30 * time.Second
```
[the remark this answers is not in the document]
acruikshank: Updated.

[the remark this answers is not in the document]
acruikshank: The app is dead in the water until the chain syncs. The appropriate max sync time is as long as it takes. 30 seconds seems about right for a network glitch.

chain/default_syncer.go
acruikshank: @phritz I discussed this with @ZenGround0 and we went through root causes, potential DOS vectors and what happens when the syncer encounters valid forks. In the end we decided it was better to move the lock back where it was. We don't see any real consequences of stacking HandleNewBlock (now HandleNewTipset) calls pending a long chain sync, and it seems safer. If you identify problems that the longer timeout will cause with this locking, we should probably get on a call to discuss.

chain/default_syncer.go
```diff
@@ -317,17 +324,8 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
 // attempt to validate and caches invalid blocks it has encountered to
 // help prevent DOS.
 func (syncer *DefaultSyncer) HandleNewBlocks(ctx context.Context, blkCids []cid.Cid) error {
```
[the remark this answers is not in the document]
acruikshank: This is a good idea. I went ahead and implemented it and renamed HandleNewBlocks to HandleNewTipset. There a few places where the code is significantly cleaner.

chain/default_syncer.go
```diff
@@ -312,31 +315,24 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
...
-	// This concurrency model is flawed.  The mutex is held during a possibly
-	// long call to the network.  TODO: re-evaluate / re-design the concurrency
-	// model to allow for collectChain to be called outside the lock.
-	// FYI the two biggest concurrency concerns at present would be addressed
-	// by locking after collectChain completes, so my hunch is this can be
-	// fixed by simply moving the lock call below collectChain.
-	logSyncer.Debugf("trying to sync %v\n", blkCids)
+func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids types.SortedCidSet) error {
+	logSyncer.Debugf("trying to sync %v\n", tipsetCids)
 	syncer.mu.Lock()
```
phritz: I would definitely add a comment here talking about the tradeoffs of moving this lock, otherwise you will just be having this conversation with someone over again when they try to change it in the future.

node/node.go
```diff
@@ -503,9 +503,10 @@ func (node *Node) Start(ctx context.Context) error {
 		// TODO it is possible the syncer interface should be modified to
```
phritz: can we just kill this TODO?

PR #2439, opened 2019-03-28
WIP Reject deals whose piece is larger than the sector size

### LINKED ISSUE #2034: Client and miner should reject attempts to store a piece larger than sector size
### Description
A client can attempt to propose a storage deal for a file that's larger than the miner sector size. This apparently just times out at the client.

See https://github.com/filecoin-project/go-filecoin/issues/1965#issuecomment-464187940

(There may have been other circumstances related to invalid nonces that caused the particular failure observed above)

### Acceptance criteria
- The client rejects attempts to store a too-large file without consulting the miner/network
- In case a client misbehaves, a miner immediately rejects a proposal for a too-large file

### Risks + pitfalls
None

### Where to begin
protocol/storage/client.go

rkowalick:
WIP

Resolves #2034

### DIFF
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -167,6 +167,15 @@ func (sm *Miner) receiveStorageProposal(ctx context.Context, sp *storagedeal.Sig
 		return sm.proposalRejector(sm, p, err.Error())
 	}
 
+	var sectorSize uint64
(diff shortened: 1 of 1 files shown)


rkowalick:
Closing this in favor of a new strategy based on #2492

PR #2444, opened 2019-03-29
FAST: add devnet environment, and network retrieval test

travisperson:
Adds a devnet environment for FAST allowing the use of FAST with deployment devnets.

#2457


tools/fast/process.go
```diff
@@ -91,9 +94,14 @@ func NewFilecoinProcess(ctx context.Context, c IPTBCoreExt, eo EnvironmentOpts)
 		ctx:        ctx,
 		initOpts:   eo.InitOpts,
 		daemonOpts: eo.DaemonOpts,
+		detached:   detached,
 	}
 }
 
+func (f *Filecoin) Detach() {
+	f.external = true
```
travisperson: `detached`

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,183 @@
...
+			break
+		}
+	}
+
+	// remove the provess from the process list
+	return os.RemoveAll(p.core.Dir())
+}
+
+// GetFunds retrieves a fixed amount of tokens from an environment
+func GetFunds(ctx context.Context, env Environment, p *Filecoin) error {
```
travisperson: Would be interested in thoughts on how to get this to work with the MemoryGenesis environment. My first thought was another function which would set some value on the environment. Not sure if it's really needed. Would be nice to provide reusability though.

laser:
@travisperson - What's gating the merge of this PR? Need any reviewers?

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,183 @@
...
+	"net/url"
+	"os"
+	"sync"
+
+	cid "github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log"
+
+	"github.com/filecoin-project/go-filecoin/address"
+
+	iptb "github.com/ipfs/iptb/testbed"
```
frrist: nit: import paths

tools/fast/environment_devnet.go
frrist: What do you think of having the MemoryGenesis run a faucet process? We could reuse almost all of this code if that was the case.

Another idea would be having the MemoryGenesis keep a reference to the genesis node process and adding separate logic to pull funds from that.

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,183 @@
...
+		var toAddr address.Address
+		if err := p.ConfigGet(ctx, "wallet.defaultAddress", &toAddr); err != nil {
+			return err
+		}
+
+		data := url.Values{}
+		data.Set("target", toAddr.String())
+
+		uri := url.URL{
+			Host:   fmt.Sprintf("faucet.%s.kittyhawk.wtf", devenv.network),
```
frrist: should this be 
```suggestion
			Host:   fmt.Sprintf("%s.kittyhawk.wtf:9797", devenv.network),

```

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,183 @@
...
+			Path:   "tap",
+			Scheme: "https",
```
frrist: I think the faucet is still on http :disappointed:

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		if _, err := p.MessageWait(ctx, mcid); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("environment does not support GetFunds")
```
frrist: might be helpful to print the type of environment here

tools/fast/tests/retrieval_test.go
```diff
@@ -108,18 +108,87 @@ func TestRetrieval(t *testing.T) {
...
+	err = series.InitAndStart(ctx, miner)
+	require.NoError(t, err)
+
+	// Start Client
+	err = series.InitAndStart(ctx, client)
+	require.NoError(t, err)
+
+	// Everyone needs FIL to deal with gas costs and make sure their wallets
+	// exists (sending FIL to a wallet addr creates it)
+	err = fast.GetFunds(ctx, env, miner)
```
frrist: Ahh I see why we'd want the MemoryGenesis to support this method now, would make the testing a lot cleaner, yeah?

tools/fast/environment_devnet.go
travisperson: Ya, I think we'd need to have a way to attach a node to the environment, which could just be a stand alone function that would store a pointer to a `process`. I'll do that in a follow up issue I think. Once this merges I'll write it up.

tools/fast/environment_devnet.go
travisperson: The `https` version doesn't use the port, where as the http uses port 9797.

https://faucet.user.kittyhawk.wtf/

tools/fast/tests/retrieval_test.go
travisperson: Ya, at least for the functional style tests to make more of the code reusable for both running tests locally or against one of the devnets.

tools/fast/environment_devnet.go
frrist: Pretty sure I made a typo earlier when trying to access this URL, looks good my bad lol

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,182 @@
...
+
+	cid "github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log"
+	iptb "github.com/ipfs/iptb/testbed"
+
+	"github.com/filecoin-project/go-filecoin/address"
+)
+
+// EnvironmentDevnet is a FAST lib environment that is meant to be used
+// when working with kittyhawk devnets.
```
anorth: Since this code is public and fairly generic, it would be better to comment about specifically what features of an infra this relies on.

More generally, this file would benefit from a top-level comment outlining its intended functionality/benefit, its environmental expectations, access needs etc.

tools/fast/environment_devnet.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, err
+	}
+
+	return env, nil
+}
+
+// GenesisCar provides a url where the genesis file can be fetched from
+func (e *EnvironmentDevnet) GenesisCar() string {
+	uri := url.URL{
+		Host:   fmt.Sprintf("genesis.%s.kittyhawk.wtf", e.network),
```
anorth: Can we factor out the specific hostname to a parameter or struct member? Ditto at line 158.

tools/fast/tests/retrieval_test.go
```diff
@@ -108,18 +108,87 @@ func TestRetrieval(t *testing.T) {
...
+}
+
+// TestRetrieval exercises storing and retreiving with the filecoin protocols
+func TestNetworkRetrieval(t *testing.T) {
+	tf.FunctionalTest(t)
+
+	// Skip the test so it doesn't run
+	t.SkipNow()
+
+	blocktime := time.Second * 30
```
anorth: It seems to me like we could reduce this. Even though a functional test is known to be "slow", when running locally there is still great advantage to it being faster.

travisperson: This is a network retrieval, so it should be running the same blocktime as the network. I don't know how we could reduce it. Once we have enforcement in consensus it can't be changed anyways.

tools/fast/environment_devnet.go
travisperson: I don't understand the immediate benefit in doing this. Can you elaborate on what use case this would enable?

anorth: The use case is that someone who is not us could use this code, or we could use it against a network that was not specifically at this domain, e.g. on some local or other infrastructure. I think this is an unnecessary coupling to Protocol Labs' infrastructure.

tools/fast/tests/retrieval_test.go
anorth: What does "network retrieval" mean, and what's a retrieval that doesn't involve a network?

I might be misunderstanding the intent. The name and comment make it sound like this should be fairly generic, not coupled to any particular network configuration. And if not, I suggest we should refactor to separate the generic from the specific so that the same test code can operate in multiple environments. I have just now realised that this bears a very strong similarity to code immediately above it. Is this intended to be essentially the same test but against a network of already-running nodes? The comments don't really differentiate. Can we factor that out (more)?

travisperson: Ya, I forgot to update those comments. I've updated them now.

This test is refactored to be generic as it can be at the moment, and may become more generic as FAST matures. The `RunRetrievalTest` test at the bottom of this file is the actual test which takes in a set of resources. These top functions now `TestRetrievalLocalNetwork` and `TestRetrievalDevnet` do the required setup that is not common between the two.

The `TestRetrievalLocalNetwork` is probably the test you want to be looking at. It uses a 5 second block time, which might be able to be reduced but I haven't played with the number much to determine if there is a lower bound. 5 second seems to work consistently though.

travisperson: More of this could be made generic once we figure out the correct way to deal with the `GetFunds` concept. Comment thread above talks about this a bit.

tools/fast/environment_devnet.go
travisperson: I'd gladly accept PRs when someone setups up their own network and wants to use this code, or if they have something custom they can use the interface and create there own environment. This one is for the kittyhawk devnets.

PR #2455, opened 2019-03-29
updates to templates and README

shannonwells:
This is an effort to direct people first to the discussion forums and the specs for things like: asking how go-filecoin works, and problems with their install, rather than filing bugs/issues. 

I put this info at the top and made the language very short and simple to increase the chance it will register.


travisperson:
@shannonwells could we also put near the top something about:

_If you wish to connect to one of the developer networks (devnets), please refer to the [usage] section below_

I think this is still going to be an issue users run into because we only mention the devnets at the end of the README, after we talk about building  / installing and I think users start down that route first when what they really want is to download the binary release or build it from the tag.

PR #2462, opened 2019-04-01
fix protocol version checking

whyrusleeping:
Checking this based on commit hash is really not a good idea (I told multiple different people we shouldnt do this).

Here is a better-but-still-bad way of doing this.

### DIFF
--- a/protocol/hello/hello.go
+++ b/protocol/hello/hello.go
@@ -24,6 +24,8 @@ func init() {
 // Protocol is the libp2p protocol identifier for the hello protocol.
 const protocol = "/fil/hello/1.0.0"
 
+const FilecoinProtocolVersion = "0.0.2"
+
 var log = logging.Logger("/fil/hello")
 
 // Message is the data structure of a single message in the hello protocol.
@@ -31,7 +33,7 @@ type Message struct {
 	HeaviestTipSetCids   []cid.Cid
 	HeaviestTipSetHeight uint64
 	GenesisHash          cid.Cid
-	CommitSha            string
+	Version              string
 }
 
 type syncCallback func(from peer.ID, cids []cid.Cid, height uint64)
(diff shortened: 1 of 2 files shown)


acruikshank:
@whyrusleeping I am not amongst the people to which you've explained why this is a bad idea. Would you mind fleshing your thinking out a bit here?

The commit SHA based protocol checking is a targeted, temporary solution to avoid confusion to new users connecting usernet. It only affects people connecting with the `devnet-test` or `devnet-user` flag. Since we have not yet completed model for change work, we the go-filecoin team will only support connecting to usernet with the specific binary it's bootstrap nodes are built upon. The commit SHA check will be removed as part of model for change work prior to testnet.

whyrusleeping:
@acruikshank its a bad idea because now we fork the entire network every time someone pushes a commit. This means that as a dev, i can't connect to the usernet to debug why (for example) chain sync is horribly slow and takes an absurd amount of space on disk.

Blocking versions based on connectivity misses a lot of the point of what the point of running this network is.

travisperson:
I think using something other than the commit SHA might be a better way to handle this issue, as the commit SHA is fairly arbitrary anyways. As @acruikshank pointed out, this is a temporary solution to a UX problem we were facing around users try to connect to the user devnet using, generally, the latest master branch that resulted in confusion around why things didn't work, generally because the protocol (data structures) changed. I think the problem in general is due to a lack of proper documentation and it's something we should try to focus on. 

For the issue at hand though, anyone can override the value used for this check by setting the [` FILECOIN_OVERRIDE_BUILD_SHA`](
https://github.com/filecoin-project/go-filecoin/blob/f5786177b7d25e51dcd6d86dd594c99fb75aa534/build/main.go#L238-L240). The value can found by going to the project README and following the devnet badges, which link to release and contains the commit SHA. Anyone wanting to build against the devnets should branch from this commit ref as well. For the user devnet there is a branch which tracks the release as well `release-0.1.0` (probably should be `release-0.1.x`, as the last digit change with any cherry-picked updates, but we don't make a branch, just a new tag)

2019-04-01-124140_1479x256_scrot

---

Something we could also do to help resolve some of the confusion, and remove the need for the explicit version check, is use build flag to enable / disable the different devnet initialization flags. Particularly when we are publishing binaries, as having a `--devnet-test`, and `--devnet-nightly` on the binary we release for the user devnet just leads to user confusion as the moment due to the constant change in the protocol. The same idea can be applied to the other binary releases (nightly). All of this should come with proper documentation as well.

anorth:
A version number like this was considered when version checking was initially implemented, but dismissed. Note that if we had used it, we would have immediately bumped the version in the commit following a release in order to prevent inadvertent connection of development code to the user devnet. We do not want the default behaviour to allow community members to connect divergent code (`master`) to the user devnet, but do require a developer override to be available. I believe that's the current situation.

@rkowalick thanks for the documentation. While getting a SHA into an environment variable is a rather inconvenient way of opting out of the check, the primary point is that it is possible for a developer to do so.

mishmosh:
Although noone loves the status quo, I'm going to close for now so we can keep moving forward. There's a documented workaround and a plan to fix it for real in ~1 month. If real fix doesn't materialize in 0.2.0 let's come back to this.

Issue #2458, opened 2019-03-31
[bug] Can't connect to the network using tag 0.1.4


NatoBoram:
**Describe the bug**

I can't join the Filecoin swarm and the `devnet-user` because my Filecoin node doesn't know what's its version.

**To Reproduce**

1. Checkout `0.1.4`
2. Build `go-filecoin`
3. Initialize `go-filecoin` according to [Getting-Started#start-running-filecoin](https://github.com/filecoin-project/go-filecoin/wiki/Getting-Started#start-running-filecoin)

**Expected behavior**

I should be able to join the network.

**Screenshots**

```log
My peer ID is QmdrPHuzaQjg8RKkZMoU6dRvEpx1QFfoniBJ2TML3jMuMX
Swarm listening on: /ip4/127.0.0.1/tcp/6000
Swarm listening on: /ip4/192.168.1.3/tcp/6000
Swarm listening on: /ip4/192.168.1.139/tcp/6000
00:39:32.966  INFO chain.stor: start loading chain at tipset: { zDPWYqFCu9VdMN9pUSSttbjJsxd2sTJGZjXv8AGZvquxnsMRFh9e }, height: 0 default_store.go:114
00:39:32.971  INFO chain.stor: finished loading 0 tipsets from { zDPWYqFCu9VdMN9pUSSttbjJsxd2sTJGZjXv8AGZvquxnsMRFh9e } default_store.go:156
00:39:43.391 WARNI /fil/hello: bad hello message from peer <peer.ID Qm*vwBpp4>: stream reset hello.go:87
00:39:44.661 ERROR /fil/hello: code not at same version: 966fe75919755074c7c2724f2c9634c2b61d4212 does not match , disconnecting from peer: <peer.ID Qm*vwBpp4> hello.go:97
00:39:53.368 ERROR /fil/hello: code not at same version: 966fe75919755074c7c2724f2c9634c2b61d4212 does not match , disconnecting from peer: <peer.ID Qm*2DX9XY> hello.go:97
00:39:58.242  INFO    autonat: Discovered AutoNAT peer QmXhxqTKzBKHA5FcMuiKZv8YaMPwpbKGXHRVZcFB2DX9XY notify.go:33
00:40:03.239 ERROR /fil/hello: code not at same version: 966fe75919755074c7c2724f2c9634c2b61d4212 does not match , disconnecting from peer: <peer.ID Qm*2DX9XY> hello.go:97
[...]

PR #2465, opened 2019-04-01
Fix install command in build script

rkowalick:
Fix the install command in the build script, as it currently:
- Doesn't work at all
- Won't include the code version during the build process

This may help resolve issues like #2458 in the future.


NatoBoram:
With this, I can confirm `go run ./build build` was fixed. However, `go run ./build install` still doesn't work.

PR #2468, opened 2019-04-01
Feat/fast process config

### LINKED ISSUE #2464: FAST supports filecoin config modification without the daemon running
### Description
The FAST library should provide a canonical means for modifying the config files of the nodes it controls before they have been started.

### Acceptance criteria
I can modify a FAST nodes config before its been started

### Risks + pitfalls
Remote nodes controlled via docker will need special handling or this issue will need to be completed in a way that works for all use cases.

frrist:
This PR allows for FAST process config file changes before starting the daemon. Closes #2464

### DIFF
--- a/tools/fast/bin/localnet/main.go
+++ b/tools/fast/bin/localnet/main.go
@@ -18,6 +18,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime/debug"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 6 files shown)


tools/fast/bin/localnet/main.go
```diff
@@ -161,6 +162,11 @@ func main() {
 	}()
 
 	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("recovered from panic", r)
+			fmt.Println("stacktrace from panic: \n" + string(debug.Stack()))
+			os.Exit(exitcode)
```
travisperson: This line is redundant as is, but I think we want to either set the exitcode, or just explicitly exit with a 1.

Issue #1701, opened 2019-01-25
Enable Filecoin iptb plugins to stream output


travisperson:
IPTB provides a simple implementation of the Output interface, which is used to
return information (mostly for RunCmd) such as stdout, stderr, and an exit code.

It does not support streaming data, as it is backed by byte slice. This is an
issue when dealing with commands which may return large amounts of data, such
as executing `client cat`.

We will need a better implementation of the Output interface that will allow us
to properly deal with a large amount of data coming back from stdout and stderr.

We currently have two plugins that need a new implementation of the Output
interface.

Here is updated documentation for the Output interface, needs a bit of polish,
but should provide a basic idea of the expectations for plugins and users.

_Note: This isn't really how iptb cli is currently implemented as it prints
the exit code and the error prior to even starting to read either stdout, or
stderr. The issue here is that we can't know the exit code, or if there is an
error until we have read all of the stdout and stderr which means buffering
or consuming it all so that the underlying command can properly exit to give
us the exit code, and that all output has been read and we know that there were
not errors._

```go
type Output interface {
  // Args returns the exact arguments used to generate the output.
  Args() []string

  // ExitCode returns an integer representing the exit status. This is
  // generally a zero (0) for success, or one (1) for an error, that is
  // the process exited with some none success. A value of -1 means that
  // no exit code was able to be retrieved. The user should check Error().
  // Note: ExitCode will block until an exit code from running the arguments
  // is available, this might require that all data is read from Stdout and
  // Stderr.
  ExitCode() int

  // Error returns any errors that occurred as part of the execution. This can
  // include io errors, timeouts, etc. Errors that occur prior to execution
  // should be returned by the error value along side a nil output.
  // See the Core Interface methods Init, Start, RunCmd.
  // Note: Error may block until both Stdout and Stderr are closed due to an
  // error, or EOF
  Error() error

  // It's the responsibility of the caller to the method that returns an Output
  // to call the Close method on both the returned values of Stderr() and
  // Stdout().
  // i.e) If you call RunCmd, and it returns a non nil Output, you should also
  // call Stdout().Close() and Stderr().Close()

  // Stderr is the stderr output from execution of Args.
  Stderr() io.ReadCloser

  // Stdout is the stdout output from execution of Args.
  Stdout() io.ReadCloser
}
```

### Local Filecoin Plugin

The local filecoin plugin implements the RunCmd by using `exec.CommandContext`.
It provides `Wait` which "waits for the command to exit and waits for any
[...]

PR #2469, opened 2019-04-01
localplugin: Read output in go routines

travisperson:
If the data returned from stdout exceeds that of the kernal buffer, reading from stderr will block forever because the stdout buffer will not drain and the command will deadlock waiting on stderr to close. Moving the reading into a go routine allows for the data to be buffered in user-space. This is a temporary fix to enable working with large data until #1701 can be resolved.

### DIFF
--- a/tools/iptb-plugins/filecoin/local/localfilecoin.go
+++ b/tools/iptb-plugins/filecoin/local/localfilecoin.go
@@ -22,6 +22,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin"
 	"github.com/ipfs/iptb/testbed/interfaces"
 	"github.com/ipfs/iptb/util"
+	"golang.org/x/sync/errgroup"
 )
 
 // PluginName is the name of the plugin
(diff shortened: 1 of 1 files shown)


tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -327,13 +328,32 @@ func (l *Localfilecoin) RunCmd(ctx context.Context, stdin io.Reader, args ...str
...
 
-	stdoutbytes, err := ioutil.ReadAll(stdout)
-	if err != nil {
+	var stderrbytes []byte
+	var stdoutbytes []byte
+
+	g.Go(func() error {
+		var err error
+		stderrbytes, err = ioutil.ReadAll(stderr)
+		if err != nil {
```
frrist: we can shorten this to just `return err` I think, but i'm happy with it either way.

travisperson: Ya, it will be remove hopefully near the end of the week / next week so I'll probably just leave it so I can merge this PR.

PR #2478, opened 2019-04-02
circle: build out of GOPATH

Kubuxu:
It enables go modules by defaul


Kubuxu:
Hmm, this doesn't work because tests require in GOPATH go-filecoin repo.

Issue #2485, opened 2019-04-03
[bug]go-filecoin panic and exit after runing about 24 hours


yaohcn:
**Describe the bug**
go-filecoin panic and exit after runing about 24 hours

**Screenshots**

> fatal error: concurrent map iteration and map write
goroutine 5988 [running]:
runtime.throw(0x1573205, 0x26)
    /usr/local/go/src/runtime/panic.go:617 +0x72 fp=0xc089c87c68 sp=0xc089c87c38 pc=0x4dff82
runtime.mapiternext(0xc089c87d48)
    /usr/local/go/src/runtime/map.go:860 +0x597 fp=0xc089c87cf0 sp=0xc089c87c68 pc=0x4c1777
github.com/filecoin-project/go-filecoin/core.(*MessagePool).timeoutMessages(0xc0000c39b0, 0x17c5260, 0xc032bc00c0,                0x7f3acef8d020, 0xc0000a8540, 0xc0a3da7770, 0x40, 0x80)
    /go/src/github.com/filecoin-project/go-filecoin/core/message_pool.go:185 +0x203 fp=0xc089c87db8 sp=0xc089c87cf0 pc=0xbb8183
github.com/filecoin-project/go-filecoin/core.(*MessagePool).UpdateMessagePool(0xc0000c39b0, 0x17c5260, 0xc032bc00c0,              0x7f3acef8d020, 0xc0000a8540, 0xc071d1b920, 0xc0a3da7770, 0x0, 0x0)
    /go/src/github.com/filecoin-project/go-filecoin/core/message_pool.go:155 +0x465 fp=0xc089c87ec8 sp=0xc089c87db8 pc=0xbb7f25
github.com/filecoin-project/go-filecoin/node.(*Node).handleNewHeaviestTipSet(0xc00000c1e0, 0x17c5260, 0xc032bc00c0, 0xc071d1b920, 0xc025015590)
    /go/src/github.com/filecoin-project/go-filecoin/node/node.go:660 +0x346 fp=0xc089c87fb8 sp=0xc089c87ec8 pc=0xcc8286
runtime.goexit()
    /usr/local/go/src/runtime/asm_amd64.s:1337 +0x1 fp=0xc089c87fc0 sp=0xc089c87fb8 pc=0x50f521
created by github.com/filecoin-project/go-filecoin/node.(*Node).Start
    /go/src/github.com/filecoin-project/go-filecoin/node/node.go:543 +0x7f1

[...]

PR #2486, opened 2019-04-03
fix bug #2485

yaohcn:
fix bug #2485

### DIFF
--- a/core/message_pool.go
+++ b/core/message_pool.go
@@ -181,10 +181,13 @@ func (pool *MessagePool) timeoutMessages(ctx context.Context, store chain.BlockP
 		}
 	}
 
+	pool.lk.Lock()
+	defer pool.lk.Unlock()
+
 	// remove all messages added before minimumHeight
 	for cid, msg := range pool.pending {
 		if msg.addedAt < minimumHeight {
-			pool.Remove(cid)
+			delete(pool.pending, cid)
 		}
 	}
 


mishmosh:
Thanks @yaohcn. Assigning to @ZenGround0 for first review.

core/message_pool.go
```diff
@@ -181,10 +181,13 @@ func (pool *MessagePool) timeoutMessages(ctx context.Context, store chain.BlockP
...
 	}
 
+	pool.lk.Lock()
+	defer pool.lk.Unlock()
+
 	// remove all messages added before minimumHeight
 	for cid, msg := range pool.pending {
 		if msg.addedAt < minimumHeight {
-			pool.Remove(cid)
+			delete(pool.pending, cid)
```
anorth: This misses the metric update in `Remove()`.

It is unfortunate that `timeoutMessages` uses the pool's private API, which makes errors like this all too easy (not your fault). I think a better fix here is to use the exported `Pending()` at line 188 then stick with `Remove` here (and no need for the extra lock).

anorth:
I think this may be accidentally(?) fixed by #2522 too

core/message_pool.go
yaohcn: Pending() return only field: message *types.SignedMessage. but here need the field: addedAt uint64.

yaohcn:
> I think this may be accidentally(?) fixed by #2522 too

yep, closed.

PR #2489, opened 2019-04-03
Return descriptive error message upon self dial

ZenGround0:
- Wrap the libp2p pingService with a check for self dial
- Abstract pingMiner function into porcelain
- Handle self dial errors as part of pinging storage / retrieval miners

Want to check if this is roughly right before adding in testing cc @anorth

### DIFF
--- a/commands/client_daemon_test.go
+++ b/commands/client_daemon_test.go
@@ -2,16 +2,23 @@ package commands_test
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"math/big"
 	"strings"
 	"testing"
 
+	"github.com/ipfs/go-ipfs-files"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/fixtures"
 	"github.com/filecoin-project/go-filecoin/protocol/storage"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 )
 
 func TestListAsks(t *testing.T) {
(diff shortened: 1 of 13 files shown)


porcelain/network.go
```diff
@@ -0,0 +1,41 @@
...
+
+// PingMinerWithTimeout pings a storage or retrieval miner, waiting the given
+// timeout and returning descriptive errors.
+func PingMinerWithTimeout(ctx context.Context, minerPID peer.ID, timeout time.Duration, plumbing netPlumbing) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	res, err := netPlumbing.NetworkPing(plumbing, ctx, minerPID)
+	if err == net.ErrPingSelf {
+		return fmt.Errorf("attempting to make deal with self.  This is currently unsupported.  Please use a separate go-filecoin node as client")
```
anorth: This level is too low for the message to be about dealing with self. Wrap it one level higher.

anorth:
Yes, this looks great. 💯

commands/client_daemon_test.go
```diff
@@ -196,3 +204,39 @@ func TestVoucherPersistenceAndPayments(t *testing.T) {
...
+	require.NoError(env.GenesisMiner.MiningStart(ctx))
+
+	// Start mining.
+	miningNode := env.RequireNewNodeWithFunds(1000)
+	pledge := uint64(10)
+	collateral := big.NewInt(int64(1))
+	price := big.NewFloat(float64(0.001))
+	expiry := big.NewInt(int64(500))
+	fmt.Printf("creatin miner\n")
+	ask, err := series.CreateMinerWithAsk(ctx, miningNode, pledge, collateral, price, expiry)
```
travisperson: `CreateMinerWithAsk` is a bit outdated. The idea is that any series should be written such that it can be used in both a synchronous (`mining once`) and asynchronous (`mining start`). Currently `CreateMinerWithAsk` starts mining, and it shouldn't do this. Anywhere that a block is required to be mining should have a call to `CtxMiningOnce`. This might be a little more difficult at the moment as `mining start` is what registers all the protocols, so it may not even be possible at the moment. There was talk of have a `mining enable` or similar which would setup everything up, but not schedule the miner itself.

ZenGround0:
@acruikshank switched maintainer review to you because this is in your territory.

porcelain/network.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/pkg/errors"
+)
+
+type netPlumbing interface {
+	NetworkPing(ctx context.Context, pid peer.ID) (<-chan time.Duration, error)
+}
+
+// PingMinerWithTimeout pings a storage or retrieval miner, waiting the given
+// timeout and returning descriptive errors.
+func PingMinerWithTimeout(ctx context.Context, minerPID peer.ID, timeout time.Duration, plumbing netPlumbing) error {
```
rosalinekarr: Let's add some tests for this method.

anorth:
@ZenGround0 what's the status here? Can we land something or dismiss it?

ZenGround0:
@anorth tests now in!  Just waiting on maintainer approval strictly speaking, though I'd like @travisperson to take a look at the FAST testing and slight changes to the retrieval action.

commands/retrieval_client_daemon_test.go
```diff
@@ -0,0 +1,65 @@
...
+	}()
+
+	// Update genesis miner's peerid
+	var minerAddr address.Address
+	err := env.GenesisMiner.ConfigGet(ctx, "mining.minerAddress", &minerAddr)
+	require.NoError(err)
+	details, err := env.GenesisMiner.ID(ctx)
+	require.NoError(err)
+	msgCid, err := env.GenesisMiner.MinerUpdatePeerid(ctx, minerAddr, details.ID, fast.AOPrice(big.NewFloat(1.0)), fast.AOLimit(300))
+	require.NoError(err)
```
travisperson: All of this should already have happened. The environment setups the `GenesisMiner` by calling  `SetupGenesisNode` which calls `MinerUpdatePeerid`. Where you running to some issue or just added this out of habit?

https://github.com/filecoin-project/go-filecoin/blob/master/tools/fast/fastesting/basic.go#L69
https://github.com/filecoin-project/go-filecoin/blob/master/tools/fast/series/setup_genesis_node.go#L39

commands/retrieval_client_daemon_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Add data to Genesis Miner.
+	f := files.NewBytesFile([]byte("satyamevajayate"))
+	cid, err := env.GenesisMiner.ClientImport(ctx, f)
+	require.NoError(err)
+
+	// Genesis Miner fails on self dial when retrieving from itself.
+	_, err = env.GenesisMiner.RetrievalClientRetrievePiece(ctx, cid, minerAddr)
+	env.GenesisMiner.DumpLastOutput(os.Stdout)
```
travisperson: Left over from debugging I'd assume

commands/retrieval_client_daemon_test.go
ZenGround0: I ran into an issue.  The miner actor and genesis pid differed.

travisperson: I'll toy with this a bit and see if I can find the issue. We shouldn't have to do this.

commands/client_daemon_test.go
```diff
@@ -230,3 +237,57 @@ func TestPieceRejectionInProposeStorageDeal(t *testing.T) {
...
+	}()
+
+	// Start mining.
+	miningNode := env.RequireNewNodeWithFunds(1000)
+
+	// This is what mining start should do, but FAST uses mining once
+	// for some very helpful builtins and because of issue 2579 we need to
+	// mine once in a loop instead of calling start.  Once #2579 is fixed
+	// this can be replaced with start.
+	minerCreateDoneCh := make(chan struct{})
```
rosalinekarr: Non-blocking, but is there a reason you don't want to use a context here?

commands/client_daemon_test.go
```diff
@@ -230,3 +237,57 @@ func TestPieceRejectionInProposeStorageDeal(t *testing.T) {
...
+	pledge := uint64(10)
+	collateral := big.NewInt(int64(1))
+	price := big.NewFloat(float64(0.001))
+	expiry := big.NewInt(int64(500))
+	ask, err := series.CreateMinerWithAsk(ctx, miningNode, pledge, collateral, price, expiry)
+	minerCreateDoneCh <- struct{}{}
+	require.NoError(err)
+
+	// Try to make a storage deal with self and fail on self dial.
+	f := files.NewBytesFile([]byte("satyamevajayate"))
```
rosalinekarr: obscure movie reference. i like it

commands/client_daemon_test.go
ZenGround0: No I would also be happy to use a new context and cancel function here.

PR #2492, opened 2019-04-03
Configure proofs mode (test or live) in genesis block

rkowalick:
We store the sector size in the initial state of the storage market actor, thereby putting this information into the genesis block. This will allow any filecoin node to retrieve the accurate sector size without needing to create a miner actor.


node/node.go
```diff
@@ -585,10 +585,7 @@ func (node *Node) setupHeartbeatServices(ctx context.Context) error {
 
 func (node *Node) setupMining(ctx context.Context) error {
 	// configure the underlying sector store, defaulting to the non-test version
-	sectorStoreType := proofs.Live
-	if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
-		sectorStoreType = proofs.Test
-	}
+	sectorStoreType := proofs.GetSectorStoreType()
```
laser: BLOCKING: This should come from the blockchain - specifically from the `StorageMarketActor`. Agree? Disagree?

proofs/interface.go
```diff
@@ -46,3 +48,11 @@ const (
 	// Test configures the SectorBuilder to be used with large sectors, in tests.
 	Test
 )
+
+func GetSectorStoreType() SectorStoreType {
```
laser: BLOCKING: I don't think this function belongs here. I think that it should be part of the `gengen` package and should not be exported. I think folks need to be creating a query message to get this value from the `StorageMarketActor`, via a query message, if they need it at runtime.

proofs/sectorbuilder/interface.go
```diff
@@ -39,10 +39,6 @@ type SectorBuilder interface {
 	// regardless of the number of times SectorSealResults is called.
 	SectorSealResults() <-chan SectorSealResult
 
-	// GetMaxUserBytesPerStagedSector produces the number of user piece-bytes
-	// which will fit into a newly-provisioned staged sector.
-	GetMaxUserBytesPerStagedSector() (uint64, error)
```
laser: Nice!

travisperson:
What is the actual impact of this change? Eg, do we need to remove all reference of `FIL_USE_SMALL_SECTOR` from the wiki, deployment scripts, etc as it will need to be configured at `genesis.car` generation?

actor/builtin/miner/miner.go
```diff
@@ -58,6 +57,8 @@ const (
 	ErrAskNotFound = 40
 	// ErrInvalidSealProof signals that the passed in seal proof was invalid.
 	ErrInvalidSealProof = 41
+	// ErrGetProofsModeFailed indicates the call to get the sector store type failed.
```
laser: suggestion: "ErrGetProofsModeFailed indicates the call to get the proofs mode failed."

actor/builtin/miner/miner.go
```diff
@@ -460,12 +462,12 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		// the PoRep verification operation needs to know some things (e.g. size)
 		// about the sector for which the proof was generated in order to verify.
 		//
-		// It is undefined behavior for a miner in "Live" mode to verify a proof
+		// It is undefined behavior for a miner in "LiveMode" mode to verify a proof
 		// created by a miner in "ProofsTest" mode (and vice-versa).
```
laser: typo: "created by a miner in "TestMode"

gengen/main.go
```diff
@@ -69,6 +69,7 @@ func main() {
 	var defaultSeed = time.Now().Unix()
 
 	jsonout := flag.Bool("json", false, "sets output to be json")
+	useSmallSectors := flag.Bool("use-small-sectors", false, "use small sectors instead of full size ones")
```
laser: SUGGESTION: In a very short time, `use-small-sectors` will need to change to `use-test-sectors`. There will be several sector sizes, so "small" won't be particularly useful in disambiguating.

consensus/genesis.go
```diff
@@ -142,7 +143,7 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
 				return nil, err
 			}
 		}
-		if err := SetupDefaultActors(ctx, st, storageMap); err != nil {
+		if err := SetupDefaultActors(ctx, st, storageMap, proofs.TestMode); err != nil {
```
laser: BLOCKING: Are you sure that you want to default this to `proofs.TestMode`? It seems like this should be configured via a `GenOption`.

abi/abi.go
```diff
@@ -246,6 +252,13 @@ func (av *Value) Serialize() ([]byte, error) {
 		}
 
 		return []byte{b}, nil
+	case ProofsMode:
+		v, ok := av.Val.(proofs.Mode)
+		if !ok {
+			return nil, &typeError{proofs.TestMode, av.Val}
```
laser: BLOCKING: You've got an `Unknown` enum value in your `proofs.TestMode`, but you're defaulting to `TestMode` here. You should use `Unknown` instead.

rkowalick:
@travisperson 

These changes will affect all of the things you mentioned (maybe even more things?). Once this PR is a more "done" state, I can and will enumerate the changes needed.

consensus/genesis.go
rkowalick: This function is only run in test scenarios, so it makes sense to default to `TestMode`

abi/abi.go
rkowalick: Resolved by `TestMode = 0`

build/main.go
```diff
@@ -165,12 +165,17 @@ func forceBuildFC() {
 
 func generateGenesis() {
 	log.Println("Generating genesis...")
+	useSmallSectorsOpt := ""
+	if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
+		useSmallSectorsOpt = "--use-small-sectors"
```
laser: BLOCKING: Please do update this option.

travisperson:
These are all the places I think we need to make updates, some of these may already be covered in this PR.

#### Faucet

https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/genesis-file-server/main.go#L13-L17

#### Test Daemon

https://github.com/filecoin-project/go-filecoin/blob/a45c0f1a70c2e8a61d6d5b4e01e641bedb344a06/testhelpers/commands.go#L849-L852

#### Updating FAST memory genesis environment

New field should be added to the structure (would be great to use the 
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/environment_memory_genesis.go#L28-L42

Defined by an argument to
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/environment_memory_genesis.go#L47

Usage of `NewEnvironmentMemoryGenesis`
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/tests/retrieval_test.go#L47
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/bin/localnet/main.go#L187

Remove previous configuration
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/tests/retrieval_test.go#L61
https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/bin/localnet/main.go#L200

Remove reference to `FIL_USE_SMALL_SECTORS` from IPTB (basically remove changed added by #1644 around the env var)

#### Updates to wiki / readme 

- https://github.com/filecoin-project/go-filecoin/wiki/Devnets#configuration
- https://github.com/filecoin-project/go-filecoin/wiki/Commands#helpful-environmental-variables

Should also add updates to the following to reference how to build each type of genesis
- https://github.com/filecoin-project/go-filecoin/wiki/Running-a-network-locally
- https://github.com/filecoin-project/go-filecoin/tree/master/gengen

travisperson:
We'll also want to coordinate around a PR for the infra side.

See https://github.com/filecoin-project/go-filecoin-infra/issues/225

tools/fast/bin/localnet/main.go
```diff
@@ -184,7 +174,7 @@ func main() {
 		return
 	}
 
-	env, err := fast.NewEnvironmentMemoryGenesis(&balance, workdir)
+	env, err := fast.NewEnvironmentMemoryGenesis(&balance, workdir, proofs.TestMode)
```
travisperson: This need to follow the configuration flag `smallSectors`

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -40,8 +41,8 @@ var DefaultFilecoinBinary = "go-filecoin"
 // DefaultLogLevel is the value that will be used for GO_FILECOIN_LOG_LEVEL
 var DefaultLogLevel = "3"
 
-// DefaultUseSmallSectors is the value that will be used for FIL_USE_SMALL_SECTORS
-var DefaultUseSmallSectors = "false"
+// DefaultTestProofsMode defaults to the live proof configuration
+var DefaultTestProofsMode = "false"
```
travisperson: I don't believe this is needed

tools/fast/tests/retrieval_test.go
```diff
@@ -58,7 +58,6 @@ func TestRetrieval(t *testing.T) {
 	options := make(map[string]string)
 	options[localplugin.AttrLogJSON] = "0"                               // Disable JSON logs
 	options[localplugin.AttrLogLevel] = "4"                              // Set log level to Info
-	options[localplugin.AttrUseSmallSectors] = "true"                    // Enable small sectors
```
travisperson: I don't see a change to the `EnvironmentMemoryGenesis`, I assume this doesn't build at the moment.

tools/fast/bin/localnet/main.go
```diff
@@ -394,37 +383,16 @@ func isEmpty(name string) (bool, error) {
 	return false, err // Either not empty or error, suits both cases
 }
 
-func getSectorSize(smallSectors bool) (uint64, error) {
-	rp := repo.NewInMemoryRepo()
-	blockstore := bstore.NewBlockstore(rp.Datastore())
-	blockservice := bserv.New(blockstore, offline.Exchange(blockstore))
-
+func getProofsMode(smallSectors bool) uint64 {
```
travisperson: Can we change this back to `getSectorSize`? It's more descriptive of the method.

travisperson: Or you can keep the name, and change the return type, and call `proofs.SectorSize` after the call to `getProofsMode`, you can then use the value to pass into `NewEnvironmentMemoryGenesis`.

CODEWALK.md
```diff
@@ -455,12 +455,11 @@ The `install-rust-proofs.sh` script fetches or generates these Groth parameters
...
-### Sector size configuration
-Sealing a sector is a compute-intensive process that grows with the size of the sector being sealed (replication grows linearly, and proof-generation logarithmically).
-For ease of development, go-filecoin can be configured to use tiny sectors, holding only 1016 bytes of user data, which are faster to seal. 
-This mode is triggered by the `FIL_USE_SMALL_SECTORS` environment variable. 
-Sector size is a parameter on which nodes in a network must agree, and determines the Groth parameters; 
-nodes with different sector sizes cannot verify each other’s proofs.
+### Proof mode configuration
+Sealing a sector and generating PoSts is a compute-intensive process that grows with the size of the sector being sealed (replication grows linearly, and proof-generation logarithmically).
+For ease of development, go-filecoin can be configured to use a test proofs mode, which reduces sectors to only 1016 bytes of user data to lower computation burden of sealing and PoSt generation.
+This mode is triggered by using the `genesis.car` file from inside the `fixtures/test/` directory.
```
acruikshank: I'd like this to be a little more clear. 
"The sector size is configured in the genesis block. The `genesis.car` in `fixtures/test/` is configured to use the smaller 1016 byte sectors."  ?

actor/builtin/miner/miner.go
```diff
@@ -460,12 +462,12 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
 		//
-		proofsMode := proofs.LiveMode
-		if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
-			proofsMode = proofs.TestMode
+		proofsMode, err := getProofsMode(ctx)
+		if err != nil {
+			return ErrGetProofsModeFailed, Errors[ErrGetProofsModeFailed]
 		}
 
 		req := proofs.VerifySealRequest{}
```
acruikshank: Non-blocking. This code to `VerifySeal` really shouldn't live here. I've created an issue for it. #2555

CODEWALK.md
laser: FWIW: The proofs mode is configured in the genesis block. Today, proofs mode has a one to one relationship with sector size. In a couple of days, proofs mode will simply constrain the miner's choice of sector sizes to commit.

So, my recommendation (to future proof this file) would be to say:

"For ease of development, go-filecoin can be configured to use a test proofs mode, which will cause storage miners to use sectors into which only 1016 bytes of user data can be written. This lowers the computational burden of sealing and generating PoSts.

The genesis.car in `fixtures/test/` is configured to use proofs mode."

tools/fast/tests/retrieval_test.go
rkowalick: I have added the appropriate argument to `NewEnvironmentMemoryGenesis`.

node/node.go
```diff
@@ -585,10 +586,19 @@ func (node *Node) setupHeartbeatServices(ctx context.Context) error {
 }
 
 func (node *Node) setupMining(ctx context.Context) error {
-	// configure the underlying sector store, defaulting to the non-test version
-	proofsMode := proofs.LiveMode
-	if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
-		proofsMode = proofs.TestMode
+	var proofsMode proofs.Mode
+	values, _, err := node.PorcelainAPI.MessageQuery(
```

proofs/sectorsize.go
```diff
@@ -0,0 +1,19 @@
...
+
+// #cgo LDFLAGS: -L${SRCDIR}/../proofs/lib -lfilecoin_proofs
+// #cgo pkg-config: ${SRCDIR}/../proofs/lib/pkgconfig/libfilecoin_proofs.pc
+// #include "../proofs/include/libfilecoin_proofs.h"
+import "C"
+import (
+	"unsafe"
+)
+
+func SectorSize(ssType Mode) uint64 {
```
acruikshank: Is this only used in test code? Is there future utility in it for production code? If not, it would be nice if this could be moved into a test package. I'm not sure how hard that would be.

CC @laser

tools/fast/bin/localnet/main.go
```diff
@@ -394,37 +379,11 @@ func isEmpty(name string) (bool, error) {
...
-		StagedSectorDir:  "",
-	}
-
-	sb, err := sectorbuilder.NewRustSectorBuilder(cfg)
-	if err != nil {
-		return 0, err
-	}
-
-	return sb.GetMaxUserBytesPerStagedSector()
+	return proofs.LiveMode
```
acruikshank: The simplification in this file is impressive!

rkowalick: Building a SectorBuilder is pretty tough in most parts of the application

proofs/sectorsize.go
rkowalick: This is definitely a useful production artifact of this PR that will put into use to resolve #2034

laser: > Is this only used in test code?

No, this will be used by storage clients to validate that the piece they'd like to make a deal for will fit into the sector size offered by the miner which owns the ask.

It will also be used by the storage miner to validate that the bytes that they were given by a storage client will fit into the sector that they've offered.

It will also be used in test code to create pieces which pack in a certain way (i.e. trigger sealing by filling up a freshly-provisioned yet-to-be-sealed sector).

actor/builtin/miner/miner.go
```diff
@@ -460,12 +462,12 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
 		//
-		// It is undefined behavior for a miner using "LiveMode" to verify a
-		// proof created by a miner using "TestMode" (and vice-versa).
+		// It is undefined behavior for a miner using "LiveMode" to verify a proof
+		// created by a miner in "TestMode"(and vice-versa).
 		//
-		proofsMode := proofs.LiveMode
-		if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
-			proofsMode = proofs.TestMode
+		proofsMode, err := getProofsMode(ctx)
```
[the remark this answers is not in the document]
rkowalick: done

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/proofs"
```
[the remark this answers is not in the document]
rkowalick: Discussed offline. Will be changed soon.

PR #2501, opened 2019-04-05
Support decimal FIL amount

### LINKED ISSUE #2265: Support non-integer values for message send
### Description

Currently, the `--value` option for `message send` only accepts integers. People will likely want to transfer non-integer amounts of FIL, so we should changes this to accept a float-as-string amount of filecoin.

### Acceptance criteria

The ability to send 0.5 filecoin to another address

### Risks + pitfalls

none known

### Where to begin

The `cmdkit.IntOption("value"...)` inside the message send command.

mslipper:
Closes #2265.

### DIFF
--- a/commands/message.go
+++ b/commands/message.go
@@ -60,7 +60,11 @@ var msgSendCmd = &cmds.Command{
 			return err
 		}
 
-		val, ok := types.NewAttoFILFromFILString(req.Options["value"].(string))
+		rawVal := req.Options["value"]
+		if rawVal == nil {
+			return errors.New("value cannot be nil")
+		}
+		val, ok := types.NewAttoFILFromFILString(rawVal.(string))
 		if !ok {
 			return errors.New("mal-formed value")
 		}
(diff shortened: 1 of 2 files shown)


commands/message.go
```diff
@@ -60,9 +60,9 @@ var msgSendCmd = &cmds.Command{
 			return err
 		}
 
-		val, ok := req.Options["value"].(int)
+		val, ok := types.NewAttoFILFromFILString(req.Options["value"].(string))
 		if !ok {
-			val = 0
+			val = types.NewZeroAttoFIL()
```
rkowalick: I think it makes sense to actually return an error in this case. Decimal FIL amounts can be malformed in a bunch of ways, and having the user send a message whose value is `0` seems bad.

mslipper:
Sorry - my editor keeps inserting broken imports.

rkowalick:
I think it makes sense to set the value to `0` if it is `nil`. There are a lot of messages in filecoin that don't require a transfer of value.

mslipper:
@rkowalick yep, this is also why the tests are failing. I'm pushing a fix for this now.

PR #2512, opened 2019-04-08
Update bug report template to include version info request

travisperson:
When bugs deal with building a lot of the time the build errors are due to incorrect versions of dependencies, and requests for this information is commonly the first comment. To help reduce the time to resolving the issue I think we should request this information up-front in the issue template.

Another common issue is around connecting to the network, and most issues do not properly fill out the steps to reproduce. I've added explicit request for the `init` and `daemon` commands in the steps.

Are there any additional information we would like to add to the bug report form to help reduce the time to resolve issues?

/cc @anorth @laser @mishmosh

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,11 +30,21 @@ Steps to reproduce the behavior:
 <!-- If applicable, add logging output or screenshots to help explain your problem. -->
 
 **Version information**
+<!-- please fill out this information below completely, it will help us solve your issue faster -->
 <!-- if you are trying to build go-filecoin -->
 - Go: <!-- go version -->
 - Rust: <!-- rustc --version -->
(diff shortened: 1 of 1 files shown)


mishmosh:
LGTM. Related: what do you think about a separate "Help" issue template? There will always be some blur, but it might make triage somewhat easier.

travisperson:
I think that would be a good idea. It would be nice to differentiate between bugs in go-filecoin and issues around building, etc.

anorth:
An idea I had was to add a "help" issue template which contained
instructions to ask for help in the forum instead. At the moment I think
we're in an awkward spot where some requests turn out to be legitimate
bugs, so it's good they were filed here, but many are not.

mishmosh:
Following the principle of "enable progress", @travisperson should go ahead and merge since it has 2 approvals. I'll follow up with a separate PR for the "help" bug template.

PR #2514, opened 2019-04-08
FAST: replace global sleep value with sleep from context

travisperson:
When first implementing FAST we had a global variable for use in series which would define a value that could be used in series when waiting around for things that could not be checked on chain, such as the status of a deal. This is problematical when it come to running test in parallel.

This PR removes this pattern and instead uses the context passed into series to define this sleep duration.


tools/fast/series/sleep_delay_from_ctx.go
```diff
@@ -0,0 +1,34 @@
...
+	defaultSleepDelay = mining.DefaultBlockTime
+)
+
+// SetSleepDelay returns a context with `d` set in the context. To sleep with
+// the value, call SleepDelayFromCtx with the context.
+func SetSleepDelay(ctx context.Context, d time.Duration) context.Context {
+	return context.WithValue(ctx, sleepDelayKey, d)
+}
+
+// SleepDelay is a helper method to make sure people don't call `time.Sleep`
```
frrist: Should be "SleepDelayFromCtx" to match function name (a bit concerned this passed limiting).

The naming here leads me to believe `SleepDelayFromCtx` will return a `duration` from the context, could we either:
1. Rename this to indicate calling the method will perform the sleeping: `SleepForCtxDelay` or similar
2. Return the duration from the context and leave the method naming as it. A very similar pattern is [used here](https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L5-L27). My preference is this option with the method explicitly wrapped in a sleep when called.

travisperson: What about `CtxSleep`?

frrist: Either sound good to me (rename or wrap in sleep call), I will leave this decision up to you.

PR #2515, opened 2019-04-08
FAST: support synchronous mining in FAST series

travisperson:
FAST is design to be a multi-purpose library for controlling Filecoin daemon for use in automation and testing. When writing some test we need explicit control over when mining occurs. To enable the use of series across synchronous and asynchronous mining (`mining start` vs `mining once`) we need to be able to call `mining once` inside of series when a block is required to be mined.

This PR supports this by providing a way for consumers to set a function on a context which when called will mine a block. Series can use the `MiningOnceFromCtx` to ensure that a block is mined when required.

### DIFF
--- a/tools/fast/bin/localnet/main.go
+++ b/tools/fast/bin/localnet/main.go
@@ -228,8 +228,7 @@ func main() {
 		return
 	}
 
-	err = genesis.MiningStart(ctx)
-	if err != nil {
+	if err := genesis.MiningStart(ctx); err != nil {
 		exitcode = handleError(err, "failed to start mining on genesis node;")
 		return
 	}
--- a/tools/fast/bin/localnet/main.go
+++ b/tools/fast/bin/localnet/main.go
@@ -228,6 +228,12 @@ func main() {
 		return
 	}
 
+	err = genesis.MiningStart(ctx)
+	if err != nil {
+		exitcode = handleError(err, "failed to start mining on genesis node;")
+		return
+	}
+
 	// Create the processes that we will use to become miners
 	var miners []*fast.Filecoin
 	for i := 0; i < minerCount; i++ {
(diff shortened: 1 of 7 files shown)


tools/fast/bin/localnet/main.go
```diff
@@ -228,6 +228,12 @@ func main() {
 		return
 	}
 
+	err = genesis.MiningStart(ctx)
```
frrist: nit:
```suggestion
if err := genesis.MiningStart(ctx); err != nil {
		exitcode = handleError(err, "failed to start mining on genesis node;")
		return
}

File: tools/fast/tests/retrieval_test.go [hunk 2]
```diff
@@ -87,6 +87,9 @@ func TestRetrieval(t *testing.T) {
 	err = series.SetupGenesisNode(ctx, genesis, genesisMiner.Address, files.NewReaderFile(genesisMiner.Owner))
 	require.NoError(err)
 
+	err = genesis.MiningStart(ctx)
```
frrist: nit: 
```suggestion
require.NoError(genesis.MiningStart(ctx))
```

tools/fast/series/ctx_mining_once.go
```diff
@@ -0,0 +1,28 @@
+package series
+
+import (
+	"context"
+)
+
+type ctxMiningOnceKey struct{}
+
+// Key used to store the MiningOnceFunc in the context
+var miningOnceKey = ctxMiningOnceKey{}
```
anorth: It looks like this pattern of putting something in context is repeated a bit. Consider putting a single FAST interface object in the context with methods corresponding to the various contextual bits. Usage would then look like `ctxSeries(ctx).MineOnce()`

travisperson: Ya, I think when we add the price / limit we may move to this pattern.

Issue #2500, opened 2019-04-04
[bug] Flood of spam messages crashes nodes without limits on mpool size and content


gmasgras:
**Describe the bug**
All of the filecoin nodes on the user devnet are being flooded by very long messages, causing a restart loop. 
The messages looks like:
```
[...]

PR #2519, opened 2019-04-08
Fix/mpool queue limit

frrist:
### Description
This PR:
1. Adds a Message Pool config that contains a limit field, this is used to control the max size of the message pool.
2. Passes the Message Pool config to the message pool when it is created ensure the message pool cannot grow indefinitely.

### Notes:
I am unsure if the is the exact solution we want to go with, however since https://github.com/filecoin-project/go-filecoin/issues/2500 is a `P0` I wanted to get something up now that makes progress towards addressing this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,15 +17,16 @@ import (
 
 // Config is an in memory representation of the filecoin configuration file
 type Config struct {
-	API       *APIConfig       `json:"api"`
-	Bootstrap *BootstrapConfig `json:"bootstrap"`
-	Datastore *DatastoreConfig `json:"datastore"`
-	Swarm     *SwarmConfig     `json:"swarm"`
-	Mining    *MiningConfig    `json:"mining"`
-	Wallet    *WalletConfig    `json:"wallet"`
-	Heartbeat *HeartbeatConfig `json:"heartbeat"`
-	Net       string           `json:"net"`
-	Metrics   *MetricsConfig   `json:"metrics"`
+	API       *APIConfig         `json:"api"`
+	Bootstrap *BootstrapConfig   `json:"bootstrap"`
+	Datastore *DatastoreConfig   `json:"datastore"`
+	Swarm     *SwarmConfig       `json:"swarm"`
+	Mining    *MiningConfig      `json:"mining"`
+	Wallet    *WalletConfig      `json:"wallet"`
+	Heartbeat *HeartbeatConfig   `json:"heartbeat"`
+	Net       string             `json:"net"`
+	Metrics   *MetricsConfig     `json:"metrics"`
+	Mpool     *MessagePoolConfig `json:"mpool"`
 }
 
 // APIConfig holds all configuration options related to the api.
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -167,6 +168,16 @@ func newDefaultMetricsConfig() *MetricsConfig {
 	}
 }
 
+type MessagePoolConfig struct {
+	Limit int `json:"limit"`
```
anorth: Use a more descriptive name, e.g. "maxMessageCount". There will be other limits in the future.

config/config.go
```diff
@@ -167,6 +168,16 @@ func newDefaultMetricsConfig() *MetricsConfig {
...
+}
+
+func newDefaultMessagePoolConfig() *MessagePoolConfig {
+	return &MessagePoolConfig{
+		Limit: 100,
```
anorth: I think we can start larger than this, like a few thousand. Perhaps find the max message size and then set this to allow 256MB of messages (and document that calculation here).

core/message_pool.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/config"
```
anorth: Not your immediate problem, but having everything depend on a package that has the config for everything else is ick. Can you move the MessagePoolConfig here and reverse the dependency?

core/message_pool.go
```diff
@@ -61,6 +63,10 @@ func (pool *MessagePool) addTimedMessage(msg *timedmessage) (cid.Cid, error) {
 		return cid.Undef, errors.Wrap(err, "failed to create CID")
 	}
 
+	if len(pool.pending) == pool.cfg.Limit {
```
anorth: Use `>=`

core/message_pool_test.go
```diff
@@ -114,6 +151,34 @@ func TestMessagePoolAsync(t *testing.T) {
...
+				}
+			}
+			wg.Done()
+		}(i)
+	}
+
+	wg.Wait()
+	assert.Len(pool.Pending(), limit)
+}
+
```
anorth: I don't think this test is needed.

plumbing/msg/sender_test.go
```diff
@@ -33,7 +40,7 @@ func TestSend(t *testing.T) {
 		addr := w.Addresses()[0]
 		timer := testhelpers.NewTestBlockTimer(1000)
 		queue := core.NewMessageQueue()
-		pool := core.NewMessagePool(timer)
+		pool := core.NewMessagePool(newTestMessagePoolConfig(100), timer)
```
anorth: Extract a single instance like you did in other tests.

config/config.go
travisperson: Max object sizes are 1MiB<sub>[0](https://github.com/filecoin-project/specs/blob/master/data-structures.md#encoding-considerations)</sub> (2<sup>20</sup>), that would put us to ~256 messages.

frrist:
Going to wait for https://github.com/filecoin-project/go-filecoin/pull/2522 to land. This PR will serve as a follow on to make it config based.

anorth:
Significant changes have been requested to #2522. This PR is much tighter scope, so I recommend landing it first (but please coordinate with @acruikshank).

frrist:
This was addressed in https://github.com/filecoin-project/go-filecoin/pull/2522, will open a follow on PR to make size a configuration option.

PR #2522, opened 2019-04-09
Add pre-validation to message pool

### LINKED ISSUE #2256: Validate message going into the message pool
### Description

We currently validate signatures prior to admitting messages to the message pool. It's not entirely clear why we do this since we do a much more robust validation including signature checking prior to processing. If this check is useful, it's probably a security check to keep unauthorized messages from filling the message queue. 

If this is the case, however, we really should be checking more than just the signature. It's easy to create messages with valid signatures that cannot be processed. If the goal is to keep unprocessable messages (messages that don't earn gas) out of the message pool we should use the inbound message validator to gate entry to the pool.

### Acceptance criteria

1. Limit total number of messages allowed in message pool at any time.
2. Drop messages sent to self.
3. Drop messages with gas limit greater than block limit.
4. Drop messages where actor has insufficient funds to cover gas.
5. Drop messages with invalid actor cid.
6. Drop messages without valid signatures.
7. Drop messages with duplicate nonce.
8. Drop messages with delta between nonce and current nonce greater than some limit

Note that this is all the checks that DefaultMessageValidator makes with additional check for message pool size, message length, duplicate nonces and maximum nonce height.

It excludes a check for minimum price because that is going to break a lot of things. It has its own issue #2523.

It also excludes a check for message size as that should be pushed back further into the network layer (libp2p already enforces 1Mb limit) #2525

### Risks & Pitfalls

The from actor is required for validation. It may be difficult to access from the message pool.

### LINKED ISSUE #2517: The go-filecoin message pool has a limit to hedge against DoS attacks
### Description
As part of addressing https://github.com/filecoin-project/go-filecoin/issues/2500 the message pool needs a size limit to help hedge against DoS attacks.

### Acceptance criteria
- Message pool has a limit that prevents it from growing in size indefinitely.
- Tests for said limiting.

### Risks + pitfalls
This does not completely prevent DoS attacks on message pool, more thought is needed as pointed out here: https://github.com/filecoin-project/go-filecoin/issues/2500#issuecomment-480701567

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/master/core/message_pool.go

acruikshank:
closes #2517
partially addresses #2500
closes #2256

### Problem

We have no pre-validation on the message pool. This means any random message can land in the message pool and waste space. Because there are several types of message error that prevent us charging gas, an attacker can easily spam the message pool with message that they know will never cost any Filecoin.

### Solution

Add validation to the message pool. This:

1. Limit total number of messages allowed in message pool at any time.
2. Drop messages sent to self.
3. Drop messages with gas limit greater than block limit.
4. Drop messages where actor has insufficient funds to cover gas.
5. Drop messages with invalid actor cid.
6.  Drop messages without valid signatures.
7. Drop messages with duplicate nonce.
8. Drop messages with delta between nonce and current nonce greater than some limit

Most of this is accomplished by using the same validation we use when validating messages for processing.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -2,6 +2,7 @@ package consensus
 
 import (
 	"context"
+	"github.com/filecoin-project/go-filecoin/address"
 	"math/big"
 
 	"github.com/filecoin-project/go-filecoin/actor"
@@ -95,7 +96,7 @@ func canCoverGasLimit(msg *types.SignedMessage, actor *actor.Actor) bool {
 
 // IngestionValidatorAPI allows the validator to access latest state
 type ingestionValidatorAPI interface {
-	LatestState(ctx context.Context) (state.Tree, error)
+	ActorFromLatestState(ctx context.Context, address address.Address) (*actor.Actor, error)
 }
 
 // IngestionValidator can access latest state and runs additional checks to mitigate DoS attacks
@@ -116,12 +117,7 @@ func NewIngestionValidator(api ingestionValidatorAPI) *IngestionValidator {
 // Errors probably mean the validation failed, but possibly indicate a failure to retrieve state
 func (v *IngestionValidator) Validate(ctx context.Context, msg *types.SignedMessage) error {
 	// retrieve from actor
-	st, err := v.api.LatestState(ctx)
-	if err != nil {
-		return err
-	}
-
-	fromActor, err := st.GetActor(ctx, msg.From)
+	fromActor, err := v.api.ActorFromLatestState(ctx, msg.From)
 	if err != nil {
 		if state.IsActorNotFoundError(err) {
 			fromActor = &actor.Actor{}
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -6,8 +6,13 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/account"
+	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/filecoin-project/go-filecoin/vm/errors"
 )
+
+// MaxNonceGap is the maximum nonce of a message past the last received on chain
+const MaxNonceGap = 100
 
 // SignedMessageValidator validates incoming signed messages.
 type SignedMessageValidator interface {
@@ -86,4 +91,49 @@ func (v *defaultMessageValidator) Validate(ctx context.Context, msg *types.Signe
 func canCoverGasLimit(msg *types.SignedMessage, actor *actor.Actor) bool {
 	maximumGasCharge := msg.GasPrice.MulBigInt(big.NewInt(int64(msg.GasLimit)))
 	return maximumGasCharge.LessEqual(actor.Balance.Sub(msg.Value))
+}
+
+// IngestionValidatorAPI allows the validator to access latest state
+type ingestionValidatorAPI interface {
+	LatestState(ctx context.Context) (state.Tree, error)
+}
+
+// IngestionValidator can access latest state and runs additional checks to mitigate DoS attacks
+type IngestionValidator struct {
+	api       ingestionValidatorAPI
+	validator defaultMessageValidator
+}
+
+// NewIngestionValidator creates a new validator with an api
+func NewIngestionValidator(api ingestionValidatorAPI) *IngestionValidator {
+	return &IngestionValidator{
+		api:       api,
+		validator: defaultMessageValidator{allowHighNonce: true},
+	}
+}
+
+// Validate validates the signed message.
+// Errors probably mean the validation failed, but possibly indicate a failure to retrieve state
+func (v *IngestionValidator) Validate(ctx context.Context, msg *types.SignedMessage) error {
+	// retrieve from actor
+	st, err := v.api.LatestState(ctx)
+	if err != nil {
+		return err
+	}
+
+	fromActor, err := st.GetActor(ctx, msg.From)
+	if err != nil {
+		if state.IsActorNotFoundError(err) {
+			fromActor = &actor.Actor{}
+		} else {
+			return err
+		}
+	}
+
+	// check that message nonce is not too high
+	if msg.Nonce > fromActor.Nonce && msg.Nonce-fromActor.Nonce > MaxNonceGap {
+		return errors.NewRevertErrorf("message nonce (%d) is too much greater than actor nonce (%d)", msg.Nonce, fromActor.Nonce)
+	}
+
+	return v.validator.Validate(ctx, msg, fromActor)
 }
--- a/core/message_pool.go
+++ b/core/message_pool.go
@@ -7,12 +7,9 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/pkg/errors"
 
-	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
-	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/metrics"
-	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -21,27 +18,22 @@ var mpSize = metrics.NewInt64Gauge("message_pool_size", "The size of the message
 // MaxMessagePoolSize is the maximum number of pending messages will will allow in the message pool at any time
 const MaxMessagePoolSize = 10000
 
-// MaxNonceGap is the maximum nonce of a message past the last received on chain
-const MaxNonceGap = 100
-
 // MessageTimeOut is the number of tipsets we should receive before timing out messages
 const MessageTimeOut = 6
-
-var (
-	ErrMessagePoolFull  = errors.New("message pool is full")
-	ErrNonceGapExceeded = errors.New("message nonce is too much greater than highest nonce on chain")
-	ErrDuplicateNonce   = errors.New("different message with same from address and nonce is already in pool")
-)
 
 type timedmessage struct {
 	message *types.SignedMessage
 	addedAt uint64
 }
 
-// MessagePoolAPI defines a interface api resources the message pool needs.
+// MessagePoolAPI defines an interface to api resources the message pool needs.
 type MessagePoolAPI interface {
 	BlockHeight() (uint64, error)
-	LatestState(ctx context.Context) (state.Tree, error)
+}
+
+// MessagePoolValidator defines a validator that ensures a message can go through the pool.
+type MessagePoolValidator interface {
+	Validate(ctx context.Context, msg *types.SignedMessage) error
 }
 
 type addressNonce struct {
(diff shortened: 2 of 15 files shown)


mining/worker_test.go
```diff
@@ -269,29 +269,42 @@ func TestGeneratePoolBlockResults(t *testing.T) {
...
-	_, err = pool.Add(smsg3)
+	_, err = pool.Add(ctx, smsg3)
 	assert.NoError(err)
-	_, err = pool.Add(smsg4)
+	_, err = pool.Add(ctx, smsg4)
 	assert.NoError(err)
 
 	assert.Len(pool.Pending(), 4)
+
+	// Wet actor nonce past nonce of message in pool.
```
frrist: is "Wet" a typo here?

core/message_pool.go
```diff
@@ -209,3 +246,42 @@ func (pool *MessagePool) LargestNonce(address address.Address) (largest uint64,
...
+			return err
+		}
+	}
+
+	// check that message nonce is not too high
+	if message.Nonce > fromActor.Nonce && message.Nonce-fromActor.Nonce > MaxNonceGap {
+		return ErrNonceGapExceeded
+	}
+
+	if err = consensus.NewOutboundMessageValidator().Validate(ctx, message, fromActor); err != nil {
```
frrist: I think we can skip the err check here:
```suggestion
return consensus.NewOutboundMessageValidator().Validate(ctx, message, fromActor)
```

core/message_pool_test.go
```diff
@@ -571,3 +774,24 @@ func (p *storeBlockProvider) GetBlock(ctx context.Context, cid cid.Cid) (*types.
...
+	})
+}
+
+func resignMessage(signer types.Signer, message *types.SignedMessage, f func(*types.Message)) *types.SignedMessage {
+	var msg types.Message
+	msg = message.Message
+	f(&msg)
+	smg, err := signMessage(signer, msg)
+	if err != nil {
+		panic("Error signing message")
```
frrist: nit: if we are going to panic here let's call this `mustResignMessage`

core/message_pool.go
```diff
@@ -7,25 +7,50 @@ import (
...
+
+// MaxNonceGap is the maximum nonce of a message past the last received on chain
+const MaxNonceGap = 100
+
 // MessageTimeOut is the number of tipsets we should receive before timing out messages
 const MessageTimeOut = 6
 
+var (
+	ErrMessagePoolFull  = errors.New("message pool is full")
+	ErrNonceGapExceeded = errors.New("message nonce is too much greater than highest nonce on chain")
```
phritz: engrish

core/message_pool.go
```diff
@@ -7,25 +7,50 @@ import (
...
+)
+
 type timedmessage struct {
 	message *types.SignedMessage
 	addedAt uint64
 }
 
-// BlockTimer defines a interface to a struct that can give the current block height.
-type BlockTimer interface {
+// MessagePoolAPI defines a interface api resources the message pool needs.
```
phritz: english, please. also this isn't really the message pool api, it's the message pool's dependencies.

core/message_pool.go
```diff
@@ -38,21 +63,22 @@ type BlockTimer interface {
 type MessagePool struct {
 	lk sync.RWMutex
```
phritz: fwiw i dont see this lock used as a rwlock, i see it used as a regular exclusive lock

core/message_pool.go
```diff
@@ -102,15 +133,20 @@ func (pool *MessagePool) Remove(c cid.Cid) {
...
-func NewMessagePool(timer BlockTimer) *MessagePool {
+func NewMessagePool(api MessagePoolAPI) *MessagePool {
 	return &MessagePool{
-		timer:   timer,
-		pending: make(map[cid.Cid]*timedmessage),
+		api:           api,
+		pending:       make(map[cid.Cid]*timedmessage),
+		addressNonces: make(map[addressNonce]bool),
 	}
 }
```
phritz: maybe remove the todo below, or turn it into issues if there is something specific still missing (i think not)

core/message_pool.go
```diff
@@ -38,21 +63,22 @@ type BlockTimer interface {
...
 	if err != nil {
 		return cid.Undef, err
 	}
 
-	return pool.addTimedMessage(&timedmessage{message: msg, addedAt: blockTime})
+	return pool.addTimedMessage(ctx, &timedmessage{message: msg, addedAt: blockTime})
 }
 
-func (pool *MessagePool) addTimedMessage(msg *timedmessage) (cid.Cid, error) {
+func (pool *MessagePool) addTimedMessage(ctx context.Context, msg *timedmessage) (cid.Cid, error) {
```
phritz: prolly worth a comment that if this thing returns error it could be because of something Bad eg can't compute a cid or it could be because the message is not valid which is expected. if you wanted to distinguish we could move validation out of addtimedmessage at the cost of a little bit of danger that someone adds unvalidated messages.

core/message_pool.go
```diff
@@ -133,9 +169,9 @@ func (pool *MessagePool) UpdateMessagePool(ctx context.Context, store chain.Bloc
 	// Add all message from the old blocks to the message pool, so they can be mined again.
 	for _, blk := range oldBlocks {
 		for _, msg := range blk.Messages {
-			_, err = pool.addTimedMessage(&timedmessage{message: msg, addedAt: uint64(blk.Height)})
+			_, err = pool.addTimedMessage(ctx, &timedmessage{message: msg, addedAt: uint64(blk.Height)})
```
phritz: blocking: just noticed line 226 which is not your bug but it is accessing pool.pending without holding the lock, needs to be fixed probably to pool.Pending()

plumbing/msg/sender.go
```diff
@@ -29,6 +29,11 @@ type chainState interface {
 	LatestState(ctx context.Context) (state.Tree, error)
 }
 
+// BlockTimer defines a interface to a struct that can give the current block height.
+type BlockTimer interface {
```
phritz: blockheighter would be more clear

core/message_pool.go
frrist: If you are going to change this can you give it a little [hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2) -- something like `poolMu`?

core/message_pool.go
ZenGround0: very comment

core/message_pool.go
```diff
@@ -7,25 +7,50 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/pkg/errors"
 
+	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/metrics"
+	"github.com/filecoin-project/go-filecoin/state"
```
anorth: Blocking: these dependencies are really unfortunate. The message pool is a low level component, nearly a leaf of an ideal dependency tree. This is bringing too much into scope.

Instead, inject the validation logic from outside. The consensus package already, necessarily has this set of deps and other validation logic. Test the validation logic independently of the pool, and for the pool merely test that the pool respects the decision of a (fake) validator.

core/message_pool.go
anorth: These error messages would be more useful constructed in place, where they can include helpful information about the pool size and limit, actor nonce etc.

core/message_pool.go
```diff
@@ -209,3 +245,38 @@ func (pool *MessagePool) LargestNonce(address address.Address) (largest uint64,
...
+			return err
+		}
+	}
+
+	// check that message nonce is not too high
+	if message.Nonce > fromActor.Nonce && message.Nonce-fromActor.Nonce > MaxNonceGap {
+		return ErrNonceGapExceeded
+	}
+
+	return consensus.NewOutboundMessageValidator().Validate(ctx, message, fromActor)
```
anorth: "Outbound" is unexpected here, so add a comment describing why it's the right thing.

core/message_pool.go
acruikshank: Resolved.

consensus/validation.go
```diff
@@ -87,3 +92,48 @@ func canCoverGasLimit(msg *types.SignedMessage, actor *actor.Actor) bool {
 	maximumGasCharge := msg.GasPrice.MulBigInt(big.NewInt(int64(msg.GasLimit)))
 	return maximumGasCharge.LessEqual(actor.Balance.Sub(msg.Value))
 }
+
+// IngestionValidatorAPI allows the validator to access latest state
+type ingestionValidatorAPI interface {
+	LatestState(ctx context.Context) (state.Tree, error)
```
anorth: It looks like this could be even higher level, with `GetActor`.

core/message_pool.go
```diff
@@ -38,21 +55,25 @@ type BlockTimer interface {
 type MessagePool struct {
 	lk sync.RWMutex
 
-	timer   BlockTimer
-	pending map[cid.Cid]*timedmessage // all pending messages
+	api           MessagePoolAPI
+	validator     MessagePoolValidator
+	pending       map[cid.Cid]*timedmessage // all pending messages
+	addressNonces map[addressNonce]bool
```
anorth: Document the intended use of this as a set of addressNonce corresponding to all the pending messages.

core/message_pool.go
```diff
@@ -102,15 +128,21 @@ func (pool *MessagePool) Remove(c cid.Cid) {
 	pool.lk.Lock()
 	defer pool.lk.Unlock()
 
+	msg, ok := pool.pending[c]
+	if ok {
+		delete(pool.addressNonces, newAddressNonce(msg.message))
+	}
 	delete(pool.pending, c)
```
anorth: May as well move this inside the `if`

core/message_pool.go
```diff
@@ -187,13 +219,25 @@ func (pool *MessagePool) timeoutMessages(ctx context.Context, store chain.BlockP
...
 
 	// remove all messages added before minimumHeight
+	for _, cid := range pool.messagesToTimeOut(minimumHeight) {
+		pool.Remove(cid)
+	}
+
+	return nil
+}
+
+// identify all messages that need to be within a read lock
```
anorth: "need to be removed"

core/message_pool.go
```diff
@@ -209,3 +253,20 @@ func (pool *MessagePool) LargestNonce(address address.Address) (largest uint64,
 	}
 	return
 }
+
+// validateMessage in message pool is a mechanism for preventing memory based DoS attacks.
+// As such, it will often fail silently.
```
anorth: I don't understand this comment.

testhelpers/core.go
```diff
@@ -88,18 +89,33 @@ func RequireRandomPeerID(require *require.Assertions) peer.ID {
...
+// NewTestMessagePoolAPI creates a new TestMessagePoolAPI.
+func NewTestMessagePoolAPI(h uint64) *TestMessagePoolAPI {
+	return &TestMessagePoolAPI{Height: h}
+}
+
+type MockMessagePoolValidator struct {
+	Valid bool
+}
+
+func NewMockMessagePoolValidator() *MockMessagePoolValidator {
```
anorth: I would call this a fake rather than a mock, since it doesn't record or check interactions.

core/message_pool.go
acruikshank: done

core/message_pool_test.go
acruikshank: done

PR #2528, opened 2019-04-09
FAST payment channel tests

travisperson:
~Depends on #2515~ merged.
Previously #2319

### DIFF
--- a/commands/payment_channel_daemon_test.go
+++ b/commands/payment_channel_daemon_test.go
@@ -1,120 +1,153 @@
 package commands_test
 
 import (
-	"fmt"
-	"strings"
-	"sync"
+	"context"
+	"math/big"
 	"testing"
+	"time"
 
-	"github.com/ipfs/go-cid"
-	cbor "github.com/ipfs/go-ipld-cbor"
-	"github.com/multiformats/go-multibase"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker"
 	"github.com/filecoin-project/go-filecoin/address"
-	"github.com/filecoin-project/go-filecoin/fixtures"
-	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 func TestPaymentChannelCreateSuccess(t *testing.T) {
 	tf.IntegrationTest(t)
 
-	assert := assert.New(t)
+	require := require.New(t)
 
-	d := makeTestDaemonWithMinerAndStart(t)
-	defer d.ShutdownSuccess()
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+	defer cancel()
 
-	args := []string{"paych", "create"}
-	args = append(args, "--from", fixtures.TestAddresses[0], "--gas-price", "0", "--gas-limit", "300")
-	args = append(args, fixtures.TestAddresses[1], "10000", "20")
+	// Get basic testing environment
+	ctx, env := fastesting.NewTestEnvironment(ctx, t, fast.EnvironmentOpts{})
 
-	paymentChannelCmd := d.RunSuccess(args...)
-	messageCid, err := cid.Parse(strings.Trim(paymentChannelCmd.ReadStdout(), "\n"))
-	require.NoError(t, err)
-
-	var wg sync.WaitGroup
-
-	wg.Add(1)
-	go func() {
-		wait := d.RunSuccess("message", "wait",
-			"--return",
-			"--message=false",
-			"--receipt=false",
-			messageCid.String(),
-		)
-		_, ok := types.NewChannelIDFromString(strings.Trim(wait.ReadStdout(), "\n"), 10)
-		assert.True(ok)
-		wg.Done()
+	// Teardown after test ends
+	defer func() {
+		err := env.Teardown(ctx)
+		require.NoError(err)
 	}()
 
-	d.RunSuccess("mining once")
+	// Start test
+	rsrc := requireNewPaychResource(ctx, t, env)
 
-	wg.Wait()
+	channelExpiry := types.NewBlockHeight(20)
+	channelAmount := types.NewAttoFILFromFIL(1000)
+
+	rsrc.requirePaymentChannel(ctx, channelAmount, channelExpiry)
 }
 
 func TestPaymentChannelLs(t *testing.T) {
 	tf.IntegrationTest(t)
 
-	assert := assert.New(t)
-	require := require.New(t)
-
 	t.Run("Works with default payer", func(t *testing.T) {
+		require := require.New(t)
+		assert := assert.New(t)
 
-		payer, err := address.NewFromString(fixtures.TestAddresses[2])
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
+		defer cancel()
+
+		// Get basic testing environment
+		ctx, env := fastesting.NewTestEnvironment(ctx, t, fast.EnvironmentOpts{})
+
+		// Teardown after test ends
+		defer func() {
+			err := env.Teardown(ctx)
+			require.NoError(err)
+		}()
+
+		// Start test
+		rsrc := requireNewPaychResource(ctx, t, env)
+
+		channelExpiry := types.NewBlockHeight(20)
+		channelAmount := types.NewAttoFILFromFIL(1000)
+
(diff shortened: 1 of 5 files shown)


tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,132 @@
+package fasting
```
frrist: :thinking: not the biggest fan of this name, but I don't have a better idea, `fastesting`?

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,132 @@
...
+	p := env.NewNodeConnected()
+
+	err := series.SendFilecoinDefaults(env.ctx, env.GenesisMiner, p, funds)
+	require.NoError(err)
+
+	return p
+}
+
+// BasicFastSetup creates a environment and a single node, and environment options
+func BasicFastSetup(ctx context.Context, t *testing.T, fastenvOpts fast.EnvironmentOpts) (*Environment, context.Context) {
```
frrist: Tiny nit: 
``` suggestion
func BasicFastSetup(ctx context.Context, t *testing.T, fastenvOpts fast.EnvironmentOpts) (context.Context, *Environment) {

```

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,132 @@
...
+	require := require.New(t)
+
+	// Create a directory for the test using the test name (mostly for FAST)
+	// Replace the forward slash as tempdir can't handle them
+	dir, err := ioutil.TempDir("", strings.Replace(t.Name(), "/", ".", -1))
```
frrist: I think devs would find FAST tests easier to debug if we put this repo in the same directory as the test being ran. I believe this is what go-ipfs does for sharness tests.

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,132 @@
...
+	ctx = series.SetCtxSleepDelay(ctx, time.Second)
+
+	return &Environment{
+		Environment:  env,
+		t:            t,
+		ctx:          ctx,
+		pluginName:   localplugin.PluginName,
+		pluginOpts:   options,
+		fastenvOpts:  fastenvOpts,
+		GenesisMiner: genesis,
```
frrist: BLOCKING: Maybe this is a different thing, but I think the Environment [already has this value](https://github.com/filecoin-project/go-filecoin/blob/1968f9a3d5c27f122ddcf48e404504b4c5d84050/tools/fast/fasting/basic.go#L31)? As well as the others, with the exception of `testing.T`. Lets talk about this more before merging.

tools/fast/fasting/basic.go
travisperson: Any idea how to get that?

tools/fast/fasting/basic.go
travisperson: The linked environment is the struct that this is filling out. Did you mean on the [`environment` interface?

https://github.com/filecoin-project/go-filecoin/blob/e3f7439e2ac63e555820a8ec7e0b290974042786/tools/fast/environment.go#L36-L46

The value returned by GenesisMiner function is not a filecoin node, but struct representing the miner in the genesis block with power, basically an account actor / wallet private key with the address of the miner in the block.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,528 @@
...
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 func TestPaymentChannelCreateSuccess(t *testing.T) {
 	t.Parallel()
-	assert := assert.New(t)
+	require := require.New(t)
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
```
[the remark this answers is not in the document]
travisperson: It's more of a coincidence that they are the same, and I do not expect all tests to have the same execution deadline. I'd be hesitant to move this into a variable as it may accidentally enforce the pattern and remove the actual usefulness of a deadline, as the `go test` command it self already has a globally enforced deadline. A shared variable will most likely just be incremented arbitrary.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,528 @@
...
+	rsrc.CreatePaymentChannel(ctx, t, channelAmount, channelExpiry)
 }
 
 func TestPaymentChannelLs(t *testing.T) {
 	t.Parallel()
-	assert := assert.New(t)
-	require := require.New(t)
 
 	t.Run("Works with default payer", func(t *testing.T) {
 		t.Parallel()
```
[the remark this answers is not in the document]
travisperson: All good, as each test is using an independent instance of a FAST environment. I think the issue is around running many commands against a single Filecoin process that we need to be warry of.

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,528 @@
...
+		targetAddr: targetAddr,
 
-	wg.Wait()
+		payer:     payerDaemon,
+		payerAddr: payerAddr,
+	}
 }
 
-func mustEncodeVoucherStr(t *testing.T, voucher paymentbroker.PaymentVoucher) string {
+func (rsrc *PaychResources) CreatePaymentChannel(ctx context.Context, t *testing.T, amt *types.AttoFIL, eol *types.BlockHeight) *types.ChannelID {
```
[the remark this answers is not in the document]
travisperson: I don't have any strong opinion here, but every method will be at some point end up with a `Must` prefix as they are part of the test it self and it feels a bit redundant.

tools/fast/fasting/basic.go
```diff
@@ -0,0 +1,132 @@
...
+
+	pluginName string
+	pluginOpts map[string]string
+
+	fastenvOpts fast.EnvironmentOpts
+
+	GenesisMiner *fast.Filecoin
+}
+
+func (env *Environment) NewNode() *fast.Filecoin {
```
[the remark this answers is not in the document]
travisperson: What about `RequireNewNode`? It will help follow the pattern already established in the [message queue tests](https://github.com/filecoin-project/go-filecoin/blob/master/core/message_queue_test.go), and reads better I think.

commands/payment_channel_daemon_test.go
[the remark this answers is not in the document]
travisperson: Maybe `require` as done in the message queue tests https://github.com/filecoin-project/go-filecoin/blob/master/core/message_queue_test.go?

tools/fast/fasting/basic.go
frrist: Ahh I see what's happening now, I got confused by the naming, all good here.

tools/fast/fasting/basic.go
frrist: oops, I didn't realize how complicated this ask was. Yes, but it's hacky and involves the runtime package -- probably better not to do it.

...But [this](https://github.com/sirupsen/logrus/blob/a6c0064cfaf982707445a1c90368f956421ebcf0/entry.go#L166-L196) is what I would reference if I were to attempt it

commands/payment_channel_daemon_test.go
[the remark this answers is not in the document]
frrist: As long as the method name indicates it will either "do the thing" or "fail the test" I think we are good.

tools/fast/fasting/basic.go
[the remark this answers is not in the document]
frrist: SGTM same answer here as above

commands/payment_channel_daemon_test.go
```diff
@@ -1,547 +1,532 @@
...
-	"github.com/filecoin-project/go-filecoin/fixtures"
-	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 func TestPaymentChannelCreateSuccess(t *testing.T) {
 	t.Parallel()
```
anorth: Don't forget to rebase against the new test flags.

tools/fast/fastesting/basic.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/ipfs/go-ipfs-files"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-filecoin/testhelpers"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
+	localplugin "github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin/local"
+)
+
+type TestEnvironment struct {
```
anorth: Please add documentation comments to these types and methods, as it looks like they're intended to be widely used by test writers.

In fact, why did the linter not catch this?

commands/payment_channel_daemon_test.go
travisperson: I'll wait for it to land.

tools/fast/fastesting/basic.go
travisperson: I'm a little surprised that this is passing CI now. The linter should of said something about these missing.

Issue #2530, opened 2019-04-09
A miner selects a sector size when pledging


laser:
### Description

A miner should select a sector size when pledging, i.e. when `miner create` runs. To start with, we should support `Test` and `TwoFiftySixMiB`. The choice of sector size should be constrained to some enumeration of values in Go code.

### Acceptance criteria

[ ] - update `miner create` command to require a new sector size-argument
[ ] - rust-fil-proofs uses selected sector size when packing pieces into sectors and sealing
[ ] - go-filecoin actors use selected sector size when verifying seal and PoSt proofs
[ ] - modify gengen `setup.json` to include miner's sector size
[ ] - make sure that 1KiB sectors are used when in test proofs mode and prevent 1KiB sectors from being used when not in test proofs mode

PR #2534, opened 2019-04-09
rename "sector store type" to "proofs mode"

laser:

## What's in this PR?

- rename "SectorStoreType" to "ProofsMode"

## Why is this PR needed?

- the old name didn't make sense

### DIFF
--- a/proofs/interface.go
+++ b/proofs/interface.go
@@ -36,7 +36,9 @@ type Verifier interface {
 	VerifySeal(VerifySealRequest) (VerifySealResponse, error)
 }
 
-// Mode configures the behavior of the SectorStore used by the SectorBuilder.
+// Mode configures sealing, sector packing, PoSt generation and other behaviors
+// of libfilecoin_proofs. Use Test mode to seal and generate PoSts quickly over
+// tiny sectors. Use Live when operating a real Filecoin node.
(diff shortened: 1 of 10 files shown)


laser:
cc @rkowalick

proofs/interface.go
```diff
@@ -36,7 +36,9 @@ type Verifier interface {
 	VerifySeal(VerifySealRequest) (VerifySealResponse, error)
 }
 
-// Mode configures the behavior of the SectorStore used by the SectorBuilder.
+// Mode configures sealing, sector packing, PoSt generation and other behaviors
```
ZenGround0: Great comment

rkowalick:
As per the quote below, can you rename `proofs.Test` and `proofs.Live`?
> Could I convince you to use the term "proofs mode" and type `proofs.Mode` instead of "sector store type" (`proofs.SectorStoreType`) in your PR?
>
> The two members of that type would be `LiveMode` and `TestMode`.

proofs/interface.go
laser: thnx sir

laser:
@rkowalick - 

> As per the quote below, can you rename proofs.Test and proofs.Live?

done!

PR #2537, opened 2019-04-10
[WIP] add Unit & Integration Testing flags

frrist:
:construction: comments are welcome, but don't put too much effort into a review at this time :construction:


commands/bitswap_stats_daemon_test.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 func TestBitswapStats(t *testing.T) {
-	t.Parallel()
+	th.UnitTest(t)
+
 	assert := assert.New(t)
 
 	d := th.NewDaemon(t).Start()
```
frrist: Renamed because this is a **daemon** test

commands/init_test.go
```diff
@@ -1,34 +0,0 @@
...
-		ctx,
-		th.ProjectRoot("tools/genesis-file-server/genesis-file-server"),
-		"--genesis-file-path",
-		th.ProjectRoot("fixtures/genesis.car"),
-		"--port",
-		strconv.Itoa(port),
-	).Start()
-	require.NoError(t, err)
-
-	td := th.NewDaemon(t, th.GenesisFile(fmt.Sprintf("http://127.0.0.1:%d/genesis.car", port))).Start()
```
frrist: this entire test was moved to `init_daemon_test.go` because its a **daemon** test.

testhelpers/flags.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"flag"
+	"testing"
+)
+
+var integrationTest = flag.Bool("integration", false, "Run the integration go tests")
+var unitTest = flag.Bool("unit", false, "Run the unit go tests")
+
+var concurrent = flag.Bool("concurrent", true, "Run the go tests concurrently")
```
frrist: Note to Reviewer(s):

All `t.Parallel()` calls were removed from tests. The call is now made inside functions that determines whether or not a test is ran (currently `UnitTest` or `IntegrationTest`), this means that sub tests (`t.Run()`) will be ran sequentially. It is my _opinion_ that this makes the testing code easier to read, feel free to push back.

The `concurrent` (avoid collision with `test.parallel`) flag has been added with a default value of `true`. This means tests are ran in parallel by default -- the original behaviour. This flag allows us to configure certain categories of tests to run in parallel or serially which I believe could prove useful in debugging test performance down the road, again feel free to push back on this.

testhelpers/flags.go
```diff
@@ -0,0 +1,33 @@
+package testhelpers
+
+import (
+	"flag"
+	"testing"
+)
+
+var integrationTest = flag.Bool("integration", false, "Run the integration go tests")
+var unitTest = flag.Bool("unit", false, "Run the unit go tests")
```
frrist: Note to Reviewer(s):

We will need to pick a "default" set of tests to run. Currently (as @travisperson has pointed out in a comment) no tests are run unless the `-unit` or `-integration` flags are passed to them. Once we decided on a default set of tests we can set their flags default value to `true`.

travisperson: I like the idea of having all tests instrumented. It provides an easy way to find tests which are not instrumented, as we can pass false values to all flags and expect nothing to run :+1:

frrist: The plan is to instrument all tests with these flags, the question I was getting at was: which flags should default to `true`?

travisperson: yep

frrist:
:balloon: this is getting to big for a sane review, going to close this and open a few smaller ones.

Issue #2274, opened 2019-03-14
ensure that CommRs passed to PoSt generate and verify CGO calls are passed are ordered identically


laser:
### Description

Ensure that the CommRs sent to generate_post from go-filecoin are ordered in the same way as the CommRs sent to verify_post. These commitments are stored in a Golang map keyed by stringified sector id.

PR #2539, opened 2019-04-10
ensure CommRs are ordered

sidke:
fixes: https://github.com/filecoin-project/go-filecoin/issues/2274

sorts CommRs when verifying and generating PoSTs


protocol/storage/miner.go
```diff
@@ -806,7 +806,9 @@ func (sm *Miner) submitPoSt(start, end *types.BlockHeight, seed proofs.PoStChall
 		commRs[i] = input.commR
 	}
 
-	proofs, faults, err := sm.generatePoSt(commRs, seed)
+	sortedCommRSlice := types.NewSortedCommRSlice(commRs...)
+
+	proofs, faults, err := sm.generatePoSt(sortedCommRSlice.Values(), seed)
```
laser: BLOCKING: `generatePoSt` should be modified to accept a value of type `SortedCommRSlice`, and the `GeneratePoStRequest` should be modified to look like this:

```
type GeneratePoStRequest struct {
	CommRs        SortedCommRSlice
	ChallengeSeed proofs.PoStChallengeSeed
}
```

Similarly, the `VerifyPoStRequest` type should be modified to look like:

```
type VerifyPoSTRequest struct {
	ChallengeSeed PoStChallengeSeed
	CommRs        SortedCommRSlice
	Faults        []uint64
	Proofs        []PoStProof
	StoreType     SectorStoreType // used to control sealing/verification performance
}
```

Rationale: The type enforces an invariant, which is that these CommRs are sorted identically in both PoSt generation and verification flows. This type should to make its way as close to the CGO call as possible to ensure that we're enforcing the invariant.

proofs/interface.go
```diff
@@ -48,3 +53,26 @@ const (
 	// Filecoin node.
 	LiveMode
 )
+
+// SortedCommRs is a slice of CommRs that has deterministic ordering.
+type SortedCommRs struct {
+	c []CommR
```
rosalinekarr: (Non-blocking) Can we avoid cryptic variable names like this on the Golang side? `c` is not especially clear.

proofs/rustverifier.go
```diff
@@ -81,8 +81,9 @@ func (rp *RustVerifier) VerifyPoST(req VerifyPoSTRequest) (VerifyPoSTResponse, e
 	defer elapsed("VerifyPoST")()
 
 	// flattening the byte slice makes it easier to copy into the C heap
-	flattened := make([]byte, 32*len(req.CommRs))
-	for idx, commR := range req.CommRs {
+	commRs := req.SortedCommRs.Values()
+	flattened := make([]byte, 32*len(commRs))
```
rosalinekarr: (Non-blocking) This isn't your change, but I'd really prefer if we just used the actual type for the first argument to `make` instead of `[]byte` and multiplying the length by 32. This could be an issue if commRs ever change to a type that isn't 32 bytes in size.

proofs/rustverifier.go
```diff
@@ -81,8 +81,9 @@ func (rp *RustVerifier) VerifyPoST(req VerifyPoSTRequest) (VerifyPoSTResponse, e
...
+	for idx, commR := range commRs {
 		copy(flattened[(32*idx):(32*(1+idx))], commR[:])
```
rosalinekarr: This logic could be much simpler too, if we didn't have the `32` silliness.

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -376,8 +376,9 @@ func (sb *RustSectorBuilder) GeneratePoSt(req GeneratePoStRequest) (GeneratePoSt
 	defer elapsed("GeneratePoSt")()
 
 	// flattening the byte slice makes it easier to copy into the C heap
-	flattened := make([]byte, 32*len(req.CommRs))
-	for idx, commR := range req.CommRs {
+	commRs := req.SortedCommRs.Values()
+	flattened := make([]byte, 32*len(commRs))
```
rosalinekarr: This 32 could go away too.

rosalinekarr:
Added an issue for the pointer math weirdness: https://github.com/filecoin-project/go-filecoin/issues/2563

proofs/rustverifier.go
laser: > This could be an issue if commRs ever change to a type that isn't 32 bytes in size.

We have to flatten our slice of fixed-length arrays into a byte slice in order to work with `C.CBytes`. 

Also, `CommR` is a [specified type](https://github.com/filecoin-project/specs/blob/99f42e76e9b3e4f309578dc1a2f1c6e4bfabfb48/proofs.md) and will likely not change from a 32-byte array.

PR #2542, opened 2019-04-11
Add Functional Integration and Unit test flags

frrist:
### Description
This PR adds 3 flags that may be used to control when certain categories of tests run.
- `-functional`: used to run tests like the Retrieval test
-  `-integration`: used to run tests like `*test_daemon.go` or faucet tests
-  `-unit`: used to run tests like `address_test.go`. The `-short` flag may be used interchangeably here.

There are 4 methods that may be called from a test that allow it to run when one of these flags is passed.
- [`FunctionalTest`](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-1333f33210e4ae54e705a0b204ce2632R21)
  - [Example Usage](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-798479df3f7c6278d71f2744dd572c03R32)
- [`IntegrationTest`](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-1333f33210e4ae54e705a0b204ce2632R34)
  - [Example Usage](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-476c93ae2e2d8ffdcd28b21d05dd5ad0R20)
- [`UnitTest`](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-1333f33210e4ae54e705a0b204ce2632R47)
  - [Example Usage](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-b8caae3b8bfb04c2370daade503829d7R26)
- [`UnitTestWithSideEffectsThatIsBad`](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-1333f33210e4ae54e705a0b204ce2632R60)
  - [Example Usage](https://github.com/filecoin-project/go-filecoin/pull/2542/files#diff-7ad10fdb35d42f8a5450bdffbd5c20d3R44)

### Proposal 

All methods except `UnitTestWithSideEffectsThatIsBad` will cause the tests they are placed in to run in parallel. This means we no longer need to decorate tests with calls to `t.Parallel()`, I believe this makes the testing code less cluttered, it also serves as a motivation to write tests that can be parallelized. 
This does mean that the tests inside `t.Run` calls will no longer run in parallel, I believe this is a reasonable tradeoff as those tests are typically short and don't benefit from the parallelism. 

Tests that us `UnitTestWithSideEffectsThatIsBad` will be run serially, while tinkering with https://github.com/filecoin-project/go-filecoin/pull/2537 I observed that there is a subset of our unit tests that cannot be ran in parallel due to shared global state. The flag is named this way to discourage the writing of tests that behave this way. It also allows us to easily see what test should be optimized.

I propose we instrument all tests with these method, and that we default the `-unit` flag to `true`. This means that only the unit tests will run when the command `go test` is issued. All other tests must be called out explicitly by passing their corresponding flags.

### DIFF
--- a/testhelpers/testflags/flags.go
+++ b/testhelpers/testflags/flags.go
@@ -1,0 +1,64 @@
+package testflags
+
+import (
+	"flag"
+	"testing"
+)
+
+// Test enablement flags
+var functionalTest = flag.Bool("functional", false, "Run the functional go tests")
+var integrationTest = flag.Bool("integration", false, "Run the integration go tests")
+var unitTest = flag.Bool("unit", false, "Run the unit go tests")
+
+// Test modifiers
+// concurrent flag will cause tests to be ran in parallel, its default value is `true`.
+var concurrent = flag.Bool("concurrent", true, "Runs the go test in parallel")
+
+// FunctionalTest will run the test its called from iff the `-functional` flag
+// is passed when calling `go test`. Otherwise the test will be skipped. FunctionalTest
+// will run the test its called from in parallel, to disallow this behaviour see
+// the concurrent flag.
+func FunctionalTest(t *testing.T) {
+	if !*functionalTest {
+		t.SkipNow()
+	}
+	if *concurrent {
+		t.Parallel()
+	}
+}
+
+// IntegrationTest will run the test its called from iff the `-integration` flag
+// is passed when calling `go test`. Otherwise the test will be skipped. IntegrationTest
+// will run the test its called from in parallel, to disallow this behaviour see
+// the concurrent flag.
+func IntegrationTest(t *testing.T) {
+	if !*integrationTest {
+		t.SkipNow()
+	}
+	if *concurrent {
+		t.Parallel()
+	}
+}
+
+// UnitTest will run the test its called from iff the `-unit` or `-short` flag
+// is passed when calling `go test`. Otherwise the test will be skipped. UnitTest
+// will run the test its called from in parallel, to disallow this behaviour see
+// the concurrent flag.
+func UnitTest(t *testing.T) {
+	if !*unitTest && !testing.Short() {
+		t.SkipNow()
+	}
+	if *concurrent {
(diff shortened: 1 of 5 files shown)


testhelpers/testflags/flags.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// Test enablement flags
+var functionalTest = flag.Bool("functional", false, "Run the functional go tests")
+var integrationTest = flag.Bool("integration", false, "Run the integration go tests")
+var unitTest = flag.Bool("unit", false, "Run the unit go tests")
+
+// Test modifiers
+// concurrent flag will cause tests to be ran in parallel, its default value is `true`.
+var concurrent = flag.Bool("concurrent", true, "Runs the go test in parallel")
```
anorth: The built in `-parallel=1` flag can be used to disable concurrency, so I don't think this is needed. https://golang.org/cmd/go/#hdr-Testing_flags

testhelpers/testflags/flags.go
```diff
@@ -0,0 +1,64 @@
...
+	if *concurrent {
+		t.Parallel()
+	}
+}
+
+// UnitTestWithSideEffectsThatIsBad will run the test its called from iff the
+// `-unit` or `-short` flag is passed when calling `go test`. Otherwise the test
+// will be skipped. UnitTestWithSideEffectsThatIsBad will run the test its called
+// serially. Tests that use this flag are bad an should feel bad.
+func UnitTestWithSideEffectsThatIsBad(t *testing.T) {
```
acruikshank: `BadUnitTestWithSideEffects`? I get what you're trying to do here, but the name reads like you're calling the side effects bad in a grammatically challenged way.

testhelpers/testflags/flags.go
frrist: Good point, I'll remove it.

PR #2544, opened 2019-04-11
Prefactor/chain traversal context check

rosalinekarr:
# Problem

The chain traversal iterator can potentially take a very long time, but it does not halt on a cancelled context.

# Solution

Change traversal's `Next()` method to halt and return an error on a cancelled context.

This PR was originally part of #2422, but @anorth [suggested](https://github.com/filecoin-project/go-filecoin/pull/2422#discussion_r274221952) moving it to a separate PR for easier review.


anorth:
Thanks!

PR #2546, opened 2019-04-11
message pool max size config based

frrist:
This PR is a follow on to https://github.com/filecoin-project/go-filecoin/pull/2522, it allows the message pools max size to be set via the configuration file.

### DIFF
--- a/core/message_pool_test.go
+++ b/core/message_pool_test.go
@@ -25,7 +25,7 @@ func TestMessagePoolAddRemove(t *testing.T) {
 	assert := assert.New(t)
 	ctx := context.Background()
 
-	pool := NewMessagePool(th.NewTestMessagePoolAPI(0), th.NewTestMessagePoolConfig(10), th.NewMockMessagePoolValidator())
+	pool := NewMessagePool(th.NewTestMessagePoolAPI(0), config.NewDefaultConfig().Mpool, th.NewMockMessagePoolValidator())
 	msg1 := newSignedMessage()
 	msg2 := mustSetNonce(mockSigner, newSignedMessage(), 1)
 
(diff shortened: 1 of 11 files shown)


node/node.go
```diff
@@ -403,7 +403,7 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 
 	// only the syncer gets the storage which is online connected
 	chainSyncer := chain.NewDefaultSyncer(&cstOffline, nodeConsensus, chainStore, fetcher)
-	msgPool := core.NewMessagePool(chainStore, consensus.NewIngestionValidator(chainStore))
+	msgPool := core.NewMessagePool(chainStore, nc.Repo.Config().Mpool, consensus.NewIngestionValidator(chainStore))
```
acruikshank: If we are going to add config to specify the maximum message pool size, we should probably also add a parameter for the maximum nonce gap and pass it into the `IngestionValidator`. This could be done in this PR or added as an issue.

core/message_pool_test.go
```diff
@@ -65,31 +65,33 @@ func TestMessagePoolValidate(t *testing.T) {
 		require := require.New(t)
 		assert := assert.New(t)
 
+		// pull the default size from the default config value
+		maxMessagePoolSize := config.NewDefaultConfig().Mpool.MaxPoolSize
 		ctx := context.Background()
-		pool := NewMessagePool(th.NewTestMessagePoolAPI(0), th.NewMockMessagePoolValidator())
+		pool := NewMessagePool(th.NewTestMessagePoolAPI(0), th.NewTestMessagePoolConfig(maxMessagePoolSize, types.Uint64(10)), th.NewMockMessagePoolValidator())
```
anorth: The test config construction generates a lot of noise. I recommend:
- move `NewMessagePoolConfig` next to the struct definition (drop `Test`), and do the `Uint64` conversion inside it.
- if you want, add `NewConfig()` as a private function here, with no parameters.

PR #2547, opened 2019-04-11
add integration test flag to all integration tests

frrist:
- rename tests that use a daemon
- move daemon tests to commands_test package
- fixup messy import paths


commands/bitswap_stats_daemon_test.go
```diff
@@ -1,14 +1,16 @@
-package commands
```
frrist: this test forks, reflecting that with the name change

commands/init_test.go
```diff
@@ -1,34 +0,0 @@
-package commands
```
frrist: moved the testing logic to `init_daemon_test.go` since this forks

commands/main_test.go
```diff
@@ -28,28 +20,3 @@ func TestRequiresDaemon(t *testing.T) {
 	assert.True(requiresDaemon(reqWithDaemon))
 	assert.False(requiresDaemon(reqWithoutDaemon))
 }
-
-func TestNoDaemonNoHang(t *testing.T) {
```
frrist: moved to `main_daemon_test.go`

tools/fast/process_stderr_recorder_test.go
```diff
@@ -9,10 +9,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	mockplugin "github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin/mock"
 )
 
 func TestStartLogCapture(t *testing.T) {
+	tf.IntegrationTest(t)
```
travisperson: I don't know if I'd call any of these FAST tests integration tests. They are unit tests for FAST. They use a mock plugin which doesn't actually start any external processes, etc.

travisperson: At least they shouldn't be bundled with the daemon tests.

travisperson: nm, lets leave this as is. If it becomes a problem we can figure it out later.

frrist: I didn't read into these test too far when making the changes, if these don't fork I think they should be unit tests, thoughts?

anorth: I have previously implied that forking automatically excludes something from being a unit test, but that is too harsh. If the forked process does almost no work and/or communication and the whole test completes in 10s of milliseconds then it can remain a unit test. But if the forked process does work like initialising a node, it's probably an integration test. It's a judgement call.

PR #2549, opened 2019-04-12
consume new multiple sector sizes API

### LINKED ISSUE #585: encode token amounts in scientific notation
Token amounts (AttoFil) are currently encoded as an integer number of units, each of which represents 10**-18 FIL. Since most values aren't going to be that small it makes sense to serialize to scientific notation to save space. Note that this issue does not refer to changing the AttoFil representation which is a big.Int quantity of attofilecoin. This issue refers to changing the serialized format to be a (varint-encoded signficand, uint8 exponent) pair where the value represented is equal to `significand * 10 ** exponent` units (where each unit is still 10**-18). You basically just clear trailing zeros of the number to be serialized and add them to the exponent (that is, while the quantity mod 10 is 0, divide it by ten and add one to the exponent).

https://docs.google.com/document/d/12xNPzVCPSC2bTv7myxNRoMGx79AqFHQb89LfmRcFpGc/edit#bookmark=id.ufk10xbo7xb

### LINKED ISSUE #2530: A miner selects a sector size when pledging

laser:

~Blocked on [this upstream rust-fil-proofs PR](https://github.com/filecoin-project/rust-fil-proofs/pull/600/files).~

## What's in this PR?

- verify_seal and verify_post accept sector size through FFI (set via proofs mode)
- init_sector_builder accepts sector class (also set via proofs mode)

## rust-fil-proofs commits in this submodule update

```
-----------------------------------------------
commit d80fae26ecbb2d8ad7149de6f955ec69f2099d46
Author: Erin Swenson-Healey <l@s3r.me>
Date:   Sun Apr 14 16:47:53 2019 -0700

    support sector class + dynamic PoSt and PoRep configuration (#600)
    
    * introduce sector class
    
    * remove now-dead ProofsMode enum
    
    * both test and live PoRep configurations use 2 partitions


-----------------------------------------------
commit 502c3d218b6aef72f768768df59a920cacd9ac2b
Author: B YI <uuuuuu@protonmail.com>
Date:   Thu Apr 11 18:44:21 2019 +0800

    feat(config): make NUM_PROVING_THREADS configurable
    
    Closes #585
```


types/sector_size.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+// SectorSize is the amount of bytes in a sector. This amount will be slightly
+// greater than the number of user bytes which can be written to a sector due to
+// bit-padding.
+type SectorSize uint64
+
+const (
+	OneKiBSectorSize = SectorSize(iota)
+	TwoHundredFiftySixMiBSectorSize
```
acruikshank: QuarterGiBSectorSize? SectorSize265MiB? Spelling out 256 seems overly verbose.

actor/builtin/miner/miner.go
```diff
@@ -463,9 +463,11 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		// It is undefined behavior for a miner using "LiveMode" to verify a
 		// proof created by a miner using "TestMode" (and vice-versa).
 		//
-		proofsMode := proofs.LiveMode
+		var sectorSize types.SectorSize
 		if os.Getenv("FIL_USE_SMALL_SECTORS") == "true" {
-			proofsMode = proofs.TestMode
+			sectorSize = types.OneKiBSectorSize
+		} else {
```
acruikshank: non-blocking: Does this conditional go away once the miner has sector size in state?

types/sector_size.go
laser: Our names will get better once we figure out what the LCD of all current and future supported sector sizes. I promise.

actor/builtin/miner/miner.go
laser: Yep. The proofs-mode switching will happen only in the `StorageMarketActor#CreateMiner` method (to ensure that the selected sector size is a valid option for the currently-configured mode). Also, the switching will happen on a property _of_ the `StorageMarketActor` instead of an environment variable.

proofs/interface.go
```diff
@@ -1,23 +1,27 @@
...
-	ProverID   [31]byte  // uniquely identifies miner
-	SectorID   [31]byte  // uniquely identifies sector
-	ProofsMode Mode      // used to control sealing/verification performance
+	CommD      types.CommD     // returned from seal
+	CommR      types.CommR     // returned from seal
+	CommRStar  types.CommRStar // returned from seal
+	Proof      types.SealProof // returned from seal
+	ProverID   [31]byte        // uniquely identifies miner
+	SectorID   [31]byte        // uniquely identifies sector
+	SectorSize types.SectorSize
```
ZenGround0: Is size the only aspect of sectorClass that matters for verifying seals?

proofs/interface.go
```diff
@@ -1,23 +1,27 @@
...
 type VerifyPoSTRequest struct {
-	ChallengeSeed PoStChallengeSeed
-	CommRs        []CommR
+	ChallengeSeed types.PoStChallengeSeed
+	CommRs        []types.CommR
 	Faults        []uint64
-	Proofs        []PoStProof
-	ProofsMode    Mode // used to control sealing/verification performance
+	Proofs        []types.PoStProof
+	SectorSize    types.SectorSize
```
ZenGround0: Why not pass SectorClass in here?  Is it not the case that PoSt verification requires information on the other aspects of sector class?

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -77,8 +78,9 @@ type RustSectorBuilderConfig struct {
 	MetadataDir      string
 	MinerAddr        address.Address
 	SealedSectorDir  string
-	ProofsMode       proofs.Mode
+	SectorStoreType  proofs.Mode
```
ZenGround0: nit: why make this name change?  To me the previous name was easier to understand.

types/commitments.go
```diff
@@ -13,7 +11,40 @@ func init() {
...
 // when sealing a sector.
 type Commitments struct {
-	CommD     proofs.CommD
-	CommR     proofs.CommR
-	CommRStar proofs.CommRStar
+	CommD     CommD
+	CommR     CommR
+	CommRStar CommRStar
 }
+
```
ZenGround0: As a go-filecoin person I am in love with this refactoring of abstractions.

types/sector_class.go
```diff
@@ -0,0 +1,25 @@
+package types
+
+// SectorClass represents the miner's chosen sector size and PoSt/seal proof
+// partitions.
+type SectorClass struct {
```
ZenGround0: I'm feeling the love for this abstraction.

types/sector_class.go
```diff
@@ -0,0 +1,25 @@
...
+	PostProofPartitions PoStProofPartitions
+	SealProofPartitions SealProofPartitions
+	SectorSize          SectorSize
```
ZenGround0: Question from curiosity: do you imagine a PoStClass to capture things like number of sectors aggregated under a post?  Is that stuff totally separate from sector class?

types/sector_size.go
```diff
@@ -1,6 +1,6 @@
 package types
 
-// SectorSize is
+// sectorSize is
```
ZenGround0: nit: we are dying to know what it is

laser: It just _is_, man! Nice catch ;)

types/sector_class.go
laser: Nice!

proofs/interface.go
laser: PoSt verification will require the sector size only. The number of PoSt proof partitions is a function of the number of bytes in the provided proof. Today, there are only 192-byte PoSt proofs, which corresponds to PoSt proof partitions = 1. As part of the multiple sector sizes epic, we'll make the length of a PoSt (and PoRep) proof variable.

proofs/interface.go
laser: Yep. The PoRep partitions can be derived from the number of bytes in the provided PoRep/seal proof.

proofs/sectorbuilder/rustsectorbuilder.go
laser: Oops. That was a good catch. `ProofsMode` is no longer a field of the `RustSectorBuilderConfig` struct.

types/sector_class.go
laser: @ZenGround0 there's no direct relationship between the number of sectors which are challenged during PoSt generation and the number of PoSt proof partitions. 

So, 

> Is that stuff totally separate from sector class?

Yep, that stuff is separate from sector class.

Issue #1976, opened 2019-02-15
Skip very-slow proofs tests by default, run occasionally on CI


anorth:
### Description
Recently-added tests under `proofs/sectorbuilder/testing` are very slow, taking some minutes to complete. Running these as part of standard developer workflows with `go test` is a drag on getting PRs ready for review. Running these on every CI build adds latency to CI results which block merging of PRs, and can clog up the backlog.

A vast majority of code changes are very unlikely to affect these tests. Let's exclude by default from developer workflows and CI, with a flag to opt-in. We should run them on CI occasionally.

### Acceptance criteria
- Very slow proofs tests are excluded from `go ... best` and default invocations of `go test ./...`, and these invocations take <120s on a representative dev laptop
- The same tests are excluded from default CI builds
- There's a mechanism to opt-in to the excluded tests
- A new CI workflow exercises these tests once every few hours

### Risks + pitfalls
- The CI workflow to test these adds a little operational complexity
- Developers working on code that may break these tests must be aware of the need to opt-in to them.
- Don't use the existing `-short` flag to `go test`; we'll reserve this for opt-in skipping of only moderately slow tests

### Where to begin
See https://github.com/filecoin-project/go-filecoin/blob/master/functional-tests/faucet_test.go for another place we've flagged some tests out by default (but that flag is set for CI every time).

phritz:
note this issue is already in progress, rkowalick is just out sick

PR #2558, opened 2019-04-12
Feat/instrument unit testflags

frrist:
### Description
This PR adds unit and functional test flags to the appropriate tests

### DIFF
--- a/proofs/sectorbuilder/testing/interface_test.go
+++ b/proofs/sectorbuilder/testing/interface_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 )
 
 // MaxTimeToSealASector represents the maximum amount of time the test should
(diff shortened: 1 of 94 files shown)


proofs/sectorbuilder/testing/interface_test.go
```diff
@@ -28,6 +29,9 @@ const MaxTimeToSealASector = time.Second * 360
 const MaxTimeToGenerateSectorPoSt = time.Second * 360
 
 func TestSectorBuilder(t *testing.T) {
+	tf.FunctionalTest(t)
+
+	// TODO DO NOT MERGE THERE BE DRAGONS!! WHAT THIS BE?
```
travisperson: _Comment in reference to the following code (for discussion so you don't have to scroll so far :-)_
```go
if os.Getenv("FILECOIN_RUN_SECTOR_BUILDER_TESTS") != "true" {
	t.SkipNow()
}
```

I remember the reason for this was to skip this test in CI and have it run instead on a schedule due to it's very expensive execution time (issue #1976). I think the question is do we want to keep this pattern or extend the the flag pattern in some way.

frrist: Thanks for pointing this out. To remain consistent with other tests I'd like to replace the env var with a flag. Maybe something like `SectorTest` or `ProofsTest`, then [modify](https://github.com/filecoin-project/go-filecoin/blob/2bfb648484544f2a3b3f3f2e8afbe3d26a6fb2a1/.circleci/config.yml#L233) CI runs to work with it as it does now.

anorth: I thought that existing functional tests already run only occasionally, but may be misremembering. I don't think we need a configuration that doesn't run in dev workflows but does run always in CI. Is that what `-functional` is now? I think we want doesn't run in dev workflows or regular CI, but does run occasionally. We could repurpose the functional flag for that. Let's get rid of the env vars.

Don't block this PR on figuring that out, let's discuss elsewhere.

ognots:
Awesome work!
I only have one request - can the final commit message describe how this changes the invocation of tests. 
Also maybe some documentation in the README?

frrist:
@ognots I have updated the commit message to better reflect the changes. As the code stands now tests still run the same, in a follow on PR I will update the docs and changes the flags so they are not all on by default.

PR #2565, opened 2019-04-15
Features/auto seal interval command 2394

### LINKED ISSUE #2394: surface auto seal interval in plumbing and with a command
we should surface auto seal setting in plumbing, eg in a protocolparams plumbing call like https://github.com/filecoin-project/go-filecoin/issues/1932. all params can just be returned in one struct for now

rosalinekarr:
# Problem

We want some protocol parameters surfaced in plumbing and commands.

# Solution

Surface them.

Resolves #2394

### DIFF
--- a/plumbing/cfg/config.go
+++ b/plumbing/cfg/config.go
@@ -2,7 +2,6 @@ package cfg
 
 import (
 	"github.com/filecoin-project/go-filecoin/repo"
-	"github.com/pkg/errors"
 	"sync"
 )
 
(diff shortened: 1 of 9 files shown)


plumbing/cfg/config.go
```diff
@@ -33,3 +34,23 @@ func (s *Config) Set(dottedKey string, jsonString string) error {
 func (s *Config) Get(dottedKey string) (interface{}, error) {
 	return s.repo.Config().Get(dottedKey)
 }
+
+type ProtocolParams struct {
+	AutoSealInterval uint
+}
+
+func (s *Config) ProtocolParams() (*ProtocolParams, error) {
```
acruikshank: BLOCKING: This should be porcelain

Issue #2538, opened 2019-04-10
Move FAST retrieval test to functional-tests, delete shell test


travisperson:
### Description

I believe the test written in [`tools/fast/tests/retrieval_test.go`](https://github.com/filecoin-project/go-filecoin/blob/master/tools/fast/tests/retrieval_test.go) covers the same surface area as the shell script we current have. I think FAST is at a place we can remove the shell script and helper library.

### Acceptance criteria

- Remove current bash retrieval script
- Move FAST retrieval test into `functional-tests` directory

### Risks + pitfalls

- Risk of loosing test coverage if the tests differ wild in execution. We should ensure the functionality is the same prior to doing this work.

### Where to begin

- https://github.com/filecoin-project/go-filecoin/tree/master/functional-tests
- https://github.com/filecoin-project/go-filecoin/tree/master/tools/fast/tests

anorth:
Go go go!

anorth:
@travisperson I sensed some resistance when removing the shell test came up in discussion. Can we address that here? In particular it seems that a few people use the functional test frequently as a sanity check. While in the long term I think that's an unhealthy habit (unit and integration tests should provide sufficient confidence), we should retain support for that use case.

I expect that the issue is that people don't know how to run the FAST version or interpret its output. Will it be as simple as `go test -functional`? Can you speak to how output can be inspected in the FAST version vs the shell version?

If anyone has other objections, please raise them here now.

laser:
I use this retrieval test when developing locally. I also use this retrieval test to validate downloaded go-filecoin binaries after a release (local network), and I use it (with some modifications) to test our devnets too.

I'd be quite bummed out to lose it, but I am aware that the software that makes up the "retrieval test" has a carrying cost. I've seen a very high ROI on that software to-date (which is to say: it's caught a lot of bugs during development and hasn't required a lot of maintenance).

As a compromise: I can agree to try supplant my usage of the retrieval script for the next couple of weeks, replacing it with the FAST retrieval test. If after a couple of weeks I haven't found any cause for complaint (e.g. it's not difficult to see the log streams from the various go-filecoin nodes being tested, errors aren't swallowed, etc.) then I will support your efforts to delete my beloved retrieval test.

How's that sound?

laser:
cc @acruikshank @rosalinekarr @rkowalick @shannonwells ^

rosalinekarr:
[...]

PR #2567, opened 2019-04-15
Feat/ognots/testing docs flags

ognots:
implements changes based on https://github.com/filecoin-project/go-filecoin/pull/2564#issuecomment-483392661

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,12 +53,21 @@ commands:
   go_test:
     parameters:
       <<: *sector_builder_tests_param
+      unit:
+        type: string
+        default: "true"
+      integration:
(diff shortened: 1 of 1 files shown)


ognots:
I rebased this change adding back the `functional-tests/run` script until https://github.com/filecoin-project/go-filecoin/issues/2538 lands and the script can be fully deprecated

frrist:
@ognots does this add the parallelization we were discussing earlier?

ognots:
@frrist no this doesn't add the parallelization. it only makes the circleci `go_test` command parameterized so it will support parallelization (in an incoming separate PR).

the parameterization is the only change in this PR

PR #2575, opened 2019-04-15
Client and Miner Rejects Pieces Over Sector Size

### LINKED ISSUE #2034: Client and miner should reject attempts to store a piece larger than sector size

rkowalick:
Resolves #2034

### DIFF
--- a/commands/client_daemon_test.go
+++ b/commands/client_daemon_test.go
@@ -1,6 +1,7 @@
 package commands_test
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 	"testing"
@@ -195,4 +196,37 @@ func TestVoucherPersistenceAndPayments(t *testing.T) {
 	assert.Contains(result, "0\t240000\t10")
 	assert.Contains(result, "0\t480000\t20")
 	assert.Contains(result, "0\t720000\t30")
+}
+
+func TestPieceRejectionInProposeStorageDeal(t *testing.T) {
+	tf.IntegrationTest(t)
+
+	assert := assert.New(t)
+	minerDaemon := th.NewDaemon(t,
+		th.WithMiner(fixtures.TestMiners[0]),
+		th.KeyFile(fixtures.KeyFilePaths()[0]),
+		th.DefaultAddress(fixtures.TestAddresses[0]),
+		th.AutoSealInterval("1"),
+	).Start()
+	defer minerDaemon.Shutdown()
+
+	clientDaemon := th.NewDaemon(t,
+		th.KeyFile(fixtures.KeyFilePaths()[1]),
+		th.DefaultAddress(fixtures.TestAddresses[1]),
+	).Start()
+	defer clientDaemon.ShutdownSuccess()
+
+	minerDaemon.RunSuccess("mining", "start")
+	minerDaemon.UpdatePeerID()
+
+	minerDaemon.ConnectSuccess(clientDaemon)
+
+	addAskCid := minerDaemon.MinerSetPrice(fixtures.TestMiners[0], fixtures.TestAddresses[0], "20", "10")
+	clientDaemon.WaitForMessageRequireSuccess(addAskCid)
+
+	dataCid := clientDaemon.RunWithStdin(bytes.NewReader(make([]byte, 3000)), "client", "import").ReadStdoutTrimNewlines()
+
+	proposeDealErrors := clientDaemon.Run("client", "propose-storage-deal", fixtures.TestMiners[0], dataCid, "0", "5").ReadStderr()
+
+	assert.Contains(proposeDealErrors, "piece is 3000 bytes but sector size is 1016 bytes")
 }
--- a/protocol/storage/client_test.go
+++ b/protocol/storage/client_test.go
@@ -168,7 +168,7 @@ func (ctp *clientTestAPI) CreatePayments(ctx context.Context, config porcelain.C
 }
 
 func (ctp *clientTestAPI) DAGGetFileSize(context.Context, cid.Cid) (uint64, error) {
-	return 1000000000, nil
+	return 1016, nil
 }
 
 func (ctp *clientTestAPI) MinerGetAsk(ctx context.Context, minerAddr address.Address, askID uint64) (miner.Ask, error) {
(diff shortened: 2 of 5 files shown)


protocol/storage/miner.go
```diff
@@ -890,3 +899,17 @@ func (sm *Miner) handleQueryDeal(s inet.Stream) {
 		log.Errorf("failed to write query response: %s", err)
 	}
 }
+
+func (sm *Miner) getSectorSize(ctx context.Context) (uint64, error) {
```
rkowalick: dry this

protocol/storage/miner_test.go
```diff
@@ -491,12 +523,12 @@ func testPaymentVouchers(porcelainAPI *minerTestPorcelain, voucherInterval int,
...
 }
 
-func testSignedDealProposal(porcelainAPI *minerTestPorcelain, vouchers []*paymentbroker.PaymentVoucher, addr address.Address) *storagedeal.SignedDealProposal {
+func testSignedDealProposal(porcelainAPI *minerTestPorcelain, vouchers []*paymentbroker.PaymentVoucher, size uint64) *storagedeal.SignedDealProposal {
 	proposal := &storagedeal.Proposal{
 		MinerAddress: porcelainAPI.targetAddress,
 		PieceRef:     types.NewCidForTestGetter()(),
-		TotalPrice:   types.NewAttoFILFromFIL(2500),
-		Size:         types.NewBytesAmount(1000),
+		TotalPrice:   types.NewAttoFILFromFIL(size / 1000 * 2500),
```
rkowalick: this math isn't quite right

protocol/storage/client_test.go
```diff
@@ -250,3 +250,7 @@ func (ctp *clientTestAPI) DealPut(storageDeal *storagedeal.Deal) error {
 	ctp.deals[storageDeal.Response.ProposalCid] = storageDeal
 	return nil
 }
+
+func (ctp *clientTestAPI) MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
+	return [][]byte{{0}}, nil
```
rkowalick: use `TestMode` here

commands/client_daemon_test.go
```diff
@@ -196,3 +197,40 @@ func TestVoucherPersistenceAndPayments(t *testing.T) {
...
+	minerDaemon.RunSuccess("mining", "start")
+	minerDaemon.UpdatePeerID()
+
+	minerDaemon.ConnectSuccess(clientDaemon)
+
+	addAskCid := minerDaemon.MinerSetPrice(fixtures.TestMiners[0], fixtures.TestAddresses[0], "20", "10")
+	clientDaemon.WaitForMessageRequireSuccess(addAskCid)
+	var byteSlice []byte
+	for i := 0; i < 3000; i++ {
+		byteSlice = append(byteSlice, byte('a'))
```
acruikshank: Nit: Since the data is unimportant, you could replace this loop with `byteSlice := make([]byte, 3000)`

commands/client_daemon_test.go
```diff
@@ -196,3 +197,40 @@ func TestVoucherPersistenceAndPayments(t *testing.T) {
...
+	}
+
+	dataCid := clientDaemon.RunWithStdin(bytes.NewReader(byteSlice), "client", "import").ReadStdoutTrimNewlines()
+
+	proposeDealOutput := clientDaemon.Run("client", "propose-storage-deal", fixtures.TestMiners[0], dataCid, "0", "5").ReadStdoutTrimNewlines()
+
+	assert.NotContains(proposeDealOutput, "accepted")
```
acruikshank: This is a weak assertion. Is it possible to recover the error or at least check for "rejected"?

protocol/storage/miner_test.go
```diff
@@ -491,12 +523,12 @@ func testPaymentVouchers(porcelainAPI *minerTestPorcelain, voucherInterval int,
...
 }
 
-func testSignedDealProposal(porcelainAPI *minerTestPorcelain, vouchers []*paymentbroker.PaymentVoucher, addr address.Address) *storagedeal.SignedDealProposal {
+func testSignedDealProposal(porcelainAPI *minerTestPorcelain, vouchers []*paymentbroker.PaymentVoucher, size uint64) *storagedeal.SignedDealProposal {
 	proposal := &storagedeal.Proposal{
 		MinerAddress: porcelainAPI.targetAddress,
 		PieceRef:     types.NewCidForTestGetter()(),
-		TotalPrice:   types.NewAttoFILFromFIL(2500),
-		Size:         types.NewBytesAmount(1000),
+		TotalPrice:   types.NewAttoFILFromFIL(2500 * uint64(math.Ceil(float64(size)/float64(1016)))),
```
acruikshank: Why is TotalPrice = 2500 * size / sector size (please add a comment)? Also 1016 should be a variable named sectorSize instead of magic number. Also (2500*size)/1016 is more accurate and more concise unless you need it to be a multiple of 2500 for some reason.

protocol/storage/miner_test.go
```diff
@@ -52,7 +53,7 @@ func TestReceiveStorageProposal(t *testing.T) {
 		}
 
 		vouchers := testPaymentVouchers(porcelainAPI, VoucherInterval, defaultAmountInc)
-		proposal := testSignedDealProposal(porcelainAPI, vouchers, porcelainAPI.targetAddress)
+		proposal := testSignedDealProposal(porcelainAPI, vouchers, 1000)
```
acruikshank: Nit: It would be good to add a `defaultSize` constant in this file rather than all these magic numbers.

protocol/storage/miner_test.go
rkowalick: I changed the derivation of total price to have no more magic numbers, since there aren't supposed to be any.

Issue #2524, opened 2019-04-09
Separate staging and sealed sectors from node repo


anorth:
### Description
At present, all on-disk storage for a filecoin node is under the "repo" directory, `~/.filecoin`. This directory contains config, structured stores of block data, and subdirectories `staging` and `sealed` for client data. The staging/sealed data is quite different to the protocol store data and the internals are managed by code in `rust-fil-proofs` which is simply given directory paths to use.

We are currently designing a migration scheme for node's repo data and will likely implement revertible migrations by operating on a copy of the repo. Migrating the sector storage is something we'd like to call out of scope for the immediate efforts since:
- its schema is unlikely to change as frequently
- it's largely orthogonal to the data core to the protocol
- it has potential to be large enough to prohibit copying as a rollback strategy
- we expect large-scale miners to customise sector storage to their architecture, but more likely use our protocol implementation

If we imagine a future separation between a library responsible for the protocol, and a sector builder responsible for storage and proofs, this separation would also be reflected in independent data storage schemes.

Thus, I propose we move the staging and sector directories out of the repo directory, so the repo directory is specialised to storage related directly to the protocol and node operation. 

Concretely, this could be a `~/.filecoin-sectors` directory, or we could nest all the protocol data at `~/.filecoin/protocol` and call that the "repo", but other suggestions welcome.

@whyrusleeping @laser we'd appreciate any thoughts you have here.

### Acceptance criteria
- Staging and sealed sectors are stored outside the directory tree containing protocol-related node storage
- Sector storage is split from the Repo interface
- The location of the repo and the sector store are configurable (the sector paths may be configured by the config contained inside the repo)

### Risks + pitfalls
Implementing this change will require some manual intervention by any node that wishes to retain sectors across the update. These actions can be performed in a shell (which is a property we value for the storage migration processes in general).

### Where to begin
Discuss here to determine any unwanted side-effects of this split. Implementation begins in https://github.com/filecoin-project/go-filecoin/tree/master/repo

laser:
The rust-fil-proofs `SectorBuilder` is created with two paths. One of the paths corresponds to the sealed sector directory and the other to the staged sector directory. I don't see any problems with moving those directories out of the "repo" directory. We'll need to augment some tests with additional teardown code; many tests clean up by deleting their "repo" directory.

> The location of the repo and the sector store are configurable (the sector paths may be configured by the config contained inside the repo)

[...]

PR #2576, opened 2019-04-16
Spike/mfc repos:  build code, basic CLI skeleton, first migration stubs, logger stubs, layout

### LINKED ISSUE #2582: Repo migrations can be built and installed using go-filecoin build command
### Description
Repo migrations need a build and possibly an install command.

### Acceptance criteria
`go run build/main.go build-migrations` builds the migration tool in `tools/migration` and outputs the binary as `go-filecoin-migrate`

All migrations should be included in the new binary.  The binary should be executable from the command line without having to run chmod, and should exit normally.

Optionally:
`go run build/main.go install-migrations` installs the migration tool in `$GOPATH/bin`

### Risks + pitfalls
none known; should be done like other build targets.

### Where to begin
build/main.go

### LINKED ISSUE #2589: Spike on repo migrations
### Description
Issue for the current spike for figuring out basics of repo migration code: layout, relationships, etc.

shannonwells:
Closes #2589 Spike on repo migrations
Also closes #2582 Repo migrations can be built and installed with the go-filecoin builder
Also foundational work for #2581 CLI for repo migrations and unit testing.

Subject to later change, this sets up the basis for completing the first repo migration, a version bump.


build/main.go
```diff
@@ -247,6 +253,8 @@ func main() {
 		buildGengen()
 	case "generate-genesis":
 		generateGenesis()
+	case "build-migrations":
+		buildMigrations()
```
anorth: Please roll this into `build` so that migrations are always built. Similarly I expect their tests to be always run.

go.mod
```diff
@@ -26,10 +20,10 @@ require (
 	github.com/ipfs/go-ipfs-blockstore v0.0.1
 	github.com/ipfs/go-ipfs-chunker v0.0.1
 	github.com/ipfs/go-ipfs-cmdkit v0.0.1
-	github.com/ipfs/go-ipfs-cmds v0.0.1
+	github.com/ipfs/go-ipfs-cmds v0.0.5
 	github.com/ipfs/go-ipfs-exchange-interface v0.0.1
 	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
-	github.com/ipfs/go-ipfs-files v0.0.1
+	github.com/ipfs/go-ipfs-files v0.0.2
```
anorth: This is a surprisingly large diff and these version bumps look independent? To the extent that it's reasonable, please land these upgrades separately.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/filecoin-project/go-filecoin/tools/migration/internal"
+	"github.com/filecoin-project/go-filecoin/tools/migration/migrate_1-to-2"
+)
+
+type Migrator interface {
+	Migrate(oldRepo, newRepo *os.File) error
+	Describe()
+	Validate(oldRepo, newRepo *os.File) error
+}
```
anorth: Blocking: document this interface extensively, including the semantics of the old and new repos. I am still confused about why two repos are provided to `Migrate`: if the new repo is non-empty, then it should be migrated in place and have no opportunity to mutate the original. Otherwise, is the migration responsible for doing the copy?

The name `Migrator` is a bit misleading, not clear whether it's the orchestrator or the migration logic. I suggest `Migration` instead.

I expect `Describe()` to return a string. If not, provide an output writer explicitly rather than assuming anything.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Run runs the given migration command
+func Run(cmd string, verbose bool) error {
+	var err error
+	miglog, err := makeMigl()
+	logpath := "/tmp/migration_log.txt"
```
anorth: Blocking: turn this run method into a `MigrationRunner` struct and extract all these paths as member fields, configurable at construction.

It is confusing that as a file in the `cmd` package this looks like a command, but is in fact a router to other commands. I like the idea of identifying the commands clearly, but in this case they probably map to simple method calls on `Migrator`. This file might be better named `migration/migrator.go`.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+
+	oldRepo, err := getOldRepo()
+	if err != nil {
+		miglog.Error(err)
+		return err
+	}
+
+	// open repodir read-only
+	// create new repodir & open read-write
+	// TODO: decide where to put the temp new repo.
```
anorth: Don't merge TODO without issues. In this case I propose extracting the constant to a field/parameter and requiring it to be specified explicitly for now. My understanding from #2524 is that this will be inferred from the containing directory and involve a timestamp and version number.

One way to go here would be to pass the container `.filecoin` directory here and give this method the responsibility of computing a subdirectory name for the copy.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+		miglog.Error(err)
+		return err
+	}
+
+	var migrator Migrator
+	if err != nil {
+		miglog.Error(err)
+		return err
+	}
+	// TODO see Issue #2585
```
anorth: Please give a one-sentence summary of the TODO here, which I guess is to inspect the repo to determine its version. I don't think we should land without that bit of logic being implemented.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	migrator = migrate_1_to_2.NewMigrator_1_2(miglog)
+
+	switch cmd {
+	case "describe":
+		migrator.Describe()
+	case "buildonly":
+		err = migrator.Migrate(oldRepo, newRepo)
```
anorth: Blocking: this should validate too. You could argue that `install` should do the validation, but one of them must.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return err
+	}
+	miglog.Print(fmt.Sprintf("Done.  Please see log for more detail: ", logpath))
+	return nil
+}
+
+// swap old repo for new
+func install(oldRepo, newRepo *os.File) error {
+	return nil
```
anorth: Document not yet implemented, and probably log that too.

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	var oldRepo *os.File
+	var oldpath string
+	filpath := os.Getenv("FIL_PATH")
+	if filpath == "" {
+		homeDir := os.Getenv("HOME")
+		oldpath = fmt.Sprintf("%s/.filecoin", homeDir)
+	} else {
+		oldpath = filpath
+	}
+	oldRepo, err := os.Open(oldpath)
```
anorth: Blocking: this must be configurable. Inject this environment lookup default from up in `main`.

tools/migration/main.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func main() { // nolint: deadcode
+	if len(os.Args) < 2 {
+		showUsageAndExit(1)
+	}
+
+	if !version.Check(runtime.Version()) {
+		exitErr(fmt.Sprintf("Invalid go version: %s", runtime.Version()))
+	}
```
anorth: Why do we need this?

tools/migration/cmd/go-filecoin-migrate.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return &os.File{}, err
+	}
+	return oldRepo, nil
+}
```
anorth: This file needs tests

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: I ask that this part be out of scope. This PR was meant to just be a foundation for implementing features more fully, and if we wait until everything is fully implemented then itw will take longer to split the work with others.  E.g. IMO the temp migration directory ought to be configurable too eventually, but for the first go-round we are aiming for simplicity.

tools/migration/main.go
shannonwells: This is done in build/main.go and I pulled it from there because I assumed we want to make sure we are running the version of go required by go-filecoin.

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: I'm confused between your objection to Migrator here vs your request for it below. I'd like to take the below suggestion to move it to migration/migrator.go, however that is where main.go is and go doesn't allow >1 package name at the same level.  So if you prefer moving it, I suggest moving it to `migration/internal/migrator.go` .   The pattern I was attempting to follow, somewhat, is this one: [golang-standards/project-layout](https://github.com/golang-standards/project-layout/tree/master/cmd):

> `/cmd`
> Main applications for this project.
> The directory name for each application should match the name of the executable you want to have (e.g., /cmd/myapp).

go.mod
shannonwells: This kept being updated after adding go.mod and go.sum in tools/migration, and a tag for the go-filecoin/tools/migration module.  I'll try to separate it as much as possible.

build/main.go
shannonwells: ✔️

go.mod
shannonwells: ✔️

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: @anorth I think this has been mostly addressed in #2614 with intention of adding more in a follow-up.

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: @anorth  in #2614 all this code has been ripped out and the file moved, but there is a stubbed-out MigrationRunner now for filling out in a follow-up PR.

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: see above

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: Never mind -- it's in #2614 , more or less.

tools/migration/main.go
shannonwells: left in the new PR

tools/migration/cmd/go-filecoin-migrate.go
shannonwells: will add in follow up to #2614 (these comments are also for my own tracking that I made requested changes)

PR #2602, opened 2019-04-17
add stat for oldest message in queue

### LINKED ISSUE #2147: Instrument metrics on message pool and outbound message queue
### Description
Add instrumentation to the message pool and message queue (#2146) and optionally post those metrics in heartbeat to the dashboards. Metrics should include 

- [x] queue size (addressed in https://github.com/filecoin-project/go-filecoin/pull/2407)
- [ ] age of oldest un-mined outbound
- [x] count of send failures (addressed in https://github.com/filecoin-project/go-filecoin/pull/2407)
- [x] count of queue timeouts (addressed in https://github.com/filecoin-project/go-filecoin/pull/2407)

### Acceptance criteria
- Message pool/queue metrics are accessible to a user (a new command?)
- Message pool metrics can be posted to dashboards

### Risks + pitfalls
None?

### Where to begin
- https://github.com/filecoin-project/go-filecoin/blob/master/core/message_pool.go
- https://github.com/filecoin-project/go-filecoin/blob/master/metrics/heartbeat.go

frrist:
closes #2147

### DIFF
--- a/core/message_queue.go
+++ b/core/message_queue.go
@@ -13,6 +13,7 @@ import (
 
 var (
 	mqSizeGa   = metrics.NewInt64Gauge("message_queue_size", "The size of the message queue")
+	mqOldestGa = metrics.NewInt64Gauge("message_queue_oldest", "The age of the oldest message in the queue or zero when empty")
 	mqExpireCt = metrics.NewInt64Counter("message_queue_expire", "The number messages expired from the queue")
 )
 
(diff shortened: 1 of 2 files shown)


core/message_queue.go
```diff
@@ -174,6 +183,30 @@ func (mq *MessageQueue) Size() int64 {
 	return l
 }
 
+// Oldest returns the oldest message stamp in the MessageQueue.
+// Oldest returns 0 if the queue is emptystarts
+func (mq *MessageQueue) Oldest() uint64 {
```
anorth: Don't export this method without a reason to. Testing is an ok reason, but document this with "Exported for testing".

Note that this is a very neat case where what we really want to test is that the metric has the right value. If the metric registry was injected, we could do this directly. Let this be a prompt to follow through with that work!

core/message_queue.go
```diff
@@ -174,6 +183,30 @@ func (mq *MessageQueue) Size() int64 {
...
+	var all []*QueuedMessage
+	for _, qm := range mq.queues {
+		for _, m := range qm {
+			all = append(all, m)
+		}
+	}
+
+	sort.Slice(all, func(i, j int) bool {
+		return all[i].Stamp < all[j].Stamp
+	})
```
anorth: This is way more work than necessary for finding the single smallest value. Just loop and maintain the least so far, perhaps in the named return value. The early exit for `Size() == 0` will then be more expensive than the normal case, I expect, so drop it.

PR #2603, opened 2019-04-17
Dockerfile genesis fix and tweaks

gmasgras:
- copy test and live genesis files to final image
- add build-time args (default to yes) for using precompiled rust-fil-proofs and bls-signatures
- reuse groth params if already downloaded under `$SRC_DIR/tmp/filecoin-proof-parameters
- don't build faucet and genesis-file-server since those have their own Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1.12.1-stretch AS builder
 MAINTAINER Filecoin Dev Team
 
-RUN apt-get update && apt-get install -y ca-certificates file sudo clang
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.12.1-stretch AS builder
 MAINTAINER Filecoin Dev Team
 
-RUN apt-get update && apt-get install -y ca-certificates file sudo clang
+RUN apt-get update && apt-get install -y ca-certificates file sudo clang jq
```
frrist: :pray:

PR #2607, opened 2019-04-17
support variable length PoSt and PoRep proofs

### LINKED ISSUE #2591: Support variable length PoSt and PoRep proofs
### Description

The miner's selection of PoSt proof partitions count and/or seal proof partitions count will impact the number of bytes in the proofs. The current `commitSector` and `submitPoSt` messages expect `types.SealProof` and `types.PoStProof` today, and those proofs' sizes are hard-coded to 192 and 384 bytes, respectively.

### Acceptance Criteria

1. `submitPoSt` and `commitSector` messages need to include one (or more, in the case of PoSt) proofs of type `[]byte`
1. The `generate_post` CGO call will need to be provided the configured `PoStProofPartitions` value

laser:
Blocked by [this upstream rust-fil-proofs PR](https://github.com/filecoin-project/rust-fil-proofs/pull/623).

Fixes #2591.

## Why's this PR needed?

Miners can choose proof partitions in order to trade off between RAM usage and on-chain proof size. What this means for go-filecoin is that the number of bytes in a PoSt or PoRep proof is no longer fixed.

## What's in this PR?

- Make PoSt and PoRep (seal) proofs `[]byte` instead of `[192]byte` and `[384]byte`
- **BREAKING CHANGE**: JSON encode the `types.PoStProof` in the `types.Block`, using [the default encoder](https://golang.org/pkg/encoding/json/#Marshal). This will produce a base64-encoded string (or null) instead of a JSON array.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -209,7 +209,7 @@ var minerExports = exec.Exports{
 		Return: []abi.Type{abi.SectorID},
 	},
 	"commitSector": &exec.FunctionSignature{
-		Params: []abi.Type{abi.SectorID, abi.Bytes, abi.Bytes, abi.Bytes, abi.Bytes},
+		Params: []abi.Type{abi.SectorID, abi.Bytes, abi.Bytes, abi.Bytes, abi.PoRepProof},
 		Return: []abi.Type{},
 	},
 	"getKey": &exec.FunctionSignature{
@@ -451,7 +451,7 @@ func (ma *Actor) GetSectorCommitments(ctx exec.VMContext) (map[string]types.Comm
 
 // CommitSector adds a commitment to the specified sector. The sector must not
 // already be committed.
-func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR, commRStar, proof []byte) (uint8, error) {
+func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR, commRStar []byte, proof types.PoRepProof) (uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -493,7 +493,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		copy(req.CommD[:], commD)
 		copy(req.CommR[:], commR)
 		copy(req.CommRStar[:], commRStar)
-		req.Proof = append(types.PoRepProof{}, proof...)
+		req.Proof = proof
 		req.ProverID = sectorbuilder.AddressToProverID(ctx.Message().To)
 		req.SectorID = sectorbuilder.SectorIDToBytes(sectorID)
 		req.SectorSize = sectorSize
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -711,7 +711,7 @@ func (ma *Actor) GetPower(ctx exec.VMContext) (*big.Int, uint8, error) {
 
 // SubmitPoSt is used to submit a coalesced PoST to the chain to convince the chain
 // that you have been actually storing the files you claim to be.
-func (ma *Actor) SubmitPoSt(ctx exec.VMContext, postProofs []types.PoStProof) (uint8, error) {
+func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -770,7 +770,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, postProofs []types.PoStProof) (u
 				ChallengeSeed: seed,
 				SortedCommRs:  sortedCommRs,
 				Faults:        []uint64{},
-				Proofs:        postProofs,
+				Proofs:        poStProofs,
 				SectorSize:    sectorSize,
 			}
 
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -493,7 +493,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		copy(req.CommD[:], commD)
 		copy(req.CommR[:], commR)
 		copy(req.CommRStar[:], commRStar)
-		copy(req.Proof[:], proof)
+		req.Proof = append(types.PoRepProof{}, proof...)
 		req.ProverID = sectorbuilder.AddressToProverID(ctx.Message().To)
 		req.SectorID = sectorbuilder.SectorIDToBytes(sectorID)
 		req.SectorSize = sectorSize
--- a/proofs/rustverifier.go
+++ b/proofs/rustverifier.go
@@ -52,15 +52,9 @@ func (rp *RustVerifier) VerifySeal(req VerifySealRequest) (VerifySealResponse, e
 	sectorIDCbytes := C.CBytes(req.SectorID[:])
 	defer C.free(sectorIDCbytes)
 
-	ppp, err := req.Proof.ProofPartitions()
-	if err != nil {
-		return VerifySealResponse{}, errors.Wrap(err, "failed to get proof partitions")
-	}
-
 	// a mutable pointer to a VerifySealResponse C-struct
 	resPtr := (*C.VerifySealResponse)(unsafe.Pointer(C.verify_seal(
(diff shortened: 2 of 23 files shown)


proofs/rustverifier.go
```diff
@@ -126,10 +127,7 @@ func (rp *RustVerifier) VerifyPoST(req VerifyPoSTRequest) (VerifyPoSTResponse, e
 	}
 
 	return VerifyPoSTResponse{
-		// TODO: change this to the bool statement
-		// See https://github.com/filecoin-project/go-filecoin/issues/1302
-		// bool(resPtr.is_valid),
-		IsValid: true,
+		IsValid: bool(resPtr.is_valid),
```
shannonwells: does this always return true or are we really validating the `PoSTResponse` now?

types/post_proof_partitions.go
```diff
@@ -1,8 +1,40 @@
...
-	TestPoStPartitions = PoStProofPartitions(iota)
-	OnePoStPartition
+	UnknownPoStProofPartitions = PoStProofPartitions(iota)
+	OnePoStProofPartition
 )
+
+// Int returns an integer representing the number of PoSt partitions
+func (p PoStProofPartitions) Int() int {
+	switch p {
+	case OnePoStProofPartition:
```
shannonwells: I assume this is because at some point >1 partition may be expected, but maybe include a comment to that extent so nobody is tempted to convert these all to if-elses.

types/proofs.go
```diff
@@ -0,0 +1,24 @@
...
+// PoStProof is the byte representation of the Proof of SpaceTime proof
+type PoStProof []byte
+
+// PoRepProof is the byte representation of the Seal Proof of Replication
+type PoRepProof []byte
+
+// ProofPartitions returns the number of partitions used to create the PoRep
+// proof.
+func (s PoRepProof) ProofPartitions() PoRepProofPartitions {
+	return NewPoRepProofPartitions(len(s) / SinglePartitionProofLen)
```
acruikshank: This should really fail here if `len(s)` is not `2*SinglePartitionProofLen`. Getting a `UnknownPoStProofPartitions` that's going to produce bad numbers somewhere else will just make it hard to track bugs.

Also, using integer math, this is going to produce:

* `UnknownPoStProofPartitions` for 0 <= len(s) < 384
* `TwoPoRepProofPartitions` for 384 <= len(s) < 576
* `UnknownPoStProofPartitions`  for 576 <= len(s) < ♾

Is this what you want? Is 500 bytes a valid TwoPoRepProofPartitions?

types/sector_class.go
```diff
@@ -21,8 +21,8 @@ func NewTestSectorClass() SectorClass {
 // go-filecoin node.
 func NewLiveSectorClass() SectorClass {
 	return SectorClass{
-		postProofPartitions:  OnePoStPartition,
-		poRepProofPartitions: TwoPoRepPartitions,
+		poStProofPartitions:  OnePoStProofPartition,
+		poRepProofPartitions: TwoPoRepProofPartitions,
```
acruikshank: The order changed from poRep to poSt to poSt to poRep here. It's a little confusing when comparing Test to Live.

actor/builtin/miner/miner.go
```diff
@@ -711,7 +711,7 @@ func (ma *Actor) GetPower(ctx exec.VMContext) (*big.Int, uint8, error) {
 
 // SubmitPoSt is used to submit a coalesced PoST to the chain to convince the chain
 // that you have been actually storing the files you claim to be.
-func (ma *Actor) SubmitPoSt(ctx exec.VMContext, postProofs []types.PoStProof) (uint8, error) {
+func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (uint8, error) {
```
ZenGround0: it's the little things

proofs/rustverifier.go
```diff
@@ -82,8 +79,16 @@ func (rp *RustVerifier) VerifySeal(req VerifySealRequest) (VerifySealResponse, e
...
+	// validate verification request
+	if len(req.Proofs) == 0 {
+		return VerifyPoSTResponse{}, errors.New("must provide at least one proof to verify")
+	}
+
+	// CommRs must be provided to C.verify_post in the same order that they were
+	// provided to the C.generate_post
 	commRs := req.SortedCommRs.Values()
+
+	// flattening the byte slice makes it easier to copy into the C heap
```
ZenGround0: @phritz nit: capitalize and period at the end.

actor/builtin/miner/miner.go
laser: It was driving me nuts

types/proofs.go
laser: I really wanted the type of this function to be `(s PoRepProof) ProofPartitions() PoRepProofPartitions` and not `(s PoRepProof) ProofPartitions() (PoRepProofPartitions, error)`. 

> Is 500 bytes a valid TwoPoRepProofPartitions?

The rust-fil-proofs system will reject (with an error) a 500 byte PoRep proof. The rust-fil-proofs system will never produce a 500 byte PoRep proof, either. By relying on the rust-fil-proofs system to reject the buggy proof, I could avoid duplicating the runtime validation into go-filecoin.

That said, it's not much software to enforce the invariant on both sides of the FFI wall. Would you agree to a `panic` in the event that we can't map the PoRepProof bytes cleanly to a PoRepProofPartitions value?

proofs/sectorbuilder/transforms.go
```diff
@@ -0,0 +1,94 @@
...
+// #cgo LDFLAGS: -L${SRCDIR}/../lib -lfilecoin_proofs
+// #cgo pkg-config: ${SRCDIR}/../lib/pkgconfig/libfilecoin_proofs.pc
+// #include "../include/libfilecoin_proofs.h"
+import "C"
+
+func goBytes(src *C.uint8_t, size C.size_t) []byte {
+	return C.GoBytes(unsafe.Pointer(src), C.int(size))
+}
+
+func goPoStProofs(partitions C.uint8_t, src *C.uint8_t, size C.size_t) ([]types.PoStProof, error) {
```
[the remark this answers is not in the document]
laser: > I start to feel uncomfortable having them private

Unfortunately, I can't export these things (and share them across packages, for instance). This is due to a limitation in CGO:

> Cgo translates C types into equivalent unexported Go types. Because the translations are unexported, a Go package should not expose C types in its exported API: a C type used in one Go package is different from the same C type used in another.

They aren't documented because they aren't exported. They aren't tested because they aren't part of the public (exported) API.

That said, I am to write some documentation for these unexported functions.

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -360,7 +358,7 @@ func (sb *RustSectorBuilder) Close() error {
 
 // GeneratePoSt produces a proof-of-spacetime for the provided commitment replicas.
 func (sb *RustSectorBuilder) GeneratePoSt(req GeneratePoStRequest) (GeneratePoStResponse, error) {
-	defer elapsed("GeneratePoSt")()
+	defer elapsed("Generate  PoSt")()
```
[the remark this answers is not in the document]
laser: I accept!

types/post_proof_partitions.go
laser: Yeah, we will definitely allow miners to select from a set of PoSt and PoRep proof partitions.

types/sector_class.go
laser: Agreed. I've ordered the fields identically for both structs.

proofs/rustverifier.go
laser: @ZenGround0 - I don't see such a rule in any of the go-filecoin markdown docs. Does one exist in the Wiki? 

Note that I always capitalize and add a period if either of these two conditions are met:

1. I'm documenting an exported type
1. I'm documenting an unexported statement with more than one sentence

The line which you have called out meets neither of these two criteria, so I politely decline your request to apply sentencification.

proofs/transforms.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// #cgo LDFLAGS: -L${SRCDIR}/lib -lfilecoin_proofs
+// #cgo pkg-config: ${SRCDIR}/lib/pkgconfig/libfilecoin_proofs.pc
+// #include "./include/libfilecoin_proofs.h"
+import "C"
+
+func cPoStProofs(src []types.PoStProof) (C.uint8_t, *C.uint8_t, C.size_t) {
```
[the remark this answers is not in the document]
laser: These functions aren't exported, so they're not tested directly. Under what circumstances is it appropriate to test unexported functions/methods? 

Note that these functions are tested indirectly through integration tests.

proofs/rustverifier.go
ZenGround0: NP sounds reasonable, I'll stop bugging people about this.

laser: I do appreciate that you evoked the Spirit of Phritz

PR #2608, opened 2019-04-18
FAST IPTB streaming

travisperson:
This still uses the `cmd.StdnnnPipe()` as I think the issue I was covering in #1701 around this isn't so much an issue because we have to read everything before Wait anyways. It will become clearer I think now. However, moving the buffering out of the plugin required a lot of stuff to move around as we can't call `ExitCode` before reading any output. This changed a few things around actions which used `RunCmdWithStdin` directly and `RunCmdLDJSONWithStdin`.

To around around this issue, I've made `RetrievalClientRetrievePiece` use the `io.ReadCloser` to fetch the exit code, and added a new interface `DecodeCloser` which exposes a `Decode` and `Close` method for use with LDJSON stuff (list-asks, chain ls, etc). When a user is done with the decode, they can close it to get any errors (error for exit code, or other errors returned by `output.Error()`).

In this Draft PR I'm looking for high level feedback on the interfaces changes / design, or high level implementation details. I'm not looking for any nit feedback as once I get feedback on the overall changes actual implementation may change.

### DIFF
--- a/tools/fast/process.go
+++ b/tools/fast/process.go
@@ -1,16 +1,19 @@
 package fast
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 
 	logging "github.com/ipfs/go-log"
 	iptb "github.com/ipfs/iptb/testbed"
 	"github.com/ipfs/iptb/testbed/interfaces"
 	"github.com/libp2p/go-libp2p-peer"
+	"golang.org/x/sync/errgroup"
 
 	fcconfig "github.com/filecoin-project/go-filecoin/config"
 	"github.com/filecoin-project/go-filecoin/tools/fast/fastutil"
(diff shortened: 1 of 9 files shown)


tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -497,3 +467,41 @@ func (l *Localfilecoin) WriteConfig(cfg interface{}) error {
 	lcfg := cfg.(*config.Config)
 	return lcfg.WriteFile(filepath.Join(l.dir, "config.json"))
 }
+
+func readAllOutput(ctx context.Context, output testbedi.Output) ([]byte, []byte, error) {
```
frrist: Can this be a method on `Output`?

tools/iptb-plugins/filecoin/local/output.go
```diff
@@ -0,0 +1,75 @@
+package pluginlocalfilecoin
+
+import (
+	"context"
+	"io"
+	"os/exec"
+)
+
+type Output struct {
```
frrist: I'd like a constructor method for this, e.g. `NewOutput`

tools/iptb-plugins/filecoin/local/output.go
```diff
@@ -0,0 +1,75 @@
...
+	args     []string
+	err      error
+	exitcode int
+	exitset  bool
```
frrist: What does this do? Based on this file it looks like it might not be needed.

tools/fast/process.go
```diff
@@ -193,45 +335,98 @@ func (f *Filecoin) RunCmdWithStdin(ctx context.Context, stdin io.Reader, args ..
...
+	// For the process to exit, we have to read all of the stdout / stderr, as
+	// the process may be blocked from finishing if the kernel buffer fills.
+	g, _ := errgroup.WithContext(ctx)
+
+	stdout := output.Stdout()
+	stderr := output.Stderr()
+
+	var bstdout []byte
+
+	g.Go(func() (err error) {
```
frrist: is this a duplication of `readAllOutput`?

tools/fast/process.go
```diff
@@ -182,6 +185,145 @@ func (f *Filecoin) DumpLastOutputJSON(w io.Writer) {
 	}
 }
 
+// outputMirror is used to store a copy of output lazily
+type outputMirror struct {
```
frrist: this was added to make `LastOutput` work, correct?

tools/iptb-plugins/filecoin/local/output.go
travisperson: It's used to short cut in the `ExitCode` function after it's been called once so we don't call `cmd.Wait()` twice.

tools/fast/process.go
travisperson: yes

PR #2609, opened 2019-04-18
CircleCI Concurrency to Unblock Forked Build Issues

ognots:
Uses CircleCI concurrency capabilities rather than specifying a high parallel setting in `go test`

See linked issue for background on the problem this solves and read commit messages for change details.


ognots:
Notice the checks section of this PR lists all components that were contained in `build_linux` individually. Definitely should be an enhancement to quickly identify test failures by type.
Once this has merged to master, go-filecoin team should decide of all components should be required to merge to master, and github repo settings set accordingly.

checks

travisperson:
No tests where run in these workflows. Is this a cache issue (as no code changed, so no tests need to run) or a configuration issue someplace?

ognots:
@travisperson thanks for catching that, I accidentally introduced a regression. it's working again and almost ready for final review

.circleci/config.yml
```diff
@@ -376,68 +411,215 @@ jobs:
...
+      - restore_cache:
+          key: v0-proof-params-{{ arch }}-{{ checksum "/tmp/rust-fil-proofs-checksum.txt" }}
+
+      - attach_workspace:
+          at: "."
+
+      - mkdir_test_results
+
+      - go_test:
+          sector_builder_tests: "<< parameters.sector_builder_tests >>"
```
frrist: Since this is only for functional tests should `sector_builder_tests` be `false`?

ognots: @frrist thanks for bringing this up!
the `sector_builder_tests` parameter is set to false by default. I've updated the config to only include the `sector_builder_tests` parameter in the functional test. it will only evaluate to `true` in the [run_sector_builder_tests workflow](https://github.com/filecoin-project/go-filecoin/blob/6d86d052ceb3aa0f236a7d14708800fcb47120b7/.circleci/config.yml#L803-L819).
this brings up another great a new level of granularity this parallelization brings - the sector build tests no longer have to run the unit and integration tests cc @laser
this is the [relevant commit](https://github.com/filecoin-project/go-filecoin/pull/2609/commits/6d86d052ceb3aa0f236a7d14708800fcb47120b7) - it also adjusts the nightly parameter slightly.

frrist: At the top of this file could we add a comment that goes over these testing workflows? Just something high level that states when they run and gives some pointers on how to parse this file when reading?

.circleci/config.yml
```diff
@@ -345,15 +388,6 @@ jobs:
       - go_build:
```
frrist: Someone less experienced with CI (me :smile: :disappointed:) added this to make tests work that required the go-filecoin binary to be in the repo, and [added this](https://github.com/filecoin-project/go-filecoin/pull/2609/files#diff-1d37e48f9ceff6d8030570cd36286a61R384) to satisfy tests that require the go-filecoin binary to be in `$GOPATH/bin`. I think we could shave 20-40sec off our build time by replacing this extra build step with a copy command. I believe this pattern is also [used here](https://github.com/filecoin-project/go-filecoin/pull/2609/files#diff-1d37e48f9ceff6d8030570cd36286a61R231).

frrist:
I don't believe any tests ran in this job:
https://circleci.com/gh/filecoin-project/go-filecoin/14312?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Inspected the `go test` command given it appears we are omitting the `github.com/filecoin-project/go-filecoin/commands` package. This is where a majority of our integration tests are.

```
go test -timeout 30m -parallel 4 github.com/filecoin-project/go-filecoin github.com/filecoin-project/go-filecoin/actor/builtin github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker github.com/filecoin-project/go-filecoin/bls-signatures github.com/filecoin-project/go-filecoin/chain github.com/filecoin-project/go-filecoin/consensus github.com/filecoin-project/go-filecoin/exec github.com/filecoin-project/go-filecoin/functional-tests github.com/filecoin-project/go-filecoin/metrics github.com/filecoin-project/go-filecoin/net/pubsub github.com/filecoin-project/go-filecoin/plumbing/bcf github.com/filecoin-project/go-filecoin/plumbing/msg github.com/filecoin-project/go-filecoin/proofs github.com/filecoin-project/go-filecoin/proofs/sectorbuilder/testing github.com/filecoin-project/go-filecoin/protocol/retrieval github.com/filecoin-project/go-filecoin/repo github.com/filecoin-project/go-filecoin/testhelpers github.com/filecoin-project/go-filecoin/tools/fast github.com/filecoin-project/go-filecoin/tools/fast/fastutil github.com/filecoin-project/go-filecoin/tools/faucet github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin github.com/filecoin-project/go-filecoin/tools/iptb-plugins/filecoin/local github.com/filecoin-project/go-filecoin/types github.com/filecoin-project/go-filecoin/vm github.com/filecoin-project/go-filecoin/wallet/util -functional=false -integration=true -sectorbuilder=false -unit=false -v
```

ognots:
there's 3 containers in that job, the second container ran the commands package https://circleci.com/gh/filecoin-project/go-filecoin/14312#tests/containers/1

@frrist this brings up an interesting point. can we expect integration and functional tests to be limited to a single directory or spread out amongst package directories?

keeping integration and functional tests in single directories would make it easier to split apart for concurrency.

frrist:
> there's 3 containers in that job, the second container ran the commands package 

Ahh that's right you told me how this works: We list all packages, divide them into sets, and test each set in parallel on different container, correct? This is something that I think we should have documented somewhere (maybe we already do and I am just unaware of it?)

> can we expect integration and functional tests to be limited to a single directory or spread out amongst package directories?

Currently integration tests are spread across at least 2 directories (`commands/` and `gengen/`). These test will be hard to move under one directory, unless we reclassify `gengen`, its considered an Integration test because it requires a filecoin binary to run.

Functional tests are also spread across 2 as well (`tools/fast/tests/` and `functional-tests/`), these may be easier to move under one dir -- cc @travisperson how would you feel about moving the FAST functional tests under `functional-tests`?

ognots:
I want to record a screen cast going through new setup and write some documentation once this is approved. Will be much easier to catch up filecoin team and contributors with a solid screen cast.

travisperson:
> how would you feel about moving the FAST functional tests under `functional-tests`?

@frrist  there is an issue for it https://github.com/filecoin-project/go-filecoin/issues/2538

anorth:
@ognots @frrist no we should not expect integration tests to converge under one directory. I expect more tests classified as integration to appear in other packages as we get better at in-process integration testing. I am under the impression that some FAST tests will be integration tests (those that replace daemon tests), but expect them to remain in the `commands` package.

Functional tests, maybe we could arrange that.

ognots:
@anorth are you suggesting we don't run functional tests every time?
to run unit and integration tests, the dependency and build steps would be required, but it would definitely save two and a half minutes of build time per CI run. 
we could easily move functional tests to a scheduled job like we do for sector builder tests.

ognots:
This commit https://github.com/filecoin-project/go-filecoin/pull/2609/commits/2164213b57000776fd4fc7cf444f68b51c5deada removes some extra commands and uses more reusable commands. also cleans up formatting in efforts to make things easier to read. 
There's a few more gains to be made in moving some bash logic to scripts outside the CircleCI config file, but not going to include that in this PR.
The size of the configuration won't be too easy to knock down much more, because YAML doesn't merging arrays to re-use some of the sequence of steps shared between jobs. There's also currently no way to break up the CircleCI config file in to smaller files and include them. It's a feature request for CircleCI but not there yet.
We could roll our own modularization of the configuration and have a git post-commit hook assemble a new `.circleci/config` file, but that's not the highest priority at the moment.

anorth:
I was under the impression that the current setup did not run functional tests, but it appears I was mistaken. Let's not change that now, so the setup you propose is fine.

Thanks for the extra DRY. Let's not trade complexity in the YAML for complexity in added commit steps at the moment, but a shell script might be worthwhile.

This is GTG so long as someone can update the branch protection checks when it lands.

PR #2610, opened 2019-04-18
Support running FAST Shell in go tests

travisperson:
This PR adds support for running the FAST Shell method in golang tests by adding support for the `TTY` environment variable to enable the use of the FAST Shell in golang tests.


frrist:
In order to make these changes impactful to users I think it's important that we provide some documentation on how to use this, preferably with some example code too. The PR description is a great start on this, but will likely go overlooked once these changes merge. Can you distill it to a markdown file and put it in something like `/tools/fast/docs`?

travisperson:
Ya, I'm going to take some of it and add it as a comment on the plugin method itself, and a bit on the FAST filecoin process method, and the rest will end up in a document.

tools/fast/docs/SHELL.md
```diff
@@ -0,0 +1,81 @@
...
+| `FIL_PATH` |  The value is set to the repository directory for the Filecoin node. Any `go-filecoin` commands ran in the shell will be executed against the Filecoin process for which the Shell method was invoked on in FAST. |
+| `FIL_PID` | The value is set to the PID for the Filecoin daemon. |
+| `FIL_BINARY` | The value is set to the binary running the Filecoin daemon. Please refer to the section below on `PATH` for more details. |
+| `PATH` | The users `PATH` will be updated to include a location that contains the binary used for executing all `go-filecoin` commands. The `go-filecoin` binary included in this location itself is defined by either the value of the `localplugin.AttrFilecoinBinary`, or the first `go-filecoin` binary found in the users `PATH`. <br/> <br/>_Note: The value of `FIL_BINARY` will not be the exact value. During node setup, the binary is copied to ensure it does not change during execution. `FIL_BINARY` will be this new path._ |
+
+The information around the `PATH` seems a little complex, but it's to ensure that there are no issues as a result of mixing binaries.
+This has the advantage that while using FAST, users can re-compile `go-filecoin`.
```
anorth: Clarify "... without affecting the running FAST environment", as opposed to "... and use those changes immediately in the FAST environment".

tools/fast/docs/SHELL.md
```diff
@@ -0,0 +1,81 @@
...
+It should be noted that the copying of the binary occurs during the call to `NewProcess`.
+
+It should also be noted that users shell configuration will be ran when the shell opens.
+If shell configuration updates the `PATH` by appending to the front, if any of those directories contain `go-filecoin`, then the `go-filecoin` command inside of the FAST Shell will **not** point to the correct binary.
```
anorth: This seems like a trap that many people will hit, in particular because `go install` installs a binary to a path that many people will have intentionally prepended.

Is it possible to add some kind of check inside the shell if this has happened? Alternatively, update all docs to use `$FIL_BINARY` rather than `go-filecoin` to avoid path ambiguity?

FWIW I never `go install`, preferring to use `./go-filecoin` to be sure. Is this a reason to remove suggestions to explicitly install in this way, e.g. from the wiki?

tools/fast/docs/SHELL.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Using a FAST Shell in _go test_
+
+The FAST Shell should only be used when running a single test.
+
+_Note: When using the FAST Shell, it's best to increase deadlines set for tests, as it's very easy to exceed them, and you will be kicked out of your shell._
```
anorth: Demonstrate how to do that here.

tools/fast/docs/SHELL.md
```diff
@@ -0,0 +1,81 @@
...
+
+When using FAST inside of go tests, `stdin` is not properly configured to the users shell.
+This results in the shell exiting immediately, with no error.
+To resolve this issue the Filecoin localplugin will attach the fd defined by the environment variables `TTY` to the shell it opens.
+This will allows the user to define which TTY to use for stdin. Generally this should be the same TTY of the shell `go test` is ran in.
+This can easily be set by using the `tty` program.
+
+```shell
+$ env TTY=$(tty) go test -v -run TestFilecoin
```
anorth: Include test timeout wrangling in this example too

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -397,7 +412,25 @@ func (l *Localfilecoin) Shell(ctx context.Context, ns []testbedi.Core) error {
 
 	cmd := exec.CommandContext(ctx, shell)
 	cmd.Env = nenvs
-	cmd.Stdin = os.Stdin
+
+	stdin := os.Stdin
+
+	// When running code with `go test`, the os.Stdin is not connected to the shell
+	// where `go test` was ran. This makes the shell exit immeidately and it's not
```
anorth: typo "immediately"

tools/fast/docs/SHELL.md
travisperson: > Is it possible to add some kind of check inside the shell if this has happened?

A check could be added in the plugin itself to determine if `go-filecoin` already exists in the `PATH`, similarly to the check for `FIL_PATH`. Referencing `$FIL_PATH` might be the best option here.

Issue #2581, opened 2019-04-16
CLI for repo migrations


shannonwells:
## Description
Repo migration should have a CLI that calls the appropriate migration command. This story is only for the interface, not the underlying functionality.

### Questions:
Should errors in CLI processing go to the same logs as the migration log?

## Acceptance criteria
- [ ] The following commands can be run from command line without error, and respond to a help option (-h or --help both work):
    * describe
    * buildonly
    * install
    * migrate
- [ ] The CLI should accept the following options:
   *   --verbose : whether to output all messages to stdout
   * --old-repo  : what directory the old repo lives in:
       * if not specified, looks at `$FIL_PATH` then at `$HOME/.filecoin`
   * --new-repo : where to put the temporary directory for repo copy to be migrated
       * if not specified, is named the same as old repo + `_<version>_<timestamp>`, for example, if the old Repo is located at $HOME/.filecoin, the newrepo should be at: `$HOME/.filecoin_0.1.5_2019.05.21_165134`  However the filename is formatted, these should be ordered with version and then timestamp to keep migrated directories with the same timestamp listed next to each other, and the file name should be easily parsed by a human.

- [ ] Keep imports to a minimum by striving to use golang standard packages.

It is ok if the commands just display the help text if the migration code is not implemented.  

### Where to begin
tools/migration/main.go

shannonwells:
Closing with merge of #2708

PR #2614, opened 2019-04-18
Build, test and skeleton CLI for repo migrations

### LINKED ISSUE #2582: Repo migrations can be built and installed using go-filecoin build command

### LINKED ISSUE #2589: Spike on repo migrations

shannonwells:
# Smaller PR after [spike](https://github.com/filecoin-project/go-filecoin/pull/2576)
* Closes #2589 
* Closes #2582 , 
* Also foundational work for #2581 CLI for repo migrations and unit testing.

### DOES NOT Include:
* CLI option for old repo location(except showing in the usage)
* CLI option for new repo location (except showing in the usage)
* actual migration code (except for stubs and interface), including doing anything with the old and new repo options
* logging of any kind

These will be implemented in later PRs.


tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+
+func getCommand() string {
+	return os.Args[1:][0]
+}
+
+func getVerbose() bool {
+	if _, found := findString("-v", os.Args); found {
+		return true
+	}
+	_, res := findString("--verbose", os.Args)
```
shannonwells: I expect to implement a getOldRepoOpt and getNewRepoOpt which will use the values returned by findString.

build/main.go
```diff
@@ -247,6 +260,8 @@ func main() {
 		buildGengen()
 	case "generate-genesis":
 		generateGenesis()
+	case "build-migrations":
```
ZenGround0: I am good with not adding `install-migrations` (it matches up with other tool options).  Just checking that you made the choice to leave it off explicitly as install shows up as an optional acceptance criteria in 2582.

tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+	os.Exit(1)
+}
+
+func showUsageAndExit(code int) {
+	fmt.Println(`    Usage:  go-filecoin-migrate (describe|buildonly|migrate) [--verbose][--oldRepo][--newRepo]`)
+	os.Exit(code)
+}
+
+func getCommand() string {
+	return os.Args[1:][0]
```
ZenGround0: nit: I might be missing something but couldn't we say `os.Args[1]` instead?

tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// TODO: should this go to the same logs as the migration log? Maybe not.
+func exitErr(errstr string) {
+	log.New(os.Stderr, "", 0).Println(errstr)
+	os.Exit(1)
+}
+
+func showUsageAndExit(code int) {
+	fmt.Println(`    Usage:  go-filecoin-migrate (describe|buildonly|migrate) [--verbose][--oldRepo][--newRepo]`)
```
ZenGround0: Really like how simple this is.

tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+	switch command {
+	case "-h", "--help":
+		showUsageAndExit(0)
+	case "describe", "buildonly", "migrate", "install":
+		mr := internal.MigrationRunner{}
+		if err := mr.Run(command, getVerbose()); err != nil {
+			exitErr(err.Error())
+		}
+	default:
+		exitErr(fmt.Sprintf("Error: Invalid command: %s", command))
```
ZenGround0: Would it be helpful to print out usage in this case?

tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+// TODO: should this go to the same logs as the migration log? Maybe not.
```
ZenGround0: my .02 is that we should log all errors to the same place until we have a compelling reason to get more complicated.

tools/migration/main.go
```diff
@@ -0,0 +1,67 @@
...
+
+func getVerbose() bool {
+	if _, found := findString("-v", os.Args); found {
+		return true
+	}
+	_, res := findString("--verbose", os.Args)
+	return res
+}
+
+func findString(str string, args []string) (string, bool) {
```
ZenGround0: nit: what is the purpose of returning the arg that matches the input string?  The caller must already have that string.

tools/migration/internal/migration.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+import (
+	"os"
+)
+
+type Migrator interface {
+	Migrate(newRepo *os.File) error
```
ZenGround0: nit: it doesn't matter much now as this is all just a stub but these signatures don't match the interfaces in the design doc.

tools/migration/main.go
ZenGround0: super small nit: it might be worth separating this out into a constant.  For example this would make it slightly easier to test as you wouldn't have to rewrite the constant in the tests.

shannonwells: So, I debated separating it as a constant, however, I generally don't like pulling out constants as tests; I prefer forcing the test to manually match content, to help force extra proofreading, but also it makes it a little easier if there is string formatting (which is one of the things I was debating).

tools/migration/main.go
shannonwells: Yeah, good catch, it's actually because it's supposed to return the value of an option, too. I'll fix it.

tools/migration/internal/migration.go
shannonwells: The design doc has to be changed actually; this was from one of the comments in the closed PR.  I'll update the doc.

build/main.go
shannonwells: thanks, it was intentional.

tools/migration/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	ast "github.com/stretchr/testify/assert"
+	req "github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-filecoin/testhelpers"
+)
+
+func TestUsage(t *testing.T) {
+	tf.UnitTest(t)
```
frrist: If these tests are calling `exec.Command` we should label them as `Integration` tests, there is an (undocumented..sorry) expectation that unit tests don't require anything to be built when running.

tools/migration/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"log"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/filecoin-project/go-filecoin/tools/migration/internal"
+	"github.com/filecoin-project/go-filecoin/util/version"
+)
+
+const USAGE = `Usage:  go-filecoin-migrate (describe|buildonly|migrate) [--verbose][--oldRepo][--newRepo]`
```
frrist: If more flags are to be added consider https://github.com/urfave/cli

shannonwells: will do, thanks

tools/migration/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+func main() { // nolint: deadcode
+	if len(os.Args) < 2 {
+		showUsageAndExit(1)
+	}
+
+	if !version.Check(runtime.Version()) {
+		exitErr(fmt.Sprintf("Invalid go version: %s", runtime.Version()))
+	}
```
anorth: If we can't articulate a specific need for this, I think we should remove it.

tools/migration/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+func getVerbose() bool {
+	if _, found := findOpt("-v", os.Args); found {
+		return true
+	}
+	_, res := findOpt("--verbose", os.Args)
+	return res
+}
+
+func findOpt(str string, args []string) (string, bool) {
```
anorth: If we keep this, please add a comment, especially what the return value are meant to mean.

PR #2615, opened 2019-04-19
Feat/jaeger tracing

frrist:
### Description
This PR adds:
- Observability config that contains the Metrics config and Tracing config
- Tracing calls to the `chain/`, `node/`, and `consensus/` packages
- Documentation to the codewalk explaining filecoin observability options and configuration

### DIFF
--- a/CODEWALK.md
+++ b/CODEWALK.md
@@ -39,7 +39,7 @@ It is complemented by specs (link forthcoming) that describe the key concepts im
 - [Sector builder & proofs](#sector-builder--proofs)
   - [Building and distribution.](#building-and-distribution)
   - [Groth parameters](#groth-parameters)
-  - [Sector size configuration](#sector-size-configuration)
+  - [Proof mode configuration](#proof-mode-configuration)
 - [Networking](#networking)
 - [Filesystem storage](#filesystem-storage)
   - [JSON Config](#json-config)
@@ -50,10 +50,14 @@ It is complemented by specs (link forthcoming) that describe the key concepts im
       - [Unit Tests (`-unit`)](#unit-tests--unit)
       - [Integration Tests (`-integration`)](#integration-tests--integration)
       - [Functional Tests (`-functional`)](#functional-tests--functional)
+      - [Sector Builder Tests (`-sectorbuilder`)](#sector-builder-tests--sectorbuilder)
 - [Dependencies](#dependencies)
 - [Patterns](#patterns)
   - [Plumbing and porcelain](#plumbing-and-porcelain)
   - [Consumer-defined interfaces](#consumer-defined-interfaces)
+  - [Observability](#observability)
+    - [Metrics](#metrics)
+    - [Tracing](#tracing)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -636,3 +640,20 @@ Our implementation of [plumbing and porcelain](#plumbing-and-porcelain) embraces
 
 This idiom is unfortunately hidden away in a [wiki page about code review](https://github.com/golang/go/wiki/CodeReviewComments#interfaces).
 See also Dave Cheney on [SOLID Go Design](https://dave.cheney.net/2016/08/20/solid-go-design).
+
+### Observability
+
+go-filecoin uses [Opencensus-go](https://github.com/census-instrumentation/opencensus-go) for stats collection and distributed tracing instrumentation.
+Stats are exported for consumption via [Prometheus](https://prometheus.io/) and traces are exported for consumption via [Jaeger](https://www.jaegertracing.io/docs/1.11/).
+
+#### Metrics
+
+go-filecoin can be configured to collect and export metrics to Prometheus via the `MetricsConfig`.
+The details of this can be found inside the [`config/`](https://godoc.org/github.com/filecoin-project/go-filecoin/config#ObservabilityConfig) package.
+To view metrics from your filecoin node using the default configuration options set the `prometheusEnabled` value to `true`, start the filecoin daemon, then visit `localhost:9400/metrics` in your web-browser. 
+
+#### Tracing
+
+go-filecoin can be configured to collect and export traces to Jaeger via the `TraceConfig`.
+The details of this can be found inside the [`config/`](https://godoc.org/github.com/filecoin-project/go-filecoin/config#ObservabilityConfig) package.
+To collect traces from your filecoin node using the default configuration options set the `jaegerTracingEnabled` value to `true`, start the filecoin daemon, then follow the [Jaeger Getting](https://www.jaegertracing.io/docs/1.11/getting-started/#all-in-one) started guide.
--- a/config/config.go
+++ b/config/config.go
@@ -151,6 +151,7 @@ func newDefaultHeartbeatConfig() *HeartbeatConfig {
 	}
 }
 
+// ObservabilityConfig is a container for configuration related to observables.
 type ObservabilityConfig struct {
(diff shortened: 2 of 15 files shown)


ZenGround0:
Ay this is cool!   What I like least about this is all of the added lines.  One way to cut down each invocation by 5 lines is to [define a defer function generator somewhere else and invoke it in a single line](https://play.golang.org/p/6KjakxcxMpD).  Because every span does the same thing in the defer you can easily use a single defer function generator which will dedup code too.

frrist:
@ZenGround0 thats a great idea, I can add that shortly.

CODEWALK.md
```diff
@@ -39,7 +39,7 @@ It is complemented by specs (link forthcoming) that describe the key concepts im
 - [Sector builder & proofs](#sector-builder--proofs)
   - [Building and distribution.](#building-and-distribution)
   - [Groth parameters](#groth-parameters)
-  - [Sector size configuration](#sector-size-configuration)
```
frrist: Changes made from running `doctoc`

CODEWALK.md
```diff
@@ -636,3 +640,15 @@ Our implementation of [plumbing and porcelain](#plumbing-and-porcelain) embraces
...
+
+#### Metrics
+
+go-filecoin can be configured to collect and export metrics to Prometheus via the `MetricsConfig`. The details of this can be found inside the [`config/`](https://github.com/filecoin-project/go-filecoin/blob/ce1f739b1569f95279a15a6a4e9978186c680d38/config/config.go) package. To view metrics from your filecoin node using the default configuration options set the `prometheusEnabled` value to `true`, start the filecoin daemon, then visit `localhost:9400/metrics` in your web-browser. 
+
+#### Tracing
+
+go-filecoin can be configured to collect and export traces to Jaeger via the `TraceingConfig`. The details of this can be found inside the [`config/`](https://github.com/filecoin-project/go-filecoin/blob/ce1f739b1569f95279a15a6a4e9978186c680d38/config/config.go) package. To collect traces from your filecoin node using the default configuration options set the `jaegerTracingEnabled` value to `true`, start the filecoin daemon, then follow the [Jaeger Getting](https://www.jaegertracing.io/docs/1.11/getting-started/#all-in-one) started guide.
```
ZenGround0: nit: `TraceingConfig` should be `TraceConfig` and the link provided does not have this implemented yet.

consensus/expected.go
```diff
@@ -238,9 +238,12 @@ func (c *Expected) IsHeavier(ctx context.Context, a, b types.TipSet, aSt, bSt st
...
 // results in an error.
-func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ancestors []types.TipSet, pSt state.Tree) (state.Tree, error) {
-	err := c.validateMining(ctx, pSt, ts, ancestors[0])
-	if err != nil {
+func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ancestors []types.TipSet, pSt state.Tree) (st state.Tree, err error) {
+	ctx, span := trace.StartSpan(ctx, "Expected.RunStateTransition")
+	span.AddAttributes(trace.StringAttribute("tipset", ts.String()))
+	defer tracing.AddErrorEndSpan(ctx, span, &err)
+
+	if err := c.validateMining(ctx, pSt, ts, ancestors[0]); err != nil {
```
ZenGround0: Blocking: I'm suspicious that this `:=` means that this error is not properly being set to the named return value preventing the span from properly reporting the error.

I could be wrong, ignore if you're sure this is fine.

consensus/processor.go
```diff
@@ -268,14 +279,21 @@ func (p *DefaultProcessor) ProcessTipSet(ctx context.Context, st state.Tree, vms
 //       revert errors.
 //   - everything else: successfully applied (include, keep changes)
 //
-func (p *DefaultProcessor) ApplyMessage(ctx context.Context, st state.Tree, vms vm.StorageMap, msg *types.SignedMessage, minerOwnerAddr address.Address, bh *types.BlockHeight, gasTracker *vm.GasTracker, ancestors []types.TipSet) (*ApplicationResult, error) {
+func (p *DefaultProcessor) ApplyMessage(ctx context.Context, st state.Tree,
```
ZenGround0: nit: when moving from horizontal to vertical I like to commit all the way and put one argument per line:
```
ApplyMessage(
  ctx context.Context,
  st state.Tree,
...
  ancestors []types.TipSet,
)
```

config/config.go
```diff
@@ -168,6 +180,23 @@ func newDefaultMetricsConfig() *MetricsConfig {
 	}
 }
 
+type TraceConfig struct {
+	// JaegerEndpoint is the URL traces are collected on.
+	JaegerEndpoint string `json:"jaegerEndpoint"`
```
ZenGround0: unimportant nit: because of similarities with this and metrics it could be nice to have a consistent ordering among the three fields across both configs.

consensus/expected.go
frrist: I think its fine: https://play.golang.org/p/9vJGxElgkZu

travisperson: Ah this is super interesting. It looks like if you name a variable, regardless of how it's returned (by assignment, or directly in a return state), it gets correctly assigned.

metrics/tracing/util.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// AddErrorEndSpan will end `span` and adds `err` to `span` iff err is not nil.
+// This is a helper method to cut down on boiler plate.
+func AddErrorEndSpan(ctx context.Context, span *trace.Span, err *error) {
+	if *err != nil {
+		span.AddAttributes(trace.StringAttribute("error", (*err).Error()))
+	}
+	span.End()
+}
```
anorth: 👍 removing boilerplate. You could also add a `StartSpan` with varargs to inline string attributes.

CODEWALK.md
```diff
@@ -636,3 +640,20 @@ Our implementation of [plumbing and porcelain](#plumbing-and-porcelain) embraces
...
+### Observability
+
+go-filecoin uses [Opencensus-go](https://github.com/census-instrumentation/opencensus-go) for stats collection and distributed tracing instrumentation.
+Stats are exported for consumption via [Prometheus](https://prometheus.io/) and traces are exported for consumption via [Jaeger](https://www.jaegertracing.io/docs/1.11/).
+
+#### Metrics
+
+go-filecoin can be configured to collect and export metrics to Prometheus via the `MetricsConfig`.
+The details of this can be found inside the [`config/`](https://godoc.org/github.com/filecoin-project/go-filecoin/config#ObservabilityConfig) package.
+To view metrics from your filecoin node using the default configuration options set the `prometheusEnabled` value to `true`, start the filecoin daemon, then visit `localhost:9400/metrics` in your web-browser.
```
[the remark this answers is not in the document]
frrist: I would prefer to set it to `true` during the deployment. Because we run all integration tests in parallel, setting this to true by default will cause them to fail due to port conflicts, i.e:
`ERROR metrics: failed to serve /metrics endpoint on listen tcp 0.0.0.0:9400: bind: address already in use export.go:58`. Some solutions could be, using port `0` by default, randomly assign the port at testing time, or disabling metrics during testing. I believe defaulting to `false` is the simplest solution here.

metrics/tracing/util.go
frrist: I agree there is a lot of boilerplate with this. However, I would prefer to leave this as is until a clearer pattern emerges, I will address it in a follow on at that point.

Issue #1887, opened 2019-02-12
Storage market penalization of late submitters


ZenGround0:
### Description
This is a stub tracking work to sync with the spec on penalization of late post submitters in the storage market.  This includes setting proving period and grace period parameters.

We need appropriate values for:

1. `miner.Actor.GracePeriodBlocks`: The number of blocks after a proving period over which a miner can still submit a post at a penalty.
2. `miner.Actor.ClientProofOfStorageTimeout`: the number of blocks between LastPoSt and the current block height after which the miner is no longer considered to be storing the client's piece and they are entitled to a refund.
3. `paymentbroker.Actor.CancelDelayBlockTime`: the number of rounds given to the miner to respond after the channel is canceled before it expires.

anorth:
@acruikshank this may be relevant for epic #2481

acruikshank:
@ZenGround0 Does this issue refer to the slashing of storage miner by the client for failing to prove a particular deal or the the slashing by the network for failing to prove committed storage?

acruikshank:
I'm going to answer yes, because I need this too.

When a miner fails to submit PoSts they will hit two separate deadlines. The first is a consensus fault and subjects the miner to slashing because they are claiming storage power they do not have. The second indicates the miner is no longer fulfilling their storage deals with clients. This deadline allows clients to reclaim their storage payments and slash the miner for damages. The second is considered to have more serious consequences, so the miner is given more time to resume their PoSts.

PR #2620, opened 2019-04-19
Feat/payment channel close 2574

### LINKED ISSUE #2574: Payment Channel Close
### Description

Storage clients need to be able to recover funds for broken deals. Currently the only mechanism for recovering funds from a payment channel is to wait for the the channel to expire. For storage deals this means waiting till after the entire duration of the proposed deal, even if the deal was immediately rejected.

To address this, when the payer of a payment channel calls Close, the Eol will be reset to a fixed amount of time after now. The fixed amount of time must be sufficient to allow the target of the payment channel to reopen the channel. This fixed amount of time should be a block time and can be set as a constant for now.

Reopening the channel is not part of this issue.

The original Eol must be retained, so that that it can be restored if the close is successfully challenged. The original eol must be stored in actor state. This will be added to the payment channel struct.

### Acceptance criteria

.1. Payment Channel payer can call `Close`
2. Calling `Close` resets the Eol to a fixed time interval from now (only if sender is payer).
3. Eol is retained.

rosalinekarr:
# Problem

Payers have no way to stop a payment channel once opened meaning they risk losing money when opening long running storage deals.

# Solution

Add a Cancel method to the PaymentBroker allowing a payer to close a payment channel early.

Resolves #2574

### DIFF
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -347,10 +347,10 @@ func (pb *Actor) Extend(vmctx exec.VMContext, chid *types.ChannelID, eol *types.
 	return 0, nil
 }
 
-// Cancel can be used to end a storage deal early. It lowers the EOL of the
-// payment channel to 1 blocktime from now and allows a client to reclaim their
-// payments. In the time before the channel is closed, a miner can potentially
-// dispute a closer.
+// Cancel can be used to end an off chain payment early. It lowers the EOL of
+// the payment channel to 1 blocktime from now and allows a caller to reclaim
+// their payments. In the time before the channel is closed, a target can
+// potentially dispute a closer.
 func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, error) {
 	if err := vmctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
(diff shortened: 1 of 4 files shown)


rosalinekarr:
@acruikshank @shannonwells, still working on improving testing, but I think the actual code should be ready for review.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -334,6 +341,49 @@ func (pb *Actor) Extend(vmctx exec.VMContext, chid *types.ChannelID, eol *types.
...
+			}
+			return errors.FaultErrorWrapf(err, "Could not retrieve payment channel with ID: %s", chid)
+		}
+
+		channel, ok := chInt.(*PaymentChannel)
+		if !ok {
+			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
+		}
+
+		channel.Eol = vmctx.BlockHeight().Add(types.NewBlockHeight(1))
```
acruikshank: BLOCKING: We're not exactly sure how much to add here, but it isn't 1. 10000 is around 3 days and seems reasonable for now. This needs to be a constant and it needs an issue to determine the correct value. #1887 is a good template.

rosalinekarr:
Closing in favor of #2628 which has already been approved.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -334,6 +347,54 @@ func (pb *Actor) Extend(vmctx exec.VMContext, chid *types.ChannelID, eol *types.
 	return 0, nil
 }
 
+// Cancel can be used to end an off chain payment early. It lowers the EOL of
+// the payment channel to 1 blocktime from now and allows a caller to reclaim
```
anorth: ...to `cancelDelayBlockTime` from now...

PR #2627, opened 2019-04-19
Dump last output on test failure in teardown

travisperson:
I've been the payment channels flake a bit, and we currently do not have any output to really debug with. Currently we just see the command that fails but no information on why it failed. This PR will dump all processes last command during Teardown if the test failed.

@frrist I think is working on something similar along with the surface panic work, but due to the increase in these test failures I want to get something in to try and resolve them.

The commands are printed for the nodes in the order they were created.


travisperson:
Example output:

_Note: this isn't the best example as it's not a failure of a command itself, but it should provide some info about how it will look when things do fail._

```
--- FAIL: TestPaymentChannelReclaimSuccess (3.50s)
    payment_channel_daemon_test.go:335: 
        	Error Trace:	payment_channel_daemon_test.go:335
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 40
        	Test:       	TestPaymentChannelReclaimSuccess
    payment_channel_daemon_test.go:342: 
        	Error Trace:	payment_channel_daemon_test.go:342
        	Error:      	Not equal: 
        	            	expected: &types.AttoFIL{val:(*big.Int)(0xc000662840)}
        	            	actual  : &types.AttoFIL{val:(*big.Int)(0xc0005102c0)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(*types.AttoFIL)(10)
        	            	+(*types.AttoFIL)(0)
        	            	 
        	Test:       	TestPaymentChannelReclaimSuccess
    payment_channel_daemon_test.go:363: 
        	Error Trace:	payment_channel_daemon_test.go:363
        	Error:      	Not equal: 
        	            	expected: &types.AttoFIL{val:(*big.Int)(0xc000632e20)}
        	            	actual  : &types.AttoFIL{val:(*big.Int)(0xc000632e40)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(*types.AttoFIL)(9800)
        	            	+(*types.AttoFIL)(9810)
        	            	 
        	Test:       	TestPaymentChannelReclaimSuccess
    print_writer.go:31: >>>> start-dump
    print_writer.go:31: ---- command     go-filecoin mining once --enc=json
    print_writer.go:31: ---- exit-code   0
    print_writer.go:31: ---- error       nil
    print_writer.go:31: ---- stdout
    print_writer.go:31: {"/":"zDPWYqFCrJSdzkF6Mq6B6D1KpdVx5b2adaYf79hdRGdT4w2nmBd6"}
    print_writer.go:31: ---- stderr
    print_writer.go:31: <<<< end-dump
    print_writer.go:31: >>>> start-dump
    print_writer.go:31: ---- command     go-filecoin paych ls --from t1uicfduhipil6fbtqe7qjetyjznd7s5kpwgaqqwi --enc=json
    print_writer.go:31: ---- exit-code   0
    print_writer.go:31: ---- error       nil
    print_writer.go:31: ---- stdout
    print_writer.go:31: {"0":{"target":"t1t5xsj7lstpzwzdwredapnvghnn6qibjiwe74jwy","amount":"1000","amount_redeemed":"0","eol":4}}
    print_writer.go:31: ---- stderr
    print_writer.go:31: <<<< end-dump
    print_writer.go:31: >>>> start-dump
    print_writer.go:31: ---- command     go-filecoin wallet balance t1uicfduhipil6fbtqe7qjetyjznd7s5kpwgaqqwi --enc=json
    print_writer.go:31: ---- exit-code   0
    print_writer.go:31: ---- error       nil
    print_writer.go:31: ---- stdout
    print_writer.go:31: "9800"
    print_writer.go:31: ---- stderr
    print_writer.go:31: <<<< end-dump
```

tools/fast/fastesting/basic.go
```diff
@@ -46,6 +46,10 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.Envi
 	env, err := fast.NewEnvironmentMemoryGenesis(big.NewInt(1000000), dir, types.TestProofsMode)
 	require.NoError(err)
 
+	defer func() {
+		dumpEnvOutputOnFail(t, env.Processes())
```
frrist: Should this be `TearDown` since it now calls this method?

Something I would like to see is, on test failure don't teardown completely instead leave the FAST repo in place with the stderr and stdout files and direct the user to that location. What do you think?

tools/fast/fastesting/basic.go
```diff
@@ -138,3 +142,22 @@ func (env *TestEnvironment) RequireNewNodeWithFunds(funds int) *fast.Filecoin {
 
 	return p
 }
+
+func (env *TestEnvironment) Teardown(ctx context.Context) error {
```
frrist: Not sure if the linter caught this, but please comment these methods

tools/fast/fastesting/print_writer.go
```diff
@@ -0,0 +1,48 @@
...
+	"sync"
+	"testing"
+)
+
+type printWriter struct {
+	pw io.WriteCloser
+	wg sync.WaitGroup
+}
+
+func newPrintWriter(t *testing.T) io.WriteCloser {
```
frrist: Please add some comments on what this method is for, it looks like it pipes the daemons last output to calls to testing.Log

tools/fast/fastesting/basic.go
travisperson: No, because if we defer Teardown it will teardown the environment in this method which we don't want to do. We just want to dump this output if any of the `require.NoError` fail so we can see output of the setup steps in this function.

tools/fast/fastesting/basic.go
travisperson: Ya, we had this issue the last time. I'm really curious why the linter isn't catching this stuff.

tools/fast/fastesting/basic.go
frrist: What if we added a method like: `TeardownOnSuccess` that would conditionally teardown the environment?

travisperson: I think we should always teardown regardless if the test passes or fails. As mentioned else where I think we can preserve the logs to makes sure they are not removed in the event that a test fails though.

frrist: Where is preserving logs on test failures being discussed? That's mainly what I am after here.

travisperson: I don't remember if there was an issue created yet, but @rosalinekarr wanted to have the daemon logs be stored in a `logs` directory alongside the project root such that they would 1) not be cleaned up on teardown, and also survive past restarts.

I think it was a discussion in slack a bit ago.

tools/fast/fastesting/basic.go
anorth: @travisperson could you please look into linter behaviour for us? We may have lost something during gomod migration.

PR #2633, opened 2019-04-22
WIP upgrade go-ds-badger

rkowalick:
wip wip wip


mishmosh:
i don't recall a specific conversation that we wanted to upgrade badger. in general, though, upgrades are good.

@anorth should we try to land this? if so, could you nominate someone to shepherd this to completion or do it yourself. if no, please close.

anorth:
Yes, let's try. @shannonwells could you take this over after migration-related work.

anorth:
Let's abandon this now. Not a priority at the moment.

Issue #2629, opened 2019-04-19
Integrate PIP verification into miner.Actor


acruikshank:
### Description

PIPs need to be verified in the miner actor. This verification occurs in rust proofs and is not currently available in go-filecoin. We will be mocking this until it is available. This issue tracks progress torwards full integration.

### Acceptance criteria

1. PIPs must be valid for `miner.Actor.verifyPIP` to return successfully.

acruikshank:
Implemented in #3066

PR #2638, opened 2019-04-22
Add VerifyPIP method to miner.Actor

acruikshank:
### Problem

Clients need miners to prove storage before collecting payments. We currently no mechanism for a miner to prove anything.

### Solution

The `VerifyPIP` method is a starting point in a larger chunk of work. It provides some of the verification the payment broker needs to allow a miner to prove storage. 

Proof verification requires:

1. The miner to provide a sector id and piece inclusion proof (PIP) along with a pre-established data commitment from the client (CommP).
2. Verification that the miner has committed the sector, `sector`, referenced by `sectorId`.
3. Verification that miner has recently submitted a proof of space time (PoSt) for the sector referenced by `sectorId`. Currently, it is sufficient to verify that the miner actors' `LastPoSt` property is within some window of the current block height.
4. Verification that the piece inclusion proof proves that the data represented by CommP is included within the data represented by `sector.CommD`. This is accomplished by making a call to rust proofs.

The last step is currently unavailable, so this version only checks that the proof is the concatenation of CommP and CommD, pending #2629.

This PR introduces the `ClientProofOfStorageTimeout` constant used to determine is the miner is sufficiently up-to-date in their proofs to verify the pip. As an additional refactor, this PR cleans up and documents some of the constants with temporary values in `miner.Actor` and `paymentbroker.Actor`.


actor/builtin/miner/miner_test.go
```diff
@@ -361,6 +361,103 @@ func TestMinerSubmitPoSt(t *testing.T) {
...
+	require.Equal(t, uint8(0), res.Receipt.ExitCode)
+
+	runVerifyPIP := func(t *testing.T, bh uint64, commP []byte, sectorId uint64, proof []byte) (bool, string) {
+		res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, bh, "verifyPIP", ancestors, commP, sectorId, proof)
+		require.NoError(t, err)
+		require.NoError(t, res.ExecutionError)
+		require.Equal(t, uint8(0), res.Receipt.ExitCode)
+
+		valid := parseAbiBoolean(res.Receipt.Return[0])
+		reason := string(res.Receipt.Return[1])
```
anorth: Make assertions about the other two return values.

actor/builtin/miner/miner_test.go
```diff
@@ -361,6 +361,103 @@ func TestMinerSubmitPoSt(t *testing.T) {
...
+
+		t.Run("PIP is invalid if proof is invalid", func(t *testing.T) {
+			wrongPIP := append([]byte{pip[0] + 1}, pip[1:]...)
+			valid, reason := runVerifyPIP(t, 3, commP, sectorId, wrongPIP)
+
+			require.False(t, valid)
+			require.Equal(t, "invalid inclusion proof", reason)
+		})
+
+		t.Run("Malformed PIP is a validation error, but not a revert error", func(t *testing.T) {
```
anorth: This name is confusing because `runVerifyPIP` cannot possibly return a revert error, only strings. Is it possible to test the return path that has an error in the fourth position of `VerifyPIP`?

actor/builtin/miner/miner_test.go
rkowalick: All of the return values of the actor's method are actually being checked, but it is nonetheless very confusing:

The `uint8` return value from the actor method is checked in the assertion on `res.Receipt.ExitCode` above.

The `errror` returned from the actor method is checked in the assert on `err` above.

actor/builtin/miner/miner_test.go
rkowalick: The code has been cleaned up to eliminate the number of return values. Hopefully this is now resolved?

rkowalick: Everything is a revert error now, so we renamed the test.

actor/builtin/miner/miner.go
```diff
@@ -542,6 +554,51 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 	return 0, nil
 }
 
+// VerifyPieceInclusion verifies that proof proves that the data represented by commP is included in the sector.
+// This method returns nothing if the verification succeeds and returns a revert error if verification fails.
+func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, sectorID uint64, proof []byte) (uint8, error) {
```
laser: COMMENT: I was surprised to see that the PIP was an input to this method and not part of the `commitSector` message. I was surprised to see this because it means that a (CommP, PIP) tuple will need to go on chain each time `VerifyPieceInclusion` is called. The `commitSector` will appear on chain only once per SectorID. I **thought** that `verifyPieceInclusion` would be included in the chain each time funds would be released. I was wrong.

What I learned is that a `verifyPieceInclusion` message will be posted on chain only when the storage client challenges the miner to demonstrate that their piece has been included in a sector for which they've generated a (recent) PoSt. Challenging will cost storage client, and we don't expect them to do this often. The best case scenario will see `verifyPieceInclusion` included in the chain 0 times per CommP/SectorID pair. Because of this, we do not think that it makes much sense to include the CommP/PIP mappings in the `commitSector` message (which, of course, will always go on-chain).

actor/builtin/miner/miner.go
```diff
@@ -542,6 +554,51 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
+
+		// If miner has not committed sector id, proof is invalid
+		sectorIDstr := strconv.FormatUint(sectorID, 10)
+		commitment, ok := state.SectorCommitments[sectorIDstr]
+		if !ok {
+			return nil, errors.NewRevertError("sector not committed")
+		}
+
+		// If miner is not up-to-date on their PoSts, proof is invalid
+		if state.LastPoSt == nil {
```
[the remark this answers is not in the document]
laser: After talking to @acruikshank - I learned that this was his intent.

PR #2642, opened 2019-04-22
Improve repo version checking

### LINKED ISSUE #2587: Make go-filecoin detect what repo version it needs, check on-disk repo version
### Description
For repo migrations, go-filecoin must be able to detect what version of a repo it needs, and gracefully shut down with an appropriate error message if the repo needs to be migrated.

### Acceptance criteria
- [x] go-filecoin runs exactly as before if there is no need for a migration
- [x] if go-filecoin detects that a repo version != its desired version, it quits with an appropriate message
- [x] go-filecoin should do version detection as early as possible
- [x] Even though it's an edge case for a binary to be behind a repo version, implement correct detection of whether the binary is ahead or behind the repo version, and output an appropriate message ("I can't predict the future, sorry" /jk), since the comparison is easy enough.

### Risks + pitfalls
??

### Where to begin
commands/daemon.go `daemonRun` is probably where this would happen, if not then close by.

ZenGround0:
Closes #2587 

A lot of this was already done.
This PR simply improves error messages in the case the version is invalid for a given binary.
(It also does some unrelated cleanup)

### DIFF
--- a/commands/daemon.go
+++ b/commands/daemon.go
@@ -55,13 +55,13 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment)
 	}
 
 	// second highest precedence is env vars.
-	if envapi := os.Getenv("FIL_API"); envapi != "" {
-		rep.Config().API.Address = envapi
+	if envAPI := os.Getenv("FIL_API"); envAPI != "" {
+		rep.Config().API.Address = envAPI
 	}
 
 	// highest precedence is cmd line flag.
-	if apiAddress, ok := req.Options[OptionAPI].(string); ok && apiAddress != "" {
-		rep.Config().API.Address = apiAddress
+	if flagAPI, ok := req.Options[OptionAPI].(string); ok && flagAPI != "" {
+		rep.Config().API.Address = flagAPI
 	}
 
 	if swarmAddress, ok := req.Options[SwarmAddress].(string); ok && swarmAddress != "" {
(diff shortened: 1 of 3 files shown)


commands/daemon.go
```diff
@@ -55,13 +55,13 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment)
 	}
 
 	// second highest precedence is env vars.
-	if envapi := os.Getenv("FIL_API"); envapi != "" {
-		rep.Config().API.Address = envapi
+	if envAPI := os.Getenv("FIL_API"); envAPI != "" {
```
ZenGround0: I thought this was more go-y ¯\_(ツ)_/¯

commands/daemon.go
```diff
@@ -55,13 +55,13 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment)
...
+		rep.Config().API.Address = envAPI
 	}
 
 	// highest precedence is cmd line flag.
-	if apiAddress, ok := req.Options[OptionAPI].(string); ok && apiAddress != "" {
-		rep.Config().API.Address = apiAddress
+	if flagAPI, ok := req.Options[OptionAPI].(string); ok && flagAPI != "" {
```
ZenGround0: This matches `envAPI`'s variable naming convention better.

repo/fsrepo.go
```diff
@@ -336,7 +340,7 @@ func (r *FSRepo) loadVersion() (uint, error) {
 
 	version, err := strconv.Atoi(strings.Trim(string(file), "\n"))
 	if err != nil {
-		return 0, err
+		return 0, errors.New("corrupt version file: version is not an integer")
```
frrist: Do we want to `Wrap` this error instead? Might be wise unless we are sure `not an integer` is the only type of error we can get back from `Atoi`. 

The assertion in the test below can still be made using `Contains`.

ZenGround0: Good point.  It's not exactly the only type you can get back from [Atoi](https://golang.org/src/strconv/atoi.go?s=5007:5039#L192).  You can also get a range error if the integer is too big.

I am plowing ahead with this because filecoin will not have a repo version bigger than maxint in my lifetime and being slightly less precise is worth the cost of hiding the internally messy `strconv.Atoi: parsing "xxx": invalid syntax` error message from the user.

PR #2644, opened 2019-04-22
Add condition to payment voucher

acruikshank:
### Problem

In general the payers of payment channels should be able to make payments conditional on something provable on chain. In specific clients need to make payments to miner conditional on whether miners are storing their files.

### Solution

This PR lays the groundwork for conditional vouchers by introducing a `condition` field to payment vouchers. It saves testing conditions and creating meaningful conditions for future work.

All of the domain specific work of proving storage will happen in the miner actor, so the payment broker (and payment voucher) don't need to know any of the details about storing files. The payment voucher just needs to contain enough information so that the payment broker can send a message to another actor (the miner.Actor) and get a true/false value back to let it know if the condition is true.

This PR introduces the `Predicate` type that contains the minimal amount of information needed to send a message from within the VM (to address, method, and parameters). Since it is a parameter to the `Redeem` and `Close` methods, it needs to be an `abi` type.

### DIFF
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -63,15 +63,15 @@ func init() {
 
 // PaymentChannel records the intent to pay funds to a target account.
 type PaymentChannel struct {
-	Target         address.Address    `json:"target"`
-	Amount         *types.AttoFIL     `json:"amount"`
-	AmountRedeemed *types.AttoFIL     `json:"amount_redeemed"`
+	Target         address.Address `json:"target"`
+	Amount         *types.AttoFIL  `json:"amount"`
+	AmountRedeemed *types.AttoFIL  `json:"amount_redeemed"`
 	// AgreedEol is the expiration for the payment channel agreed upon by the
 	// payer and payee upon initialization or extension
-	AgreedEol      *types.BlockHeight `json:"agreed_eol"`
+	AgreedEol *types.BlockHeight `json:"agreed_eol"`
 	// Eol is the actual expiration for the payment channel which can differ from
 	// AgreedEol when the payment channel is in dispute
-	Eol            *types.BlockHeight `json:"eol"`
+	Eol *types.BlockHeight `json:"eol"`
 }
 
 // Actor provides a mechanism for off chain payments.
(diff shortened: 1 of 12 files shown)


actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -753,10 +755,11 @@ func (sys *system) applySignatureMessage(target address.Address, amtInt uint64,
 	require := require.New(sys.t)
 
 	amt := types.NewAttoFILFromFIL(amtInt)
-	signature, err := sys.Signature(amt, validAt)
+	signature, err := sys.Signature(amt, validAt, nil)
 	require.NoError(err)
 
-	pdata := core.MustConvertParams(sys.payer, sys.channelID, amt, validAt, signature)
+	var predicate *types.Predicate
+	pdata := core.MustConvertParams(sys.payer, sys.channelID, amt, validAt, predicate, signature)
```
anorth: Blocking: some test somewhere here must verify that non-empty condition is correctly encoded in the signed voucher.

porcelain/payment_channel_test.go
```diff
@@ -100,6 +100,7 @@ func TestPaymentChannelVoucher(t *testing.T) {
 			types.NewChannelID(5),
 			types.NewAttoFILFromFIL(10),
 			types.NewBlockHeight(0),
+			nil,
```
anorth: Blocking: populate a non-empty value and check it below.

types/payment_voucher.go
```diff
@@ -19,6 +26,7 @@ type PaymentVoucher struct {
 	Amount    AttoFIL         `json:"amount"`
 	ValidAt   BlockHeight     `json:"valid_at"`
 	Signature Signature       `json:"signature"`
+	Condition *Predicate      `json:"condition"`
```
anorth: Please take this opportunity to document all the fields in these structs.

actor/builtin/paymentbroker/paymentbroker_test.go
acruikshank: I've added tests to ensure vouchers are generated correctly with conditions, and that voucher signing includes the condition.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -194,12 +194,12 @@ func (pb *Actor) CreateChannel(vmctx exec.VMContext, target address.Address, eol
 // target Redeem(200)          -> Payer: 1000, Target: 200, Channel: 800
 // target Close(500)           -> Payer: 1500, Target: 500, Channel: 0
 //
-func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, sig []byte) (uint8, error) {
```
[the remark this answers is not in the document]
acruikshank: Currently the voucher parameter does nothing in this method aside from allowing the signature to validate. This will change very soon. I'll be sure to document the condition parameter once its are actually used.

PR #2649, opened 2019-04-23
Badge update shouldn't be point of failure in CircleCI

ognots:
Moves badge update functionality to a bash script and re-orders CircleCI jobs. See commit messages and linked issue for details.


scripts/update-badge.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
anorth: Please add a comment explaining the intent of this script.

PR #2650, opened 2019-04-23
Do not exclude any linter errors

travisperson:
golangci-lint chooses to ignore a few linter errors by default. One of these was the doc comments from golint which is why we haven't seen these issues fail ci. This PR disables this configuration and doesn't silent any linter errors.

There are a things that still need proper comments, I've noted these with a `TODO(author)`. Comment on the line with whatever I should replace the placeholder with and I'll update the PR.

### DIFF
--- a/porcelain/protocol.go
+++ b/porcelain/protocol.go
@@ -11,6 +11,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// ProtocolParams TODO(rosa)
 type ProtocolParams struct {
 	AutoSealInterval uint
 	SectorSizes      []uint64
@@ -21,6 +22,7 @@ type protocolParamsPlumbing interface {
 	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
 }
 
+// ProtocolParameters TODO(rosa)
 func ProtocolParameters(ctx context.Context, plumbing protocolParamsPlumbing) (*ProtocolParams, error) {
 	autoSealIntervalInterface, err := plumbing.ConfigGet("mining.autoSealIntervalSeconds")
 	if err != nil {
--- a/repo/fsrepo_test.go
+++ b/repo/fsrepo_test.go
@@ -80,8 +80,9 @@ func TestFSRepoInit(t *testing.T) {
 
 	dir, err := ioutil.TempDir("", "")
 	assert.NoError(t, err)
-
-	defer os.RemoveAll(dir)
+	defer func() {
+		require.NoError(t, os.RemoveAll(dir))
+	}()
 
 	t.Log("init FSRepo")
 	assert.NoError(t, InitFSRepo(dir, config.NewDefaultConfig()))
@@ -125,7 +126,9 @@ func TestFSRepoOpen(t *testing.T) {
 	t.Run("[fail] repo version newer than binary", func(t *testing.T) {
 		dir, err := ioutil.TempDir("", "")
 		assert.NoError(t, err)
-		defer os.RemoveAll(dir)
+		defer func() {
+			require.NoError(t, os.RemoveAll(dir))
+		}()
 
 		assert.NoError(t, InitFSRepo(dir, config.NewDefaultConfig()))
 		// set wrong version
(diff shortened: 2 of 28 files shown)


types/porep_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
 package types
 
+// PoRepProofPartitions TODO(laser)
```
laser: PoRepProofPartitions represents the number of partitions used when creating a PoRep proof, and impacts the size of the proof.

types/porep_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
...
 type PoRepProofPartitions uint64
 
 const (
+	// TestPoRepProofPartitions TODO(laser)
```
laser: TestPoRepProofPartitions is an opaque value signaling that an unknown number of partitions were used when creating a PoRep proof in test mode.

types/porep_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
...
 	TestPoRepProofPartitions = PoRepProofPartitions(iota)
+	// TwoPoRepPartitions TODO(laser)
```
laser: TwoPoRepPartitions indicates that two partitions were used to create a PoRep proof.

types/sector_size.go
```diff
@@ -6,6 +6,9 @@ package types
 type SectorSize uint64
 
 const (
+	// OneKiBSectorSize TODO(laser)
```
laser: OneKiBSectorSize indicates a sector which, after sealing, contains 1024 bytes.

types/sector_size.go
```diff
@@ -6,6 +6,9 @@ package types
...
 	OneKiBSectorSize = SectorSize(iota)
+
+	// TwoHundredFiftySixMiBSectorSize TODO(laser)
```
laser: TwoHundredFiftySixMiBSectorSize indicates a sector which, after sealing, contains 256MiB.

types/post_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
 package types
 
+// PoStProofPartitions TODO(laser)
```
laser: PoStProofPartitions represents the number of partitions used when creating a PoSt proof, and impacts the size of the proof.

types/post_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
...
 type PoStProofPartitions uint64
 
 const (
+	// TestPoStPartitions TODO(laser)
```
laser: TestPoStPartitions is an opaque value signaling that an unknown number of partitions were used when creating a PoSt proof in test mode.

types/post_proof_partitions.go
```diff
@@ -1,8 +1,11 @@
...
 	TestPoStPartitions = PoStProofPartitions(iota)
+	// OnePoStPartition TODO(laser)
```
laser: OnePoStPartition indicates that a single partition was used to create a PoSt proof.

porcelain/protocol.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// ProtocolParams TODO(rosa)
```
rosalinekarr: ProtocolParams are a commonly inspected set of parameters for the storage and retrieval protocols that we make available for the user to inspect via the `go-filecoin protocol` command.

porcelain/protocol.go
```diff
@@ -21,6 +22,7 @@ type protocolParamsPlumbing interface {
 	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
 }
 
+// ProtocolParameters TODO(rosa)
```
rosalinekarr: ProtocolParameters exposes certain commonly inspected parameters of the storage and retrieval protocols for use by the user and elsewhere in the application.

repo/fsrepo_test.go
```diff
@@ -80,8 +80,9 @@ func TestFSRepoInit(t *testing.T) {
 
 	dir, err := ioutil.TempDir("", "")
 	assert.NoError(t, err)
-
-	defer os.RemoveAll(dir)
+	defer func() {
+		require.NoError(t, os.RemoveAll(dir))
+	}()
```
anorth: It would be worth factoring this out to a function, but you don't have to do that here.

ZenGround0: Along the same lines I have a slight preference for putting this all on one lines but that's purely a style opinion.

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -520,7 +520,7 @@ func (l *Localfilecoin) SwarmAddrs() ([]string, error) {
 
 /** Config Interface **/
 
-// GetConfig returns the nodes config.
+// Config returns the nodes config.
```
ZenGround0: super nit: throw an apostrophe in there

repo/fsrepo_test.go
travisperson: Ya, I think we can come back and make a good-first-issue to do this work.

PR #2651, opened 2019-04-24
Repo migrations: CLI support for repo options

shannonwells:
# Problem
Repo migrations CLI needs to accept configuration for location of old and new fs repos, and if these options are not set, then the right fallbacks should be used.

Edit:  This is part one for copying the repo to the new directory to be migrated. The actual copying functionality will come in the second PR.

# Solution
* Greatly expanded usage
* Support CLI options for old repo and new repo prefix.
* EDIT:  --old-repo is now _REQUIRED_
* Since we want to timestamp and migration-version stamp a repo copy to migrate, the new-repo option is used as a prefix for the directory name.
* Create a RepoWrangler interface,  as deps for MigrationRunner, to manage access to the file system, and also to be replaced for testing MigrationRunner.   
<del>* The default RepoHelper, RepoMigrationHelper, uses `repo.GetRepoDir(cliOpt)` the same way FSRepo does</del>


tools/migration/internal/repo_migration_helper.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// GetOldRepo returns the old repo dir, opened as read-only.
+func (rmh *RepoMigrationHelper) GetOldRepo() (*os.File, error) {
+	return os.Open(getOldRepoPath(rmh.oldRepoPath))
+}
+
+// MakeNewRepo creates the new repo dir and returns it with Read/Write access.
+func (rmh *RepoMigrationHelper) MakeNewRepo() (*os.File, error) {
+	if err := os.Mkdir(rmh.newRepoPath, os.ModeDir|0744); err != nil {
```
shannonwells: should this be 0700?

tools/migration/internal/migration.go
```diff
@@ -30,17 +30,27 @@ type Migration interface {
...
+	rmh     RepoHelper
 }
 
-func NewMigrationRunner(verb bool) *MigrationRunner {
-	return &MigrationRunner{verbose: verb}
+func NewMigrationRunner(verb bool, command string, rmh RepoHelper) *MigrationRunner {
+	return &MigrationRunner{
+		verbose: verb,
+		command: command,
+		rmh:     rmh,
```
anorth: `rmh` confused me for a long time, because the term "repo migration helper" only appears in the concrete implementation, not the interfaces. Suggest `helper` or `repos`.

tools/migration/internal/rmh_whitebox_test.go
```diff
@@ -0,0 +1,46 @@
...
+		assert.Regexp(rgx, newpath)
+	})
+
+	t.Run("Adds a timestamp to the new repo dir", func(t *testing.T) {
+		dirname := "/tmp/myfilecoindir"
+		newpath := getNewRepoPath(dirname, "", "1", "2")
+		rgx, err := regexp.Compile("/tmp/myfilecoindir_1_2_[0-9]{4}-[0-9]{2}-[0-9]{2}_[0-9]{6}")
+		require.NoError(err)
+		assert.Regexp(rgx, newpath)
+	})
```
anorth: Separating these "white box" tests from the interface tests makes things a bit harder to follow, and I don't think is achieving much. They're exercising functionality that is declared by the interfaces/comments: i.e. externally observable. The directory names are observable from `File.Name()`.

Roll these tests in with `repo_migration_helper_test.go` so one place provides a comprehensive exercise of the functionality.

tools/migration/internal/repo_migration_helper.go
anorth: I think it's fine.

Maybe be explicit that this will fail if such a directory already exists.

tools/migration/internal/migration.go
```diff
@@ -30,17 +30,27 @@ type Migration interface {
 	Validate(oldRepo, newRepo *os.File) error
 }
 
+type RepoHelper interface {
+	GetOldRepo() (*os.File, error)
```
ZenGround0: I'm concerned that this method / general pattern is not providing the right thing for our migrations.  Most migrations will need to do their own thing with a repo and an `*os.File` is too opinionated an abstraction to write migrations simply.  For example, in the first simple migration I've been thinking about converting chain metadata from json to cbor.  To do this the migrating function should open a datastore, read all of a certain kind of key and rewrite the data at that key in the new format.  To do this the migration really needs a path to get a datastore object from `badgerds.NewDatastore`, an os.File *could* work, but the most straightforward thing you would do with it is read the path using `Read` or `Readdir`.

Not saying we need to do exactly what ipfs does, but the prior art here makes a lot of sense to me.  [Migration interfaces](https://github.com/ipfs/fs-repo-migrations/blob/master/go-migrate/migrate.go#L14) are given the power to [read paths](https://github.com/ipfs/fs-repo-migrations/blob/master/go-migrate/cli.go#L12)

ZenGround0: Correcting myself: the keys should be read from the old datastore and written to the new datastore.  This implies that it is also better to define access to the new repo in terms of a path.

tools/migration/internal/repo_migration_helper.go
```diff
@@ -0,0 +1,64 @@
...
+//  oldVersion:  old repo version
+//  newVersion:  version to migrate to
+// Returns:  a path generated using all of the above information plus a timestamp.
+// Example output:   /Users/davonte/.filecoin_1_2_2019-08-06_150455
+func getNewRepoPath(oldPath, newRepoOpt, oldVersion, newVersion string) string {
+	var newRepoDir string
+	if newRepoOpt != "" {
+		newRepoDir = newRepoOpt
+	} else {
+		newRepoDir = oldPath
```
ZenGround0: I think that this nests the new repo within the old one which seems like a problem.

I think it is ok to leave in a bad state until after #2524 lands because that work will make this work out as intended.

tools/migration/internal/migration.go
ZenGround0: Going back to the design doc I think this isn't such a big change to the interface.  I think we just need to change
```golang
// Migrate runs the migration steps on a copy of the repo
func (mig *Migrator) Migrate(newRepo os.File,) error
// Validate returns error if migration tests failed, describing why
func (mig *Migrator) Validate(oldRepo, newRepo os.File) error
```

to 

```
// Migrate runs the migration steps on a copy of the repo
func (mig *Migrator) Migrate(newRepoPath string,) error
// Validate returns error if migration tests failed, describing why
func (mig *Migrator) Validate(oldRepoPath, newRepoPath string) error
```

tools/migration/internal/repo_migration_helper.go
ZenGround0: Note that work in #2524 will break the assumption that the repo to be migrated will be at the path returned by `getOldRepoPath` as written, we'll need to add on a `/repo` somewhere either directly in `getOldRepoPath` or in a new function of the `go-fileocin/repo` module

tools/migration/internal/utils.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"os/user"
+	"strings"
+	"time"
+)
+
+// NowString creates an ISO8601-like timestamp string
+func NowString() string {
+	now := time.Now()
+	return now.Format("2006-01-02_150405")
```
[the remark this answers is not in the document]
shannonwells: I was going to argue in favor of keeping it, but having the dashes left out of the date actually isn't terrible to read.  I have a tremendous dislike for names that are not instantly parseable by humans.

tools/migration/internal/migration.go
shannonwells: Ok, I guess I misunderstood a previous comment from @anorth  (which I'm too lazy to hunt down) -- I thought we specifically did not want to give individual migrations access to the file system, but control the type of access and hand them file pointers.

tools/migration/internal/migration.go
```diff
@@ -30,17 +30,27 @@ type Migration interface {
...
+	MakeNewRepo() (*os.File, error)
+}
```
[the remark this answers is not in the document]
shannonwells: I created this helper because of the outcome of this thread: https://filecoinproject.slack.com/archives/CEHHJNJS3/p1555977375188400

tools/migration/internal/rmh_whitebox_test.go
shannonwells: I initially wrote unit tests trying to observe from `File.Name()` and one of the problems this causes, is you must spend a lot of time ensuring that files are created, ensuring that they are deleted afterward. Fully testing this function through an exported function is what drove the question about testing with environment variables, when what I really wanted to test was the behavior of getNewRepoPath, not the ability to create and destroy files.

tools/migration/internal/repo_migration_helper.go
```diff
@@ -0,0 +1,64 @@
...
+		newRepoPath: getNewRepoPath(oldPath, newRepoOpt, oldVersion, newVersion),
+		newVersion:  newVersion,
+		oldRepoPath: oldPath,
+		oldVersion:  oldVersion,
+	}
+}
+
+// GetOldRepo returns the old repo dir, opened as read-only.
+func (rmh *RepoMigrationHelper) GetOldRepo() (*os.File, error) {
+	return os.Open(getOldRepoPath(rmh.oldRepoPath))
```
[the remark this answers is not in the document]
shannonwells: Whoops, good catch, thanks

tools/migration/internal/utils.go
```diff
@@ -0,0 +1,25 @@
...
+// if there is no initial ~ , it returns dirname.
+func ExpandHomedir(dirname string) string {
+	if strings.LastIndex(dirname, "~") != 0 {
+		return dirname
+	}
+
+	usr, _ := user.Current()
+	home := usr.HomeDir
+	return strings.Replace(dirname, "~", home, 1)
+}
```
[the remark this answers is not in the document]
shannonwells: I had looked at go-homedir, but I just now saw a note in the README about how os/user requires `cgo`, which hurts cross-platform compatibility, so that's an additional good reason.

tools/migration/internal/migration.go
ZenGround0: I vaguely remember that too and I can see us wanting to do something to restrict what the migration can do.  If we want to go down that road we'll need to fine tune which interfaces the migration has access to so that we can restrict but in a way that is readily consumed by the migration.  For example the interface would need to return a collection of datastores, files etc.  This gets pretty messy because the interface might end up needing to become migration dependent.

I would argue that the whole song and dance around copying the entire repo and only giving the path of the copy to the migration gets us both isolation and the needed flexibility.

shannonwells: If it only gets the path to the copy, how does it validate the results though?

PR #2657, opened 2019-04-24
Remove assert.New and require.New

rkowalick:
Remove `assert.New` and `require.New` from the codebase.


ZenGround0:
I don't like it but will not block.

travisperson:
Whats the reason for not wanting to use `assert.New` and `require.New`?

acruikshank:
Here are a few justifications for this PR:

1. Defining a variable that shadows the package is bad practice.
2. using `assert.New` and `require.New` is adding hundreds of lines to our test code.
3. We're doing it wrong in many of our tests. If you include a `t.Run` within a test you must define new `assert` and `require` for a new value of `t`, but we often use the `assert` and `require` in closure scope. Using it incorrectly makes our tests unsafe under parallelization (I fixed a lot of this in our flaky test phase). Using these correctly adds a lot of boilerplate when you're trying to write a lot of quick, idiomatic, well documented test cases off and existing setup.
4. The people that established this convention are no longer on the project. I was never given any explanation for why we standardized on the OO versions of this code. If we default to the functional style, there will be nothing wrong with using the OO style occasionally (though I challenge you to provide a reason beyond the extra 3 characters in the assertions).
5. It's tedious. Writing a test you start by adding the `assert.New`, then you start writing your test and realize you need a require, so you go back up and add the `require.New`, then you realize that you want all `require`s so you have to go up and delete the `assert.New`. It's a really small amount of effort, it should be a small amount of effort multiplied over a lot of testing. This is a great place to reduce friction.

This is something I've been wanting to propose for a long time.

PR #2658, opened 2019-04-24
WIP Separate sector data from repo

ZenGround0:
- separate repo interface into Repo and SectorRepo
- fsrepo handles repo data at the new <root>/repo path

### DIFF
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -1,0 +1,62 @@
+package paths
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// node repo path defaults
+const filPathVar = "FIL_PATH"
+const defaultHomeDir = "~/.filecoin"
+const defaultRepoDir = "repo"
+
+// node sector storage path defaults
+const filSectorPathVar = "FIL_SECTOR_PATH"
+const defaultSectorDir = "sectors"
+const defaultSectorStagingDir = "staging"
+const defaultSectorSealingDir = "sealed"
+
+// GetRepoPath returns the path of the filecoin repo from a potential override
+// string, the FIL_PATH environment variable and a default of ~/.filecoin/repo.
+func GetRepoPath(override string) string {
+	// override is first precedence
+	if override != "" {
+		return override
+	}
+	// Environment variable is second precedence
+	envRepoDir := os.Getenv(filPathVar)
+	if envRepoDir != "" {
+		return envRepoDir
+	}
+	// Default is third precedence
+	return filepath.Join(defaultHomeDir, defaultRepoDir)
+}
+
+// GetSectorPath returns the path of the filecoin sector storage from a
+// potential override string, the FIL_SECTOR_PATH environment variable and a
+// default of ~/.filecoin/sectors.
+func GetSectorPath(override string) string {
+	// override is first precedence
+	if override != "" {
+		return override
+	}
+	// Environment variable is second precedence
+	envRepoDir := os.Getenv(filSectorPathVar)
+	if envRepoDir != "" {
+		return envRepoDir
+	}
+	// Default is third precedence
+	return filepath.Join(defaultHomeDir, defaultSectorDir)
+}
+
+// StagingDir returns the path to the sector staging directory given the sector
+// storage directory path.
+func StagingDir(sectorPath string) string {
+	return filepath.Join(sectorPath, defaultSectorStagingDir)
+}
+
+// StagingDir returns the path to the sector sealed directory given the sector
+// storage directory path.
+func SealedDir(sectorPath string) string {
+	return filepath.Join(sectorPath, defaultSectorSealingDir)
+}
--- a/repo/fsrepo.go
+++ b/repo/fsrepo.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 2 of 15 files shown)


ZenGround0:
@anorth @shannonwells it seemed like all of this was covered heavily enough by tests but please let me know if there are areas that you would like to see tested.

ZenGround0:
@laser just added you to make sure I'm not reking any rust-fil-proofs guarantees, no need to look at the whole thing.

repo/fsrepo_test.go
```diff
@@ -178,9 +186,10 @@ func TestFSRepoRoundtrip(t *testing.T) {
 	cfg := config.NewDefaultConfig()
 	cfg.API.Address = "foo" // testing that what we get back isnt just the default
 	assert.NoError(err, InitFSRepo(dir, cfg))
-
+	fmt.Printf("expected dir %v\n", dir)
```
shannonwells: remove debug print

repo/fsrepo_test.go
```diff
@@ -178,9 +186,10 @@ func TestFSRepoRoundtrip(t *testing.T) {
...
 	r, err := OpenFSRepo(dir)
 	assert.NoError(err)
+	fmt.Printf("r.repoPath: %s, r.rootPath: %s\n", r.repoPath, r.rootPath)
```
shannonwells: same

repo/fsrepo.go
```diff
@@ -493,12 +518,17 @@ func (r *FSRepo) SetAPIAddr(maddr string) error {
 	return nil
 }
 
+func APIAddrOfRoot(rootDir string) (string, error) {
```
shannonwells: There were some other checks in this code you abstracted -- were they not necessary?
```
rawPath := filepath.Join(filepath.Clean(repoDir), repo.APIFile)	
		apiFilePath, err := homedir.Expand(rawPath)	
		if err != nil {	
			return "", errors.Wrap(err, fmt.Sprintf("can't resolve local repo path %s", rawPath))	
		}
```

laser:
@ZenGround0 

> @laser just added you to make sure I'm not reking any rust-fil-proofs guarantees, no need to look at the whole thing.

I don't see anything that's going to wreck us. Thanks for asking.

ZenGround0:
@travisperson tagging to make sure I didn't ruin any iptb stuff.  No need to look at the whole thing.

repo/fsrepo.go
ZenGround0: I think this is a bug, thanks for catching.

repo/fsrepo.go
```diff
@@ -67,41 +69,42 @@ type FSRepo struct {
...
+	rootDir = GetHomeDir(rootDir)
+	if err := InitFSRepo(rootDir, cfg); err != nil {
 		return nil, err
 	}
-	return OpenFSRepo(repoDir)
+	return OpenFSRepo(rootDir)
 }
 
 // OpenFSRepo opens an already initialized fsrepo at the given path
 func OpenFSRepo(p string) (*FSRepo, error) {
```
anorth: Change `p` and the comment to clearly refer to `homePath` not the repo path.

Possibly better would be to take a repo path here and push the expansion and joining up to the caller.

repo/fsrepo.go
```diff
@@ -169,12 +172,18 @@ func (r *FSRepo) loadFromDisk() error {
 
 // InitFSRepo initializes an fsrepo at the given path using the given configuration
 func InitFSRepo(p string, cfg *config.Config) error {
-	expath, err := homedir.Expand(p)
+	homePath, err := homedir.Expand(p)
```
anorth: Ditto

repo/fsrepo.go
```diff
@@ -458,19 +467,35 @@ func fileExists(file string) bool {
...
+func (r *FSRepo) SealedDir() (string, error) {
+	// Default is to keep sector data alongside node repo data.
+	if r.cfg.SectorBase.RootDir == "" {
+		return path.Join(r.homePath, "sectors/sealed"), nil
+	}
+	// If not using default check for writeability
+	if err := checkWritable(r.cfg.SectorBase.RootDir); err != nil {
+		return "", nil
+	}
+	return path.Join(r.cfg.SectorBase.RootDir, "sealed"), nil
```
anorth: As noted elsewhere, remove this responsibility from "fsrepo". I would suggest the permission checks are out of place here, too.

ZenGround0:
@anorth I've pushed us to use a model where users specify the repo dir directly as opposed to the home dir.  If this is still has not gone far enough lets sync on what we want for users when it comes to default directories and configurations of home/repo/sector dirs.

paths/paths.go
```diff
@@ -0,0 +1,62 @@
+package paths
```
anorth: This probably shouldn't be a top-level package, but thanks for breaking it out. I don't have a concrete better suggestion at the moment.

repo/mem.go
```diff
@@ -114,26 +90,9 @@ func (mr *MemRepo) Version() uint {
...
-func (mr *MemRepo) SealedDir() string {
-	return mr.sealedDir
-}
-
-// CleanupSectorDirs removes all sector directories and their contents.
-func (mr *MemRepo) CleanupSectorDirs() {
-	os.RemoveAll(mr.StagingDir()) // nolint: errcheck
-	os.RemoveAll(mr.SealedDir())  // nolint:errcheck
-}
-
```
anorth: 👍

Issue #2625, opened 2019-04-19
Miner Actor verifies piece inclusion


acruikshank:
### Description

Our payment channels will now require miners to prove storage before collecting payment. This proof needs to happen in the miner actor. Add a `VerifyPiece` method that given a client data commitment (CommP), A miner data commitment (CommD), and a piece inclusion proof (PIP) to `miner.Actor` needs to confirm the following:

1. The piece inclusion proof proves that the data represented by CommP is in the data represented by CommD.
2. The miner includes CommD among their committed sectors.
3. The difference between the miner's `LastPoSt` and now does not differ by more than `ClientProofOfStorageTimeout` blocks.

The method must return nil if the proof is verified and a string verification error otherwise.

### Acceptance criteria

1. The method accurately verifies the three points listed above.
2. The method returns a short, useful error message when verification fails.
3. All of this is well tested.

PR #2668, opened 2019-04-25
Check predicate on voucher redeem

### LINKED ISSUE #2656: check condition payment channel redeem and close
### Description

If a payment voucher contains a condition, the payment broker must check it before allowing the redeem or close to continue. This involves sending a message to another actor. If the message does not error, the check is considered to be valid. If it returns an error, the voucher is considered to be in an invalid state.

The first target of this functionality is the `verifyPieceInclusion` method of miner actors.

### Acceptance criteria

1. Given a voucher with a condition, redeem and close only succeed if the actor and method and params specified by the condition can be called successfully.
2. A condition can be created that calls `verifyPieceInclusion`. This will require some joint testing of paymentbroker and a miner actor.

rkowalick:
Resolves #2656

### Problem

In order to create a payment system in a low trust environment, we must support payers tying payment vouchers to conditions that can be verified on chain. The first example of this is that storage clients should be able to require to prove piece inclusion before redeeming storage payments.

### Solution

The mechanism is to add a specification to the payment voucher for a message to be sent to another actor which can determine the validity of the condition by erroring out when it fails. Conditions will be based on 3 sources:

1. Parameters about the condition supplied by the payer.
2. Additional parameters supplied by the redeemer. This probably always includes a proof that can be verified by the other actor.
3. Chain state. This will usually be actor state of the actor asked to validate the condition.

In the storage client case:

1. The storage client creates a condition to ask the storage miner's miner actor to verify that the miner is storing their piece (`verifyPieceInclusion`).
2. The storage client provides `CommP` for the piece.
3. When the storage miner redeems the voucher, they need to supply two additional parameters: the sector id of the sector that contains the piece and a piece inclusion proof (supplied as a byte slice) that will prove that `CommP` is contained within the sector.
4. When this message is sent to the miner actor the actor will check its state to find whether the sector exists and the miner's proofs are up-to-date. It then checks whether the pip is valid. It returns a revert error if any of these things are false (implemented in #2625).

#### This PR includes the following

1. Changes `Predicate` to take an `[]interface{}` instead of a `[]byte` for parameters. This keeps the actor from having to decode and reencode parameter lists.
2. Adds parameter lists ([]interface{}) to abi.
3. Adds a `redeemerSuppliedParams` parameter to `paymentbroker.Redeem` and `paymentbroker.Close`
4. Has `paymentbroker.Redeem` and `paymentbroker.Close` concatenate parameters from the voucher condition with the `redeemerSuppliedParams` and send a message to the address and method supplied by the condition (if there is a condition). If the message fails, the `Redeem` or `Close` fails
5. Adds payment broker tests to confirm this mechanism works against a test actor.
6. Adds a `miner_redeem` test (a unit test that integrates `paymentbroker` with `miner.Actor`) to confirm that this mechanism works with the `miner.Actor.verifyPieceInclusion`.

### DIFF
--- a/actor/builtin/integration_test.go
+++ b/actor/builtin/integration_test.go
@@ -6,10 +6,9 @@ import (
 
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-hamt-ipld"
-	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-ipfs-blockstore"
 	"github.com/stretchr/testify/require"
 
-	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/actor/builtin"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker"
@@ -33,31 +32,37 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
 	addrGetter := address.NewForTestGetter()
 	target := addrGetter()
 	defaultValidAt := types.NewBlockHeight(uint64(0))
-	//sectorID := uint64(123)
 	_, st, vms := requireGenesis(ctx, t, target)
 
-	minerActor := actor.NewActor(types.MinerActorCodeCid, types.ZeroAttoFIL)
-	storage := vms.NewStorage(address.TestAddress, minerActor)
-	minerState := miner.State{SectorCommitments: {"STRING": {CommD: [32]byte{}, CommR: [32]byte{}, CommRStar: [32]byte{}}}}
-	require.NoError(t, miner.Actor{}.InitializeState(storage, minerState))
-	require.NoError(t, st.SetActor(ctx, address.TestAddress, minerActor))
+	minerAddr := addrGetter()
+	sectorID := uint64(123)
+	commP := []byte{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
+	commD := []byte{0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7}
+	lastPoSt := types.NewBlockHeight(10)
+
+	require.NoError(t, createMinerWithCommitment(ctx, st, vms, minerAddr, sectorID, commD, lastPoSt))
 
 	payerActor := th.RequireNewAccountActor(require.New(t), types.NewAttoFILFromFIL(50000))
 	state.MustSetActor(st, payer, payerActor)
 
 	channelID := establishChannel(ctx, st, vms, payer, target, 0, types.NewAttoFILFromFIL(1000), types.NewBlockHeight(20000))
 
-	toAddress := addrGetter()
-
-	t.Run("Redeem should succeed if condition is met", func(t *testing.T) {
-		condition := &types.Predicate{To: toAddress, Method: "checkParams", Params: []interface{}{addrGetter(), uint64(6)}}
+	t.Run("Voucher with piece inclusion condition and correct proof succeeds", func(t *testing.T) {
+		// create voucher with piece inclusion condition
+		condition := &types.Predicate{To: minerAddr, Method: "verifyPieceInclusion", Params: []interface{}{commP}}
 
 		amt := types.NewAttoFILFromFIL(100)
 		sig, err := paymentbroker.SignVoucher(channelID, amt, defaultValidAt, payer, condition, mockSigner)
 		require.NoError(t, err)
 		signature := ([]byte)(sig)
 
-		pdata := core.MustConvertParams(payer, channelID, amt, types.NewBlockHeight(0), condition, signature, []interface{}{types.NewBlockHeight(43)})
+		// make redeem request with correct sector id and PIP
+		// TODO: this pip is very fake
+		pip := []byte{}
+		pip = append(pip, commP[:]...)
+		pip = append(pip, commD[:]...)
+		suppliedParams := []interface{}{sectorID, pip}
+		pdata := core.MustConvertParams(payer, channelID, amt, types.NewBlockHeight(0), condition, signature, suppliedParams)
 		msg := types.NewMessage(target, address.PaymentBrokerAddress, 0, types.NewAttoFILFromFIL(0), "redeem", pdata)
 
 		appResult, err := th.ApplyTestMessage(st, vms, msg, types.NewBlockHeight(0))
@@ -65,6 +70,25 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, appResult.ExecutionError)
 	})
+}
+
+func createMinerWithCommitment(ctx context.Context, st state.Tree, vms vm.StorageMap, minerAddr address.Address, sectorId uint64, commD []byte, lastPoSt *types.BlockHeight) error {
+	minerActor := miner.NewActor()
+	storage := vms.NewStorage(minerAddr, minerActor)
+
+	commitments := map[string]types.Commitments{}
+	commD32 := [32]byte{}
+	copy(commD32[:], commD)
+	commitments[string(sectorId)] = types.Commitments{CommD: commD32, CommR: [32]byte{}, CommRStar: [32]byte{}}
+	minerState := &miner.State{
+		SectorCommitments: commitments,
+		LastPoSt:          lastPoSt,
+	}
+	executableActor := miner.Actor{}
+	if err := executableActor.InitializeState(storage, minerState); err != nil {
+		return err
+	}
+	return st.SetActor(ctx, minerAddr, minerActor)
 }
 
 func requireGenesis(ctx context.Context, t *testing.T, targetAddresses ...address.Address) (*hamt.CborIpldStore, state.Tree, vm.StorageMap) {
--- a/actor/builtin/integration_test.go
+++ b/actor/builtin/integration_test.go
@@ -2,6 +2,7 @@ package builtin_test
 
 import (
 	"context"
+	"strconv"
 	"testing"
 
 	"github.com/ipfs/go-datastore"
@@ -72,14 +73,15 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
 	})
 }
 
-func createMinerWithCommitment(ctx context.Context, st state.Tree, vms vm.StorageMap, minerAddr address.Address, sectorId uint64, commD []byte, lastPoSt *types.BlockHeight) error {
+func createMinerWithCommitment(ctx context.Context, st state.Tree, vms vm.StorageMap, minerAddr address.Address, sectorID uint64, commD []byte, lastPoSt *types.BlockHeight) error {
 	minerActor := miner.NewActor()
 	storage := vms.NewStorage(minerAddr, minerActor)
 
 	commitments := map[string]types.Commitments{}
 	commD32 := [32]byte{}
 	copy(commD32[:], commD)
-	commitments[string(sectorId)] = types.Commitments{CommD: commD32, CommR: [32]byte{}, CommRStar: [32]byte{}}
+	sectorIDstr := strconv.FormatUint(sectorID, 10)
+	commitments[sectorIDstr] = types.Commitments{CommD: commD32, CommR: [32]byte{}, CommRStar: [32]byte{}}
 	minerState := &miner.State{
 		SectorCommitments: commitments,
 		LastPoSt:          lastPoSt,
--- a/actor/builtin/integration_test.go
+++ b/actor/builtin/integration_test.go
@@ -46,7 +46,7 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
 	payerActor := th.RequireNewAccountActor(require.New(t), types.NewAttoFILFromFIL(50000))
 	state.MustSetActor(st, payer, payerActor)
 
-	channelID := establishChannel(ctx, st, vms, payer, target, 0, types.NewAttoFILFromFIL(1000), types.NewBlockHeight(20000))
+	channelID := establishChannel(st, vms, payer, target, 0, types.NewAttoFILFromFIL(1000), types.NewBlockHeight(20000))
 
 	t.Run("Voucher with piece inclusion condition and correct proof succeeds", func(t *testing.T) {
 		// create voucher with piece inclusion condition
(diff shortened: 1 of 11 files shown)


actor/builtin/integration_test.go
```diff
@@ -0,0 +1,213 @@
+package builtin_test
```
rkowalick: Maybe change the name of this file?!

actor/builtin/integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	commD := []byte{0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7}
+	lastPoSt := types.NewBlockHeight(10)
+
+	require.NoError(t, createMinerWithCommitment(ctx, st, vms, minerAddr, sectorID, commD, lastPoSt))
+
+	payerActor := th.RequireNewAccountActor(require.New(t), types.NewAttoFILFromFIL(50000))
+	state.MustSetActor(st, payer, payerActor)
+
+	channelID := establishChannel(st, vms, payer, target, 0, types.NewAttoFILFromFIL(1000), types.NewBlockHeight(20000))
+	pip := []byte{}
```
rkowalick: Comment above this line with `// TODO: this pip is very fake`

actor/builtin/integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	defaultValidAt := types.NewBlockHeight(uint64(0))
+	_, st, vms := requireGenesis(ctx, t, target)
+
+	minerAddr := addrGetter()
+	sectorID := uint64(123)
+	commP := []byte{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
+	commD := []byte{0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7}
+	lastPoSt := types.NewBlockHeight(10)
+
+	require.NoError(t, createMinerWithCommitment(ctx, st, vms, minerAddr, sectorID, commD, lastPoSt))
```
rkowalick: This through the establishChannel channel should be right before the `t.Run`s

actor/builtin/integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	tf.UnitTest(t)
+
+	var mockSigner, _ = types.NewMockSignersAndKeyInfo(10)
+
+	ctx := context.Background()
+	payer := mockSigner.Addresses[0]
+	addrGetter := address.NewForTestGetter()
+	target := addrGetter()
+	defaultValidAt := types.NewBlockHeight(uint64(0))
+	_, st, vms := requireGenesis(ctx, t, target)
```
rkowalick: Move this down too

actor/builtin/integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/filecoin-project/go-filecoin/vm"
+)
+
+func TestVerifyPieceInclusionInRedeem(t *testing.T) {
+	tf.UnitTest(t)
+
+	var mockSigner, _ = types.NewMockSignersAndKeyInfo(10)
+
+	ctx := context.Background()
```
rkowalick: describe what all of this setup does

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -38,6 +38,8 @@ const (
 	ErrInvalidSignature = 42
 	//ErrTooEarly indicates that the block height is too low to satisfy a voucher
 	ErrTooEarly = 43
+	//ErrConditionInvalid indicates that the condition attached to a voucher did not execute successfully
+	ErrConditionInvalid = 43
```
rkowalick: iota?

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -205,7 +207,9 @@ func (pb *Actor) CreateChannel(vmctx exec.VMContext, target address.Address, eol
 // target Redeem(200)          -> Payer: 1000, Target: 200, Channel: 800
 // target Close(500)           -> Payer: 1500, Target: 500, Channel: 0
 //
-func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate, sig []byte) (uint8, error) {
+// If a condition is provided in the voucher, concatenate its params with supplied params to send a message.
+// Any non-fault error is considered a validation failure.
```
rkowalick: `Any non-fault error when the condition is run is considered a validation failure.`

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/vm/errors"
+	"github.com/ipfs/go-cid"
```
rkowalick: wrong order

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -107,6 +112,68 @@ func TestPaymentBrokerUpdate(t *testing.T) {
...
+	tf.UnitTest(t)
+
+	addrGetter := address.NewForTestGetter()
+	toAddress := addrGetter()
+
+	sys := setup(t)
+	require.NoError(t, sys.st.SetActor(context.TODO(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
+
+	t.Run("Redeem should succeed if condition is met", func(t *testing.T) {
+		condition := &types.Predicate{To: toAddress, Method: "checkParams", Params: []interface{}{addrGetter(), uint64(6)}}
```
rkowalick: ¿Don't put the changes that break the test inline?

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -814,17 +913,16 @@ func (sys *system) retrieveChannel(paymentBroker *actor.Actor) *PaymentChannel {
 	return channel
 }
 
-func (sys *system) applySignatureMessage(target address.Address, amtInt uint64, validAt *types.BlockHeight, nonce uint64, method string, height uint64) (*consensus.ApplicationResult, error) {
+func (sys *system) applySignatureMessage(target address.Address, amtInt uint64, validAt *types.BlockHeight, nonce uint64, method string, height uint64, condition *types.Predicate, suppliedParams ...interface{}) (*consensus.ApplicationResult, error) {
```
anorth: These helper methods are really hard to follow and poorly documented. Could you leave these in a better state than you found them by documenting the ones touched? Adding more args here has made things slightly less transparent.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -205,7 +207,9 @@ func (pb *Actor) CreateChannel(vmctx exec.VMContext, target address.Address, eol
 // target Redeem(200)          -> Payer: 1000, Target: 200, Channel: 800
 // target Close(500)           -> Payer: 1500, Target: 500, Channel: 0
 //
-func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate, sig []byte) (uint8, error) {
+// If a condition is provided in the voucher, concatenate its params with supplied params to send a message.
```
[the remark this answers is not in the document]
ZenGround0: I think this comment is trying to explain that internally the Redeem function will send a message that includes the combination of redeemer supplied parameters and the payer supplied parameters within the input condition.  I think putting this comment about internals is misleading, at least without painstakingly explaining the condition checking message call which feels out of scope of this comment, and should be moved down to the `checkCondition` call.

`redeemerSuppliedParams` might be better called `redeemerWitness` (or `redeemerProof` but proof is overloaded everywhere) because params is too general and confusing because these params are getting passed into the `Redeem` function as params but the intent is that they are params of the condition checking internal message call.  If we really want to keep params we could be more specific: `redeemerConditionCheckingParams`

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -255,7 +263,10 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 
 // Close first executes the logic performed in the the Update method, then returns all
 // funds remaining in the channel to the payer account and deletes the channel.
-func (pb *Actor) Close(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate, sig []byte) (uint8, error) {
+//
+// If a condition is provided in the voucher, concatenate its params with supplied params to send a message.
```
ZenGround0: Same comment out of scope issue here.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -626,26 +641,37 @@ const separator = 0x0
...
-func createVoucherSignatureData(channelID *types.ChannelID, amount *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate) []byte {
+func createVoucherSignatureData(channelID *types.ChannelID, amount *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate) ([]byte, error) {
 	data := append(channelID.Bytes(), separator)
 	data = append(data, amount.Bytes()...)
 	data = append(data, separator)
 	if condition != nil {
 		data = append(data, condition.To.Bytes()...)
 		data = append(data, []byte(condition.Method)...)
-		data = append(data, condition.Params...)
+		encodedParams, err := abi.ToEncodedValues(condition.Params...)
```
ZenGround0: Out of curiosity: isn't this an argument for just encoding params as bytes once and for all on entrance to the VM?  That way we don't have to deal with this re-encoding and possible error?

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -709,3 +735,19 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
 
 	return actor.LoadTypedLookup(ctx, storage, byChannelCID, &PaymentChannel{})
 }
+
+func checkCondition(vmctx exec.VMContext, condition *types.Predicate, redeemerSuppliedParams []interface{}) (uint8, error) {
+	if condition == nil {
+		return 0, nil
+	}
+	params := append(condition.Params[:0:0], condition.Params...)
```
ZenGround0: huh cool, I erroneously thought the python slice copy: `params := condition.Params[:]` would work in go.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -709,3 +735,19 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
...
+	params = append(params, redeemerSuppliedParams...)
+	_, _, err := vmctx.Send(condition.To, condition.Method, types.NewZeroAttoFIL(), params)
+	if err != nil {
+		if errors.IsFault(err) {
+			return errors.CodeError(err), err
+		}
+		return ErrConditionInvalid, errors.RevertErrorWrap(err, "failed to validate voucher condition")
```
ZenGround0: Dumb question: is it ok for this error to absorb out of gas errors?  I guess the error is wrapped so the user will see that the voucher isn't necessarily invalid, its just they ran out of gas?

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -205,7 +207,11 @@ func (pb *Actor) CreateChannel(vmctx exec.VMContext, target address.Address, eol
 // target Redeem(200)          -> Payer: 1000, Target: 200, Channel: 800
 // target Close(500)           -> Payer: 1500, Target: 500, Channel: 0
 //
-func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate, sig []byte) (uint8, error) {
+// If a condition is provided in the voucher:
+// - The parameters provided in the condition will be combined with redeemerSuppliedParams
+// - A message will be sent to the the condition.To address using the condition.Method with the combined params
+// - If the message returns an error the condition is considered to be false and the redeem will fail
+func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types.ChannelID, amt *types.AttoFIL, validAt *types.BlockHeight, condition *types.Predicate, sig []byte, redeemerSuppliedParams []interface{}) (uint8, error) {
```
ZenGround0: The name of `redeemerSuppliedParams` could be improved.  The new comment goes a long way at explaining how they are used but IMO the name itself should specify what these are the params of.  `redeemerWitness` (or `redeemerProof` but proof is too overloaded).  If you want to keep params we could be more specific `redeemerSuppliedConditionCheckingParams`, or maybe `redeemerConditionWitness`.

actor/builtin/miner_redeem_test.go
```diff
@@ -0,0 +1,220 @@
...
+	var mockSigner, _ = types.NewMockSignersAndKeyInfo(10)
+	payer := mockSigner.Addresses[0]
+	payerActor := th.RequireNewAccountActor(require.New(t), types.NewAttoFILFromFIL(50000))
+	state.MustSetActor(st, payer, payerActor)
+
+	// Create a payment channel from payer -> target
+	channelID := establishChannel(st, vms, payer, target, 0, types.NewAttoFILFromFIL(1000), types.NewBlockHeight(20000))
+
+	// Make a pip
+	// TODO: This pip is very Fake
```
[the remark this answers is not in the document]
acruikshank: referenced.

actor/builtin/paymentbroker/paymentbroker.go
acruikshank: This is pretty confusing, but we don't actually call this method within the VM. Signing happens outside the VM to save on chain computation, but it lives in this file to keep related code together. 

We do only encode and decode parameters once as they cross the VM boundary. We use this method to encode prior to signing, because the fact that these parameters will be encoded this way as they cross the boundary ensures that this mechanism will work for signing.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -709,3 +735,19 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
 
 	return actor.LoadTypedLookup(ctx, storage, byChannelCID, &PaymentChannel{})
 }
+
+func checkCondition(vmctx exec.VMContext, condition *types.Predicate, redeemerSuppliedParams []interface{}) (uint8, error) {
```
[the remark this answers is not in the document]
acruikshank: documented

actor/builtin/paymentbroker/paymentbroker.go
acruikshank: Yeah, we had to research this too.

actor/builtin/paymentbroker/paymentbroker.go
acruikshank: If the condition runs out of gas, we should see something like "failed to validate voucher condition: Insufficient gas". I think this is a pretty accurate description of the problem.

There are actually a lot of ways this message can fail that don't involve the condition being invalid. Hopefully we've covered most of them in the tests. The burden is on the target to ensure that payment vouchers have valid conditions and they call the methods with enough gas. We'll have two features to support this:

1. We'll add a method to payment broker to test a condition off chain with a query method.
2. We'll have storage miners automatically check all payment vouchers have valid conditions once they have all the data (correct miner address, correct method, and correct CommP).

actor/builtin/paymentbroker/paymentbroker_test.go
acruikshank: documented

PR #2672, opened 2019-04-26
Repo migration functionality: Copy old repo, install new repo #2643

### LINKED ISSUE #2643: Implement repo duplication & installation in repo migration runner
### Description
Basic structure + stubs for migration runner, the layer under the repo migration CLI, plus functionality for copying and swapping out the "migrated" copy

### Acceptance criteria
- [x] CLI instantiates and calls MigrationRunner.Run when called with a valid command
- [x] MigrationRunner duplicates repo
- [x] MigrationRunner correctly swaps out the duplicated copy

It's fine if this is demonstrated only in the tests.

### Risks + pitfalls
None

### Where to begin
tools/migration/internal/migration.go

shannonwells:
Closes #2643 
This updates and adds more file system wrangling functionality to `RepoFSWrangler`:

* Changes `MakeNewRepo` to do the actual copying and permissions setting on the new repo.
* Adds `InstallNewRepo` which renames the old repo and symlinks the new repo to the old repo dir.  The new name for the old repo dir is returned so that it can be logged.
* Updates the require/assert usage to stop calling `New`
* changes some of the unit tests to use different functions for file system operations
* deletes the `ExpandHomedir` function since it's not used any more, because the `--old-repo` CLI param is required now.

Open questions:
* @ZenGround0 and I tried to come up with less vague names than `RepoMigrationHelper` and `RepoFSWrangler` is what we "settled" on as good enough for now. It does cover what it's doing: convenience functions for touching the file system, but seriously, ??? 
* I feel like I'm trying to predict what will be required of the wrangler before it has been used.  This is my best guess at the interface.  Input requested.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,6 +69,7 @@ require (
 	github.com/multiformats/go-multistream v0.0.1
 	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/opencontainers/image-spec v1.0.1 // indirect
+	github.com/otiai10/copy v1.0.1
 	github.com/pkg/errors v0.8.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/polydawn/refmt v0.0.0-20190221155625-df39d6c2d992
--- a/tools/migration/internal/repo_fs_wrangler.go
+++ b/tools/migration/internal/repo_fs_wrangler.go
@@ -40,30 +40,33 @@ func (rmh *RepoFSWrangler) GetNewRepoPath() string {
 	return rmh.newRepoPath
 }
 
-// MakeNewRepo creates the new repo dir and returns it with Read/Write access.
-//   This returns an error if the directory exists.
-func (rmh *RepoFSWrangler) MakeNewRepo() (*os.File, error) {
-	if err := os.Mkdir(rmh.newRepoPath, os.ModeDir|0744); err != nil {
-		return nil, err
-	}
-	return os.Open(rmh.newRepoPath)
-}
-
-func (rmh *RepoFSWrangler) CopyRepo() error {
-	return rcopy.Copy(rmh.oldRepoPath, rmh.newRepoPath)
-}
-
-func (rmh *RepoFSWrangler) InstallNewRepo() error {
-	// rename the old dir:
-	archiveName := strings.Join([]string{rmh.oldRepoPath, rmh.oldVersion, NowString()}, "-")
-	if err := os.Rename(rmh.oldRepoPath, archiveName); err != nil {
+// MakeNewRepo copies the old repo to the new repo dir with Read/Write access.
+//   Returns an error if the directory exists.
+func (rmh *RepoFSWrangler) MakeNewRepo() error {
+	if err := rcopy.Copy(rmh.oldRepoPath, rmh.newRepoPath); err != nil {
 		return err
 	}
-	// symlink the new dir
+
+	return os.Chmod(rmh.newRepoPath, os.ModeDir|0744)
+}
+
+// InstallNewRepo renames the old repo and symlinks the new repo at the old name.
+// returns the new path to the old repo and any error.
+func (rmh *RepoFSWrangler) InstallNewRepo() (string, error) {
+	archivedRepo := strings.Join([]string{rmh.oldRepoPath, rmh.oldVersion, NowString()}, "-")
+
+	if err := os.Rename(rmh.oldRepoPath, archivedRepo); err != nil {
+		return archivedRepo, err
+	}
+	if err := os.Chmod(archivedRepo, os.ModeDir|0444); err != nil {
+		return archivedRepo, err
+	}
+
 	if err := os.Symlink(rmh.newRepoPath, rmh.oldRepoPath); err != nil {
-		return err
+		return archivedRepo, err
 	}
-	return nil
+
+	return archivedRepo, nil
 }
 
 // getNewRepoPath generates a new repo path for a migration.
--- a/tools/migration/internal/repo_fs_wrangler.go
+++ b/tools/migration/internal/repo_fs_wrangler.go
@@ -3,12 +3,14 @@ package internal
 import (
 	"os"
 	"strings"
+
+	rcopy "github.com/otiai10/copy"
 )
 
 // RepoFSWrangler manages filesystem operations and figures out what the correct paths
 // are for everything.
 type RepoFSWrangler struct {
-	oldRepoPath, newRepoPath string
+	oldVersion, oldRepoPath, newRepoPath string
 }
 
 // NewRepoFSWrangler takes options for old and new repo paths, figures out
@@ -18,6 +20,7 @@ func NewRepoFSWrangler(oldRepoOpt, newRepoPrefixOpt, oldVersion, newVersion stri
 	return &RepoFSWrangler{
 		newRepoPath: getNewRepoPath(oldRepoOpt, newRepoPrefixOpt, oldVersion, newVersion),
 		oldRepoPath: oldRepoOpt,
+		oldVersion:  oldVersion,
 	}
 }
 
@@ -44,6 +47,23 @@ func (rmh *RepoFSWrangler) MakeNewRepo() (*os.File, error) {
 		return nil, err
 	}
 	return os.Open(rmh.newRepoPath)
+}
+
+func (rmh *RepoFSWrangler) CopyRepo() error {
+	return rcopy.Copy(rmh.oldRepoPath, rmh.newRepoPath)
+}
+
+func (rmh *RepoFSWrangler) InstallNewRepo() error {
+	// rename the old dir:
+	archiveName := strings.Join([]string{rmh.oldRepoPath, rmh.oldVersion, NowString()}, "-")
+	if err := os.Rename(rmh.oldRepoPath, archiveName); err != nil {
+		return err
+	}
+	// symlink the new dir
+	if err := os.Symlink(rmh.newRepoPath, rmh.oldRepoPath); err != nil {
+		return err
+	}
+	return nil
 }
 
 // getNewRepoPath generates a new repo path for a migration.
--- a/tools/migration/internal/repo_fs_wrangler.go
+++ b/tools/migration/internal/repo_fs_wrangler.go
@@ -1,6 +1,8 @@
 package internal
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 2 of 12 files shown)


tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -3,12 +3,14 @@ package internal
...
+
+	rcopy "github.com/otiai10/copy"
 )
 
 // RepoFSWrangler manages filesystem operations and figures out what the correct paths
 // are for everything.
 type RepoFSWrangler struct {
-	oldRepoPath, newRepoPath string
+	oldVersion, oldRepoPath, newRepoPath string
 }
```
shannonwells: oldVersion is retained in order to rename/archive the old repo.

tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -3,12 +3,14 @@ package internal
...
 	"strings"
+
+	rcopy "github.com/otiai10/copy"
 )
 
 // RepoFSWrangler manages filesystem operations and figures out what the correct paths
 // are for everything.
 type RepoFSWrangler struct {
-	oldRepoPath, newRepoPath string
+	oldVersion, oldRepoPath, newRepoPath string
```
shannonwells: oldVersion is used for renaming/archiving the old repo.

tools/migration/internal/repo_fs_wrangler_test.go
```diff
@@ -1,13 +1,14 @@
...
+	"io/ioutil"
 	"os"
+	"path"
 	"regexp"
 	"testing"
 
-	"github.com/golangci/golangci-lint/pkg/fsutils"
-	ast "github.com/stretchr/testify/assert"
-	req "github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
shannonwells: this was a bad Goland auto-import that I missed in the previous PR.  Thanks to @rkowalick and @acruikshank for catching it

tools/migration/internal/repo_fs_wrangler_test.go
```diff
@@ -16,71 +17,110 @@ import (
...
+	require.NoError(t, rmh.MakeNewRepo())
+	dir, err := os.Open(rmh.GetNewRepoPath())
+	require.NoError(t, err)
+	stat, err := dir.Stat()
+	require.NoError(t, err)
+	expectedPerms := "drwxr--r--"
+	assert.Equal(t, expectedPerms, stat.Mode().String())
+}
+
+func TestRepoFSWrangler_InstallNewRepo(t *testing.T) {
```
shannonwells: Mainly this test is to make sure I correctly used go file system calls, but also that permissions were set correctly.

tools/migration/main_test.go
```diff
@@ -7,80 +7,76 @@ import (
 	"testing"
 
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
-	ast "github.com/stretchr/testify/assert"
-	req "github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
shannonwells: Following the new testing conventions for assertions.

go.mod
```diff
@@ -69,6 +69,7 @@ require (
 	github.com/multiformats/go-multistream v0.0.1
 	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/opencontainers/image-spec v1.0.1 // indirect
+	github.com/otiai10/copy v1.0.1
```
shannonwells: I looked over this package and it seems legit, not much to it and does the right thing simply and seems to be properly done.

tools/migration/internal/repo_fs_wrangler_test.go
```diff
@@ -16,71 +17,110 @@ import (
...
-		require.NoError(err)
-		assert.Regexp(rgx, newpath)
+		require.NoError(t, err)
+		assert.Regexp(t, rgx, newpath)
 	})
 }
 
-func mustMakeTmpDir(require *req.Assertions, dirname string) {
-	require.NoError(os.Mkdir(dirname, os.ModeDir|os.ModeTemporary|os.ModePerm))
+func TestRepoFSWrangler_MakeNewRepo(t *testing.T) {
```
anorth: `tf.UnitTest(t)` here and below

tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -37,13 +40,33 @@ func (rmh *RepoFSWrangler) GetNewRepoPath() string {
...
 }
 
-// MakeNewRepo creates the new repo dir and returns it with Read/Write access.
-//   This returns an error if the directory exists.
-func (rmh *RepoFSWrangler) MakeNewRepo() (*os.File, error) {
-	if err := os.Mkdir(rmh.newRepoPath, os.ModeDir|0744); err != nil {
-		return nil, err
+// MakeNewRepo copies the old repo to the new repo dir with Read/Write access.
+//   Returns an error if the directory exists.
+func (rmh *RepoFSWrangler) MakeNewRepo() error {
```
anorth: Rename this to more clearly imply the copy operation, e.g. `CloneRepo`. Return the path to the clone. As I noted above, I think this should find a free directory name rather than fail, but if not then take the intended migration version number here rather than at construction.

tools/migration/internal/runner.go
```diff
@@ -36,7 +36,7 @@ type RepoWrangler interface {
 	// GetOldRepo opens and returns the old repo with read-only access
 	GetOldRepo() (*os.File, error)
 	// MakeNewRepo creates and returns the new repo dir with read/write permissions
-	MakeNewRepo() (*os.File, error)
+	MakeNewRepo() error
 	// GetOldRepoPath returns the full path of the old repo
 	GetOldRepoPath() string
```
anorth: The name "old" might cause confusion here. After installation, the "old" repo is at an archived path, and the new migrated repo is at the old path. Maybe something like "target" or "canonical"?

tools/migration/internal/repo_fs_wrangler.go
ZenGround0: +1 `CloneRepo` is clearer
+1 version number seems unimportant in name of the copy
+1 that if we do want the version number here the wrangler can read it within this subfunction, probably by calling a `RepoVersion` method on the wrangler

tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -37,13 +40,33 @@ func (rmh *RepoFSWrangler) GetNewRepoPath() string {
...
+
+	if err := os.Rename(rmh.oldRepoPath, archivedRepo); err != nil {
+		return archivedRepo, err
+	}
+	if err := os.Chmod(archivedRepo, os.ModeDir|0444); err != nil {
+		return archivedRepo, err
 	}
-	return os.Open(rmh.newRepoPath)
+
+	if err := os.Symlink(rmh.newRepoPath, rmh.oldRepoPath); err != nil {
```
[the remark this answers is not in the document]
shannonwells: > Just rename the archived copy to the canonical repo name after archiving the old one out of the way.
If we are going to require `--old-repo` as a parameter, and we are not going to use `FIL_PATH` or `~/.filecoin`, then I think we have to respect that this is where they want their filecoin directory to be located, and putting the new one in `~/.filecoin` (the "canonical repo name") will result in unexpected and possibly deleterious behavior.

> Was there another reason to use symlinks?
Yes:  hard links cannot be applied to directories:
```
$ ln bin bin2
ln: bin: Is a directory
```

tools/migration/internal/repo_fs_wrangler_test.go
```diff
@@ -16,71 +17,110 @@ import (
...
+
+	// check that the new repo is at the old location.
+	dir, err = os.Open(rmh.GetOldRepoPath())
+	require.NoError(t, err)
+	contents, err = dir.Readdirnames(0)
+	require.NoError(t, err)
+	assert.Contains(t, contents, "newRepoFile")
+}
+
+func mustMakeTmpDir(t *testing.T, dirname string) string {
```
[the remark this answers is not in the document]
shannonwells: right thanks, I keep getting that backward.

tools/migration/internal/repo_fs_wrangler.go
shannonwells: >One unspoken design decision is whether the wrangler should logically be re-usable, or tightly coupled to a specific migration process – I hadn't seen this decision before, am now angling for the former, but can be convinced the latter is better.

I think the wrangler code ought to be reusable; otherwise there will be a lot of needlessly duplicate code.

shannonwells: > But can't it read it out of the repo? What happens if they conflict? Is it checked?
Do we want the wrangler to try to parse anything in the repo?  I figured that checking would happen outside of the wrangler.

tools/migration/internal/runner.go
shannonwells: +1 to calling it tmp first and live after

travisperson:
How does issue #2684 impact these changes?

shannonwells:
@travisperson this issue drove the creation of #2684; @anorth and I were discussing the mechanics of setting a new repo path, and decided to expect that the repo path is a symlink which points to a version-named directory; then the node can keep the same value for FIL_PATH and we just swap out what it points to after migration, the same way that other packages are done, e.g. postgres.

tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -1,68 +1,88 @@
...
+	archivedRepoPath := strings.Join([]string{realRepoPath, "archived", NowString()}, "-")
 
-// MakeNewRepo creates the new repo dir and returns it with Read/Write access.
-//   This returns an error if the directory exists.
-func (rmh *RepoFSWrangler) MakeNewRepo() (*os.File, error) {
-	if err := os.Mkdir(rmh.newRepoPath, os.ModeDir|0744); err != nil {
-		return nil, err
+	if err := os.Rename(realRepoPath, archivedRepoPath); err != nil {
+		return archivedRepoPath, err
+	}
```
anorth: Blocking: this should happen *after* the symlink is switched, else subsequent failure leaves the link broken.

shannonwells: I assume you mean the rename and not the link removal, so if we're not renaming then it's not an issue.

tools/migration/internal/repo_fs_wrangler.go
```diff
@@ -1,68 +1,88 @@
...
 //     newRepoOpt:  whatever was passed in by the CLI (can be blank)
-//     oldVersion:  old repo version
-//     newVersion:  version to migrate to
 // Returns:
-//     a path generated using all of the above information plus a timestamp.
+//     a path generated using the above information plus tmp_<timestamp>.
 // Example output:
-//     /Users/davonte/.filecoin_1_2_20190806-150455
-func getNewRepoPath(oldPath, newRepoOpt, oldVersion, newVersion string) string {
+//     /Users/davonte/.filecoin_tmp_20190806-150455-001
```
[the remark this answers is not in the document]
shannonwells: It's not included, I just forgot to update the comment.

PR #2681, opened 2019-04-29
Feat/uncancel payment channel 2622

### LINKED ISSUE #2622: Payment Channel Redeem uncancels payment channel
### Description

Payment channels can now be canceled by the payer. This is a challenge to the target to prove something (usually piece inclusion). The target can uncancel the channel by redeeming a valid voucher. 

The challenge is resolved by requiring a proof for a voucher to be valid, but this is not a requirement for this work. Any valid voucher will undo the cancel. 

It may be convenient to uncancel the channel on 'close' as well, but since the channel will close in any case, its not a requirement.

### Acceptance criteria

1. If a payment channel has been canceled, and a subsequent redeem is successful, the payment channel's eol is restored.
2. The payment channel does not expire early

rosalinekarr:
# Problem

Payment Channel Cancellation currently has no recourse. If a client cancels an accepted payment channel, the miner has no way to dispute said cancellation when complying with the original agreement.

# Solution

Change Payment Channel Redeem to automatically reset the EOL of a channel to the originally agreed upon EOL as long as the piece inclusion proof for said redemption is valid.

Resolves #2622

### DIFF
--- a/actor/builtin/paymentbroker/paymentbroker_test.go
+++ b/actor/builtin/paymentbroker/paymentbroker_test.go
@@ -200,20 +200,28 @@ func TestPaymentBrokerRedeemReversesCancellations(t *testing.T) {
 
 	sys := setup(t)
 
+	// Cancel the payment channel
 	pdata := core.MustConvertParams(sys.channelID)
 	msg := types.NewMessage(sys.payer, address.PaymentBrokerAddress, 1, types.NewAttoFILFromFIL(1000), "cancel", pdata)
-
 	result, err := sys.ApplyMessage(msg, 100)
 	require.NoError(t, result.ExecutionError)
 	require.NoError(t, err)
 	require.Equal(t, uint8(0), result.Receipt.ExitCode)
 
+	// Expect that the EOL of the payment channel now reflects the cancellation
+	paymentBroker := state.MustGetActor(sys.st, address.PaymentBrokerAddress)
+	channel := sys.retrieveChannel(paymentBroker)
+	assert.Equal(t, types.NewBlockHeight(20000), channel.AgreedEol)
+	assert.Equal(t, types.NewBlockHeight(10100), channel.Eol)
+
+	// Redeem the payment channel
 	result, err = sys.ApplyRedeemMessageWithBlockHeight(sys.target, 500, 0, 10000)
 	require.NoError(t, err)
 
-	paymentBroker := state.MustGetActor(sys.st, address.PaymentBrokerAddress)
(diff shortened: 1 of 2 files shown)


actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -195,6 +195,52 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
+		require.NoError(t, err)
+
+		paymentBroker := state.MustGetActor(sys.st, address.PaymentBrokerAddress)
+		channel := sys.retrieveChannel(paymentBroker)
+
+		assert.Equal(t, types.NewBlockHeight(20000), channel.AgreedEol)
+		assert.Equal(t, types.NewBlockHeight(20000), channel.Eol)
+	})
+
+	t.Run("does not reset EOL on failure", func(t *testing.T) {
```
shannonwells: I don't understand enough about payment channels to see why this setup fails and the other one succeeds.

rosalinekarr: The `updateChannel` call prior to setting the EOL on line 256 of `paymentbroker.go` does the actual validation. I just added this test since my original implementation set the EOL before the `updateChannel` which breaks functionality. This test just checks that we're doing things in the right order.

acruikshank: This test is unnecessary and and somewhat confusing. The `AgreedEol` and `Eol` are stored in actor state, and actor state is always rolled back for any error condition. This test is only testing the rollback logic.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -195,6 +195,52 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
 	})
 }
 
+func TestPaymentBrokerRedeemReversesCancellations(t *testing.T) {
+	tf.UnitTest(t)
+
+	t.Run("resets EOL on success", func(t *testing.T) {
```
acruikshank: This test is correct, but requires a lot of internal knowledge to understand why. These tests should serve as documentation. Please comment the purpose of the steps, and add an intermediate assertion after the cancel for `AgreedEol` and `Eol` to demonstrate that redeem is actually resetting those values.

actor/builtin/paymentbroker/paymentbroker_test.go
rosalinekarr: Removed

PR #2682, opened 2019-04-29
expand homedir in paths

ZenGround0:
Should I go ahead and implement a `--homedir` flag right now and move everything (iptb, fast, daemon test, wiki) to specify both repo and sectors base with this flag?

Alternatively I could implement and move callers to specify a sectors dir with `--sectordir`.

We should probably do this sooner rather than later because I'm realizing that lots of different testing daemons are using the same sealed and staged dirs and this is probably not great.

cc @anorth @shannonwells @travisperson

### DIFF
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -3,6 +3,8 @@ package paths
 import (
 	"os"
 	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
 )
 
 // node repo path defaults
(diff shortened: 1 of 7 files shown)


laser:
I have a slight preference for the additional control that a `--sectordir` flag would provide.

shannonwells:
See also: https://github.com/filecoin-project/go-filecoin/issues/2684

travisperson:
> Should I go ahead and implement a `--homedir` flag right now and move everything (iptb, fast, daemon test, wiki) to specify both repo and sectors base with this flag?

> Yes, we should probably switch the flag to refer to ~/.filecoin, the home dir, and have that be the only flag or env var that most users care about.

I really do not want to do this on a whim to fix a bug. I strongly believe `FIL_PATH` should only ever point to the the `config.json` and `api` files. The configuration is used to construct a repo. The `api` file defines where talk to a daemon using the configuration to operate on the repo it defines.

Sector storage, chain storage, block storage, deal storage, etc are properties of a repo, defined by a configuration. It just so happens that these default to a location relative to the repo configuration at the moment, and at the moment are not defined by the configuration.

No where was this `--homedir` flag discussed in a design doc, it came out of an issue (#2524) on a whim due to what I preserve was due to the lack of clarity in the design doc itself (which still has TODO comments but is accepted and was never actually sent out on any official communication [per out communication guidelines](https://github.com/filecoin-project/pm/blob/master/go-filecoin/comms.md#howto-communicate-within-the-dev-team)).

This needs to be discussed before we start breaking things.

This is not a change to make on the week before we do a release.

paths/paths.go
```diff
@@ -3,11 +3,26 @@ package paths
...
 
+var log logging.EventLogger
+var defaultHomeDir string
+
+func init() {
+	log = logging.Logger("paths")
+	var err error
+	defaultHomeDir, err = homedir.Expand(defaultHomeDirUnexpanded)
+	if err != nil {
+		log.Errorf("could not expand ~/.filecoin default home directory: %s", err)
```
travisperson: Panic. We are not in a valid configuration if this check fails. Either we panic or we move this error check someplace where we can shutdown correctly.

ZenGround0:
Burned a few hours getting the FAST retrieval test working on CI and I am still running into a failure that doesn't show up on local.  Further debugging is needed to get everything working on CI with the homedir and I am  deleting the homedir as my next task.  So I'm closing this down and putting proper homedir expansion in a PR for #2689.

PR #2683, opened 2019-04-29
init takes --sectordir flag

### LINKED ISSUE #2680: [bug] Sector path config is not being setup correctly
**Describe the bug**
Slack discussion: https://filecoinproject.slack.com/archives/CEHHJNJS3/p1556559696046900

There are two parts to this bug

1) The sector path is not expanded which results in a `~/.filecoin/sector` directory being created in the current working directory where the daemon is started.

2) The sector path should default to inside of the filecoin home directory. This means if we supply a `--repodir` flag or use `FIL_PATH` during a call to `init`, the sector path should default inside of it, and only ever be moved out if explicitly defined in the configuration file.

laser:
## Why does this PR exist?

Our tests are all writing sector and sector metadata to `~/.filecoin/sectors` in spite of the fact that the test daemons are being initialized with unique repo directories. 

This bug prevents me from updating the rust-fil-proofs submodule. The new head of rust-fil-proofs contains commits which replace the JSON/filesystem sector builder metadata store with a RocksDB metadata store. The RocksDB metadata store writes a `LOCK` to the directory into which it was initialized, and since all go-filecoin test daemons were using the same directory, most go-filecoin tests fail after the update.

## What's in this PR?

Allow callers of `init` to pass `--sectordir` flag which specifies the directory into which sealed and staged sector data will be written (and sector builder metadata, too).

## Future work

This changeset is deliberately small - just enough to unblock me from integrating rust-fil-proofs. Further work is required to ensure that the `"sectors"` string isn't duplicated across the `paths` and `commands` packages.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -34,6 +34,7 @@ var initCmd = &cmds.Command{
 		cmdkit.StringOption(GenesisFile, "path of file or HTTP(S) URL containing archive of genesis block DAG data"),
 		cmdkit.StringOption(PeerKeyFile, "path of file containing key to use for new node's libp2p identity"),
 		cmdkit.StringOption(WithMiner, "when set, creates a custom genesis block with a pre generated miner account, requires running the daemon using dev mode (--dev)"),
+		cmdkit.StringOption(OptionSectorDir, "path of directory into which staged and sealed sectors will be written"),
 		cmdkit.StringOption(DefaultAddress, "when set, sets the daemons's default address to the provided address"),
 		cmdkit.UintOption(AutoSealIntervalSeconds, "when set to a number > 0, configures the daemon to check for and seal any staged sectors on an interval.").WithDefault(uint(120)),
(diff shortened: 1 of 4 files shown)


commands/init.go
```diff
@@ -81,6 +82,10 @@ var initCmd = &cmds.Command{
 func getConfigFromOptions(options cmdkit.OptMap) (*config.Config, error) {
 	newConfig := config.NewDefaultConfig()
 
+	if dir, ok := options[OptionRepoDir].(string); ok {
```
ZenGround0: Blocking: this goes against the purpose of the recent change.  `sectors` must not be under the repo dir.  It is no one's fault but mine that we are not synced on mental models but we need to converge fast or this is going to bite us.  See #2682 for ways to improve our CLI that keep the sector separate from the repo (either have an OptionSectorsDir that we pass int, or an OptionHomeDir that specifies a home path for initing sector storage and repo)

laser: That's a good point. You just did a bunch of work which allows sector stuff to live outside of the repo dir and here I go adding it back.

I've updated the PR to reflect our earlier conversation about initializing with a `--sectordir` flag.

PR #2688, opened 2019-04-30
Feat/store condition in paych 2677

### LINKED ISSUE #2677: Store Condition in Payment Channel
### Description

The redeemer of a payment channel voucher will sometimes need to supply a significant amount of data as additional parameters to a condition, and generally speaking, that data will be the same from one voucher to the next. The payment broker should store the full condition in the payment channel.

In the case of miner payments, this means storing the `verifyPieceInclusion` condition including the sector id and piece inclusion proof. These should not change from one redeem to the next (but the condition must be replayed, because the miner's LastPoSt could go out of date).

### Acceptance criteria

1. When a `Redeem` is successful, the payment broker sets a `redeemed` flag on the payment channel and stores a `condition` predicate with the concatenation of condition and `redeemerSuppliedParams` (the parameters actually sent in the message) as `condition` in the payment channel.
2. When `Redeem` or `Close` is called, and a condition exists in the voucher, and the redeemed flag is true, and `paymentchannel.condition` exists, and `redeemerConditionParams` is empty, the payment broker will construct the condition message from the cached condition.
3. In the above case, we will not overwrite the `condition`.

NOTE There is potential problem here that the payer could provide some vouchers with a condition that requires additional params, and then later vouchers that have conditions that require no parameters. If the target redeems one of the earlier vouchers successfully, and then attempts to redeem a no parameter voucher, we will try to use the old parameters and the voucher will fail.

This is not a problem for the existing storage client, storage miner scenario, and any valid case where this might happen is far fetched. If we do feel like we need to solve this, we should store a hash of the condition and only reuse parameters if the hashes match. For now this extra data on chain isn't worth it.

rosalinekarr:
# Problem

The conditions params provided to redeem a payment channel can sometimes be very large and identical from one call to another.

# Solution

Cache them on the payment channel, so they only need to be sent once.

Resolves #2677

### DIFF
--- a/abi/encode.go
+++ b/abi/encode.go
@@ -31,6 +31,10 @@ func EncodeValues(vals []*Value) ([]byte, error) {
 // DecodeValues decodes an array of abi values from the given buffer, using the
 // provided type information.
 func DecodeValues(data []byte, types []Type) ([]*Value, error) {
+	if len(types) > 0 && len(data) == 0 {
+		return nil, fmt.Errorf("expected %d parameters, but got 0", len(types))
+	}
+
 	if len(data) == 0 {
 		return nil, nil
 	}
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -84,7 +84,7 @@ type PaymentChannel struct {
 
 	// Conditions are the set of conditions for redeeming or closing the payment
 	// channel
-	Conditions *types.Predicate
+	Conditions *types.Predicate `json:"conditions"`
 
 	// Eol is the actual expiration for the payment channel which can differ from
 	// AgreedEol when the payment channel is in dispute
@@ -93,7 +93,7 @@ type PaymentChannel struct {
 	// Redeemed is a flag indicating whether or not Redeem has been called on the
 	// payment channel yet. This is necessary because AmountRedeemed can still be
 	// zero in the event of a zero-value voucher
-	Redeemed boolean
+	Redeemed bool `json:"redeemed"`
 }
 
 // Actor provides a mechanism for off chain payments.
@@ -263,6 +263,9 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 		// Reset the EOL to the originally agreed upon EOL in the event that the
 		// channel has been cancelled.
 		channel.Eol = channel.AgreedEol
+
+		// Mark the payment channel as redeemed
+		channel.Redeemed = true
 
 		return byChannelID.Set(ctx, chid.KeyString(), channel)
 	})
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -266,7 +266,9 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 
 		// Mark the payment channel as redeemed and set the conditions
 		channel.Redeemed = true
-		if condition != nil {
+		if condition == nil {
+			channel.Conditions = nil
+		} else {
 			newConditions := *condition
 			newConditions.Params = append(newConditions.Params, redeemerConditionParams...)
 			channel.Conditions = &newConditions
@@ -793,7 +795,9 @@ func checkCondition(vmctx exec.VMContext, chid *types.ChannelID, payerAddress ad
 				return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 			}
 
-			cachedCondition = channel.Conditions
+			if channel.Redeemed {
+				cachedCondition = channel.Conditions
+			}
 			return nil
 		})
 		if err != nil {
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -82,9 +82,18 @@ type PaymentChannel struct {
 	// payer and payee upon initialization or extension
 	AgreedEol *types.BlockHeight `json:"agreed_eol"`
 
+	// Conditions are the set of conditions for redeeming or closing the payment
+	// channel
+	Conditions *types.Predicate
+
 	// Eol is the actual expiration for the payment channel which can differ from
 	// AgreedEol when the payment channel is in dispute
 	Eol *types.BlockHeight `json:"eol"`
+
+	// Redeemed is a flag indicating whether or not Redeem has been called on the
+	// payment channel yet. This is necessary because AmountRedeemed can still be
+	// zero in the event of a zero-value voucher
+	Redeemed boolean
 }
 
 // Actor provides a mechanism for off chain payments.
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -264,8 +264,13 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 		// channel has been cancelled.
 		channel.Eol = channel.AgreedEol
 
-		// Mark the payment channel as redeemed
+		// Mark the payment channel as redeemed and set the conditions
 		channel.Redeemed = true
+		if condition != nil {
+			newConditions := *condition
+			newConditions.Params = append(newConditions.Params, redeemerConditionParams...)
+			channel.Conditions = &newConditions
+		}
 
 		return byChannelID.Set(ctx, chid.KeyString(), channel)
 	})
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -766,6 +766,8 @@ func checkCondition(vmctx exec.VMContext, channel *PaymentChannel, condition *ty
 		return nil
 	}
 
+	// If new params have been provided or we don't yet have a cached condition,
+	// cache the provided params and condition on the payment channel.
 	if !channel.Redeemed || channel.Condition == nil || len(redeemerSuppliedParams) > 0 {
 		newParams := condition.Params
 		newParams = append(newParams, redeemerSuppliedParams...)
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -231,7 +231,7 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 		return errors.CodeError(Errors[ErrInvalidSignature]), Errors[ErrInvalidSignature]
 	}
 
-	if errCode, err := checkCondition(vmctx, condition, redeemerConditionParams); err != nil {
+	if errCode, err := checkCondition(vmctx, chid, payer, condition, redeemerConditionParams); err != nil {
 		return errCode, err
 	}
 
@@ -304,7 +304,7 @@ func (pb *Actor) Close(vmctx exec.VMContext, payer address.Address, chid *types.
 		return errors.CodeError(Errors[ErrInvalidSignature]), Errors[ErrInvalidSignature]
 	}
 
-	if errCode, err := checkCondition(vmctx, condition, redeemerConditionParams); err != nil {
+	if errCode, err := checkCondition(vmctx, chid, payer, condition, redeemerConditionParams); err != nil {
 		return errCode, err
 	}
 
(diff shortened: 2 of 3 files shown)


actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -82,9 +82,18 @@ type PaymentChannel struct {
 	// payer and payee upon initialization or extension
 	AgreedEol *types.BlockHeight `json:"agreed_eol"`
 
+	// Conditions are the set of conditions for redeeming or closing the payment
+	// channel
+	Conditions *types.Predicate `json:"conditions"`
```
acruikshank: // Condition is a condition for ...
Condition

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -750,10 +769,45 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
...
 	if condition == nil {
 		return 0, nil
 	}
+
+	if len(redeemerSuppliedParams) == 0 {
+		ctx := context.Background()
+		storage := vmctx.Storage()
+
+		var cachedCondition *types.Predicate
+		err := withPayerChannelsForReading(ctx, storage, payerAddress, func(byChannelID exec.Lookup) error {
```
acruikshank: Instead of fetching the payment channel twice, move the condition check into updateCondition with the other validations. We can set the cached condition there as well.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -255,6 +264,16 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 		// channel has been cancelled.
 		channel.Eol = channel.AgreedEol
 
+		// Mark the payment channel as redeemed and set the conditions
+		channel.Redeemed = true
+		if condition == nil {
+			channel.Conditions = nil
+		} else {
+			newConditions := *condition
```
acruikshank: newCondition

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -255,6 +264,16 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
...
+			newConditions.Params = append(newConditions.Params, redeemerConditionParams...)
```
acruikshank: This should use the same params that were executed in the condition. All of this (except setting redeemed) should be set in `checkCondition`.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -195,6 +195,114 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NoError(t, appResult.ExecutionError)
+
+		// Expect that the conditions are now nil
+		paymentBroker = state.MustGetActor(sys.st, address.PaymentBrokerAddress)
+		channel = sys.retrieveChannel(paymentBroker)
+		assert.Nil(t, channel.Conditions)
+	})
+
+	t.Run("Redeem uses cached conditions in subsequent calls", func(t *testing.T) {
```
acruikshank: For documentation's sake, it would be good to run the 'without params' case before running the redeem with params to establish that it fails, then run it after to show it succeeds.

Also we need an executes cached condition test for the the `Close` method.

ZenGround0:
I'm confused about something basic.  Why don't we store the condition in the payment channel when the voucher is being created?  The condition data already passes through the VM into a message at this point putting it on chain.  In fact voucher bytes (including condition) are also returned as the result of `Voucher`.  Is keeping it out of the paymentbroker's state tree somehow worth it based on the expected lack of Redeem and Close calls / the smaller data of a condition vs redeemer condition parameters?  This seems contradictory to the approach of putting any redeemer condition parameters forever on chain as soon as they get passed in the first message.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -750,10 +769,45 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
 
 // checkCondition combines params in the condition with the redeemerSuppliedParams, sends a message
 // to the actor and method specified in the condition, and returns an error if one exists.
-func checkCondition(vmctx exec.VMContext, condition *types.Predicate, redeemerSuppliedParams []interface{}) (uint8, error) {
+func checkCondition(vmctx exec.VMContext, chid *types.ChannelID, payerAddress address.Address, condition *types.Predicate, redeemerSuppliedParams []interface{}) (uint8, error) {
 	if condition == nil {
 		return 0, nil
 	}
+
+	if len(redeemerSuppliedParams) == 0 {
```
ZenGround0: A comment explaining that this is the cached condition would help readers a lot.

ZenGround0: sry cached condition is overloaded, I mean the situation in which `checkCondition` will look in its cache

acruikshank:
@ZenGround0 That's a really good question. Voucher creation is handled by the payment broker, but it actually happens off-chain using a query message. So we can't actually update payment channel there. All of this is optimized to minimize on-chain data for happy path transactions. This means we can't, for example, store the condition in the payment channel when we create the channel. This would mean that every payment channel that __might__ require a condition to be redeemed would need to put the condition on chain. 

The payer can construct some vouchers (off-chain) that contain a condition and one or more that come later that don't have a condition. If the payer does not decide to cancel the payment channel and the target doesn't need the funds immediately, the target can wait for the condition-less voucher to become valid and collect the funds without ever posting the condition or their parameters, thus saving data on chain.

ZenGround0:
Thank you, totally clear!

abi/encode.go
```diff
@@ -31,6 +31,10 @@ func EncodeValues(vals []*Value) ([]byte, error) {
 // DecodeValues decodes an array of abi values from the given buffer, using the
 // provided type information.
 func DecodeValues(data []byte, types []Type) ([]*Value, error) {
+	if len(types) > 0 && len(data) == 0 {
+		return nil, fmt.Errorf("expected %d parameters, but got 0", len(types))
+	}
+
```
rosalinekarr: This was necessary due to a bug uncovered in testing. If an actor method which requires params is called without any params at all, not returning an error here causes a panic.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -750,18 +760,29 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
 
 // checkCondition combines params in the condition with the redeemerSuppliedParams, sends a message
 // to the actor and method specified in the condition, and returns an error if one exists.
-func checkCondition(vmctx exec.VMContext, condition *types.Predicate, redeemerSuppliedParams []interface{}) (uint8, error) {
+func checkCondition(vmctx exec.VMContext, channel *PaymentChannel, condition *types.Predicate, redeemerSuppliedParams []interface{}) error {
 	if condition == nil {
-		return 0, nil
+		if channel != nil {
```
acruikshank: `checkCondition` should only be called if we've found the channel. This check then becomes unnecesary.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -750,18 +760,29 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
...
+			channel.Condition = nil
+		}
+		return nil
+	}
+
+	if !channel.Redeemed || channel.Condition == nil || len(redeemerSuppliedParams) > 0 {
```
acruikshank: BLOCKING: This will panic if channel doesn't exist (wrong channel id). `checkCondition` should only be called after we've established the existence of the channel in `updateChannel`. Also these conditions need a comment to explain the logic. e.g.:

// Update payment channel condition if this is the first redeem, we haven't had a condition previously, or the
// redeemer has provided new parameters.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -750,18 +760,29 @@ func findByChannelLookup(ctx context.Context, storage exec.Storage, byPayer exec
...
-	_, _, err := vmctx.Send(condition.To, condition.Method, types.NewZeroAttoFIL(), params)
+
+	_, _, err := vmctx.Send(condition.To, condition.Method, types.NewZeroAttoFIL(), channel.Condition.Params)
 	if err != nil {
 		if errors.IsFault(err) {
-			return errors.CodeError(err), err
+			return err
 		}
-		return ErrConditionInvalid, errors.RevertErrorWrap(err, "failed to validate voucher condition")
+		return errors.RevertErrorWrap(err, "failed to validate voucher condition")
```
acruikshank: BLOCKING: This needs to be a coded error with error code `ErrConditionInvalid`.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -117,84 +117,209 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
+
+		// Expect that the condition is set and correct
+		paymentBroker := state.MustGetActor(sys.st, address.PaymentBrokerAddress)
+		channel := sys.retrieveChannel(paymentBroker)
+		assert.NotNil(t, channel.Condition)
+		assert.Equal(t, toAddress, channel.Condition.To)
+		assert.Equal(t, method, channel.Condition.Method)
+		assert.Contains(t, channel.Condition.Params, addrParam.Bytes())
+		assert.Contains(t, channel.Condition.Params, sectorIdParam)
+		assert.Contains(t, channel.Condition.Params, blockHeightParam.Bytes())
```
acruikshank: We have other tests to assert all properties of a condition are valid. It's enough to assert the condition is not nil here.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -117,84 +117,209 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
-		appResult, err := callRedeem(condition, badRedeemerParams)
+		appResult, err := sys.applySignatureMessage(sys.target, 100, types.NewBlockHeight(0), 0, "redeem", 0, condition, badRedeemerParams...)
 
 		require.NoError(t, err)
 		require.Error(t, appResult.ExecutionError)
 		require.Contains(t, appResult.ExecutionError.Error(), "failed to validate voucher condition: invalid params")
 	})
 }
 
+func TestPaymentBrokerRedeemSetsConditionAndRedeemed(t *testing.T) {
```
acruikshank: BLOCKING: This test is a really good start. It tests that we set the condition on first redeem and reuse it on the second redeem. We need tests for when the condition is updated with new params, and we need a test that the condition is updated when the first redeem had a nil condition.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -245,6 +248,10 @@ func (pb *Actor) Redeem(vmctx exec.VMContext, payer address.Address, chid *types
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		if err := checkCondition(vmctx, channel, condition, redeemerConditionParams); err != nil {
```
acruikshank: BLOCKING: This needs to go into `updateChannel` so we run the same validation logic in redeem and close and also so that we can run the validations in the right order. It might be worth renaming `updateChannel` to `validateAndUpdateChannel`.

actor/builtin/paymentbroker/paymentbroker.go
rosalinekarr: It should no longer be possible to call `checkCondition` without a channel. The prior nil check was a mistaken leftover from some previous code.

I'll add a comment.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -117,84 +117,272 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
 		badAddressParam := address.Undef
 		badParams := []interface{}{badAddressParam, sectorIdParam}
 
 		condition := &types.Predicate{To: toAddress, Method: method, Params: badParams}
-		appResult, err := callRedeem(condition, redeemerParams)
+		appResult, err := sys.applySignatureMessage(sys.target, 100, types.NewBlockHeight(0), 0, "redeem", 0, condition, redeemerParams...)
 
 		require.NoError(t, err)
 		require.Error(t, appResult.ExecutionError)
 		require.Contains(t, appResult.ExecutionError.Error(), "failed to validate voucher condition: got undefined address")
```
acruikshank: Not your issue, but please switch these to test the error code rather than the error message now that that is possible.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -117,84 +117,272 @@ func TestPaymentBrokerRedeemWithCondition(t *testing.T) {
...
+
+	t.Run("Redeem uses cached condition in subsequent calls", func(t *testing.T) {
+		sys := setup(t)
+		require.NoError(t, sys.st.SetActor(context.TODO(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
+
+		// Redeem without params and expect a panic
+		condition := &types.Predicate{To: toAddress, Method: method}
+		appResult, err := sys.applySignatureMessage(sys.target, 200, types.NewBlockHeight(0), 0, "redeem", 0, condition)
+		require.NoError(t, err)
+		require.Error(t, appResult.ExecutionError)
```
acruikshank: Also check error code here to make  sure it's failing for the right reason,

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -446,6 +637,39 @@ func TestPaymentBrokerCloseWithCondition(t *testing.T) {
...
+	require.Error(t, result.ExecutionError)
+
+	// Successfully redeem the payment channel with params
+	condition = &types.Predicate{To: toAddress, Method: method, Params: payerParams}
+	result, err = sys.applySignatureMessage(sys.target, 100, types.NewBlockHeight(0), 0, "redeem", 0, condition, redeemerParams...)
+	require.NoError(t, err)
+	require.NoError(t, result.ExecutionError)
+
+	// Close again without params and expect no error
+	result, err = sys.ApplyCloseMessage(sys.target, 200, 0)
```
acruikshank: This passes because the incoming condition is nil, not because the previous condition was stored.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -446,6 +637,39 @@ func TestPaymentBrokerCloseWithCondition(t *testing.T) {
...
+	redeemerParams := []interface{}{blockHeightParam}
+
+	sys := setup(t)
+	require.NoError(t, sys.st.SetActor(context.TODO(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
+
+	// Close without params and expect a panic
+	condition := &types.Predicate{To: toAddress, Method: method}
+	result, err := sys.applySignatureMessage(sys.target, 100, sys.defaultValidAt, 0, "close", 0, condition)
+	require.NoError(t, err)
+	require.Error(t, result.ExecutionError)
```
acruikshank: Check error code here

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -446,6 +637,39 @@ func TestPaymentBrokerCloseWithCondition(t *testing.T) {
...
+	sectorIdParam := uint64(6)
+	payerParams := []interface{}{addrParam, sectorIdParam}
+	blockHeightParam := types.NewBlockHeight(43)
+	redeemerParams := []interface{}{blockHeightParam}
+
+	sys := setup(t)
+	require.NoError(t, sys.st.SetActor(context.TODO(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
+
+	// Close without params and expect a panic
+	condition := &types.Predicate{To: toAddress, Method: method}
```
acruikshank: Provide params here to show that the exact same command fails before the redeem and succeeds after because the condition has been stored.

PR #2690, opened 2019-04-30
downgrade hello logs to debug

### LINKED ISSUE #2667: Remove unnecessary errors and warnings during hello
### Description
Right now we output too many errors and warnings when the node tries to connect to the network and begin syncing the chain. Many of them should be debug messages.

When nodes connect through the hello protocol, but are incorrectly connecting because they're connecting to a different network or different genesis block, the errors logged from that should be debug messages instead.

Here's an example (click the "click the show"): https://github.com/filecoin-project/go-filecoin/issues/2612

### Acceptance criteria
No more hello message errors and warnings when a user starts syncing the chain.

### Risks + pitfalls
none

### Protocol Changes
none

### Where to begin
- The relevant code is probably in https://github.com/filecoin-project/go-filecoin/tree/master/protocol/hello, but please verify.
- Whoever picks this up, please request review from @ZenGround0.

frrist:
closes #2667


protocol/hello/hello.go
```diff
@@ -90,11 +90,11 @@ func (h *Handler) handleNewStream(s net.Stream) {
 
 	switch err := h.processHelloMessage(from, &hello); err {
 	case ErrBadGenesis:
-		log.Warningf("genesis cid: %s does not match: %s, disconnecting from peer: %s", &hello.GenesisHash, h.genesis, from)
+		log.Debugf("genesis cid: %s does not match: %s, disconnecting from peer: %s", &hello.GenesisHash, h.genesis, from)
```
anorth: Should the "bad hello message" case above also disconnect?

frrist:
@anorth we may consider this PR as an alternative to downgrading the logs here: https://github.com/filecoin-project/go-filecoin/pull/2690 -- currently seeking review from the libp2p on correct usage of this.

protocol/hello/hello.go
frrist: I don't know. I believe an error for that case indicates a failure to decode the cbor message, my gut says disconnecting is the correct behaviour. I can add that here as well as the metics since it will be a small change.

whyrusleeping: Yeah, I'd say it makes sense to disconnect if you send a bad hello. Not sending a hello at all should be okay though, but if youre gonna play the game, better follow the rules.

PR #2691, opened 2019-04-30
Fix/chain sync logs

### LINKED ISSUE #2669: Chain sync status messages
### Description
When we fetch the chain we periodically print out the height of the most recent block we've fetched from the network: 20,000, 15,000, 10,000. This makes users feel like something is going wrong. We say we are "syncing" blocks up to this height, we should say "fetching". Also we should count forwards, not backwards.

### Acceptance criteria
- Status messages that say "syncing" should say "fetching" instead.
- Count forwards, not backwards.

### Risks + pitfalls
none

### Protocol Changes
none

### Where to begin
- Here's the output line: https://github.com/filecoin-project/go-filecoin/blob/master/chain/default_syncer.go#L144
- Counting should happen in `collectChain: https://github.com/filecoin-project/go-filecoin/blob/master/chain/default_syncer.go#L102
- If you are picking this up, please tag @ZenGround0 to review the PR.

### LINKED ISSUE #2670: Human-readable chain sync output
### Description
After we have fetched blocks and start processing, the log output is extremely noisy. Let's improve the signal to noise ratio.

### Acceptance criteria
1. Remove most of the block processing CLI outputs (sending them to a higher log level such as debug). (This issue doesn't need to touch `rust-fil-proofs` output.)
2. Add a one-line, human-readable status and update it periodically. For example:
`Syncing the chain, this will take a while. Currently at block 5,000 of 20,000 (25%).`
or
`Syncing 5,000 of 20,000 blocks, this will take a while.`
By periodically, we mean roughly every 2,000 blocks.

### Risks + pitfalls
n/a

### Protocol Changes
n/a

### Where to begin
- Most of the logging outputs are here: https://github.com/filecoin-project/go-filecoin/blob/master/consensus/processor.go#L282 
- If you are picking this up, please tag @ZenGround0 to review the PR.

frrist:
closes #2669 
closes #2670


chain/default_syncer.go
```diff
@@ -110,7 +110,7 @@ func (syncer *DefaultSyncer) getBlksMaybeFromNet(ctx context.Context, blkCids []
 // that interacts with the network. It does NOT add tipsets to the chainStore..
 func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.SortedCidSet) ([]types.TipSet, error) {
 	var chain []types.TipSet
-	defer logSyncer.Info("chain synced")
+	defer logSyncer.Infof("chain fetch from network complete for %v", tipsetCids)
```
travisperson: This will always show the base tipset they had prior to chain fetching due to the value being overridden on 149 I believe, not where they are after collectionChain returns (input value).

```
chain = append([]types.TipSet{ts}, chain...)
tipsetCids, err = ts.Parents()
```

frrist: doh you are right (`defer`), good catch

chain/default_syncer.go
```diff
@@ -326,7 +326,7 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
 // attempt to validate and caches invalid blocks it has encountered to
 // help prevent DOS.
 func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids types.SortedCidSet) error {
-	logSyncer.Debugf("trying to sync %v\n", tipsetCids)
+	logSyncer.Debugf("trying to sync %v", tipsetCids)
```
ZenGround0: It could be helpful to make this more precise:
"Beginning fetch and sync of chain with head %v"

chain/default_syncer.go
```diff
@@ -383,6 +389,9 @@ func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids typ
 			syncer.badTipSets.AddChain(chain[i:])
 			return err
 		}
+		if (i % 500) == 0 {
+			logSyncer.Infof("processing block %d of %v", i, highestBlock)
```
ZenGround0: Blocking: this is misleading because it is conflating height and number of blocks which are not the same.  There are a few different ways you can do this.  One way would be "processing block x of y for chain with head at c" where y is the length of the fetched chain (`len(chain)`), x is `i`, and c is the `tipsetCids` above.

You could also do this strictly by height "processing block height u of v" where v is the highest block as now and u is the height of tipset at index i.

chain/default_syncer.go
```diff
@@ -110,7 +110,7 @@ func (syncer *DefaultSyncer) getBlksMaybeFromNet(ctx context.Context, blkCids []
 // that interacts with the network. It does NOT add tipsets to the chainStore..
 func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.SortedCidSet) ([]types.TipSet, error) {
 	var chain []types.TipSet
-	defer logSyncer.Info("chain synced")
+	defer logSyncer.Info("chain fetch from network complete")
 	for {
 		var blks []*types.Block
 		// check the cache for bad tipsets before doing anything
```
ZenGround0: Blocking: down at line 144 we should change the log from "syncing" to "fetching"

chain/default_syncer.go
```diff
@@ -110,7 +110,7 @@ func (syncer *DefaultSyncer) getBlksMaybeFromNet(ctx context.Context, blkCids []
 // that interacts with the network. It does NOT add tipsets to the chainStore..
 func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.SortedCidSet) ([]types.TipSet, error) {
 	var chain []types.TipSet
-	defer logSyncer.Info("chain synced")
+	defer logSyncer.Info("chain fetch from network complete")
```
ZenGround0: Could be good to include `tipsetCids` in this log so it is easy to identify the fetched chain as the same one referred to earlier.

chain/default_syncer.go
ZenGround0: Also down at line 144 we should count up in terms of blocks fetched, not down in terms of block height.

chain/default_syncer.go
ZenGround0: (original tipsetCids of the head that is, not the value this variable updates to each loop

chain/default_syncer.go
frrist: nice catch, will do.

Issue #2689, opened 2019-04-30
Move sector storage as sibling to repo, reverting "home" directory concept


ZenGround0:
### Description
Remove the homedir concept introduced in #2658.  Replace it with [proposal 2c](https://github.com/filecoin-project/go-filecoin/issues/2687) outline in the link. The sector directory defaults to a sibling of the repo directory `<repodir>/../.filecoin_sectors` without being wrapped in any go-filecoin specific wrapper.  Put another way, go-filecoin doesn't need to know anything about repo migrations.

This is an alternative resolution to https://github.com/filecoin-project/go-filecoin/issues/2524

### Acceptance criteria
Homedir is gone, sectors are properly put in a sibling.

anorth:
I think this is now resolved.

Issue #2687, opened 2019-04-30
What to do about the default repo layout?


ZenGround0:
# What is the problem?
PR #2658 introduced the model explicitly outlined in #2524.  This new model caused issues for @laser addressed here: #2683, and during the discussion @travisperson raised a bunch of [issues with the new model](https://github.com/filecoin-project/go-filecoin/pull/2682#issuecomment-487989338).

# What is this issue for?
Outlining the two main courses of action and the different branches within each, including pros and cons, and a rough set of work that will need to be done to get any path finished.

# Acceptance Criteria?
A decision is made and I know 1) what we want long term and 2) what I am going to do to safely close out this workstream for the release

# Paths forward
## Keep Latest Changes
By default there is a single directory for persistent storage (home).  One subdirectory for sector storage (sectors), one subdirectory for all data that can be subject to migrations (repo).
* Pros 
	* default separation of go-filecoin and rust-fil proofs data. (See #2524 for exhaustive list of pros)
	* particularly migrations can copy the migrate-able data easily without having to handle large amounts of out of scope sector data by default

* Cons: 
	* directory structure grows in complexity.  Right now users (tests) need to think about two directories (home and repo)
	* just specifying the repo dir with the unchanged —repodir flag isn’t enough to separate sector storage into a separate directories which means lots of different users write to the same sector storage

* TODO: 
	* All paths forward require the homedir expansion fix in PR 2682

### Path forward 1a
Allow daemon to specify homedir with a —homedir flag and change tests to use this to specify a segregated testing path for both sector storage and repo storage.  Also FIL_PATH would reference the homedir and FIL_REPO_PATH would now reference the repo dir.  No tests would need to specify the repo dir or sectors dir, we could even remove those flags.
* Pros:
	* only one flag makes it easier on users
        * config path is determined by homedir flag unless user is doing something weird
* Cons: 
	* interaction between —homedir and —repodir might confuse users
	* config file path is a little harder to infer from a —homedir flag passed in 

* TODO:
	* Fix environment variables and flags to work as described
	* Update all callers (tests) to use homedir instead of repodir.  This includes messing with paths iptb uses that assume config is directly under the passed in directory
	* Update documentation

### Path forward 1b
Maintain default separation of migrate able data and sectors but the node only references a single directory ever unless a power user messes with the config.  The repo would again by the top level directory and could be specified by the —repodir  or FIL_PATH.  By default the repo would contain two directories migrateable and sectors.
* Pros: 
[...]

Issue #2684, opened 2019-04-29
go-filecoin init symlinks to ~/.filecoin/repo to a version-named repo directory sibling


shannonwells:
### Description
Repo migrations will be taking a similar approach to traditional database migrations, where the location of the database dir is actually a symlink pointing to a version-named directory.  Because of this we want to make `go-filecoin init` set up the repo as a version-named directory and then symlink the canonical name to it.
```
~/.filecoin   -> .filecoin-201903012000-v2 # canonical repo path symlinks to repo dir
~/.filecoin-201903012000-v2 # the actual repo
~/.filecoin-201901122000-v1 # pre-migration, now unlinked
~/.filecoin-sectors
```

Then `go-filecoin-migrate` can:
- assume the repo to be migrated exists as a symlink at the canonical path `$FIL_PATH` and follow it to find the directory to clone, or
- have an optional `--old-repo` CLI arg point to a repo directory (not symlink) to copy instead

### Acceptance criteria
`go-filecoin init`, creates ~/.filecoin-v201901122000-v1 (or something like it) and symlinks to `~/.filecoin` 

The directory name should be in the same format expected to be used for migrated repos.

### Risks + pitfalls
This change is backwards incompatible, so devs will need to either re-init or manually rename and symlink the repo directory *in order to run any migration*. Note that at the next release, all user devnet nodes will need to re-init in any case.

anorth:
See also #2524 and beware of any follow-on changes to the env var, what the "repo" refers to, etc.

It would be really great to land this before release so that we can use it for migrations.

anorth:
@shannonwells I have updated the description here to better reflect the directory layout after #2524. I also clarified my understanding of the migration's expectations. I think that the --old-repo arg may be omitted in the usual case, but if specified should point to a directory.

anorth:
I have updated the description again to reflect the decision in #2687

PR #2692, opened 2019-05-01
Remove homedir

ZenGround0:
#2689 

This also closes out #2680 ( I shut down the already reviewed solution in #2682) in favor of this because these changes remove a gnarly CI-only IPTB cleanup failure during the retrieval test)

This gets rid of the homedir concept while keeping the default sector directory out of the repo as part of compromise 2c laid out here #2687.  

@travisperson I went ahead and defaulted IPTB sector directory storage to be within the repo underneath `sectors` just as it was before I started these changes (we're not migrating these repos so putting sectors inside them is nice and doesn't cause problems).  I don't know my way around the filecoin IPTB plugin well so please alert me if the way I did this is ugly / not the right way.

@shannonwells I don't expect this to get in the way of the other repo migration efforts but please let me know if this unintentionally blocks or frustrates your work.

cc @anorth please take a look too

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -938,14 +938,31 @@ func initSectorBuilderForNode(ctx context.Context, node *Node, proofsMode types.
 	// metadata in the staging directory, it should be in its own directory.
 	//
 	// Tracked here: https://github.com/filecoin-project/rust-fil-proofs/issues/402
-	sectorDir := paths.GetSectorPath(node.Repo.Config().SectorBase.RootDir)
+	repoPath, err := node.Repo.Path()
+	if err != nil {
+		return nil, err
+	}
+	sectorDir, err := paths.GetSectorPath(node.Repo.Config().SectorBase.RootDir, repoPath)
+	if err != nil {
+		return nil, err
+	}
+
+	stagingDir, err := paths.StagingDir(sectorDir)
+	if err != nil {
+		return nil, err
+	}
+
+	sealedDir, err := paths.SealedDir(sectorDir)
+	if err != nil {
+		return nil, err
+	}
 	cfg := sectorbuilder.RustSectorBuilderConfig{
 		BlockService:     node.blockservice,
 		LastUsedSectorID: lastUsedSectorID,
-		MetadataDir:      paths.StagingDir(sectorDir),
+		MetadataDir:      stagingDir,
 		MinerAddr:        minerAddr,
-		SealedSectorDir:  paths.SealedDir(sectorDir),
-		StagedSectorDir:  paths.StagingDir(sectorDir),
+		SealedSectorDir:  sealedDir,
+		StagedSectorDir:  stagingDir,
 		SectorClass:      sectorClass,
 	}
 
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -3,60 +3,61 @@ package paths
 import (
 	"os"
 	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
 )
 
 // node repo path defaults
 const filPathVar = "FIL_PATH"
-const defaultHomeDir = "~/.filecoin"
-const defaultRepoDir = "repo"
+const defaultRepoDir = "~/.filecoin"
 
 // node sector storage path defaults
 const filSectorPathVar = "FIL_SECTOR_PATH"
-const defaultSectorDir = "sectors"
+const defaultSectorDir = ".filecoin_sectors"
 const defaultSectorStagingDir = "staging"
 const defaultSectorSealingDir = "sealed"
 
 // GetRepoPath returns the path of the filecoin repo from a potential override
 // string, the FIL_PATH environment variable and a default of ~/.filecoin/repo.
-func GetRepoPath(override string) string {
+func GetRepoPath(override string) (string, error) {
 	// override is first precedence
 	if override != "" {
-		return override
+		return homedir.Expand(override)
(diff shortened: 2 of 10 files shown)


node/node.go
```diff
@@ -938,14 +938,31 @@ func initSectorBuilderForNode(ctx context.Context, node *Node, proofsMode types.
 	// metadata in the staging directory, it should be in its own directory.
 	//
 	// Tracked here: https://github.com/filecoin-project/rust-fil-proofs/issues/402
-	sectorDir := paths.GetSectorPath(node.Repo.Config().SectorBase.RootDir)
+	repoPath, err := node.Repo.Path()
```
ZenGround0: I don't love adding this method to the repo interface.  It is the cleanest way to capture repo path info passed with the `--repodir` flag so that the sectors path can default to `repodir/../filecoin_sectors`

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -53,6 +56,9 @@ var (
 
 	// AttrLogJSON is the key used to set the node to output json logs
 	AttrLogJSON = "logJSON"
+
+	// AttrSectorsPath is the key used to set the sectors path
+	AttrSectorsPath = "sectorsPath"
```
ZenGround0: cc @travisperson there are a few ways to pass in sectorsPath.  Using an attribute seemed like the cleanest way to give iptb nodes a default of sector storage under the repo path.  

We could alternatively require all iptb users that want this config pass in the `--sectorsdir` flag upon init / enforce this option as part of fast environment setup.

ZenGround0: And of course we could always write to the config before calling `mining start` as part of the FAST series that sets up a mining node.

travisperson: I think this is good, but it does make it impossible to set the sector dir via the init flag I believe because it overrides it. Does the value default to empty in the config? If so, we could check the value first to see if it was set. I think this is fine though if we run into that issue we can fix it.

ZenGround0: Nice thinking.  Pushing this up now.

node/node.go
frrist: Looks like we had [similar ideas](https://github.com/filecoin-project/go-filecoin/pull/2663/files#diff-1bae8ed1271a68f399d2ee0151eb38f6R496) here, although I only added it to `FSRepo` and kept it off the interface.  I did it this way because it is my understanding that a "repo" is an abstract concept, and therefore may not always have a path (e.g. Memrepo). What do you think?

I am flagging this because I rank merging the `inspect` command as lower priority that this MFC business. If it makes things easier I will wait on merging the `inspect` code until this goes through, what do you prefer?

ZenGround0: Your way is better in terms of clean abstractions but the node only has access to the interface and must be able to init a sector builder when initialized with a mem repo for test so this addition to the interface is needed without other intervention.

I would prefer merging this first, but the merge conflict with the inspect command should be easy to resolve.

frrist: If we are only doing this for the sake of testing then why not just type assert ([example](https://github.com/filecoin-project/go-filecoin/pull/2663/files#diff-b989720bf97c51e8761fd9483d5cedd6R82))? I don't mean to be an Armchair Expect here, I haven't read this PR in full -- this bit just caught my attention while skimming so take these comments as non-blocking.

You are good to merge from my end, the conflicts in `inspect` will be simple.

paths/paths.go
```diff
@@ -3,60 +3,61 @@ package paths
...
 // node repo path defaults
 const filPathVar = "FIL_PATH"
-const defaultHomeDir = "~/.filecoin"
-const defaultRepoDir = "repo"
+const defaultRepoDir = "~/.filecoin"
 
 // node sector storage path defaults
 const filSectorPathVar = "FIL_SECTOR_PATH"
-const defaultSectorDir = "sectors"
+const defaultSectorDir = ".filecoin_sectors"
```
anorth: opinion: underscores are harder to type than minus signs, so I prefer them outside of super-constrained contexts like identifier names in code.

tools/iptb-plugins/filecoin/local/localfilecoin.go
```diff
@@ -38,6 +38,9 @@ var errTimeout = errors.New("timeout")
 // DefaultFilecoinBinary is the name or full path of the binary that will be used
 var DefaultFilecoinBinary = "go-filecoin"
 
+// DefaultSectorsPath is the name of the sector path relative to the repo path
+var DefaultSectorsPath = "sectors"
```
anorth: Putting the sectors inside the repo differs from the way every other use case will lay things out, which could cause problems later.

node/node.go
anorth: Is this only to handle the sector dir being empty in the config? I think we should disallow that, and require it to be set in the config. `init` should set a default of `.filecoin_sectors` (where the `.filecoin` prefix is taken from the basename of the --repodir value), which can be overridden by a flag specifying some other path (or just changed in the file afterwards).

It's just mandatory configuration, with a default. I think this will remove some ugliness in the IPTB plugin too.

Note that I'm touching `init` soon to set up the symlink, so could roll this in to #2684.

PR #2693, opened 2019-05-01
add piece inclusion proofs to deals

### LINKED ISSUE #2571: Piece Inclusion Proofs in Deals
### Description

In order for a client to know their piece is included within a deal, the miner must include a piece inclusion proof. This proof allows the client to verify that their piece is included within a section, which they can then check that the miner is storing via their storage proofs. The signed deal including the proof is also important for storage slashing.

This story is just about including that proof in the deal.

### Acceptance criteria

1. Once sealing is complete, the miner creates a piece inclusion proof.
2. The deal has a field for the proof.
3. The miner includes the proof in the deal and signs it.

### Risks

Piece inclusions proofs are not yet implemented, and their exact structure is unknown. Resolution of both of these things will depend on the rust proofs team.

acruikshank:
Closes #2571

### Problem

There is currently no mechanism for miners to send piece inclusion proofs back to clients. As a convenience, we would also like storage miners to communicate the message cid of the `commitSector` message used to commit the sector containing the client's piece, so that the client can more easily track when the commitment lands on their own chain.

### Solution

This PR adds the inclusion proof and the commit message cid to the ProofInfo in the deal response sent back to the client. The piece inclusion proof is currently faked pending the availability real proofs from rust proofs (#2629).

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -838,7 +838,7 @@ func (node *Node) StartMining(ctx context.Context) error {
 					val := result.SealingResult
 					// This call can fail due to, e.g. nonce collisions. Our miners existence depends on this.
 					// We should deal with this, but MessageSendWithRetry is problematic.
-					_, err := node.PorcelainAPI.MessageSend(
+					msgCid, err := node.PorcelainAPI.MessageSend(
 						node.miningCtx,
 						minerOwnerAddr,
 						minerAddr,
@@ -857,7 +857,7 @@ func (node *Node) StartMining(ctx context.Context) error {
 						continue
 					}
 
-					node.StorageMiner.OnCommitmentAddedToChain(val, nil)
+					node.StorageMiner.OnCommitmentSent(val, msgCid, nil)
 				}
 			case <-node.miningCtx.Done():
 				return
--- a/proofs/sectorbuilder/interface.go
+++ b/proofs/sectorbuilder/interface.go
@@ -68,8 +68,9 @@ type SectorSealResult struct {
 
 // PieceInfo is information about a filecoin piece
 type PieceInfo struct {
-	Ref  cid.Cid `json:"ref"`
-	Size uint64  `json:"size"` // TODO: use BytesAmount
+	Ref            cid.Cid `json:"ref"`
+	Size           uint64  `json:"size"` // TODO: use BytesAmount
+	InclusionProof []byte  `json:"inclusionProof"`
 }
 
 // SealedSectorMetadata is a sector that has been sealed by the PoRep setup process
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -393,14 +393,14 @@ func (sm *Miner) updateDealResponse(proposalCid cid.Cid, f func(*storagedeal.Res
 	return nil
 }
 
-func (sm *Miner) processStorageDeal(c cid.Cid) {
-	log.Debugf("Miner.processStorageDeal(%s)", c.String())
+func (sm *Miner) processStorageDeal(proposalCid cid.Cid) {
+	log.Debugf("Miner.processStorageDeal(%s)", proposalCid.String())
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	d := sm.porcelainAPI.DealGet(c)
+	d := sm.porcelainAPI.DealGet(proposalCid)
 	if d == nil {
-		log.Errorf("could not retrieve deal with proposal CID %s", c.String())
+		log.Errorf("could not retrieve deal with proposal CID %s", proposalCid.String())
 	}
 	if d.Response.State != storagedeal.Accepted {
 		// TODO: handle resumption of deal processing across miner restarts
@@ -414,7 +414,7 @@ func (sm *Miner) processStorageDeal(c cid.Cid) {
 	log.Debug("Miner.processStorageDeal - FetchGraph")
 	if err := dag.FetchGraph(ctx, d.Proposal.PieceRef, dag.NewDAGService(sm.node.BlockService())); err != nil {
 		log.Errorf("failed to fetch data: %s", err)
-		err := sm.updateDealResponse(c, func(resp *storagedeal.Response) {
+		err := sm.updateDealResponse(proposalCid, func(resp *storagedeal.Response) {
 			resp.Message = "Transfer failed"
 			resp.State = storagedeal.Failed
 			// TODO: signature?
@@ -427,7 +427,7 @@ func (sm *Miner) processStorageDeal(c cid.Cid) {
 
 	fail := func(message, logerr string) {
 		log.Errorf(logerr)
-		err := sm.updateDealResponse(c, func(resp *storagedeal.Response) {
+		err := sm.updateDealResponse(proposalCid, func(resp *storagedeal.Response) {
 			resp.Message = message
 			resp.State = storagedeal.Failed
 		})
@@ -451,7 +451,7 @@ func (sm *Miner) processStorageDeal(c cid.Cid) {
 	}
 
 	// There is a race here that requires us to use dealsAwaitingSeal below. If the
-	// sector gets sealed and OnCommitmentAddedToChain is called right after
+	// sector gets sealed and OnCommitmentSent is called right after
 	// AddPiece returns but before we record the sector/deal mapping we might
 	// miss it. Hence, dealsAwaitingSealStruct. I'm told that sealing in practice is
 	// so slow that the race only exists in tests, but tests were flaky so
@@ -465,7 +465,7 @@ func (sm *Miner) processStorageDeal(c cid.Cid) {
 		return
 	}
 
-	err = sm.updateDealResponse(c, func(resp *storagedeal.Response) {
+	err = sm.updateDealResponse(proposalCid, func(resp *storagedeal.Response) {
 		resp.State = storagedeal.Staged
 	})
 	if err != nil {
@@ -474,7 +474,7 @@ func (sm *Miner) processStorageDeal(c cid.Cid) {
 
 	// Careful: this might update state to success or failure so it should go after
 	// updating state to Staged.
-	sm.dealsAwaitingSeal.add(sectorID, c)
+	sm.dealsAwaitingSeal.add(sectorID, proposalCid)
 	if err := sm.saveDealsAwaitingSeal(); err != nil {
 		log.Errorf("could not save deal awaiting seal: %s", err)
 	}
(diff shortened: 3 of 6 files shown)


protocol/storage/miner.go
```diff
@@ -493,16 +493,19 @@ type dealsAwaitingSealStruct struct {
...
 
 func (sm *Miner) loadDealsAwaitingSeal() error {
 	sm.dealsAwaitingSeal = &dealsAwaitingSealStruct{
-		SectorsToDeals:    make(map[uint64][]cid.Cid),
-		SuccessfulSectors: make(map[uint64]*sectorbuilder.SealedSectorMetadata),
-		FailedSectors:     make(map[uint64]string),
+		SectorsToDeals:     make(map[uint64][]cid.Cid),
+		SuccessfulSectors:  make(map[uint64]*sectorbuilder.SealedSectorMetadata),
+		FailedSectors:      make(map[uint64]string),
+		CommitmentMessages: make(map[uint64]cid.Cid),
```
acruikshank: This isn't a great to store this. A better solution is coming (see feat/better_testing_for_deals_awaiting_seals), but it comes with a large refactor that is better addressed in a separate PR after this.

protocol/storage/miner.go
```diff
@@ -393,14 +393,14 @@ func (sm *Miner) updateDealResponse(proposalCid cid.Cid, f func(*storagedeal.Res
 	return nil
 }
 
-func (sm *Miner) processStorageDeal(c cid.Cid) {
-	log.Debugf("Miner.processStorageDeal(%s)", c.String())
+func (sm *Miner) processStorageDeal(proposalCid cid.Cid) {
```
anorth: Thank you

protocol/storage/miner.go
```diff
@@ -493,16 +493,19 @@ type dealsAwaitingSealStruct struct {
 	SuccessfulSectors map[uint64]*sectorbuilder.SealedSectorMetadata
 	// Maps from sector id to seal failure error string.
 	FailedSectors map[uint64]string
+	// Maps from sector id to a sectors commitSector message
```
anorth: "... message CID"

protocol/storage/miner.go
```diff
@@ -530,6 +533,13 @@ func (sm *Miner) saveDealsAwaitingSeal() error {
 	return nil
 }
 
+func (dealsAwaitingSeal *dealsAwaitingSealStruct) addCommitMessageCid(sectorID uint64, msgCid cid.Cid) {
```
anorth: Prefer `addCommitmentMessageCid`

node/node.go
```diff
@@ -856,7 +856,7 @@ func (node *Node) StartMining(ctx context.Context) error {
 						continue
 					}
 
-					node.StorageMiner.OnCommitmentAddedToChain(val, nil)
+					node.StorageMiner.OnCommitmentAddedToChain(val, msgCid, nil)
```
anorth: It's not on the chain yet, so this call is misleading. Could you rename it to `OnCommitmentSent` or similar. I'd expect an `OnCommitmentAddedToChain` to be callback in response to observing it mined in a block or similar.

protocol/storage/storagedeal/types.go
```diff
@@ -124,11 +127,15 @@ type Deal struct {
...
+	// Sector id allows us to find the committed sector metadata on chain
 	SectorID uint64
-	CommR    []byte
-	CommD    []byte
+
+	// CommitmentMessage is the cid of the message that committed the sector. It's used to track when the sector goes on chain.
+	CommitmentMessage *cid.Cid
+
+	// PieceInclusionProof is a proof that a the piece is included within a sector
+	PieceInclusionProof []byte
```
anorth: Could you justify for me removing the CommR and CommD?

For CommD, I understand it's part of the PIP, so that makes sense, but CommR isn't. Even if it's in the CommitmentMessage, that's not sufficient to gaurantee the client can access it given the message CID, at least until it's mined into a block. Since this message is not on-chain data, I'd suggest leaving CommR here to simplify the client's bookkeeping.

protocol/storage/storagedeal/types.go
```diff
@@ -45,6 +45,9 @@ type Proposal struct {
 	// PieceRef is the cid of the piece being stored
 	PieceRef cid.Cid
 
+	// DataCommitment is the the data commitment of the piece used for proof of piece inclusion
+	DataCommitment types.CommP
+
```
anorth: Blocking: Including this here is a bit confusing. A proposal goes from client to miner, and the client clearly doesn't know this. If we want to add information that a miner sends in response I think we should wrap the proposal instead.

This appears to be unused in this PR, so could you revert it now, and we can discuss with more context when it's used.

proofs/sectorbuilder/interface.go
```diff
@@ -68,8 +68,9 @@ type SectorSealResult struct {
 
 // PieceInfo is information about a filecoin piece
 type PieceInfo struct {
-	Ref  cid.Cid `json:"ref"`
-	Size uint64  `json:"size"` // TODO: use BytesAmount
+	Ref            cid.Cid `json:"ref"`
+	Size           uint64  `json:"size"` // TODO: use BytesAmount
+	InclusionProof []byte  `json:"includionProof"`
```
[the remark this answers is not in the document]
acruikshank: 😊

node/node.go
laser: @anorth - You make a good point. 

@acruikshank - You can call `MessageWait` after calling `MessageSend`, and call `OnCommitmentAddedToChain` in the function passed to `MessageWait`. That would make these names not-a-lie.

protocol/storage/miner.go
```diff
@@ -592,6 +603,7 @@ func (sm *Miner) OnCommitmentAddedToChain(sector *sectorbuilder.SealedSectorMeta
 		errMsg := fmt.Sprintf("failed sealing sector: %d", sectorID)
 		sm.dealsAwaitingSeal.fail(sector.SectorID, errMsg)
 	} else {
+		sm.dealsAwaitingSeal.addCommitMessageCid(sectorID, msgCid)
```
laser: SUGGESTION: Is there any circumstance where you'd want to add a commit message but the sector had not sealed successfully? If not, I think you should just add the commit message CID to the `sm.dealsAwaitingSeal.success` call and remove `addCommitMessageCid`.

protocol/storage/miner.go
```diff
@@ -601,19 +613,46 @@ func (sm *Miner) OnCommitmentAddedToChain(sector *sectorbuilder.SealedSectorMeta
...
 
+// search the sector's piece info to find the one for the given deal's piece
+func (sm *Miner) findPieceInfo(dealCid cid.Cid, sector *sectorbuilder.SealedSectorMetadata) (*sectorbuilder.PieceInfo, error) {
+	deal := sm.porcelainAPI.DealGet(dealCid)
+	if deal == nil || deal.Response.State == storagedeal.Unknown {
+		return nil, errors.Errorf("Could not find deal with deal cid %s", dealCid)
+	}
+
+	for _, info := range sector.Pieces {
+		if info.Ref == deal.Proposal.PieceRef {
```
laser: BLOCKING: You should use `info.Ref.Equals(deal.Proposal.PieceRef)` instead of `==` (in spite of the fact that `Equals` delegates to `==`). Maybe I'm being overly paranoid, but it is possible that the `go-cid` maintainers change the implementation out from underneath you and `==` no longer does what you'd expect.

protocol/storage/storagedeal/types.go
acruikshank: Good catch. I was sketching and forgot to remove this.
It's not true that the client doesn't know this, though. In fact, its critical that the client be able to calculate CommP from their existing data. CommP should be embedded within the conditions of the payments though (#2705). 

I'm not sure why I thought it should go here.

protocol/storage/storagedeal/types.go
laser: > Could you justify for me removing the CommR and CommD?

If the client has the content id of the `commitSector` message and the `commitSector` message includes data and replica commitments, then the replica and data commitments are redundant (in this data structure).

node/node.go
acruikshank: Good call.

protocol/storage/miner.go
```diff
@@ -493,16 +493,19 @@ type dealsAwaitingSealStruct struct {
...
 	onFail    func(dealCid cid.Cid, message string)
 }
 
 func (sm *Miner) loadDealsAwaitingSeal() error {
 	sm.dealsAwaitingSeal = &dealsAwaitingSealStruct{
-		SectorsToDeals:    make(map[uint64][]cid.Cid),
-		SuccessfulSectors: make(map[uint64]*sectorbuilder.SealedSectorMetadata),
-		FailedSectors:     make(map[uint64]string),
+		SectorsToDeals:     make(map[uint64][]cid.Cid),
+		SuccessfulSectors:  make(map[uint64]*sectorbuilder.SealedSectorMetadata),
```
laser: SUGGESTION: Introduce a new type which has a `commitSector` message CID and also a `SealedSectorMetadata`, and collapse the `SuccessfulSectors` and `CommitmentMessages` maps into a single map. Call this new type `PostedSectorMetadata`.

protocol/storage/miner.go
```diff
@@ -493,16 +493,19 @@ type dealsAwaitingSealStruct struct {
...
 	onSuccess func(dealCid cid.Cid, sector *sectorbuilder.SealedSectorMetadata)
 	onFail    func(dealCid cid.Cid, message string)
 }
 
 func (sm *Miner) loadDealsAwaitingSeal() error {
 	sm.dealsAwaitingSeal = &dealsAwaitingSealStruct{
-		SectorsToDeals:    make(map[uint64][]cid.Cid),
-		SuccessfulSectors: make(map[uint64]*sectorbuilder.SealedSectorMetadata),
-		FailedSectors:     make(map[uint64]string),
+		SectorsToDeals:     make(map[uint64][]cid.Cid),
```
laser: QUESTION: Under what circumstances [do we need to stringify uint64](https://github.com/filecoin-project/go-filecoin/blob/master/actor/builtin/miner/miner.go#L510) before using as a map key?

protocol/storage/storagedeal/types.go
acruikshank: @anorth CommR is used to verify the sector is correctly sealed. The client can't use it.
At a higher level, any information about the sector (other than its id) that the miner delivers here is kind of worthless because they haven't proven anything. The appropriate way for a client to access CommD and CommR is to look it up on their local chain by sector id via a query message.

Since we already have `verifyPieceInclusion` implemented in `miner.Actor`, clients won't need `CommD` at all. They can just call `verifyPieceInclusion` with their `CommP`, the miners piece inclusion proof, and the `sectorID`.

node/node.go
acruikshank: @laser I'd rather just make the names right. At least right now. The client is going to have to MessageWait anyway to compensate for chain sync timing.

protocol/storage/miner.go
acruikshank: I had the same thought. It led to a bigger refactor (#2694), so I kept it out of this PR.

protocol/storage/miner.go
acruikshank: Done in #2694.

protocol/storage/miner.go
acruikshank: Not here. This only gets serialized to JSON as far as I can tell.

laser: Thanks.

Issue #2601, opened 2019-04-17
A miner's power reflects the amount of storage they've committed to the network


laser:
### Description

Power Table and other EC components should compute power as a function of bytes committed to network.

As per the spec, a miner's power is a function of the total number of bytes they have committed to the network. Today, the miner's power is a function of the number of _sectors_ which they have committed to the network. We need to update this part of the codebase in order to conform to the specification.

### Tasks

- Modify `StorageMarketActor` and `PowerTableView` and EC to use `BytesAmount` instead of `big.Int`
- Modify `StorageMarketActor.TotalStorage` to use `BytesAmount`
- Modify `Actor#UpdatePower` to use `BytesAmount`
- Modify `Actor#GetTotalStorage` to use `BytesAmount`
- Modify `PowerTableView` and `MarketView` to use `BytesAmount`
- Modify `Expected#Weight` to use `BytesAmount`
- Modify `consensus.IsWinningTicket` to use `BytesAmount`

PR #2697, opened 2019-05-01
pin a miner to a sector size when creating

laser:
Makes incremental progress on #2530 

## Why does this PR exist?

A storage miner operator needs to be able to select a sector size when pledging. This sector size becomes an immutable property of their storage miner actor. This PR makes incremental progress to that end by creating a miner with a sector size (derived from the proofs mode) and persisting that choice to the miner's storage.

## What's in this PR?

- remove notion of test sector class
- replace SectorSize enum with BytesAmount to [conform to spec](https://github.com/filecoin-project/specs/blob/master/actors.md#createstorageminer)
- modify the `protocol` command such that it reports writeable (pre bit-padding) and non-writable ("sector size") in its output
- modify storage market actor `CreateMiner` command to select a sector size for miner based on proofs mode

## Next steps

The `miner create` command needs to be updated to accept a sector size (in bytes), and that value should be plumbed to the storage market actor's `CreateMiner` method.

Note: Before loosening the restriction that live proofs mode must always use the 256MiB sectors, I need to complete #2601.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -138,6 +138,10 @@ type State struct {
 	LastPoSt           *types.BlockHeight
 
 	Power *big.Int
+
+	// SectorSize is the amount of space in each sector committed to the network
+	// by this miner.
+	SectorSize *types.BytesAmount
 }
 
 // NewActor returns a new miner actor
@@ -146,7 +150,7 @@ func NewActor() *actor.Actor {
 }
 
 // NewState creates a miner state struct
-func NewState(owner address.Address, key []byte, pledge *big.Int, pid peer.ID, collateral *types.AttoFIL) *State {
+func NewState(owner address.Address, key []byte, pledge *big.Int, pid peer.ID, collateral *types.AttoFIL, sectorSize *types.BytesAmount) *State {
 	return &State{
 		Owner:             owner,
 		PeerID:            pid,
@@ -156,6 +160,7 @@ func NewState(owner address.Address, key []byte, pledge *big.Int, pid peer.ID, c
 		SectorCommitments: make(map[string]types.Commitments),
 		Power:             big.NewInt(0),
 		NextAskID:         big.NewInt(0),
+		SectorSize:        sectorSize,
 	}
 }
 
@@ -251,6 +256,10 @@ var minerExports = exec.Exports{
 	"isBootstrapMiner": &exec.FunctionSignature{
 		Params: nil,
 		Return: []abi.Type{abi.Boolean},
+	},
+	"getSectorSize": &exec.FunctionSignature{
+		Params: nil,
+		Return: []abi.Type{abi.BytesAmount},
 	},
 }
 
@@ -449,6 +458,29 @@ func (ma *Actor) GetSectorCommitments(ctx exec.VMContext) (map[string]types.Comm
 	return a, 0, nil
 }
 
+// GetSectorSize returns the size of the sectors committed to the network by
+// this miner.
+func (ma *Actor) GetSectorSize(ctx exec.VMContext) (*types.BytesAmount, uint8, error) {
+	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
+		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
+	}
+
+	var state State
+	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
+		return state.SectorSize, nil
+	})
+	if err != nil {
+		return nil, errors.CodeError(err), err
+	}
+
+	amt, ok := out.(*types.BytesAmount)
+	if !ok {
+		return nil, 1, errors.NewFaultErrorf("expected a *types.BytesAmount, but got %T instead", out)
+	}
+
+	return amt, 0, nil
+}
+
 // CommitSector adds a commitment to the specified sector. The sector must not
 // already be committed.
 func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR, commRStar []byte, proof types.PoRepProof) (uint8, error) {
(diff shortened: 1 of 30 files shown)


actor/builtin/miner/miner_test.go
```diff
@@ -40,7 +40,7 @@ func createTestMinerWith(pledge int64,
 	msg := types.NewMessage(minerOwnerAddr, address.StorageMarketAddress, nonce, types.NewAttoFILFromFIL(collateral), "createMiner", pdata)
 
 	result, err := th.ApplyTestMessage(stateTree, vms, msg, types.NewBlockHeight(0))
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
laser: `assert` doesn't halt the test on failure, so `result.Receipt.Return[0]` results in an index out of bounds error.

actor/builtin/miner/miner.go
```diff
@@ -465,59 +497,40 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
+			req := proofs.VerifySealRequest{}
+			copy(req.CommD[:], commD)
+			copy(req.CommR[:], commR)
+			copy(req.CommRStar[:], commRStar)
+			req.Proof = proof
+			req.ProverID = sectorbuilder.AddressToProverID(ctx.Message().To)
+			req.SectorID = sectorbuilder.SectorIDToBytes(sectorID)
+			req.SectorSize = state.SectorSize
+
+			res, err := (&proofs.RustVerifier{}).VerifySeal(req)
```
rosalinekarr: Not your change and not blocking, but is there a reason we don't have a constructor for `RustVerifier`?

commands/protocol.go
```diff
@@ -28,13 +29,32 @@ var protocolCmd = &cmds.Command{
...
 			}
+
 			return nil
 		}),
 	},
 }
+
+func readableBytesAmount(amt float64) string {
+	unit := 0
+	units := []string{"B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB"}
```
rosalinekarr: lol, YiB sector sizes

porcelain/api.go
```diff
@@ -124,6 +124,16 @@ func (a *API) MinerGetOwnerAddress(ctx context.Context, minerAddr address.Addres
 	return MinerGetOwnerAddress(ctx, a, minerAddr)
 }
 
+// MinerGetSectorSize queries for the sector size of the given miner.
+func (a *API) MinerGetSectorSize(ctx context.Context, minerAddr address.Address) (*types.BytesAmount, error) {
+	return MinerGetSectorSize(ctx, a, minerAddr)
+}
+
+// MinerGetSectorSize queries for the sector size of the given miner.
```
rosalinekarr: Comment doesn't match method. This will fail lint.

porcelain/miner.go
```diff
@@ -283,6 +285,7 @@ func MinerPreviewSetPrice(ctx context.Context, plumbing mpspAPI, from address.Ad
 // mgoaAPI is the subset of the plumbing.API that MinerGetOwnerAddress uses.
 type mgoaAPI interface {
 	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
+	ActorGetSignature(ctx context.Context, actorAddr address.Address, method string) (*exec.FunctionSignature, error)
```
rosalinekarr: Not blocking, but it would better match our established pattern to create new interfaces for `MinerGetSectorSize` and `MinerGetLastCommittedSectorID`.

porcelain/miner.go
```diff
@@ -295,6 +298,57 @@ func MinerGetOwnerAddress(ctx context.Context, plumbing mgoaAPI, minerAddr addre
...
+	if !ok {
+		return nil, errors.New("failed to convert returned ABI value")
+	}
+
+	return sectorSize, nil
+}
+
+// MinerGetLastCommittedSectorID queries for the id of the last sector committed
+// by the given miner.
+func MinerGetLastCommittedSectorID(ctx context.Context, plumbing mgoaAPI, minerAddr address.Address) (uint64, error) {
```
rosalinekarr: Not sure I agree with it, but some arguments have been made in the past that nothing should be exposed in porcelain/plumbing that isn't explicitly useful to the end user. I think an exception is warranted here, but I'm interested to hear @acruikshank's opinion.

porcelain/miner.go
```diff
@@ -295,6 +298,57 @@ func MinerGetOwnerAddress(ctx context.Context, plumbing mgoaAPI, minerAddr addre
...
+
+	methodSignature, err := plumbing.ActorGetSignature(ctx, minerAddr, "getLastUsedSectorID")
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to acquire 'getLastUsedSectorID' signature")
+	}
+
+	lastUsedSectorIDVal, err := abi.Deserialize(rets[0], methodSignature.Return[0])
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to deserialize returned value")
+	}
```
rosalinekarr: These last 14 lines are essentially identical to the previous method except for a single method argument. It might be a good idea to extract it into a private method for any future porcelain methods that do the same thing. Alternatively, we could combine these two into a new method called `QueryAndDeserialize` or something.

porcelain/miner.go
rosalinekarr: Actually, on second thought, since this is only used in `node/node.go`, I'm inclined to say it should just live there.

node/node.go
```diff
@@ -889,49 +872,20 @@ func (node *Node) StartMining(ctx context.Context) error {
...
-		return nil, errors.Wrapf(err, "failed to get last used sector id for miner w/address %s", minerAddr.String())
+		return nil, errors.Wrapf(err, "failed to get sector size for miner w/address %s", minerAddr.String())
 	}
 
-	var sectorClass types.SectorClass
-	if proofsMode == types.TestProofsMode {
-		sectorClass = types.NewTestSectorClass()
-	} else {
-		sectorClass = types.NewLiveSectorClass()
+	lastUsedSectorID, err := node.PorcelainAPI.MinerGetLastCommittedSectorID(ctx, minerAddr)
```

node/node.go
```diff
@@ -889,49 +872,20 @@ func (node *Node) StartMining(ctx context.Context) error {
...
-	lastUsedSectorID, ok := lastUsedSectorIDVal.Val.(uint64)
-	if !ok {
-		return 0, errors.New("failed to convert returned ABI value to uint64")
-	}
-
-	return lastUsedSectorID, nil
-}
-
-func initSectorBuilderForNode(ctx context.Context, node *Node, proofsMode types.ProofsMode) (sectorbuilder.SectorBuilder, error) {
+func initSectorBuilderForNode(ctx context.Context, node *Node) (sectorbuilder.SectorBuilder, error) {
```
acruikshank: At some point this should move to a protocol. Hard to say without some research whether that should be the storage protocol or some new (sectorbuilder?) protocol. At any rate, it's not needed in this PR.

protocol/storage/client.go
```diff
@@ -110,13 +110,14 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		return nil, errors.Wrap(err, "failed to determine the size of the data")
 	}
 
-	sectorSize, err := smc.getSectorSize(ctx)
+	sectorSize, err := smc.api.MinerGetSectorSize(ctxSetup, miner)
```
acruikshank: 💯

proofs/sectorbuilder/testing/harness.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	"github.com/filecoin-project/go-filecoin/types"
```
acruikshank: import order

porcelain/miner.go
```diff
@@ -295,6 +298,57 @@ func MinerGetOwnerAddress(ctx context.Context, plumbing mgoaAPI, minerAddr addre
 	return address.NewFromBytes(res[0])
 }
 
+// MinerGetSectorSize queries for the sector size of the given miner.
+func MinerGetSectorSize(ctx context.Context, plumbing mgoaAPI, minerAddr address.Address) (*types.BytesAmount, error) {
```
acruikshank: BLOCKING: This and `MinerGetLastCommittedSectorID` below need tests. At the very least we need a test that uses a fake `MessageQuery` that creates a response with a sector size and the real `ActorGetSignature`  to deserialize the value into a sector size.

actor/builtin/miner/miner.go
```diff
@@ -449,6 +458,29 @@ func (ma *Actor) GetSectorCommitments(ctx exec.VMContext) (map[string]types.Comm
 	return a, 0, nil
 }
 
+// GetSectorSize returns the size of the sectors committed to the network by
+// this miner.
+func (ma *Actor) GetSectorSize(ctx exec.VMContext) (*types.BytesAmount, uint8, error) {
```
acruikshank: BLOCKING: This method needs tests.

porcelain/miner.go
acruikshank: We are trying to move all the code out of node. When the code that depends on this functionality moves, it will move into protocol. Helper functionality to support protocols is an appropriate use of porcelain, so I think this is a great refactor.

porcelain/miner.go
laser: Done, although I feel a little icky having introduced such a test. 

Given that the method under test uses `abi.Deserialize`, I don't see how to test these two methods without re-implementing the ABI serializers (in test code). 

I get how the DI struct allows us to write fine-grained tests, but I'm not convinced that (in this case) the benefit (low-level test coverage) is worth the cost (increased coupling to implementation-level stuff).

actor/builtin/miner/miner.go
laser: Done. See earlier gripe about `MinerGetSectorSize` test.

commands/protocol.go
laser: to the moon

actor/builtin/miner/miner.go
laser: Yeah: The struct itself is stateless. `Verifier` is an interface and `RustVerifier` satisfies the interface with a bunch of data-in, data-out CGO calls. 

Really, I should just eliminate the Java-style `Verifier` interface and `RustVerifier` struct and rewrite its methods as functions.

laser:
@rosalinekarr - assuming that I've addressed your blocking feedback, would you please approve the PR? If not, please do let me know what other things you'd like to see changed.

PR #2698, opened 2019-05-01
migration logger

ZenGround0:
@shannonwells take a look and let me know if this is near to what you had in mind.

(Patterned pretty closely on your spike branch)

### DIFF
--- a/tools/migration/internal/logger.go
+++ b/tools/migration/internal/logger.go
@@ -1,0 +1,47 @@
+package internal
+
+import (
+	"io"
+	"log"
+	"os"
+)
+
+// Logger logs migration events to disk and, if initialized with verbose,
+// stdout too.
+type Logger struct {
+	closer io.WriteCloser
+	logger *log.Logger
+}
+
+// NewLogger creates a new Logger.  All log writes go to f, the logging file.
+// If the verbose flag is true all log writes also go to stdout.
+func NewLogger(wc io.WriteCloser, verbose bool) *Logger {
+	// by default just write to file
+	var w io.Writer
+	w = wc
+	if verbose {
+		w = io.MultiWriter(wc, os.Stdout)
+	}
+	return &Logger{
+		closer: wc,
+		logger: log.New(w, "", 0),
+	}
+}
+
+// Error logs an error to the logging output.
+func (l *Logger) Error(err error) {
+	if err == nil {
+		return
+	}
+	l.logger.Printf("ERROR: %s", err.Error())
+}
+
+// Print logs a string to the logging output.
+func (l *Logger) Print(msg string) {
+	l.logger.Print(msg)
+}
+
+// Close closes the logfile backing the Logger.
+func (l *Logger) Close() error {
+	return l.closer.Close()
+}
--- a/tools/migration/internal/logger_test.go
+++ b/tools/migration/internal/logger_test.go
@@ -1,0 +1,75 @@
+package internal_test
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	. "github.com/filecoin-project/go-filecoin/tools/migration/internal"
+)
+
+func TestLoggerWritesToFile(t *testing.T) {
+	f, err := ioutil.TempFile("", "logfile")
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, os.Remove(f.Name()))
+	}()
+
+	logger := NewLogger(f, false)
+	logger.Print("testing print 1")
+	errTest := errors.New("testing error 2")
+	logger.Error(errTest)
+
+	// Reopen file so we can read new writes
+	out, err := ioutil.ReadFile(f.Name())
+	require.NoError(t, err)
+	outStr := string(out)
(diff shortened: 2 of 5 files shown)


tools/migration/main.go
```diff
@@ -60,13 +66,18 @@ func main() { // nolint: deadcode
...
-		oldRepoOpt, found := findOpt("old-repo", os.Args)
+		logFile, err := getLogFile()
+		if err != nil {
+			exitErr(err.Error())
+		}
+		logger := internal.NewLogger(logFile, getVerbose())
 
+		oldRepoOpt, found := findOpt("old-repo", os.Args)
 		if found == false {
 			exitErr(fmt.Sprintf("Error: --old-repo is required\n%s\n", USAGE))
```
ZenGround0: @shannonwells thinking about this more it seemed like logging these exit errors isn't what we want to do.  We want the user to see these on stderr not in the logfile by default, or stdout if running verbose.  Happy to log these if you have a different perspective but leaving off for now.

tools/migration/main.go
```diff
@@ -97,6 +108,22 @@ func getVerbose() bool {
...
+	path, err := getLogFilePath()
+	if err != nil {
+		return nil, err
+	}
+	return os.OpenFile(path, os.O_APPEND|os.O_CREATE, 0644)
+}
+
+func getLogFilePath() (string, error) {
+	if logPath, found := findOpt("--log-file", os.Args); found {
+		return homedir.Expand(logPath)
```
anorth: We shouldn't need to expand values from args – the shell will do it. The only reason we should be expanding ourselves is when expanding a hardcoded path.

tools/migration/internal/logger.go
```diff
@@ -0,0 +1,46 @@
...
+// NewLogger creates a new Logger.  All log writes go to f, the logging file.
+// If the verbose flag is true all log writes also go to stdout.
+func NewLogger(f *os.File, verbose bool) *Logger {
+	// by default just write to file
+	var w io.Writer
+	w = f
+	if verbose {
+		w = io.MultiWriter(f, os.Stdout)
+	}
+	log.SetOutput(w)
```
anorth: Blocking: this global side-effect is out of place here (or possibly anywhere). Constructing an object should not change process-wide logging behaviour.

I thought that the point of this object was that migrations would log to it directly, bypassing the `log` package. In that case, the methods below should write to logFile directly.

If we think it's also practically necessary to set the global log output to this stream, so that code using the `log` package gets the same behaviour, then that call should be made in the `main` entrypoint or similar. For that, add an `Install()` or similar method to Logger that can be invoked when wanted.

tools/migration/internal/logger_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	. "github.com/filecoin-project/go-filecoin/tools/migration/internal"
+)
+
+func TestLoggerWritesToFile(t *testing.T) {
+	f, err := ioutil.TempFile("", "logfile")
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, os.Remove(f.Name()))
+	}()
```
anorth: Question: is it necessary or valuable to remove tempfiles, or files created in a tempdir? Isn't the whole point that they will be unlinked automatically when the process exits? This adds a bunch of noise to our tests which I am not confident is justified.

tools/migration/internal/logger.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Logger logs migration events to disk and, if initialized with verbose,
+// stdout too.
+type Logger struct {
+	logFile *os.File
+}
+
+// NewLogger creates a new Logger.  All log writes go to f, the logging file.
+// If the verbose flag is true all log writes also go to stdout.
+func NewLogger(f *os.File, verbose bool) *Logger {
```
anorth: I suggest relaxing `os.File` to `io.Writer` here, since no other behaviour of File is needed. This could also simplify testing, where you could point this to a bytestream.

tools/migration/main.go
```diff
@@ -97,6 +108,22 @@ func getVerbose() bool {
 	return res
 }
 
+func getLogFile() (*os.File, error) {
```
anorth: `openLogFile` better describes the work that this does.

tools/migration/main.go
ZenGround0: Ah thanks, overreacting after the last bug 😅

tools/migration/internal/logger_test.go
ZenGround0: I'm going off of the advice in [ioutil](https://golang.org/pkg/io/ioutil/#example_TempFile): "It is the caller's responsibility to remove the file when no longer needed."

tools/migration/internal/logger.go
shannonwells: good point about setting global log.

ZenGround0: Good point.  I will get rid of the global set entirely.

tools/migration/main.go
shannonwells: Fine with me

tools/migration/internal/logger.go
ZenGround0: I'm going to make it a `WriteCloser` so I can still close the file from within the logger

ZenGround0: (without casting)

PR #2699, opened 2019-05-01
Display error when address is not a miner address

### LINKED ISSUE #2434: [bug] Bad error message when attempting to invoke miner commands on non-miner actor
**Describe the bug**
Running `go-filecoin miner pledge t1ebxo3dgpkawkreayhhnq3umue6g6sbcmlnji3si` returns the following error:

```
Error: unable to determine return type: missing export: getPledge
```

Ditto for `go-filecoin miner power t1ebxo3dgpkawkreayhhnq3umue6g6sbcmlnji3si`:

```
Error: unable to determine return type: missing export: getPower
```

**Expected behavior**
Power/pledge data should be returned.

**Screenshots**
If applicable, add logging output or screenshots to help explain your problem.

 - OS: OS X and Ubuntu
 - Version: User devnet binary release 0.1.4

mslipper:
Closes #2434.


mishmosh:
Apologies for the delay in routing this. Tagging @ZenGround0 for review.

@mslipper could you rebase at some point?

mslipper:
I'll check out the integration test failure today.

mishmosh:
Tagging @acruikshank for maintainer review. If approved, could you assign back to @frrist or @ZenGround0 to merge?

commands/miner.go
```diff
@@ -46,6 +46,9 @@ var minerPledgeCmd = &cmds.Command{
 		if err != nil {
 			return err
 		}
+		if minerAddr.Protocol() != address.Actor {
```
acruikshank: This isn't right.

We have 4 different types of addresses:
1. ID, which we don't really use
2. Actor, which we use when we we aren't generating the address from a public key
3. Secp256k, which we use when we do generate the address from a public key
4. BLS, which we don't use yet

In theory all the addresses are interchangeable in the sense that any actor could be addressed by any type of address.

In practice, AccountActors have addresses that represent their associated public keys and all other actors are generated from a string (address.Actor protocol). It's probably true that a MinerActor will never be addressed by a Secp256k address, but it's not true that every Actor address points to a miner, or that MinerActor's will always be addressed by an Actor address.

If we want more intuitive error messages from this command, we really want to get the actor from latest state (`ActorGet`) and assert that its code Cid equals `types.MinerActorCodeCid`.

frrist:
Closing since this has sat for sometime, @mslipper feel free to reopen once feedback is addressed.

PR #2700, opened 2019-05-01
Use a struct over global vars in default_syncer_test

### LINKED ISSUE #2116: Remove global variables from default_syncer_test and default_store_test

### LINKED ISSUE #2434: [bug] Bad error message when attempting to invoke miner commands on non-miner actor

mslipper:
Closes #2116. All `*Test` methods initialize a struct that contains the syncer params, which is in turn passed to all the other initialization/`require*` methods. Let me know if there's an alternative approach you'd like me to implement.

### DIFF
--- a/chain/default_store_test.go
+++ b/chain/default_store_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/repo"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -86,6 +87,7 @@ func requireHeadTipset(t *testing.T, chain chain.Store) types.TipSet {
 
 // Adding tipsets to the store doesn't error.
 func TestPutTipSet(t *testing.T) {
+	tf.UnitTest(t)
 	dstP := initDSTParams()
 
 	ctx := context.Background()
@@ -101,6 +103,7 @@ func TestPutTipSet(t *testing.T) {
 
 // Tipsets can be retrieved by key (all block cids).
 func TestGetByKey(t *testing.T) {
+	tf.UnitTest(t)
 	dstP := initDSTParams()
 
 	ctx := context.Background()
@@ -137,6 +140,7 @@ func TestGetByKey(t *testing.T) {
 
 // Tipsets can be retrieved by parent key (all block cids of parents).
 func TestGetByParent(t *testing.T) {
+	tf.UnitTest(t)
 	dstP := initDSTParams()
 
 	ctx := context.Background()
@@ -170,6 +174,7 @@ func TestGetByParent(t *testing.T) {
 }
 
 func TestGetMultipleByParent(t *testing.T) {
+	tf.UnitTest(t)
 	dstP := initDSTParams()
 
 	ctx := context.Background()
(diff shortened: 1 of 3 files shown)


mslipper:
Hmm, quite a few conflicts - if this is generally the right track let me know and I'll fix them. Would prefer to know the solution is OK before resolving.

ZenGround0:
@mslipper this is definitely the right approach and generally great.  As you've probably already seen the conflicts are most likely all related to our new usage pattern around the require and assert packages introduced in PR #2657.

mslipper:
Excellent. I'll resolve the conflicts shortly.

chain/default_store_test.go
```diff
@@ -84,108 +87,112 @@ func requireHeadTipset(t *testing.T, chain chain.Store) types.TipSet {
...
+		TipSetStateRoot: dstP.genStateRoot,
 	}
 	err := cs.PutTipSetAndState(ctx, genTsas)
 	assert.NoError(t, err)
 }
 
 // Tipsets can be retrieved by key (all block cids).
 func TestGetByKey(t *testing.T) {
+	dstP := initDSTParams()
 	tf.BadUnitTestWithSideEffects(t)
```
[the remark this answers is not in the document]
mslipper: This is done.

chain/default_store_test.go
```diff
@@ -16,50 +16,52 @@ import (
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/repo"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
-	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
```
frrist: BLOCKING: We will still need flags on all of these tests, if they can be ran in parallel then the `tf.UnitTest(t)` flag is appropriate, else they will need to continue using the `tf.BadUnitTestWithSideEffects(tf)` flag.

chain/default_syncer_test.go
```diff
@@ -613,7 +629,7 @@ func TestLoadFork(t *testing.T) {
 
 	fakeChildParams.Nonce = uint64(2)
 	forklink1blk3 := th.RequireMkFakeChild(t, fakeChildParams)
-	//th.FakeChildParams{Parent: forkbase, GenesisCid: genCid, StateRoot: genStateRoot, Nonce: uint64(2)})
```
shannonwells: Id say remove this commented out FakeChildParams statement rather than updating

chain/default_store_test.go
ZenGround0: Thanks @frrist.  Sorry for missing that @mslipper

chain/default_syncer_test.go
mslipper: 👍

chain/default_store_test.go
mslipper: Done.

frrist: At first pass I didn't realize this was going to allow us to remove the `BadUnitTestWithSideEffects` flag. This is massive, thank you @mslipper!!

PR #2701, opened 2019-05-02
Simplifications and cleanup to repo initialisation.

anorth:
No significant functional change is intended here, though the checks around whether a repo
directory is suitable for initialisation have become stricter (requiring it to be empty).

This is in preparation for inserting a symlink as the repo path #2684.

FYI @shannonwells - if this is approved and you want it, feel free to merge before I do.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -55,10 +55,15 @@ var initCmd = &cmds.Command{
 		if err != nil {
 			return err
 		}
-		rep, err := repo.CreateRepo(repoDir, newConfig)
+
+		if err := repo.InitFSRepo(repoDir, newConfig); err != nil {
+			return err
+		}
+		rep, err := repo.OpenFSRepo(repoDir)
 		if err != nil {
 			return err
 		}
+
 		// The only error Close can return is that the repo has already been closed
 		defer rep.Close() // nolint: errcheck
 
--- a/repo/fsrepo.go
+++ b/repo/fsrepo.go
@@ -38,26 +38,18 @@ const (
 	DefaultRepoDir = "repo"
 )
 
-// NoRepoError is returned when trying to open a repo where one does not exist
-type NoRepoError struct {
-	Path string
-}
-
 var log = logging.Logger("repo")
-
-func (err NoRepoError) Error() string {
-	return fmt.Sprintf("no filecoin repo found in %s.\nplease run: 'go-filecoin init [--repodir=%s]'", err.Path, err.Path)
-}
 
 // FSRepo is a repo implementation backed by a filesystem.
 type FSRepo struct {
-	path string
-
+	// Path to the repo root directory.
+	path    string
 	version uint
 
 	// lk protects the config file
-	lk       sync.RWMutex
-	cfg      *config.Config
+	lk  sync.RWMutex
+	cfg *config.Config
+
 	ds       Datastore
 	keystore keystore.Keystore
 	walletDs Datastore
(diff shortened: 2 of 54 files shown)


commands/init.go
```diff
@@ -55,10 +55,15 @@ var initCmd = &cmds.Command{
 		if err != nil {
 			return err
 		}
-		rep, err := repo.CreateRepo(repoDir, newConfig)
+
+		if err := repo.InitFSRepo(repoDir, newConfig); err != nil {
+			return err
+		}
+		rep, err := repo.OpenFSRepo(repoDir)
```
anorth: This was the only caller of CreateRepo, so I inlined it.

repo/fsrepo.go
```diff
@@ -36,26 +36,18 @@ const (
 	DefaultRepoDir         = "repo"
 )
 
-// NoRepoError is returned when trying to open a repo where one does not exist
-type NoRepoError struct {
-	Path string
-}
-
```
anorth: This structure was unreferenced externally.

repo/fsrepo_test.go
```diff
@@ -75,48 +75,59 @@ const (
 }`
 )
 
-func TestFSRepoInit(t *testing.T) {
+func TestInitRepo(t *testing.T) {
```
anorth: Removed duplication from two different test methods, making them more exhaustive.

ZenGround0:
The issue is that IPTB and this PR are making different assumptions about whether or not it is allowed to init a repo in a non-empty directory.  It is pretty baked into IPTB that you can do this.  While there are ways forward to make IPTB respect this new constraint I'll call out that it will be a somewhat involved redesign that will block this PR until after EOW unless (even if) we recruit @travisperson's full attention right now.

I don't have a strong opinion on which assumption is The Right Way but we can't do this as is written by release unless we treat this like an all hands on deck emergency.

ZenGround0:
Quick pointer to the opposite assumption in IPTB:
IPTB nodes run with a [binary setup](https://github.com/filecoin-project/go-filecoin/blob/master/tools/iptb-plugins/filecoin/local/localfilecoin.go#L111) in a daemon specific directory.   That daemon specific directory is the same as the repo.  

This setup needs to be called before the node gets init (so that there is a binary to run).

Thinking about this more it seems like a fairly simple change to iptb's data organization model would allow for an isolated repo.  Node initialization could create auxiliary locations for the node that are siblings of the repo.  It's a little messier but still pretty neat:

Now:
```
root
   0
      bin
      sectors
      config.json
      ...
   1
      bin
      sectors
      config.go
      ...
   ...
```

Mods to get this PR through:
```
   0_bin
   0_sectors
   0
      config.json
      ...
   1_bin
   1_sectors
   1
      config.go
      ...
   ...
```

I am an IPTB novice so I could easily be missing other assumptions elsewhere in the tooling that rely on all data being in one dir.  If I am not missing anything it would be a fairly simple change to `localfilecoin.go` to setup data this way.

anorth:
Ok, I'll revert this particular aspect of the change 😢. Stay tuned.

travisperson:
> I am an IPTB novice so I could easily be missing other assumptions elsewhere in the tooling that rely on all data being in one dir

IPTB provides each instance of a plugin with a directory to store information in. In the case of a local plugin we set the `FIL_PATH` to point to this directory. Everything must be contained inside of the directory that IPTB provides to the plugin.

The local filecoin plugin can be updated to do whatever it wants with that directory though, so we could append an additional layer to the provided path without issue.

anorth:
In #2709 I took @travisperson's suggestion to nest the repo inside the IPTB dir. Rebasing this PR on that now passes. This needs to be based back on master after that lands.

repo/fsrepo_test.go
```diff
@@ -368,69 +385,41 @@ func TestRepoAPIFile(t *testing.T) {
...
-		err = os.Mkdir(dir, 0444)
-		assert.NoError(t, err)
-		assert.False(t, ConfigExists(dir))
+	t.Log("snapshot path was created during FSRepo Init")
+	assert.True(t, fileExists(filepath.Join(path, snapshotStorePrefix)))
 
-		_, err = CreateRepo(dir, cfg)
-		assert.Contains(t, err.Error(), "permission denied")
-	})
+	// TODO: asserting the exact content here is gonna get old real quick
```
ZenGround0: Would be good to link an issue here

anorth:
Ping @travisperson. This is now based on master, which nests the repo inside the IPTB directory.

laser:
@travisperson - Would you mind reviewing this PR?

PR #2708, opened 2019-05-03
Repo migration version detection and decisioning

shannonwells:
* Check the fsrepo version and the repo.Version and figure out what migration to run
* Running the correct command.
* Validation is run before performing an install.
* Use the logger
* The version on disk is always updated on a successful validation.
* Bunch of test helper funcs

Also:
* Not finding a migration is not an error.
* MigrationRunner.Run returns a struct so main and tests can be smarter about run results
* Before running the install command we check the new repo version to try to make sure it has been migrated -- just in case.
* Updates to USAGE


tools/migration/internal/runner.go
```diff
@@ -27,22 +40,112 @@ type Migration interface {
...
+	// oldRepoOpt is the option passed  from the CLI.
 	oldRepoOpt string
+
+	// MigrationsProvider is a dependency for fetching available migrations
+	// to allow unit tests to supply test migrations without creating test fixtures.
+	MigrationsProvider func() []Migration
+
+	// RepoVersionGetter is a dependency for getting a Repo Version,
+	// to allow the binary version of the repo to be mocked in testing.
+	RepoVersionGetter func() uint
```
anorth: I see this is intended to abstract the hardcoded `repo.Version` constant. I don't think migration code should be referring to that constant at all. The version of an existing repo should be determined by inspecting it. That constant will need factoring out from almost all the places that refer to it,  its only use being to check that a repo loaded by go-filecoin matches the expected version.

tools/migration/internal/runner.go
```diff
@@ -27,22 +40,112 @@ type Migration interface {
...
+
 func (m *MigrationRunner) Run() error {
 	// TODO: Issue #2595 Implement first repo migration
+	repoVersion, err := m.loadVersion()
+	if err != nil {
+		return err
+	}
+	if repoVersion == repo.Version {
+		return errors.New("binary version = repo version; migration not run")
+	}
```
anorth: As above, the repo version at which the migration binary was built should not be relevant. Inspect the repo that was specified as the target (oldRepoOpt)

tools/migration/internal/runner.go
```diff
@@ -27,22 +40,112 @@ type Migration interface {
...
+	for _, mig := range m.MigrationsProvider() {
+		from, to := mig.Versions()
+		if from == repoVersion && to == from+1 {
```
anorth: I don't think this is the right form for this check. Clarify that it's just a policy rather than fundamentally necessary to this code that the version bump is exactly one. I can imagine losing this check with no ill effects, if we decide to.

```
if to != from + 1 {
  log.warn("Refusing multi-version migration from %d to %d", ...)
  continue
}
if (from == repoVersion) { ... }

Comment (anorth) [inline tools/migration/internal/runner.go]: Rather than immediately taking the first match, traverse all of them and confirm that there is exactly one match first, failing (for now) if there are multiple.

File: tools/migration/internal/runner.go [hunk 4]
```diff
@@ -27,22 +40,112 @@ type Migration interface {
...
+	if err != nil {
+		return 0, err
+	}
+
+	if version < 0 || version > 10000 {
+		return 0, fmt.Errorf("repo version out of range: %s", strVersion)
+	}
+
+	return uint(version), nil
+}
```
anorth: I think we're better off sharing this code (perhaps as a free function in fsrepo), if possible. Definitely remove duplication of the "version" filename

tools/migration/internal/runner.go
```diff
@@ -27,22 +40,112 @@ type Migration interface {
...
+
+func (m *MigrationRunner) validateAndUpdateVersion(toVersion uint, newRepoPath string, mig Migration) error {
+	if err := mig.Validate(m.oldRepoOpt, newRepoPath); err != nil {
+		return err
```
anorth: Wrap the error

tools/migration/internal/runner_test.go
```diff
@@ -1,18 +1,178 @@
...
+func (m *TestMigValidationFails) Describe() string {
+	return "the migration that doesn't do anything"
+}
+
+func (m *TestMigValidationFails) Migrate(newRepoPath string) error {
+	return nil
+}
+
+func (m *TestMigValidationFails) Validate(oldRepoPath, newRepoPath string) error {
+	return errors.New("validation has failed")
```
anorth: You could compress this code by defining the migration at runtime, with a TestMigration struct that has version, description, and function pointers.

tools/migration/internal/runner.go
ZenGround0: IMO duplication of the "version" filename is WAI.  I think we can assert that this is a well known and agreed upon feature of any filesystem repo just as the [fs-repo spec for IPFS does](https://github.com/ipfs/specs/tree/master/repo/fs-repo): "Any fs-repo implementation of any versions >0 MUST be able to read the version file. It MUST NOT change format between versions."

ZenGround0: Though I guess keeping a free function in fs-repo for this fits into this world view just as well

tools/migration/internal/runner.go
ZenGround0: We need some way to figure out which version is the most recent that this binary will try to get us to.  Using repo.Version seems like a good way to do this.  If we want to separate out to prevent dependency shifts from messing up rebuilds then we could have a MaxVersion or something defined in the migration binary.

shannonwells: So now we look at our migrations list and find the largest version number.  We don't need to depend on the repo version number in fsrepo.

tools/migration/internal/runner.go
shannonwells: Added the free function repo.VersionFilename

tools/migration/internal/runner_test.go
```diff
@@ -1,26 +1,223 @@
...
+
+	t.Run("Runs the right migration version", func(t *testing.T) {
+		RequireSetRepoVersion(t, 1, repoDir)
+
+		dummyLogFile, dummyLogPath := RequireOpenTempFile(t, "logfile")
+		defer RequireRemove(t, dummyLogPath)
+		logger := NewLogger(dummyLogFile, false)
+		runner := NewMigrationRunner(logger, "describe", repoSymlink, "")
+		runner.MigrationsProvider = testProviderValidationFails
+		assert.EqualError(t, runner.Run(), "binary version 1 = repo version 1; migration not run")
```
anorth: This looks like a copy-paste error from the previous test, judging from the t.run description.

tools/migration/internal/utils_test.go
```diff
@@ -14,3 +14,7 @@ func TestNowString(t *testing.T) {
 	rg, _ := regexp.Compile("^[0-9]{8}-[0-9]{6}$")
 	assert.Regexp(t, rg, adateStr)
 }
+
+func SomeTestHelper() string {
+	return "helper"
+}
```
anorth: Unused

tools/migration/internal/runner.go
```diff
@@ -26,27 +41,169 @@ type Migration interface {
...
+		m.logger.Print(fmt.Sprintf("new repo will be at %s", m.newRepoPath))
+
+		if err = mig.Migrate(m.newRepoPath); err != nil {
+			return errors.New("migration failed: " + err.Error())
+		}
+	case "install":
+		if m.newRepoPath == "" {
+			return errors.New("installation failed: new repo is missing")
+		}
+		if err = m.validateAndUpdateVersion(to, m.newRepoPath, mig); err != nil {
```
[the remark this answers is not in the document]
ZenGround0: IMO version bumping could fall under the actual `Migrate` function of different migrations.

tools/migration/internal/repo_fs_helpers.go
```diff
@@ -60,34 +60,22 @@ func InstallNewRepo(oldRepoLink, newRepoPath string) error {
...
-// OpenRepo opens repoPath
-func OpenRepo(repoPath string) (*os.File, error) {
-	return os.Open(repoPath)
-}
-
 // getNewRepoPath generates a new repo path for a migration.
 // Params:
 //     oldPath:  the actual old repo path
-//     newRepoOpt:  whatever was passed in by the CLI (can be blank)
+//     newRepoPath:  whatever was passed in by the CLI (can be blank)
```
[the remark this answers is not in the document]
shannonwells: I put it back; it's actually needed if you run just install; otherwise we have no way of knowing what the new repo path is.

tools/migration/internal/runner_test.go
shannonwells: It was a leftover test idea I meant to delete.

tools/migration/internal/runner.go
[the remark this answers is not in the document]
shannonwells: > IMO version bumping could fall under the actual Migrate function of different migrations.

I strongly feel that version bumping should be controlled by the Runner, whose job it is to also call Validation, and should not be done until Validation passes. Otherwise you leave the new repo in a half-migrated state, esp. given we use version to detect whether migration is needed.

tools/migration/internal/runner.go
```diff
@@ -26,27 +41,169 @@ type Migration interface {
...
+	if err != nil {
+		return err
+	}
+	if repoVersion == m.getTargetMigrationVersion() {
+		return fmt.Errorf("binary version %d = repo version %d; migration not run", repoVersion, m.getTargetMigrationVersion())
+	}
+
+	var mig Migration
+	if mig, err = m.getValidMigration(repoVersion); err != nil {
+		return fmt.Errorf("migration check failed: %s", err.Error())
```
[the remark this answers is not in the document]
shannonwells: If getValidMigration fails, it could be that the version file was corrupt or there is >1 valid migration, in which case there is a migration bug.  By the time of this check, we would already have detected whether the repo is up-to-date (which is not an error case), so we know migration is needed. If we can't find one and only one that applies, that sounds like a valid error to me.

tools/migration/internal/runner.go
shannonwells: My mistake, I misunderstood what you asked for.  I also still thought we were trying to minimize intrusions into go-filecoin.

tools/migration/internal/runner.go
shannonwells: I don't think I understand this comment entirely. The user will definitely want to know if they are up to date as opposed to some other version condition. It should not just say a valid migration can't be found.  The other reason to do this check is an early bail out. If there's nothing to do, then why go through the rigamarole of finding valid migrations. 

Since we know what the end migration version will be from what's in our own list, as you pointed out @anorth, technically it's nothing to do with what is at `repo.Version`, although these had better be the same. I used repo.Version initially, because it was quick to check and because it should never be out of sync.

Failing to find valid migrations otherwise, I think should actually produce an error; please see my comment below.

tools/migration/internal/repo_fs_helpers.go
```diff
@@ -60,34 +60,22 @@ func InstallNewRepo(oldRepoLink, newRepoPath string) error {
...
-	}
-
+func getNewRepoPath(oldPath string) (string, error) {
 	// unlikely to see a name collision but make sure; making it loop up to 1000
 	// ensures that even if there are 1000 calls/sec then the timestamp will change
 	// anyway.
 	var newpath string
 	for i := 1; i < 1000; i++ {
-		newpath = strings.Join([]string{newRepoPrefix, NowString(), fmt.Sprintf("%03d", i)}, "-")
+		newpath = strings.Join([]string{oldPath, NowString(), fmt.Sprintf("%03d", i)}, "-")
```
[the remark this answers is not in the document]
shannonwells: yes

tools/migration/internal/runner.go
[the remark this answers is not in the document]
shannonwells: I forgot to add, the original reasoning behind putting validation in the Install step is so you couldn't shoot yourself in the foot if something went wrong and the failure was undetected or forgotten.  Having the install command do the validate before actually swapping out the links provides some additional safety.  Maybe there should at least be a version check before the actual Install call.

tools/migration/internal/runner_buildonly_test.go
```diff
@@ -0,0 +1,33 @@
...
+		defer RequireRemove(t, dummyLogPath)
+		logger := NewLogger(dummyLogFile, false)
+		runner := NewMigrationRunner(logger, "buildonly", repoSymlink, "")
+		runner.MigrationsProvider = testProviderPasses
+
+		assert.NoError(t, runner.Run())
+
+		version, err := runner.GetNewRepoVersion()
+		require.NoError(t, err)
+		assert.Equal(t, uint(0), version)
```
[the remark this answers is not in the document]
shannonwells: Added test helpers to assert these things in multiple tests.  I believe the assertions cover implicitly or explicitly all of these points.

tools/migration/internal/runner.go
```diff
@@ -26,27 +41,169 @@ type Migration interface {
...
 
-	return m.logger.Close()
+	switch m.command {
+	case "describe":
+		m.logger.Print(mig.Describe())
+	case "migrate":
+		if m.newRepoPath, err = CloneRepo(m.oldRepoOpt); err != nil {
+			return err
+		}
+		m.logger.Print(fmt.Sprintf("new repo will be at %s", m.newRepoPath))
```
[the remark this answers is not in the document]
shannonwells: Filed issue #2740

tools/migration/main.go
```diff
@@ -70,36 +82,56 @@ func main() { // nolint: deadcode
...
+			}
 		}
 
-		runner := internal.NewMigrationRunner(logger, command, oldRepoOpt)
-		if err := runner.Run(); err != nil {
+		runner, err := internal.NewMigrationRunner(logger, command, oldRepoOpt, newRepoOpt)
+		if err != nil {
 			exitErr(err.Error())
 		}
+		runResult := runner.Run()
```
shannonwells: Use the new RunResult struct

tools/migration/internal/runner.go
```diff
@@ -83,22 +83,47 @@ func NewMigrationRunner(logger *Logger, command, oldRepoOpt, newRepoOpt string)
...
 		m.logger.Print(fmt.Sprintf("Repo up-to-date: binary version %d = repo version %d", repoVersion, m.getTargetMigrationVersion()))
-		return nil
+		return RunResult{OldVersion: repoVersion, NewVersion: repoVersion}
 	}
 
 	var mig Migration
 	if mig, err = m.getValidMigration(repoVersion); err != nil {
-		return fmt.Errorf("migration check failed: %s", err.Error())
+		return RunResult{
+			Err:        fmt.Errorf("migration check failed: %s", err.Error()),
```
anorth: Prefer `errors.Wrap()`

tools/migration/main.go
```diff
@@ -70,36 +82,56 @@ func main() { // nolint: deadcode
 		if err != nil {
 			exitErr(err.Error())
 		}
+
 		logger := internal.NewLogger(logFile, getVerbose())
```
ZenGround0: I don't see us closing the logfile anywhere with logger.Close().  We should do that.  It's actually a pain to do because I don't think we can use defers when we exit.

Feel free to open up a follow on issue if you want to handle this work after this PR.

tools/migration/internal/runner_buildonly_test.go
```diff
@@ -0,0 +1,51 @@
...
+		defer RequireRemove(t, dummyLogPath)
+		logger := NewLogger(dummyLogFile, false)
+		runner, err := NewMigrationRunner(logger, "buildonly", repoSymlink, "")
+		require.NoError(t, err)
+		runner.MigrationsProvider = testProviderPasses
+
+		runResult := runner.Run()
+		assert.NoError(t, runResult.Err)
+
+		AssertBumpedVersion(t, runResult.NewRepoPath, repoDir, 0)
```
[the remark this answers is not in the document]
shannonwells: These tests actually fail if you don't do the setup independently, because the previous test run affects the state.

tools/migration/internal/test_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+	require.NoError(t, os.Symlink(repoDir, symLink))
+
+	require.NoError(t, repo.WriteVersion(repoDir, repoVersion))
+	return repoDir, symLink
+}
+
+// CaptureOutput puts log content into
+func CaptureOutput(f func()) string {
+	var buf bytes.Buffer
+	log.SetOutput(&buf)
```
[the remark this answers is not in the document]
shannonwells: Yea it makes me itch, but it's in test only, and I'm not sure how else to capture stderr/stdout.  Ideas?

tools/migration/internal/repo_fs_helpers.go
```diff
@@ -47,10 +47,15 @@ func CloneRepo(oldRepoLink string) (string, error) {
 // InstallNewRepo archives the old repo, and symlinks the new repo in its place.
 // returns any error.
 func InstallNewRepo(oldRepoLink, newRepoPath string) error {
+	var err error
```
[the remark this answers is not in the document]
shannonwells: If you declare it at the top then you never have to remember to change the  := if you're rearranging/adding/deleting the clauses

tools/migration/internal/runner.go
```diff
@@ -26,27 +40,219 @@ type Migration interface {
...
+		oldRepoOpt:         oldPath,
+		newRepoPath:        newPath,
+		MigrationsProvider: DefaultMigrationsProvider,
+	}, nil
+}
+
+// RunResult stores the needed results of calling Run()
+type RunResult struct {
+	// full path to new repo (migrated or not).
+	// This is blank unless repo was cloned -- if it errors out to early or for "describe"
```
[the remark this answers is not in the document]
shannonwells: I can't even believe I did that.  My English teacher mother would shame me.

tools/migration/internal/runner.go
```diff
@@ -26,27 +40,219 @@ type Migration interface {
...
+	targetVersion := m.getTargetMigrationVersion()
+	if repoVersion == targetVersion {
+		m.logger.Print(fmt.Sprintf("Repo up-to-date: binary version %d = repo version %d", repoVersion, m.getTargetMigrationVersion()))
+		return RunResult{OldVersion: repoVersion, NewVersion: repoVersion}
+	}
+
+	var mig Migration
+	if mig, err = m.findMigration(repoVersion); err != nil {
+		return RunResult{
+			Err:        errors.Wrapf(err, "migration check failed"),
```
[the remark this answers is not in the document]
shannonwells: The wrapped errors will contain the versions and more explanation, but I did notice that one of those did not have the version in it, so I'll add it.

tools/migration/internal/test_helpers.go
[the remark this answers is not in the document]
ZenGround0: hmm yeah I understand better now.  This seems necessary as long as we use global logging in exitErr and doing anything else there ups complexity without a whole lot of benefit

tools/migration/main.go
shannonwells: I'll file a PR -- I'd like to discuss the best way, because logger.Close() returns an error, also we could pass the logger to exitErr, etc etc.

tools/migration/internal/runner.go
```diff
@@ -26,27 +40,219 @@ type Migration interface {
...
-		oldRepoOpt: oldRepoOpt,
+	// MigrationsProvider is a dependency for fetching available migrations
+	// to allow unit tests to supply test migrations without creating test fixtures.
+	MigrationsProvider func() []Migration
+}
+
+// NewMigrationRunner builds a MigrationRunner for the given command and repo options
+// Returns an error if homepath expansion fails for oldRepoOpt or newRepoOpt
+func NewMigrationRunner(logger *Logger, command, oldRepoOpt, newRepoOpt string) (*MigrationRunner, error) {
+	oldPath, err := homedir.Expand(oldRepoOpt)
```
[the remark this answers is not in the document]
shannonwells: It turns out that it doesn't get expanded.  When I run the the command without this code, this is what happens:
```
$ ./go-filecoin-migrate describe --old-repo=~/.filecoin-symlink
Error: open ~/.filecoin-symlink/version: no such file or directory
$ go build -o go-filecoin-migrate .
```
When I put it back:
```
$ ./go-filecoin-migrate describe --old-repo=~/.filecoin-symlink
$ 
```

tools/migration/internal/repo_fs_helpers.go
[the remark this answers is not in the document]
shannonwells: Oh never mind, I was looking in the wrong place. I hadn't realized it was the only one without the :=

tools/migration/main.go
shannonwells: #2760

PR #2712, opened 2019-05-03
Feat/paych cancel fails with valid conditions 2678

### LINKED ISSUE #2678: Payment Channel Cancel Fails if Condition exists and is valid
Blocked on #2677

### Description

We don't want payers to be able to harass targets by continually canceling a channel forcing the target to respond with more expensive redeem calls. As of #2677, we will be storing whether the target has successfully redeemed against a payment channel and the last condition they've used. We can use this information to block frivolous calls to cancel.

The `paymentbroker.Actor.Cancel` method should fail if:

1. The payment channel `redeemed` flag is true, and there is no condition.
2. The payment channel `redeemed` flag is true, a condition exists, and no revert error is returned when the condition is converted into a message and called (i.e. the condition is still valid)

### Acceptance criteria

1. The payment channel respect the conditions described above.
2. Integration tests exist for both code paths.

### Risks + pitfalls

Sending a message and continuing only when an error is return, is unnatural and a little risky. Some thought care be given to which conditions constitute a validation failure versus a general system failure. This should be the difference between a revert error and a fault, but there might be scenarios that don't fit that categorization.

rosalinekarr:
# Problem

The current implementation of payment channel cancel allows payers to harass payees by repeatedly cancelling channels and forcing payees to reopen them by redeeming.

# Solution

Only allow payers to cancel payment channels when the channel either hasn't been redeemed yet or the last redeem included invalid conditions.

Resolves #2678

### DIFF
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		if channel.Redeemed {
+			if channel.Condition == nil {
+				return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
+			} else {
+				err := checkCondition(vmctx, channel, channel.Condition, []interface{}{})
+				if err != nil && !errors.IsFault(err) {
+					return err
+				}
+				if err == nil {
+					return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
+				}
+			}
+		}
+
 		eol := vmctx.BlockHeight().Add(types.NewBlockHeight(CancelDelayBlockTime))
 
 		// eol can only be decreased
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -40,6 +40,8 @@ const (
 	ErrTooEarly = 43
 	//ErrConditionInvalid indicates that the condition attached to a voucher did not execute successfully
 	ErrConditionInvalid = 44
+	//ErrConditionValid indicates that the condition attached to a voucher did execute successfully and therefore can't be cancelled
+	ErrConditionValid = 45
 )
 
 // CancelDelayBlockTime is the number of rounds given to the target to respond after the channel
@@ -416,14 +418,14 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 
 		if channel.Redeemed {
 			if channel.Condition == nil {
-				return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
+				return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
 			} else {
 				err := checkCondition(vmctx, channel, channel.Condition, []interface{}{})
-				if err != nil && !errors.IsFault(err) {
-					return err
-				}
 				if err == nil {
-					return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
+					return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
+				}
+				if !errors.ShouldRevert(err) {
+					return err
 				}
 			}
 		}
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -416,14 +416,21 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		// If the payment channel has been successfully redeemed, check if it's
+		// valid and fail to prevent abuse
 		if channel.Redeemed {
+			// If it doesn't have a condition, it's valid, so throw an error
 			if channel.Condition == nil {
 				return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
 			} else {
+				// Otherwise, check the condition on the payment channel
 				err := checkCondition(vmctx, channel)
+				// If we receive no error, the condition is valid, so we fail
 				if err == nil {
 					return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
 				}
+				// If there's a non-revert error, we have bigger problem, so raise the
+				// error
 				if !errors.ShouldRevert(err) {
 					return err
 				}
--- a/actor/builtin/paymentbroker/paymentbroker.go
+++ b/actor/builtin/paymentbroker/paymentbroker.go
@@ -420,7 +420,7 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			if channel.Condition == nil {
 				return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
 			} else {
-				err := checkCondition(vmctx, channel, channel.Condition, []interface{}{})
+				err := checkCondition(vmctx, channel)
 				if err == nil {
 					return errors.NewCodedRevertError(ErrConditionValid, "channel cannot be cancelled due to successful redeem")
 				}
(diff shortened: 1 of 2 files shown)


actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -825,6 +825,60 @@ func TestPaymentBrokerCancel(t *testing.T) {
...
+	toAddress := addrGetter()
+	method := "paramsNotZero"
+	addrParam := addrGetter()
+	sectorIdParam := uint64(6)
+	payerParams := []interface{}{addrParam, sectorIdParam}
+	blockHeightParam := types.NewBlockHeight(43)
+	redeemerParams := []interface{}{blockHeightParam}
+
+	sys := setup(t)
+	require.NoError(t, sys.st.SetActor(context.TODO(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
```
frrist: baby nit: call this `context.Background()` unless we intend to alter this context later. Code should use context.TODO when it's unclear which Context to use or it is not yet available. Ditto this below.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -825,6 +825,60 @@ func TestPaymentBrokerCancel(t *testing.T) {
...
+
+	// Redeem the payment channel with bad params and expect invalid condition error
+	condition := &types.Predicate{To: toAddress, Method: method, Params: payerParams}
+	result, err := sys.applySignatureMessage(sys.target, 200, types.NewBlockHeight(0), 0, "redeem", 0, condition)
+	require.NoError(t, err)
+	require.Error(t, result.ExecutionError)
+	require.EqualValues(t, errors.CodeError(result.ExecutionError), ErrConditionInvalid)
```
frrist: The params are backwards:
```suggestion
require.EqualValues(t, ErrConditionInvalid, errors.CodeError(result.ExecutionError))
```

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		if channel.Redeemed {
+			if channel.Condition == nil {
+				return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
+			} else {
+				err := checkCondition(vmctx, channel, channel.Condition, []interface{}{})
+				if err != nil && !errors.IsFault(err) {
```
acruikshank: BLOCKING: If the error is not a revert error (`IsRevert`) it should be returned. Faults (and bad programmer) errors should be returned immediately.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
...
+					return err
+				}
+				if err == nil {
+					return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
```
acruikshank: BLOCKING: a passing condition here should produce a coded revert error. Faults indicate that some aspect of internal state is corrupt and block processing should be abandoned altogether.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		if channel.Redeemed {
+			if channel.Condition == nil {
+				return errors.NewFaultError("Channel cannot be cancelled due to successful redeem")
```
acruikshank: BLOCKING: This error must be a revert error and should be coded.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -825,6 +825,60 @@ func TestPaymentBrokerCancel(t *testing.T) {
...
+	payerParams := []interface{}{toAddress, sectorIdParam}
+
+	sys := setup(t)
+	require.NoError(t, sys.st.SetActor(context.Background(), toAddress, actor.NewActor(pbTestActorCid, types.NewZeroAttoFIL())))
+
+	// Redeem the payment channel with bad params and expect invalid condition error
+	condition := &types.Predicate{To: toAddress, Method: method, Params: payerParams}
+	result, err := sys.applySignatureMessage(sys.target, 200, types.NewBlockHeight(0), 0, "redeem", 0, condition)
+	require.NoError(t, err)
+	require.Error(t, result.ExecutionError)
```
acruikshank: BLOCKING: The cancel succeeds below because the redeem fails and the `Redeemed` flag never gets set. In order to demonstrate the cancel can succeed after a condition is set:

1. Redeem must be called with a condition successfully.
2. Establish that the condition and redeem flag are set in the payment channel.
3. Change internal state occur to invalidate the condition.
4. Call Redeem and establish it succeeds.

The easiest way to alter internal state is add a method to PBTestActor that succeeds or fails based on some flag. Ideally this would be in actor state, but it's probably ok to create a field on PBTestActor itself.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -825,6 +825,60 @@ func TestPaymentBrokerCancel(t *testing.T) {
...
+	require.EqualValues(t, ErrConditionInvalid, errors.CodeError(result.ExecutionError))
+
+	// Attempt to Cancel and expects success
+	pdata := core.MustConvertParams(sys.channelID)
+	msg := types.NewMessage(sys.payer, address.PaymentBrokerAddress, 1, types.NewAttoFILFromFIL(1000), "cancel", pdata)
+	_, err = sys.ApplyMessage(msg, 100)
+	assert.NoError(t, err)
+}
```
acruikshank: BLOCKING: We need a test that tests that cancel fails when the payment channel is redeemed with a nil condition.

actor/builtin/paymentbroker/paymentbroker_test.go
```diff
@@ -825,6 +825,60 @@ func TestPaymentBrokerCancel(t *testing.T) {
...
+	condition := &types.Predicate{To: toAddress, Method: method, Params: payerParams}
+	result, err := sys.applySignatureMessage(sys.target, 100, types.NewBlockHeight(0), 0, "redeem", 0, condition, redeemerParams...)
+	require.NoError(t, err)
+	require.NoError(t, result.ExecutionError)
+
+	// Attempts to Cancel and expects failure
+	pdata := core.MustConvertParams(sys.channelID)
+	msg := types.NewMessage(sys.payer, address.PaymentBrokerAddress, 1, types.NewAttoFILFromFIL(1000), "cancel", pdata)
+	_, err = sys.ApplyMessage(msg, 100)
+	assert.Error(t, err)
```
acruikshank: BLOCKING: Please assert the error code.

actor/builtin/paymentbroker/paymentbroker.go
acruikshank: Specifically
1. Check err == nil. if so return coded revert error.
2. Check if `!errors.IsRevert(err)`. If so return error unmodified.,
3. else continue.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
...
+			} else {
+				err := checkCondition(vmctx, channel, channel.Condition, []interface{}{})
```
acruikshank: This is a little confusing. It would be better to extract the part of `checkCondtion` that makes the call from the part that handles storing the condition, and just call that here.

actor/builtin/paymentbroker/paymentbroker.go
```diff
@@ -414,6 +414,20 @@ func (pb *Actor) Cancel(vmctx exec.VMContext, chid *types.ChannelID) (uint8, err
 			return errors.NewFaultError("Expected PaymentChannel from channels lookup")
 		}
 
+		if channel.Redeemed {
```
acruikshank: Please explain this logic in the method comments.

actor/builtin/paymentbroker/paymentbroker.go
rosalinekarr: `IsRevert` isn't a method implemented anywhere. Do you want me to write it?

rosalinekarr: Do you mean `ShouldRevert`?

rosalinekarr: (Answered offline: Yes)

PR #2710, opened 2019-05-03
issue-2553 GetTipSetAndState should be split into two methods v0

jscode017:
My first try for this issue.
I think that there are some question remain:
1. How to name abbreviate tipsetstateroot?
I see that most of tipset in code is named ts, so I name tipsetstateroot tssr, but it seems weird. So maybe something like tsStateRoot?
2.At [here](https://github.com/jscode017/go-filecoin/blob/34c19cb529fb27c26aad974f3897ee4462f77f80/chain/tip_index.go#L79) What I did in chain/tip_index.go is that the Get() function remains and write to new methods to access Get() function then access TipSet and TipSetStateRoot respectively, I'm not sure is this appropriate?

PR #2715, opened 2019-05-03
Refactor/extract latest state 2552

### LINKED ISSUE #2552: Extract getLatestState from the Blockchain Facade plumbing
### Description

Presently, there is a `getLatestState` private method in `plumbing/bcf/blockchain_facade.go` that should become a public method in the state package.

The functionality of this method is duplicated in several other places throughout the project where a public state package method could be used instead. (Just search for any occurrence of `LoadStateTree`. At the time of this writing, there's about 20 of them.)

Additionally, this method is the only part of the Blockchain Facade plumbing which depends on the cbor store (`cst *hamt.CborIpldStore`), so, once it's extracted, said dependency can be removed.

### Acceptance criteria

1. The `getLatestState` method is removed from `plumbing/bcf/blockchain_facade.go`
2. A new method for fetching latest state is added to the state package
3. All call sites for `LoadStateTree` are updated to use the new state package method as appropriate
4. The `cst *hamt.CborIpldStore` dependency is removed from `plumbing/bcf/blockchain_facade.go`

### Risks + pitfalls

Where in the state package this method should be added and what it should be named are open questions.

### Where to begin

`plumbing/bcf/blockchain_facade.go`

rosalinekarr:
# Problem

A lot of our plumbing and core logic is highly repetitive in fetching the latest state from the chain.

# Solution

Add some methods to the chain module and do a little refactoring to clean up the code.

Resolves #2552

### DIFF
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -15,9 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 
-	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/actor/builtin"
-	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/metrics/tracing"
 	"github.com/filecoin-project/go-filecoin/repo"
 	"github.com/filecoin-project/go-filecoin/state"
@@ -404,16 +402,6 @@ func (store *DefaultStore) LatestState(ctx context.Context) (state.Tree, error) 
 		return nil, err
 	}
 	return state.LoadStateTree(ctx, store.stateStore, stateCid, builtin.Actors)
-}
-
-// ActorFromLatestState gets the latest state and retrieves an actor from it.
-func (store *DefaultStore) ActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
-	st, err := store.LatestState(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	return st.GetActor(ctx, addr)
 }
 
 // GenesisCid returns the genesis cid of the chain tracked by the default store.
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -397,13 +397,18 @@ func (store *DefaultStore) BlockHeight() (uint64, error) {
 	return store.head.Height()
 }
 
-// ActorFromLatestState gets the latest state and retrieves an actor from it.
-func (store *DefaultStore) ActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
+// LatestState gets the latest state from the state Store.
+func (store *DefaultStore) LatestState(ctx context.Context) (state.Tree, error) {
 	stateCid, err := store.GetTipSetStateRoot(store.GetHead())
 	if err != nil {
 		return nil, err
 	}
-	st, err := state.LoadStateTree(ctx, store.stateStore, stateCid, builtin.Actors)
+	return state.LoadStateTree(ctx, store.stateStore, stateCid, builtin.Actors)
+}
+
+// ActorFromLatestState gets the latest state and retrieves an actor from it.
+func (store *DefaultStore) ActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
+	st, err := store.LatestState(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -9,16 +9,13 @@ import (
 	"github.com/cskr/pubsub"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-hamt-ipld"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	logging "github.com/ipfs/go-log"
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 
-	"github.com/filecoin-project/go-filecoin/actor/builtin"
 	"github.com/filecoin-project/go-filecoin/metrics/tracing"
 	"github.com/filecoin-project/go-filecoin/repo"
-	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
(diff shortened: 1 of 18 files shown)


rosalinekarr:
This PR will be much more useful and cleaner after #2710 is merged and this is rebased on top of it, so we don't need to worry about reviewing or merging it until then.

Also, this is not any kind of priority change. I just wrote it while blocked between tasks.

chain/default_store.go
```diff
@@ -391,13 +391,18 @@ func (store *DefaultStore) BlockHeight() (uint64, error) {
...
 	tsas, err := store.GetTipSetAndState(store.GetHead())
 	if err != nil {
 		return nil, err
 	}
-	st, err := state.LoadStateTree(ctx, store.stateStore, tsas.TipSetStateRoot, builtin.Actors)
+	return state.LoadStateTree(ctx, store.stateStore, tsas.TipSetStateRoot, builtin.Actors)
+}
+
+// ActorFromLatestState gets the latest state and retrieves an actor from it.
+func (store *DefaultStore) ActorFromLatestState(ctx context.Context, addr address.Address) (*actor.Actor, error) {
```
ZenGround0: We shouldn't be putting these on the chain store.  I don't know how actor got here but it doesn't belong here.  Long ago I put this exact `LatestState` on the function and @phritz brought up the good point that this helps make the api more redundant and adds more opportunity to footgun.

With my `LatestState` api back in the day we ended up having a tricky bug because `store.LatestState` and `store.GetTsas(store.Head())` did not match.

If we are trying to clean this up (which is a great idea) my strong opinion is we need to reduce the set of calls on the store (push them up to other layers) not add calls to the store.

ZenGround0: *on the function => on the chain store interface

shannonwells:
Based on Zen's comments it looks like this isn't quite ready yet - ping me once those comments are addressed and I'll review.

chain/default_store.go
rosalinekarr: This was added a while ago by @acruikshank. I'm planning to remove it in this PR, but first I needed to get some changes merged in under it from @jscode017.

rosalinekarr:
@ZenGround0 @shannonwells obviously this isn't a priority, but this PR should be ready for review now

node/node_test.go
```diff
@@ -151,29 +143,9 @@ func TestNodeStartMining(t *testing.T) {
...
-		Network:      net.New(minerNode.Host(), nil, nil, nil, nil, nil),
-		Wallet:       wallet.New(walletBackend),
-		Deals:        strgdls.New(minerNode.Repo.DealsDatastore()),
-	})
-	porcelainAPI := porcelain.New(plumbingAPI)
-
 	seed.GiveKey(t, minerNode, 0)
 	mineraddr, minerOwnerAddr := seed.GiveMiner(t, minerNode, 0)
-	_, err := storage.NewMiner(mineraddr, minerOwnerAddr, minerNode, minerNode.Repo.DealsDatastore(), porcelainAPI)
+	_, err := storage.NewMiner(mineraddr, minerOwnerAddr, minerNode, minerNode.Repo.DealsDatastore(), nil)
```
acruikshank: nice!

plumbing/bcf/blockchain_facade.go
```diff
@@ -4,25 +4,37 @@ import (
...
 	"github.com/filecoin-project/go-filecoin/sampling"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/ipfs/go-cid"
-	hamt "github.com/ipfs/go-hamt-ipld"
+	"github.com/ipfs/go-hamt-ipld"
 	"github.com/pkg/errors"
 )
 
+type bcfChainReader interface {
```
acruikshank: I like that you're moving away from the ReadStore interface. You should only include the methods that are used though.  It looks like this only needs 'BlockHeight`, `GetBlock`, `GetHead`, `GetTipSetStateRoot`, and `GetTipSet`.

plumbing/bcf/blockchain_facade.go
```diff
@@ -4,25 +4,37 @@ import (
...
+	GetTipSetStateRoot(tsKey types.SortedCidSet) (cid.Cid, error)
+	HeadEvents() *pubsub.PubSub
+	Load(ctx context.Context) error
+	Stop()
+}
+
 // BlockChainFacade is a facade pattern for the chain core api. It provides a
 // simple, unified interface to the complex set of calls in chain, state, exec,
 // and sampling for use in protocols and commands.
 type BlockChainFacade struct {
```
acruikshank: Non-Blocking: Incidentally, this struct isn't really acting as a façade. `plumbing.Api` and `porcelain.Api` are better examples of that pattern. This is really a collection of utility methods around chain state. `ChainState` might be a better name.

plumbing/msg/previewer.go
```diff
@@ -41,16 +47,10 @@ func (p *Previewer) Preview(ctx context.Context, optFrom, to address.Address, me
 		return types.NewGasUnits(0), errors.Wrap(err, "couldnt encode message params")
 	}
 
-	headTs := p.chainReader.GetHead()
-	stateCid, err := p.chainReader.GetTipSetStateRoot(headTs)
-	if err != nil {
-		return types.NewGasUnits(0), errors.Wrap(err, "couldnt get latest state root")
-	}
-	st, err := state.LoadStateTree(ctx, p.cst, stateCid, builtin.Actors)
+	st, err := chain.LatestState(ctx, p.chainReader, p.cst)
```
acruikshank: Great refactor.

plumbing/msg/queryer.go
```diff
@@ -44,16 +50,10 @@ func (q *Queryer) Query(ctx context.Context, optFrom, to address.Address, method
 		return nil, errors.Wrap(err, "couldnt encode message params")
 	}
 
-	headTs := q.chainReader.GetHead()
-	stateCid, err := q.chainReader.GetTipSetStateRoot(headTs)
-	if err != nil {
-		return nil, errors.Wrap(err, "couldnt get latest state root")
-	}
-	st, err := state.LoadStateTree(ctx, q.cst, stateCid, builtin.Actors)
+	st, err := chain.LatestState(ctx, q.chainReader, q.cst)
```
acruikshank: 💯

node/node.go
```diff
@@ -115,9 +127,6 @@ type Node struct {
 	AddNewlyMinedBlock newBlockFunc
 	blockTime          time.Duration
 	cancelMining       context.CancelFunc
-	GetAncestorsFunc   mining.GetAncestors
-	GetStateTreeFunc   mining.GetStateTree
-	GetWeightFunc      mining.GetWeight
```

node/node.go
```diff
@@ -81,13 +82,24 @@ var (
 
 type pubSubProcessorFunc func(ctx context.Context, msg pubsub.Message) error
 
+type nodeChainReader interface {
```
acruikshank: Non-Blocking: It's great that we're moving away from the `ChainStore` java style interface. It would be nice if this PR actually remove it, rather than leaving  both styles in the codebase, but it appears it's been replaced enough times to set the pattern. It would also be better to use the new consumer defined interfaces to stub or fake testing, but this is a step in the right direction.

rosalinekarr: That was sort of a side-goal of this PR, but I chose not to do it yet to keep the diff size down. I think it would be a great follow up task.

PR #2737, opened 2019-05-07
Fix/message pool update 0.2.0

ZenGround0:
This adds a correct FindCommonAncestor utility and updates the message pool to use it.  Tests still WIP

- [x] add correct findcommon ancestor utility
- [x] test findcommon ancestor
- [x] update message pool to use this utility
- [ ] test message pool with null block chains

### DIFF
--- a/chain/get_ancestors.go
+++ b/chain/get_ancestors.go
@@ -19,6 +19,9 @@ type recentAncestorsChainReader interface {
 	GetTipSet(tsKey types.SortedCidSet) (*types.TipSet, error)
 }
 
+// ErrNoCommonAncestor is returned when two chains assumed to have a common ancestor do not.
+var ErrNoCommonAncestor = errors.New("no common ancestor")
+
 // GetRecentAncestorsOfHeaviestChain returns the ancestors of a `TipSet` with
 // height `descendantBlockHeight` in the heaviest chain.
 func GetRecentAncestorsOfHeaviestChain(ctx context.Context, chainReader recentAncestorsChainReader, descendantBlockHeight *types.BlockHeight) ([]types.TipSet, error) {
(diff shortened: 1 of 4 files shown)


core/message_pool_test.go
```diff
@@ -221,6 +221,7 @@ func TestUpdateMessagePool(t *testing.T) {
 		MustAdd(p, m[0], m[1])
 
 		parent := types.TipSet{}
+
```
acruikshank: ?

core/message_pool_test.go
```diff
@@ -387,14 +398,21 @@ func TestUpdateMessagePool(t *testing.T) {
 		m := types.NewSignedMsgs(6, mockSigner)
 		MustAdd(p, m[3], m[5])
 
-		oldChain := NewChainWithMessages(store, types.TipSet{},
+		// TODO FIX
+		// TODO FIX
```
acruikshank: remove these

chain/get_ancestors.go
```diff
@@ -124,3 +127,43 @@ func CollectAtMostNTipSets(ctx context.Context, iterator *TipsetIterator, n uint
 	}
 	return ret, nil
 }
+
+// FindCommonAncestor returns the common ancestor of the two tipsets pointed to
+// by the input iterators.  If they share no common ancestor errIterComplete
```
anorth: ...ErrNoCommonAncestor will be returned

chain/get_ancestors.go
```diff
@@ -124,3 +127,43 @@ func CollectAtMostNTipSets(ctx context.Context, iterator *TipsetIterator, n uint
...
+// will be returned.
+func FindCommonAncestor(oldIter, newIter *TipsetIterator) (types.TipSet, error) {
```
anorth: nit: "old" and "new" are not relevant concepts within this function; names like "left" and "right" would avoid implying semantics that don't exist.

chain/get_ancestors.go
```diff
@@ -124,3 +127,43 @@ func CollectAtMostNTipSets(ctx context.Context, iterator *TipsetIterator, n uint
...
+		// Update the pointers.  Pointers move back one tipset if they
+		// point to a tipset at the same height or higher than the
+		// other pointer's tipset.
+		if newHeight >= oldHeight {
+			if err := newIter.Next(); err != nil {
+				return nil, err
+			}
+		}
+
+		if newHeight < oldHeight {
```
anorth: Blocking: why not `<=`? To make the symmetry obvious, I recommend mirroring L156: `if oldHeight >= newHeight`.

If I've misunderstood something that requires this asymmetry, it definitely needs a comment.

chain/get_ancestors_test.go
```diff
@@ -268,3 +268,129 @@ func TestGetRecentAncestorsStartingEpochIsNull(t *testing.T) {
...
+
+	afterNullBlock := th.RequireMkFakeChild(t, fakeChildParams)
+	requirePutBlocks(t, blockSource, afterNullBlock)
+	afterNullTS := th.RequireNewTipSet(t, afterNullBlock)
+	afterNullTsas := &chain.TipSetAndState{
+		TipSet:          afterNullTS,
+		TipSetStateRoot: genStateRoot,
+	}
+	th.RequirePutTsas(ctx, t, chainStore, afterNullTsas)
+	afterNullIter := chain.IterAncestors(ctx, chainStore, afterNullTS)
```
anorth: You could also check here that, even without a fork, the common ancestor of `expectedCA` and `afterNullTS` is `expectedCA`.

chain/get_ancestors.go
ZenGround0: `oldHeight >= newHeight` is better (though not strictly necessary).  Thanks for catching

PR #2742, opened 2019-05-08
Docs/update contributing guide with testing practices 1995

### LINKED ISSUE #1995: Update Contribution guide with `package <pkg>_test` pattern for all testing
### Description

This idea comes from work @shannonwells has done in two PRs

- #1939 
- #1795

This is the first time I've seen this pattern of using `<pkg>_test`, pretty cool what it provides.

- It makes it trivial to run all the tests in a single file (if test files do not share common functionality defined elsewhere, this is great for running tests on file save.
- Help ensure we are writing testable code.

Here are some resource about this kind of testing

- https://stackoverflow.com/a/31443271
- https://medium.com/@matryer/5-simple-tips-and-tricks-for-writing-unit-tests-in-golang-619653f90742

One issue of using the different package name is we lose access to private intervals of packages. However, I would say in general being dependent on the intervals of a package is not a great way to test them. Public API testing is great, and using this approach will help us make sure our code is testable, and reduce the likely hood of breaking tests when implementations change. It's also nice to be able to make changes and have the assurance that the behavior didn't change. Tests that accidentally depend on intervals make this harder.

I think we could default to using `package <pkg>_test` for all `<file>_test.go` that are testing `<file>`. When we determine that we need internal access to properly test something (hopefully few and far between) we could use the pattern `package <pkg>`, and the name for the test file would be `<file>_internal_test.go`.

This will also be a nice flag to indicate that we are depending on package internals. Authors and reviewers can use this as feedback to look into if we could improve the code to make testing it no dependent on internals of the package.

To summarize
- All tests should default to using `<file>_test.go` with `package <pkg>_test` for all exports of `<file>`.
- When interval state information is required, tests should be placed in `<file>_interval_test.go` with `package <pkg>`.

Thanks @shannonwells, I think this is something we should look into.

rosalinekarr:
# Problem

We have adopted testing practices that are not documented in the contribution guide.

# Solution

Document them.

Resolves #1995

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,11 +83,11 @@ We review every change before merging to master, using GitHub pull requests.
 
 `go-filecoin` requires 2 approvals for all PRs. We use the following process, which aims to merge code quickly and efficiently while avoiding both accidental and malicious introduction of bugs, unintended consequences, or vulnerabilities.
 
-1. The first review is done by a committer familiar with that area of the codebase. Once they deem it ready for merge, they will assign a maintainer for the second review. 
+1. The first review is done by a committer familiar with that area of the codebase. Once they deem it ready for merge, they will assign a maintainer for the second review.
 1. Once the PR has 2 Approvals and blocking comments have been addressed, the committer [rebases and merges](#landing-changes) the PR.
     - If the author has merge access as a committer or maintainer, they merge their own PRs.
 
-If your PR hasn't been reviewed in 3 days, pinging reviewers via Github or community chat is welcome and encouraged. 
+If your PR hasn't been reviewed in 3 days, pinging reviewers via Github or community chat is welcome and encouraged.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -283,6 +283,7 @@ that should not happen and that a dev should go look at.
 - Daemon tests (integration tests that run a node and send it commands):
   - Daemon tests should test integration, not comprehensive functionality
   - Daemon tests should validate that their responses conform to a JSON schema
+- Tests should be placed in a separate package, and follow the naming pattern `$PACKAGE_test`. For example, a test of the chain package should live in a package named `chain_test`.
```
anorth: LG. I think it's worth adding an escape hatch, which we will occasionally need. `In limited situations, a "white box" test may be placed in the same package as the code it tests`.

PR #2749, opened 2019-05-08
Feat/deals ls 2729

### LINKED ISSUE #2729: CLI for interacting with storage deals
### Description
Both storage miners and clients keep track of and persist the data associated with storage deals.  This data is needed to do things like redeem payments (for miners), and interact with the payment channel to arbitrate disputes.  This data is also useful to access for the purpose of managing your deals either as a client or a miner.  For example a client will want to be able to lookup when its deals are completed to know when it needs to make new deals.  A storage miner will similarly want to inspect deals to determine when it has openings for new sectors.

We should have some minimal way to access this stored data from the CLI.

### Acceptance criteria
1. There is some CLI pattern for listing all deals ever made by the node (as client and as miner).

One possibility is to add both `go-filecoin miner list-deals` and `go-filecoin client list-deals` as subcommands that print out the cids of deals.  We could then also add `go-filecoin show deal <cid>` to inspect each deal in greater detail.  

There are plenty of other ways to do this.  The above is just a suggestion that seems simple, ergonomic and respects existing CLI patterns.

Right now we store both client and miner deals together so in order to do things this way we'll need to filter all of the deals on the node by miner address or payment info's payer address.  Alternatively we could have the CLI wrap DealsLS directly and let higher level CLI tools do this filtering.

2. Deals are printed out in a way that make them easy to use (maybe default to json?)

### Risks + pitfalls
This could clutter up the CLI so try to do it in a way that respects existing patterns and does what users expect.

### Where to begin
Look at relevant plumbing calls: https://github.com/filecoin-project/go-filecoin/blob/master/plumbing/strgdls/store.go

If you go the route suggested above you can make a "filterDeals" porcelain call that uses a membership function to filter deals returned by DealsLs, and then at the commands level pass in a different closure to this porcelain function for the miner and client versions.  Then look at how the show command works add in reads from the deals datastore and ergonomic serialization of deal data.

rosalinekarr:
# Problem

We don't have any way for miners or clients to view a list of their current deals yet.

# Solution

Add that.

Resolves #2729

### DIFF
--- a/commands/client.go
+++ b/commands/client.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"strconv"
@@ -26,6 +27,7 @@ var clientCmd = &cmds.Command{
 		"propose-storage-deal": clientProposeStorageDealCmd,
 		"query-storage-deal":   clientQueryStorageDealCmd,
 		"list-asks":            clientListAsksCmd,
+		"list-deals":           clientListDealsCmd,
 		"payments":             paymentsCmd,
 	},
 }
@@ -225,6 +227,49 @@ respectively.
 		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, ask *porcelain.Ask) error {
 			fmt.Fprintf(w, "%s %.3d %s %s\n", ask.Miner, ask.ID, ask.Price, ask.Expiry) // nolint: errcheck
 			return nil
+		}),
+	},
+}
+
+type clientListDealResult struct {
+	Miner       address.Address   `json:"minerAddress"`
+	ProposalCid cid.Cid           `json:"proposalCid"`
+	State       storagedeal.State `json:"state"`
+}
+
+var clientListDealsCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "List all deals made by this client",
+		ShortDescription: `
+Lists all recorded deals made by this client with miners on the network. This
+may include pending deals, active deals, finished deals and cancelled deals.
+`,
+	},
+	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
+		deals, err := GetPorcelainAPI(env).DealsLs()
+		if err != nil {
+			return err
+		}
+
+		for _, deal := range deals {
+			out := &clientListDealResult{
+				Miner:       deal.Miner,
+				ProposalCid: deal.Response.ProposalCid,
+				State:       deal.Response.State,
+			}
+			if err = re.Emit(out); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	},
+	Type: clientListDealResult{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, res *clientListDealResult) error {
+			encoder := json.NewEncoder(w)
+			encoder.SetIndent("", "\t")
+			return encoder.Encode(res)
 		}),
 	},
 }
--- a/commands/client.go
+++ b/commands/client.go
@@ -246,16 +246,19 @@ may include pending deals, active deals, finished deals and cancelled deals.
 `,
 	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		deals, err := GetPorcelainAPI(env).DealClientLs()
+		dealsCh, err := GetPorcelainAPI(env).DealClientLs(req.Context)
 		if err != nil {
 			return err
 		}
 
-		for _, deal := range deals {
+		for deal := range dealsCh {
+			if deal.Err != nil {
+				return deal.Err
+			}
 			out := &clientListDealResult{
-				Miner:       deal.Miner,
-				ProposalCid: deal.Response.ProposalCid,
-				State:       deal.Response.State,
+				Miner:       deal.Deal.Miner,
+				ProposalCid: deal.Deal.Response.ProposalCid,
+				State:       deal.Deal.Response.State,
 			}
 			if err = re.Emit(out); err != nil {
 				return err
--- a/commands/client.go
+++ b/commands/client.go
@@ -233,6 +233,7 @@ respectively.
 
 type clientListDealResult struct {
 	Miner       address.Address   `json:"minerAddress"`
+	PieceCid    cid.Cid           `json:"pieceCid"`
 	ProposalCid cid.Cid           `json:"proposalCid"`
 	State       storagedeal.State `json:"state"`
 }
@@ -257,6 +258,7 @@ may include pending deals, active deals, finished deals and cancelled deals.
 			}
 			out := &clientListDealResult{
 				Miner:       deal.Deal.Miner,
+				PieceCid:    deal.Deal.Proposal.PieceRef,
 				ProposalCid: deal.Deal.Response.ProposalCid,
 				State:       deal.Deal.Response.State,
 			}
--- a/commands/client.go
+++ b/commands/client.go
@@ -246,7 +246,7 @@ may include pending deals, active deals, finished deals and cancelled deals.
 `,
 	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		deals, err := GetPorcelainAPI(env).DealsLs()
+		deals, err := GetPorcelainAPI(env).DealClientLs()
 		if err != nil {
 			return err
 		}
--- a/commands/miner.go
+++ b/commands/miner.go
@@ -449,16 +449,19 @@ may include pending deals, active deals, finished deals and rejected deals.
 `,
 	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		deals, err := GetPorcelainAPI(env).DealMinerLs()
+		dealsCh, err := GetPorcelainAPI(env).DealMinerLs(req.Context)
 		if err != nil {
 			return err
 		}
 
-		for _, deal := range deals {
+		for deal := range dealsCh {
+			if deal.Err != nil {
+				return deal.Err
+			}
 			out := &minerListDealResult{
-				Miner:       deal.Miner,
-				ProposalCid: deal.Response.ProposalCid,
-				State:       deal.Response.State,
+				Miner:       deal.Deal.Miner,
+				ProposalCid: deal.Deal.Response.ProposalCid,
+				State:       deal.Deal.Response.State,
 			}
 			if err = re.Emit(out); err != nil {
 				return err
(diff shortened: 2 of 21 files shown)


rosalinekarr:
Added a follow up task for dealing with the `context.TODO`s added in this PR: https://github.com/filecoin-project/go-filecoin/issues/2764

porcelain/storagedeals.go
```diff
@@ -1,25 +1,87 @@
...
+
+// DealMinerLs returns a channel with all deals received as a miner
+func DealMinerLs(ctx context.Context, plumbing dealMinerLsPlumbing) (<-chan *strgdls.StorageDealLsResult, error) {
+	results := make(chan *strgdls.StorageDealLsResult)
+
+	minerAddress, _ := plumbing.ConfigGet("mining.minerAddress")
+
+	dealCh, err := plumbing.DealsLs(ctx)
+	if err != nil {
+		return results, err
```
shannonwells: Why does this return results, err and the other return `nil, err` in the same place? 

This code is identical except for names and where the deal miner is or is not equal to the minerAddress.  Is there no way to DRY it up?  I'm sure you tried, just want to know what stopped it.

shannonwells: Also I had to extract the code into files and do a diff to understand what was different; this is a code smell to me, because it will tempt somebody to unify it, and while they'll surely fail, their time will have been blown understanding it.  I suggest trying to extract some of this into smaller function calls for readability, and/or comment.

protocol/storage/client.go
```diff
@@ -301,7 +302,7 @@ func (smc *Client) isMaybeDupDeal(p *storagedeal.Proposal) bool {
 
 // LoadVouchersForDeal loads vouchers from disk for a given deal
 func (smc *Client) LoadVouchersForDeal(dealCid cid.Cid) ([]*types.PaymentVoucher, error) {
-	storageDeal := smc.api.DealGet(dealCid)
+	storageDeal := smc.api.DealGet(context.TODO(), dealCid)
```
laser: `Client#LoadVouchersForDeal` should accept a context. `API#Payments` should pass its context to this method.

protocol/storage/miner.go
```diff
@@ -401,7 +400,7 @@ func (sm *Miner) processStorageDeal(proposalCid cid.Cid) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	d := sm.porcelainAPI.DealGet(proposalCid)
+	d := sm.porcelainAPI.DealGet(context.TODO(), proposalCid)
```
laser: `Miner#proposalAcceptor` and `Miner#proposalRejector` should be modified to accept a context. 

`Miner#handleMakeDeal` and `Miner#handleQueryDeal` should be modified to _create_ a context, and should pass this context through the call stack.

protocol/storage/miner.go
```diff
@@ -643,7 +642,7 @@ func (sm *Miner) onCommitSuccess(dealCid cid.Cid, sector *sectorbuilder.SealedSe
 
 // search the sector's piece info to find the one for the given deal's piece
 func (sm *Miner) findPieceInfo(dealCid cid.Cid, sector *sectorbuilder.SealedSectorMetadata) (*sectorbuilder.PieceInfo, error) {
-	deal := sm.porcelainAPI.DealGet(dealCid)
+	deal := sm.porcelainAPI.DealGet(context.TODO(), dealCid)
```
laser: Same story as `Miner#processStorageDeal`.

protocol/storage/miner.go
```diff
@@ -916,7 +915,7 @@ func (sm *Miner) submitPoSt(start, end *types.BlockHeight, seed types.PoStChalle
 
 // Query responds to a query for the proposal referenced by the given cid
 func (sm *Miner) Query(c cid.Cid) *storagedeal.Response {
-	storageDeal := sm.porcelainAPI.DealGet(c)
+	storageDeal := sm.porcelainAPI.DealGet(context.TODO(), c)
```
laser: `Miner#Query` should be modified to accept the context created in `Miner#handleQueryDeal`.

porcelain/storagedeals.go
rosalinekarr: The `return results, err` was a mistake. I've fixed it.

I also extracted some of the common functionality to DRY up the code.

protocol/storage/miner.go
rosalinekarr: Do you mean you want me to create a context in `handleQueryDeal`? Because there isn't currently a context in that method.

laser:
Yep, I’d create one there (if anywhere)

rosalinekarr:
@laser @shannonwells ready for re-review

commands/client.go
```diff
@@ -229,6 +231,52 @@ respectively.
...
+		dealsCh, err := GetPorcelainAPI(env).DealClientLs(req.Context)
+		if err != nil {
+			return err
+		}
+
+		for deal := range dealsCh {
+			if deal.Err != nil {
+				return deal.Err
+			}
+			out := &clientListDealResult{
```
laser: SUGGESTION: I would add the piece CID into this structure.

commands/miner.go
```diff
@@ -463,3 +466,49 @@ Values will be output as a ratio where the first number is the miner power and s
...
+		dealsCh, err := GetPorcelainAPI(env).DealMinerLs(req.Context)
+		if err != nil {
+			return err
+		}
+
+		for deal := range dealsCh {
+			if deal.Err != nil {
+				return deal.Err
+			}
+			out := &minerListDealResult{
```
laser: SUGGESTION: I would add piece CID to this structure, too.

porcelain/storagedeals.go
```diff
@@ -1,25 +1,88 @@
...
+
+	return outCh, nil
+}
+
+func filterDealChannel(dealCh <-chan *strgdls.StorageDealLsResult, filterFunc func(*storagedeal.Deal) bool) <-chan *strgdls.StorageDealLsResult {
+	outCh := make(chan *strgdls.StorageDealLsResult)
+
+	go func() {
+		for deal := range dealCh {
+			if deal.Err != nil || filterFunc(&deal.Deal) {
```
laser: QUESTION: What happens to this goroutine if the request context is done (and there are no longer `outCh` consumers) after we enter the for-loop? Will this goroutine leak?

rosalinekarr: `DealsLs` checks the context it's given in each loop, so if the context is done, the `dealCh` will be closed and this loop will stop. I thought about performing the same check here, but decided against it since it would be redundant.

laser:
@rosalinekarr - Is there anything I can do to help move this PR along?

rosalinekarr:
@laser, nope, this PR is currently blocked by an issue with retrieval testing, so I'm waiting until some fixes are merged into master to rebase and unblock it.

commands/client.go
```diff
@@ -229,6 +231,54 @@ respectively.
...
+			if err = re.Emit(out); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	},
+	Type: clientListDealResult{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, res *clientListDealResult) error {
```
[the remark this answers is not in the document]
rosalinekarr: JSON encoding is used by default, but it's not indented by default. I've been making a pattern of using indented JSON for output on commands that we don't have any better text encoding yet.

commands/miner.go
```diff
@@ -463,3 +466,51 @@ Values will be output as a ratio where the first number is the miner power and s
 		}),
 	},
 }
+
+type minerListDealResult struct {
```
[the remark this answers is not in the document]
rosalinekarr: My assumption was that the miner and clients will likely be interested in different aspects of deals in the future, but this may be premature optimization. Do you think I should combine them?

go.mod
```diff
@@ -16,6 +16,8 @@ require (
 	github.com/golang/mock v1.2.0 // indirect
 	github.com/golang/protobuf v1.3.1 // indirect
 	github.com/golangci/golangci-lint v1.15.0
+	github.com/google/go-github v17.0.0+incompatible
```
acruikshank: What is this? You don't seem to have introduced this dependency in the code.

go.sum
```diff
@@ -156,6 +156,10 @@ github.com/golangci/unconvert v0.0.0-20180507085042-28b1c447d1f4 h1:zwtduBRr5SSW
 github.com/golangci/unconvert v0.0.0-20180507085042-28b1c447d1f4/go.mod h1:Izgrg8RkN3rCIMLGE9CyYmU9pY2Jer6DgANEnZ/L/cQ=
 github.com/google/go-cmp v0.2.0 h1:+dTQ8DZQJz0Mb/HjFlkptS1FeQ4cWSnN941F8aEG4SQ=
 github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5aqRK0M=
+github.com/google/go-github v17.0.0+incompatible h1:N0LgJ1j65A7kfXrZnUDaYCs/Sf4rEjNlfyDHW9dolSY=
+github.com/google/go-github v17.0.0+incompatible/go.mod h1:zLgOLi98H3fifZn+44m+umXrS52loVEgC2AApnigrVQ=
+github.com/google/go-querystring v1.0.0 h1:Xkwi/a1rcvNg1PPYe5vI8GbeBY/jrVuDX5ASuANWTrk=
+github.com/google/go-querystring v1.0.0/go.mod h1:odCYkC5MyYFN7vkCjXpyrEuKhc/BUO6wN/zVPAxq5ck=
```
acruikshank: I don't think this change should be here.

porcelain/storagedeals.go
```diff
@@ -12,20 +16,127 @@ var (
...
+		}
+	}()
+
+	return out, nil
+}
+
+func filterDealChannel(dealCh <-chan *StorageDealLsResult, filterFunc func(*storagedeal.Deal) bool) <-chan *StorageDealLsResult {
+	outCh := make(chan *StorageDealLsResult)
+
+	go func() {
```
acruikshank: This is still creating 2 goroutines where only 1 is needed (part of the point of moving the code out of plumbing). If you moved the DealLs code in here and have it also run filterFunc before emitting the value to the output channel,  then you could implement `DealClientLs`, `DealMinerLs` and `DealLs` by calling `filterDealChannel` with different filterFuncs (where DealLs's filterFunc always returns true).

rosalinekarr: What's the advantage of using one goroutine instead of two? This seems more readable to me, and I don't see how it would have any impact on efficiency.

acruikshank: It's less code and less system complexity. Less things to go wrong. For those reasons, I think it'd also be more readable. I'm not concerned about the efficiency. I'll leave it up to you, but I think its better not to use goroutines to pipe together channels where it can be avoided.

PR #2758, opened 2019-05-09
Add `long` support for `chain head` command

nijynot:
This PR adds the `[--long | -l]` option support for `go-filecoin chain head`.

So calling `go-filecoin chain head -l` will print:

```
zDPWYqFCx9xBQHpiwanHkAsc3hBc7enrWk8VLzQwddj1iGjfQyvW	t2f4t333piq5hq4wbvult2ovhz4jofuh7ze4pfyki	zdpuAyLKsuDwDzV83fRuDczAhaAysFFNb4bpQsAhnau5ZyRuc	1146	0
zDPWYqFCz6Sy59iTcw2NKCY5itdydyuN6pEXN4wggpHa8L35PRe4	t24dp6qmnqicuz5qchpleq5qt45hprdjmutrqcc7a	zdpuAztEcGebEkE6eFgHm5yw14f5Ch9B4oEYhtUsEusYL5DFF	1146	0
```

similarly to `go-filecoin chain ls -l`.

### DIFF
--- a/commands/chain.go
+++ b/commands/chain.go
@@ -7,7 +7,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit"
 	"github.com/ipfs/go-ipfs-cmds"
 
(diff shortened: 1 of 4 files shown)


mishmosh:
Thanks for this contribution. Tagging @frrist for review.

frrist:
@nijynot thank you for this contribution! This will be really nice to have, and apologise for the delay in reviewing. Will be giving this a look today. 

I believe this change will have some (untested) effects on other projects using go-filecoin -- the [block explorer](http://user.kittyhawk.wtf:8000/) - https://github.com/filecoin-project/filecoin-explorer, and possibly [dashboard](https://stats.kittyhawk.wtf/) - https://github.com/filecoin-project/filecoin-network-stats.

@travisperson I believe this will have some effect on the block explorer since it uses a filecoin node to get the chain head and expects a CID, not a block. Can you speak to this being true? Gut says it would [happen here](https://github.com/filecoin-project/filecoin-explorer/blob/acc1c766a68d1a287acf667e9c853f8239f6c2f0/src/api/chain.js#L112-L117)


cc @mslipper, will returning a block from the command `go-filecoin chain head` have any effect on the current dashboard? 

If any of the above are true, and we wish to continue consuming a CID at those endpoints we could consider adding a `go-filecoin chain head -s` (--short) command such that only a CID is returned. Thoughts?

travisperson:
Changing the return type on this has some ramifications at the moment for the block explorer. The issue is that it's the only way we can get the CID of the head block. Technically we can recalculate it from the returned data, but I don't believe we have that code written for javascript yet to do that encoding. It might exist someplace in js-ipfs land.

I think we do want something like this though, because the current `chain head / show block` is a bit annoying to get more information about the head block. However, after reading this PR it sounds like we can work around that by doing `chain ls | head -n1`.

If we want to move a head with this PR we will have to make sure we don't break the [filecoin-explorer](https://github.com/filecoin-project/filecoin-explorer).

This PR may have to sit for a bit, or we can move it to an issue so we can figure out if we want to do this work.

anorth:
@mslipper could you comment on whether this is reasonable for us to land?

frrist:
@nijynot due to  https://github.com/filecoin-project/go-filecoin/pull/2758#issuecomment-494485527 I am going to close this (@mslipper feel free to chime in if this is something we can land). A new API is being worked on in https://github.com/filecoin-project/go-filecoin/issues/3098 that will hopfully make changes like this easier moving forward.

Issue #2532, opened 2019-04-09
Cleanly separated networks (DHT, pubsub, ...)


mishmosh:
This epic collects multiple issues around the topic of cleanly compartmentalizing networks and their DHTs. Whoever works on this, please propose a set of changes to resolve them as a group: #1700, #2357, #2421, #2426

anorth:
I believe that we're gaining useful information from the unintentional "attacks" by our old network on the new one each time we restart the user devnet. I propose that we push back fixing this for a while so that we can continue learning and fixing underlying issues.

I do think we will still want to make this separation around the time of launching a long-lived testnet though.

anorth:
Deleting this epic now after removing all the issues from it. Most were subsumed by network names (as part of protocol upgrade work).

PR #2765, opened 2019-05-09
add an approval step to user devnet deploy

ognots:
this will let us build the github release and docker image and give notice to community that they should upgrade their node before triggering user devnet deploy


mishmosh:
We should do this, but it won't full guard against the possibility of an old node adding an old block from the old chain. For that, we would need Clean DHTs (the 4 stories in this epic): https://github.com/filecoin-project/go-filecoin/issues/2532.

PR #2767, opened 2019-05-09
Added a Printf method to the migration logger

### LINKED ISSUE #2740: migration logger needs a printf
### Description
Instead of having to nest fmt.Sprintf calls, it was requested that the logger [implement a printf](https://github.com/filecoin-project/go-filecoin/pull/2708/files#r281045813)

### Acceptance criteria
can make calls like logger.Printf("foo %s", bar)

rosalinekarr:
# Problem

We want a Printf method on the migration logger for #2708 

# Solution

Add one!

Resolves #2740

### DIFF
--- a/tools/migration/internal/logger.go
+++ b/tools/migration/internal/logger.go
@@ -41,6 +41,11 @@ func (l *Logger) Print(msg string) {
 	l.logger.Print(msg)
 }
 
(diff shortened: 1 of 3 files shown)


anorth:
Could you update the existing calls that are wrapping `Sprintf` too?

PR #2768, opened 2019-05-09
A storage miner's power should be measured in an amount of bytes contributed to the network, not number of sectors

### LINKED ISSUE #2601: A miner's power reflects the amount of storage they've committed to the network

laser:
Fixes #2601 

## Why does this PR exist?

This PR brings the go-filecoin implementation a bit closer to the spec w/respect to how miner power is tracked. Today, miner power is recorded in the number of sectors which they have committed to the network. With the multiple sector sizes feature, miners X, Y, and Z on the same network may be committing sectors of different sizes and the power table needs to accommodate for that. 

## What's in this PR?

This PR modifies the power table and EC code and storage market actor (and tests, and ...) to track miner power in bytes amount instead of number of sectors committed to the network. Total storage is also tracked by bytes amount.

## Next Steps

Now that miner power (and total storage capacity) is measured in bytes, we can loosen the restriction that prevents a single network from supporting more than one sector size. My next task is to modify the create miner flow to prompt an operator to pick the appropriate sector size for their miner.

### DIFF
--- a/actor/builtin/storagemarket/storagemarket.go
+++ b/actor/builtin/storagemarket/storagemarket.go
@@ -57,9 +57,9 @@ type Actor struct{}
 type State struct {
 	Miners cid.Cid `refmt:",omitempty"`
 
-	// TotalCommitedStorage is the number of sectors that are currently committed
-	// in the whole network.
-	TotalCommittedStorage *big.Int
+	// TODO: Determine correct unit of measure. Could be denominated in the
+	// smallest sector size supported by the network.
+	TotalCommittedStorage *types.BytesAmount
 
 	ProofsMode types.ProofsMode
 }
@@ -74,7 +74,7 @@ func (sma *Actor) InitializeState(storage exec.Storage, proofsModeInterface inte
 	proofsMode := proofsModeInterface.(types.ProofsMode)
 
 	initStorage := &State{
-		TotalCommittedStorage: big.NewInt(0),
+		TotalCommittedStorage: types.NewBytesAmount(0),
 		ProofsMode:            proofsMode,
 	}
 	stateBytes, err := cbor.DumpObject(initStorage)
@@ -103,12 +103,12 @@ var storageMarketExports = exec.Exports{
 		Return: []abi.Type{abi.Address},
 	},
 	"updatePower": &exec.FunctionSignature{
-		Params: []abi.Type{abi.Integer},
+		Params: []abi.Type{abi.BytesAmount},
 		Return: nil,
 	},
 	"getTotalStorage": &exec.FunctionSignature{
 		Params: []abi.Type{},
-		Return: []abi.Type{abi.Integer},
+		Return: []abi.Type{abi.BytesAmount},
 	},
 	"getProofsMode": &exec.FunctionSignature{
 		Params: []abi.Type{},
@@ -185,7 +185,7 @@ func (sma *Actor) CreateMiner(vmctx exec.VMContext, pledge *big.Int, publicKey [
 // UpdatePower is called to reflect a change in the overall power of the network.
 // This occurs either when a miner adds a new commitment, or when one is removed
 // (via slashing or willful removal). The delta is in number of sectors.
-func (sma *Actor) UpdatePower(vmctx exec.VMContext, delta *big.Int) (uint8, error) {
+func (sma *Actor) UpdatePower(vmctx exec.VMContext, delta *types.BytesAmount) (uint8, error) {
 	if err := vmctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -208,7 +208,7 @@ func (sma *Actor) UpdatePower(vmctx exec.VMContext, delta *big.Int) (uint8, erro
 			return nil, errors.FaultErrorWrapf(err, "could not load lookup for miner with address: %s", miner)
 		}
 
-		state.TotalCommittedStorage = state.TotalCommittedStorage.Add(state.TotalCommittedStorage, delta)
+		state.TotalCommittedStorage = state.TotalCommittedStorage.Add(delta)
 
 		return nil, nil
 	})
@@ -220,7 +220,7 @@ func (sma *Actor) UpdatePower(vmctx exec.VMContext, delta *big.Int) (uint8, erro
 }
 
 // GetTotalStorage returns the total amount of proven storage in the system.
-func (sma *Actor) GetTotalStorage(vmctx exec.VMContext) (*big.Int, uint8, error) {
+func (sma *Actor) GetTotalStorage(vmctx exec.VMContext) (*types.BytesAmount, uint8, error) {
 	if err := vmctx.Charge(actor.DefaultGasCost); err != nil {
 		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -233,12 +233,12 @@ func (sma *Actor) GetTotalStorage(vmctx exec.VMContext) (*big.Int, uint8, error)
 		return nil, errors.CodeError(err), err
 	}
 
-	count, ok := ret.(*big.Int)
+	amt, ok := ret.(*types.BytesAmount)
 	if !ok {
 		return nil, 1, fmt.Errorf("expected *big.Int to be returned, but got %T instead", ret)
 	}
 
-	return count, 0, nil
+	return amt, 0, nil
 }
 
 // GetSectorSize returns the sector size of the block chain
--- a/actor/builtin/storagemarket/storagemarket.go
+++ b/actor/builtin/storagemarket/storagemarket.go
@@ -59,6 +59,9 @@ type State struct {
 
 	// TODO: Determine correct unit of measure. Could be denominated in the
 	// smallest sector size supported by the network.
+	//
+	// See: https://github.com/filecoin-project/specs/issues/6
+	//
 	TotalCommittedStorage *types.BytesAmount
 
 	ProofsMode types.ProofsMode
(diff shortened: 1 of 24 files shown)


commands/miner_daemon_test.go
```diff
@@ -443,7 +443,7 @@ func TestMinerPower(t *testing.T) {
 	power := powerOutput.ReadStdoutTrimNewlines()
 
 	assert.NoError(t, err)
-	assert.Equal(t, "3 / 6", power)
+	assert.Equal(t, "3072 / 6144", power)
```
laser: N.B. 3 * 1024 == 3072

chain/power_table_view_test.go
```diff
@@ -25,40 +25,44 @@ func TestTotal(t *testing.T) {
...
 	require.NoError(t, err)
 
-	assert.Equal(t, power, actual)
+	expected := types.OneKiBSectorSize.Mul(types.NewBytesAmount(numCommittedSectors))
+
+	assert.Equal(t, expected.Uint64(), actual.Uint64())
 }
 
-func requireMinerWithPower(ctx context.Context, t *testing.T, power uint64) (bstore.Blockstore, address.Address, state.Tree) {
+func requireMinerWithPower(ctx context.Context, t *testing.T, numCommittedSectors uint64) (bstore.Blockstore, address.Address, state.Tree) {
```
anorth: This function name is now a bit misleading about the uint value.

chain/power_table_view_test.go
```diff
@@ -25,40 +25,44 @@ func TestTotal(t *testing.T) {
...
-	power := uint64(12)
-	bs, addr, st := requireMinerWithPower(ctx, t, power)
+	numCommittedSectors := uint64(12)
+	bs, addr, st := requireMinerWithPower(ctx, t, numCommittedSectors)
 
 	actual, err := (&consensus.MarketView{}).Miner(ctx, st, bs, addr)
 	require.NoError(t, err)
 
-	assert.Equal(t, power, actual)
+	expected := types.OneKiBSectorSize.Mul(types.NewBytesAmount(numCommittedSectors))
```
anorth: Wrapping `numCommittedSectors` in a `BytesAmount` is weird. It would be less weird to convert the arguments to uint64 and then multiply those.

gengen/util/gengen.go
```diff
@@ -40,10 +40,16 @@ type Miner struct {
...
+	//
+	// TODO: This struct needs a field which represents the size of sectors
+	// that this miner has committed. For now, sector size is configured by
+	// the StorageMarketActor's ProofsMode.
+	//
 	// TODO: this will get more complicated when we actually have to
-	// prove real files
-	Power uint64
+	// prove real files.
+	NumCommittedSectors uint64
```
anorth: uint64 seems excessive here.  Should we reduce to a uint32 @whyrusleeping ?

gengen/util/gengen.go
```diff
@@ -257,14 +263,22 @@ func setupMiners(st state.Tree, sm vm.StorageMap, keys []*types.KeyInfo, miners
...
+		sectorSize := types.OneKiBSectorSize
+		if proofsMode == types.LiveProofsMode {
+			sectorSize = types.TwoHundredFiftySixMiBSectorSize
+		}
+
 		minfos = append(minfos, RenderedMinerInfo{
 			Address: maddr,
 			Owner:   m.Owner,
-			Power:   m.Power,
+			Power:   sectorSize.Mul(types.NewBytesAmount(m.NumCommittedSectors)),
```
anorth: Weird but correct again.

laser:
@anorth 

> The 18exabyte number becomes a more immediate issue with this change. Should we change it now to a bigint? Does this require spec changes? @whyrusleeping

I am glad that you have pointed this issue out. I will address the issue in this PR.

The [spec](https://github.com/filecoin-project/specs/blob/master/data-structures.md#bytesamount) states that `BytesAmount` are `Integers`. The `BytesAmount` type in go-filecoin currently wraps `big.Int`. `big.Int` (and therefore `BytesAmount`) and is [serialized as LEB128](https://github.com/filecoin-project/specs/blob/master/data-structures.md#integer). We are conformant with the specification in that regard.

The 18.44 exabyte figure is not a limitation of the specification, but rather a bug in our implementation. The bug is that we convert between arbitrary-precision `big.Int` and `BytesAmount` [to `uint64`](https://github.com/filecoin-project/go-filecoin/blob/master/consensus/power_table_view.go#L42) in many places. It is [undefined behavior](https://golang.org/pkg/math/big/#Int.Uint64) if the value of the underlying `big.Int` cannot be represented as `uint64`.

chain/power_table_view_test.go
laser: Indeed. Fixed.

chain/power_table_view_test.go
laser: Sounds good. Done.

gengen/util/gengen.go
laser: Agreed. Updated as per your suggestion.

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -57,9 +57,9 @@ type Actor struct{}
 type State struct {
 	Miners cid.Cid `refmt:",omitempty"`
 
-	// TotalCommitedStorage is the number of sectors that are currently committed
-	// in the whole network.
-	TotalCommittedStorage *big.Int
+	// TODO: Determine correct unit of measure. Could be denominated in the
```
ZenGround0: Would you mind linking an existing github issue in this TODO, or making a new one and linking it if one does not already exist?

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -233,12 +233,12 @@ func (sma *Actor) GetTotalStorage(vmctx exec.VMContext) (*big.Int, uint8, error)
 		return nil, errors.CodeError(err), err
 	}
 
-	count, ok := ret.(*big.Int)
+	amt, ok := ret.(*types.BytesAmount)
```
ZenGround0: I'm rob pike and I approve this go-y name change

chain/default_syncer_test.go
```diff
@@ -909,20 +909,22 @@ func TestTipSetWeightDeep(t *testing.T) {
...
+				NumCommittedSectors: 10,
 			},
 			{
-				Power: uint64(980),
+				NumCommittedSectors: 980,
 			},
 		},
 	}
 
+	totalPower := types.NewBytesAmount(1000).Mul(types.OneKiBSectorSize)
```
ZenGround0: 🙏

actor/builtin/storagemarket/storagemarket.go
laser: Done!

consensus/power_table_view.go
```diff
@@ -35,45 +35,34 @@ type MarketView struct{}
...
-	res.SetBytes(rets[0])
 
-	return res.Uint64(), nil
+	return types.NewBytesAmountFromBytes(rets[0]), nil
 }
 
-// Miner returns the storage that this miner has committed as a uint64.
-// If the total storage value exceeds the max value of a uint64 this method
-// errors.
-// TODO: currently power is in sectors, figure out if & how it should be converted to bytes.
```
ZenGround0: The day has come

gengen/util/gengen.go
```diff
@@ -40,10 +40,16 @@ type Miner struct {
 	// PeerID is the peer ID to set as the miners owner
 	PeerID string
 
-	// Power is the amount of power this miner should start off with
+	// NumCommittedSectors is the number of sectors that this miner has
+	// committed to the network.
+	//
+	// TODO: This struct needs a field which represents the size of sectors
```
ZenGround0: 👍 Could we link an issue saying the same thing within this TODO?

testhelpers/consensus.go
```diff
@@ -95,8 +88,8 @@ func NewValidTestBlockFromTipSet(baseTipSet types.TipSet, stateRootCid cid.Cid,
 	}
 }
 
-// MakeRandomPoSTProofForTest creates a random proof.
-func MakeRandomPoSTProofForTest() []byte {
+// MakeRandomPoStProofForTest creates a random proof.
+func MakeRandomPoStProofForTest() []byte {
```
ZenGround0: Doing God's work

types/bytes_amount.go
```diff
@@ -116,6 +116,16 @@ func (z *BytesAmount) Mul(y *BytesAmount) *BytesAmount {
 	return &newZ
```
ZenGround0: nit & not your bug: The comments for `Mul` `Sub` and `Add` mistakenly reference variable `x` when they actually mean `z`.  Could you fix this as you've fixed the other comments below?

actor/builtin/storagemarket/storagemarket.go
laser: i literally loled

types/bytes_amount.go
laser: Yep indeed. Just fixed them all.

gengen/util/gengen.go
laser: Done!

PR #2770, opened 2019-05-10
Refactor/deal get respects errors 2762

### LINKED ISSUE #2762: DealGet should respect errors
### Description

The current implementation of `DealGet` in `porcelain/storagedeals.go` swallows errors instead of returning them. It's should return an error and all invocations of it should respect said error.

### Acceptance criteria

1. `DealGet` returns errors
2. All invocations of `DealGet` are updated to handle errors

### Where to begin

`porcelain/storagedeals.go`

rosalinekarr:
# Problem

The current implementation of `DealGet` does not return errors properly in the event of datastore failure.

# Solution

Change `DealGet` to return errors.

Resolves #2762

### DIFF
--- a/porcelain/storagedeals.go
+++ b/porcelain/storagedeals.go
@@ -2,8 +2,14 @@ package porcelain
 
 import (
 	"github.com/ipfs/go-cid"
+	errors "github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
+)
+
+var (
+	// ErrDealNotFound means DealGet failed to find a matching deal
+	ErrDealNotFound = errors.New("deal not found")
 )
 
 type strgdlsPlumbing interface {
@@ -11,15 +17,15 @@ type strgdlsPlumbing interface {
 }
 
 // DealGet returns a single deal matching a given cid or an error
-func DealGet(plumbing strgdlsPlumbing, dealCid cid.Cid) *storagedeal.Deal {
+func DealGet(plumbing strgdlsPlumbing, dealCid cid.Cid) (*storagedeal.Deal, error) {
 	deals, err := plumbing.DealsLs()
 	if err != nil {
-		return nil
+		return nil, err
 	}
 	for _, storageDeal := range deals {
 		if storageDeal.Response.ProposalCid == dealCid {
-			return storageDeal
+			return storageDeal, nil
 		}
 	}
-	return nil
+	return nil, errors.New("deal not found")
 }
--- a/porcelain/storagedeals.go
+++ b/porcelain/storagedeals.go
@@ -27,5 +27,5 @@ func DealGet(plumbing strgdlsPlumbing, dealCid cid.Cid) (*storagedeal.Deal, erro
 			return storageDeal, nil
 		}
 	}
-	return nil, errors.New("deal not found")
+	return nil, ErrDealNotFound
 }
--- a/protocol/storage/client.go
+++ b/protocol/storage/client.go
@@ -249,7 +249,7 @@ func (smc *Client) checkDealResponse(ctx context.Context, resp *storagedeal.Resp
 func (smc *Client) minerForProposal(c cid.Cid) (address.Address, error) {
 	storageDeal, err := smc.api.DealGet(c)
 	if err != nil {
-		return address.Undef, fmt.Errorf("no such proposal by cid: %s", c)
+		return address.Undef, errors.Wrapf(err, "no such proposal by cid: %s", c)
 	}
 	return storageDeal.Miner, nil
 }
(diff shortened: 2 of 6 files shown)


protocol/storage/client.go
```diff
@@ -225,10 +225,6 @@ func (smc *Client) recordResponse(resp *storagedeal.Response, miner address.Addr
 	if !proposalCid.Equals(resp.ProposalCid) {
 		return fmt.Errorf("cids not equal %s %s", proposalCid, resp.ProposalCid)
 	}
-	storageDeal := smc.api.DealGet(proposalCid)
-	if storageDeal != nil {
-		return fmt.Errorf("deal [%s] is already in progress", proposalCid.String())
-	}
```
anorth: This removal looks like an unintended semantic change: if there is *no* error from DealGet, this should raise one. OTOH, if the error is "not found", this needs to continue.

porcelain/storagedeals.go
```diff
@@ -2,24 +2,30 @@ package porcelain
...
+		return nil, err
 	}
 	for _, storageDeal := range deals {
 		if storageDeal.Response.ProposalCid == dealCid {
-			return storageDeal
+			return storageDeal, nil
 		}
 	}
-	return nil
+	return nil, errors.New("deal not found")
```
anorth: Use `ErrDealNotFound`, especially so you can check for it in storage/client.go.

protocol/storage/client.go
```diff
@@ -251,11 +247,10 @@ func (smc *Client) checkDealResponse(ctx context.Context, resp *storagedeal.Resp
 }
 
 func (smc *Client) minerForProposal(c cid.Cid) (address.Address, error) {
-	storageDeal := smc.api.DealGet(c)
-	if storageDeal == nil {
+	storageDeal, err := smc.api.DealGet(c)
+	if err != nil {
 		return address.Undef, fmt.Errorf("no such proposal by cid: %s", c)
```
anorth: Wrap the error

protocol/storage/client.go
```diff
@@ -301,8 +296,8 @@ func (smc *Client) isMaybeDupDeal(p *storagedeal.Proposal) bool {
 
 // LoadVouchersForDeal loads vouchers from disk for a given deal
 func (smc *Client) LoadVouchersForDeal(dealCid cid.Cid) ([]*types.PaymentVoucher, error) {
-	storageDeal := smc.api.DealGet(dealCid)
-	if storageDeal == nil {
+	storageDeal, err := smc.api.DealGet(dealCid)
+	if err != nil {
 		return []*types.PaymentVoucher{}, fmt.Errorf("could not retrieve deal with proposal CID %s", dealCid)
```
anorth: Wrap here too

protocol/storage/client_test.go
```diff
@@ -106,7 +106,8 @@ func TestProposeDeal(t *testing.T) {
 
 		assert.Equal(t, storagedeal.Accepted, dealResponse.State)
 
-		retrievedDeal := testAPI.DealGet(dealResponse.ProposalCid)
+		retrievedDeal, err := testAPI.DealGet(dealResponse.ProposalCid)
+		require.NoError(t, err)
 
 		assert.Equal(t, retrievedDeal.Response, dealResponse)
```
anorth: Please test the error return in the case of no deal found

protocol/storage/miner.go
```diff
@@ -382,12 +383,12 @@ func rejectProposal(sm *Miner, p *storagedeal.Proposal, reason string) (*storage
 }
 
 func (sm *Miner) updateDealResponse(proposalCid cid.Cid, f func(*storagedeal.Response)) error {
-	storageDeal := sm.porcelainAPI.DealGet(proposalCid)
-	if storageDeal == nil {
+	storageDeal, err := sm.porcelainAPI.DealGet(proposalCid)
+	if err != nil {
 		return fmt.Errorf("failed to get retrive deal with proposal CID %s", proposalCid.String())
```
anorth: Wrap all the errors here

protocol/storage/miner.go
```diff
@@ -401,8 +402,8 @@ func (sm *Miner) processStorageDeal(proposalCid cid.Cid) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	d := sm.porcelainAPI.DealGet(proposalCid)
-	if d == nil {
+	d, err := sm.porcelainAPI.DealGet(proposalCid)
+	if err != nil {
 		log.Errorf("could not retrieve deal with proposal CID %s", proposalCid.String())
```
ZenGround0: We should log the actual error too.

protocol/storage/client_test.go
rosalinekarr: I'm not sure I understand your suggestion. Are you saying we should test ErrDealNotFound here, because I don't see how we could get that error when proposing a new deal?

protocol/storage/client.go
```diff
@@ -251,11 +254,10 @@ func (smc *Client) checkDealResponse(ctx context.Context, resp *storagedeal.Resp
 }
 
 func (smc *Client) minerForProposal(c cid.Cid) (address.Address, error) {
-	storageDeal := smc.api.DealGet(c)
-	if storageDeal == nil {
-		return address.Undef, fmt.Errorf("no such proposal by cid: %s", c)
+	storageDeal, err := smc.api.DealGet(c)
+	if err != nil {
+		return address.Undef, errors.Wrapf(err, "no such proposal by cid: %s", c)
```
anorth: This message could now be misleading, depending on the error.  I suggest tweaking the wording to "failed to fetch deal".

protocol/storage/client_test.go
anorth: Not right at this line, but I was wanting test coverage of the tricky error logic that was incorrect in the first version of this PR. With closer inspection, I see the case I named is the expected one, already exercised. Is it possible to add to this test to exercise the case when the error *is* nil, because deal already in progress?

PR #2771, opened 2019-05-10
Fix migration logging

ZenGround0:
- logfile closed when main exits
    - correct flags during file open to enable logging
    - correct usage of global log in runner

### DIFF
--- a/tools/migration/internal/runner_test.go
+++ b/tools/migration/internal/runner_test.go
@@ -66,11 +66,12 @@ func TestMigrationRunner_Run(t *testing.T) {
 		require.NoError(t, err)
 		runner.MigrationsProvider = testProviderPasses
 
-		out := CaptureOutput(func() {
-			assert.NoError(t, runner.Run().Err)
-		})
-		assert.Contains(t, out, "skipping migration from 0 to 1")
-		assert.Contains(t, out, "did not find valid repo migration for version 199")
+		assert.NoError(t, runner.Run().Err)
+
+		out, err := ioutil.ReadFile(dummyLogPath)
+		require.NoError(t, err)
+		assert.Contains(t, string(out), "skipping migration from 0 to 1")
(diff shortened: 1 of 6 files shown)


tools/migration/internal/runner_test.go
```diff
@@ -66,11 +66,12 @@ func TestMigrationRunner_Run(t *testing.T) {
 		require.NoError(t, err)
 		runner.MigrationsProvider = testProviderPasses
 
-		out := CaptureOutput(func() {
-			assert.NoError(t, runner.Run().Err)
-		})
-		assert.Contains(t, out, "skipping migration from 0 to 1")
-		assert.Contains(t, out, "did not find valid repo migration for version 199")
+		assert.NoError(t, runner.Run().Err)
```
shannonwells: Is CaptureOutput dead then?  Can we delete it?

tools/migration/main.go
```diff
@@ -124,6 +125,15 @@ func exitErr(errstr string) {
 	os.Exit(1)
 }
 
+// exitErrorCloseLogger closes the logger and calls exitError
+func exitErrCloseLogger(errstr string, logger *internal.Logger) {
+	err := logger.Close()
+	if err != nil {
+		errstr = fmt.Sprintf("%s. Error closing logfile when reporting this error %s", errstr, err.Error())
```
shannonwells: It's a little awkward, but it's unlikely to ever see it.  I can't think of a better solution.

tools/migration/main.go
```diff
@@ -124,6 +125,15 @@ func exitErr(errstr string) {
 	os.Exit(1)
 }
 
+// exitErrorCloseLogger closes the logger and calls exitError
+func exitErrCloseLogger(errstr string, logger *internal.Logger) {
```
rosalinekarr: can we use `log.Fatal` here instead? it might read better.

PR #2775, opened 2019-05-10
expose StdErr() from iptb.Output with a dump function

### LINKED ISSUE #2611: FAST provides an easy way to assert on stderr output of actions
### Description

When testing, a common thing to do is assert that a command exited, and assert that the correct error was received. Currently this is a bit clunky with FAST and we should provide a better way to perform this kind of work.

shannonwells:
Closes #2611 , providing an easy way to check the StdErr of the Output from running a given FAST command.

### DIFF
--- a/commands/client_daemon_test.go
+++ b/commands/client_daemon_test.go
@@ -270,10 +270,5 @@ func TestSelfDialStorageGoodError(t *testing.T) {
 	f := files.NewBytesFile([]byte("satyamevajayate"))
 	_, _, err = series.ImportAndStore(ctx, miningNode, ask, f)
 	assert.Error(t, err)
-	var cmdOutBytes []byte
-	w := bytes.NewBuffer(cmdOutBytes)
-	miningNode.DumpLastOutput(w)
-	outputStr := string(w.Bytes())
-	expectedErrStr := "attempting to make storage deal with self"
-	assert.Contains(t, outputStr, expectedErrStr)
+	fastesting.AssertStdErrContains(t, miningNode, "attempting to make storage deal with self")
 }
--- a/commands/retrieval_client_daemon_test.go
+++ b/commands/retrieval_client_daemon_test.go
@@ -1,7 +1,6 @@
 package commands_test
 
 import (
-	"bytes"
 	"context"
 	"math/big"
 	"testing"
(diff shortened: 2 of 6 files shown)


commands/retrieval_client_daemon_test.go
```diff
@@ -52,7 +52,7 @@ func TestSelfDialRetrievalGoodError(t *testing.T) {
 	assert.Error(t, err)
 	var cmdOutBytes []byte
 	w := bytes.NewBuffer(cmdOutBytes)
-	env.GenesisMiner.DumpLastOutput(w)
+	env.GenesisMiner.DumpLastStdErr(w)
```
travisperson: I was thinking instead of using a dump like this we would either just return the Stderr reader directly or even just return the bytes / string. Not 100% which is better. I think what we want is a way to get _just_ the stderr without any extra bits. `DumpLastOutput` is more for debugging than it is for making assertions.

We could also clean this up with the `AssertStdErrContains` (eg: `AssertStdErrContains(t, env.GenesisMiner, expectedErrStr)`)

shannonwells: >We could also clean this up with the AssertStdErrContains (eg: AssertStdErrContains(t, env.GenesisMiner, expectedErrStr))

Great minds think alike :)

commands/client_daemon_test.go
```diff
@@ -270,10 +270,5 @@ func TestSelfDialStorageGoodError(t *testing.T) {
 	f := files.NewBytesFile([]byte("satyamevajayate"))
 	_, _, err = series.ImportAndStore(ctx, miningNode, ask, f)
 	assert.Error(t, err)
-	var cmdOutBytes []byte
-	w := bytes.NewBuffer(cmdOutBytes)
-	miningNode.DumpLastOutput(w)
-	outputStr := string(w.Bytes())
-	expectedErrStr := "attempting to make storage deal with self"
-	assert.Contains(t, outputStr, expectedErrStr)
+	fastesting.AssertStdErrContains(t, miningNode, "attempting to make storage deal with self")
```
ZenGround0: 🎉

tools/fast/environment_memory_genesis_test.go
```diff
@@ -54,6 +54,7 @@ func TestEnvironmentMemoryGenesis(t *testing.T) {
 
 		// did we teardown correctly?
 		assert.NoError(t, env.Teardown(ctx))
+		assert.Equal(t, 0, len(env.Processes()))
```
ZenGround0: 👍

PR #2790, opened 2019-05-17
Storage client use the Has method from datastore

### LINKED ISSUE #2763: Storage client should use the Has method from go-datastore
### Description

Currently, `isMaybeDupDeal` in `protocol/storage/client.go` calls the `DealsLs` plumbing method and iterates over results to check for the existence of a deal. This is inefficient and doesn't take full advantage of the API exposed by `go-datastore`.

We should add a `DealsHas` plumbing method to `plumbing/strgdls/store.go` which uses `go-datastore`'s underlying [`Has` method ](https://godoc.org/github.com/ipfs/go-datastore#MapDatastore.Has) and can be used in the storage client in place of `isMaybeDupDeal`.

### Acceptance criteria

1. A `DealsHas` method is added to `plumbing/strgdls` for exposing `go-datastore`'s `Has` method
2. `DealHas` is tested
3. `protocol/storage/client.go` is changed to use the new `DealsHas` method
4. `isMaybeDupDeal` is removed from `protocol/storage/client.go`

### Where to begin

`plumbing/strgdls/store.go`

kishansagathiya:
Fixes #2763


anorth:
Thanks for your contribution, and sorry for the delay. It looks like you have a test failure to fix up.

@rosalinekarr would you shepherd this PR?

plumbing/strgdls/store.go
```diff
@@ -58,3 +44,45 @@ func (store *Store) Put(storageDeal *storagedeal.Deal) error {
...
+	for _, d := range deals {
+		if d.Miner == storageProposal.MinerAddress {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (store *Store) lsWithPrefix(prefix string) ([]*storagedeal.Deal, error) {
```
rosalinekarr: oof... this is going to conflict with my work in #2749

we don't need to worry about it in this PR, but we're probably going to have a nasty rebase to do later

plumbing/api.go
```diff
@@ -7,12 +7,12 @@ import (
...
 	"github.com/ipfs/go-cid"
-	"github.com/ipfs/go-ipfs-exchange-interface"
+	exchange "github.com/ipfs/go-ipfs-exchange-interface"
 	ipld "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log"
 	uio "github.com/ipfs/go-unixfs/io"
-	"github.com/libp2p/go-libp2p-metrics"
-	"github.com/libp2p/go-libp2p-peer"
+	metrics "github.com/libp2p/go-libp2p-metrics"
+	peer "github.com/libp2p/go-libp2p-peer"
```
rosalinekarr: These two changes should be unnecessary. Golang uses the last part of the path following the hyphen by default.

protocol/storage/client.go
```diff
@@ -8,9 +8,9 @@ import (
 
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
-	"github.com/libp2p/go-libp2p-host"
-	"github.com/libp2p/go-libp2p-peer"
-	"github.com/libp2p/go-libp2p-protocol"
+	host "github.com/libp2p/go-libp2p-host"
+	peer "github.com/libp2p/go-libp2p-peer"
+	protocol "github.com/libp2p/go-libp2p-protocol"
```
rosalinekarr: These are also unnecessary.

plumbing/strgdls/store.go
```diff
@@ -58,3 +44,45 @@ func (store *Store) Put(storageDeal *storagedeal.Deal) error {
 
 	return nil
 }
+
+// Has checks if a deal is already present in the datastore corresponding to
+// given proposal
+// bool should be ignored in case of a non-nil error
+func (store *Store) Has(storageProposal *storagedeal.Proposal) (bool, error) {
```
rosalinekarr: Let's add a test for this method.

plumbing/api.go
kishansagathiya: My vs code setup does this on it's own. It will come back again on saving, will have to check.

kishansagathiya: I am not sure how this exactly works, but what I saw right now by trying out is that 

`"github.com/libp2p/xyz"` would be replace by the package name, if different. say `abc "github.com/libp2p/xyz"`

if the package name is xyz it would just stay `"github.com/libp2p/xyz"`

plumbing/strgdls/store.go
kishansagathiya: added in the latest commit

plumbing/api.go
rosalinekarr: The package names of `go-libp2p-metrics` and `go-libp2p-peer` are already `metrics` and `peer`, so that shouldn't be a problem.

Either way, this isn't a blocking problem, so feel free to just leave it in.

rosalinekarr:
@anorth, this branch still needs to be rebased, but otherwise, it should be ready for review

anorth:
Since this is old, please ping me after the rebase. @kishansagathiya so sorry about the delay.

kishansagathiya:
On it. Will push soon after rebasing.

kishansagathiya:
Yet to fix `TestDuplicateDeals`

frrist:
Thanks for working on this @kishansagathiya. Going to close for now, but feel free to reopen once the above feedback is addressed.

PR #2791, opened 2019-05-17
Refactor deals awaiting seals into its own file.

acruikshank:
This PR is a refactor that ballooned out of #2693 into enough work for a new PR.

### Problem

dealsAwaitingSeals is burried within protocol/storage/miner. I found its logic confusing and under-tested. 

### Solution

This PR streamlines the logic, moves it to its own file with a separate file for tests, and adds testing for its features.

### DIFF
--- a/protocol/storage/deals_awaiting_seal.go
+++ b/protocol/storage/deals_awaiting_seal.go
@@ -15,7 +15,7 @@ func init() {
 // sectorInfo combines sector Metadata from rust proofs with go-filecoin specific data
 type sectorInfo struct {
 	Metadata         *sectorbuilder.SealedSectorMetadata
-	CommitMessageCid *cid.Cid
+	CommitMessageCid cid.Cid
 	Succeeded        bool
 	ErrorMessage     string
 }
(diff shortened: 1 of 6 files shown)


protocol/storage/deals_awaiting_seal_test.go
```diff
@@ -0,0 +1,213 @@
...
+		dealsAwaitingSeal.fail(wantSectorID, wantMessage)
+		dealsAwaitingSeal.fail(someOtherSectorID, "some message")
+		dealsAwaitingSeal.process(wantSectorID, cid0)
+		dealsAwaitingSeal.process(wantSectorID, cid1) // Shouldn't trigger a call, see process().
+
+		assert.Len(t, gotCids, 1, "onFail should've been called once")
+	})
+}
+
+func TestDealsAwaitingSealSuccess(t *testing.T) {
```
frrist: missing `UnitTest` flag, others below are as well

protocol/storage/deals_awaiting_seal.go
```diff
@@ -0,0 +1,130 @@
+package storage
+
+import (
+	"sync"
+
+	"github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
```
frrist: nit: should be its own line

PR #2793, opened 2019-05-17
Feat/change nonce too high log 2748

### LINKED ISSUE #2748: Nonce too high should be a warning, not an error
### Description

This log message should be a warning, not an error:

```
16:07:51.315 ERROR       node: processMessage(): validation error adding message to pool: message nonce (60271) is too much greater than actor nonce (0) node.go:1040
```

This log can occur under normal operation with members on a devnet running out of date code. Error logs should be reserved for events that require the attention of the miner operator.

rosalinekarr:
# Problem

We log expected block validation failures as errors despite the fact they can occur frequently during normal daemon operation.

# Solution

Change them to info logs

Resolves #2748


laser:
Pinging @travisperson for review

PR #2795, opened 2019-05-20
Factor message packaging out of plumbing into core.Outbox

anorth:
This is part of follow-up refactoring to the message queue that @acruikshank and I discussed a while ago. Sorry it's a little long: the delta is mostly moving and simplifying the old Sender test to exercise the outbox instead.

I intend to also:
- move the blocktimer out of message pool (this change means one of the `Add` callers now knows height)
- move the policy for handling new tipsets into Outbox


core/outbox.go
```diff
@@ -0,0 +1,147 @@
...
+		validator: validator,
+		queue:     queue,
+		publisher: publisher,
+		chains:    chains,
+		actors:    actors,
+	}
+}
+
+// Send marshals and sends a message, retaining it in the outbound message queue.
+func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value *types.AttoFIL,
```
anorth: This is a fairly direct move from `msg.Sender`, but with tighter dependencies and the message pool/publish extracted to `publisher`.

core/outbox_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/core"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOutbox(t *testing.T) {
```
anorth: These are ported from the `msg.Sender` tests, but with fakes instead of real dependencies (chain store, state HAMT etc).

I dropped a few internal tests for `nextNonce` and added some nonce checks to the existing tests, so this test is now in a `_test` package.

node/node.go
```diff
@@ -119,7 +119,7 @@ type Node struct {
 	// Incoming messages for block mining.
 	MsgPool *core.MessagePool
 	// Messages sent and not yet mined.
-	Outbox *core.MessageQueue
+	MsgQueue *core.MessageQueue
```
anorth: I'll replace this with the actual outbox object soon.

node/node.go
```diff
@@ -410,24 +411,27 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
...
 	}
 	fcWallet := wallet.New(backend)
 
+	// only the syncer gets the storage which is online connected
+	chainSyncer := chain.NewDefaultSyncer(&cstOffline, nodeConsensus, chainStore, fetcher)
+	msgPool := core.NewMessagePool(chainStore, nc.Repo.Config().Mpool, consensus.NewIngestionValidator(chainFacade, nc.Repo.Config().Mpool))
+	msgQueue := core.NewMessageQueue()
+
+	msgPublisher := newDefaultPublisher(pubsub.NewPublisher(fsub), core.Topic, msgPool)
+	actorProvider := newDefaultActorProvider(chainStore, &cstOffline)
```
anorth: The publisher and actor provider are the "wiring" between the outbox and other components.

core/outbox.go
```diff
@@ -0,0 +1,147 @@
...
+	gasPrice types.AttoFIL, gasLimit types.GasUnits, method string, params ...interface{}) (out cid.Cid, err error) {
+	defer func() {
+		if err != nil {
+			msgSendErrCt.Inc(ctx, 1)
```
rosalinekarr: If this is only executed in the event of no errors, why not just put it on line 122 and avoid the added complexity of the defer statement at all?

node/node.go
rosalinekarr: The `ActorProvider` looks a lot like this existing `chainFacade` plumbing. Maybe we should combine the `chainProvider` and `actorProvider` interfaces on `Outbox` and replace them both with `chainFacade`?

On a somewhat related note, I'm thinking we should also extract `chainFacade` from plumbing later on too.

core/outbox.go
```diff
@@ -0,0 +1,147 @@
...
+	}()
+
+	encodedParams, err := abi.ToEncodedValues(params...)
+	if err != nil {
+		return cid.Undef, errors.Wrap(err, "invalid params")
+	}
+
+	// Lock to avoid race for message nonce.
+	ob.nonceLock.Lock()
+	defer ob.nonceLock.Unlock()
```
rosalinekarr: Not your change, but I feel like we should lock inside the `nextNonce` method instead of here. That will simplify this code and reduce the amount of time we need to lock nonces for.

node/actor_provider.go
```diff
@@ -0,0 +1,46 @@
...
+	tree, err := state.LoadStateTree(ctx, ap.state, stateCid, builtin.Actors)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load latest state")
+	}
+
+	actr, err := tree.GetActor(ctx, addr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "no actor at address %s", addr)
+	}
+	return actr, nil
```
rosalinekarr: With regard to using the `chainFacade` instead of `chainProvider` and `actorProvider`, this method is identical to `plumbing/bcf/blockchain_facade.go` line 89.

node/msg_publisher.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	if _, err := p.pool.Add(ctx, message); err != nil {
+		return errors.Wrap(err, "failed to add message to message pool")
+	}
+
+	if err = p.network.Publish(p.topic, encoded); err != nil {
+		return errors.Wrap(err, "failed to publish message to network")
+	}
+	return nil
```
rosalinekarr: Looks great, but I'm not totally clear on why this is under `node`. Why not put it with the outbox? It feels like the `node` directory has been sort of devolving into a "misc" folder lately.

node/node.go
rosalinekarr: Also, btw `chainFacade` was a terrible choice of name on my part, and I would totally be up for calling it something like `actorChainDecorator` or `chainStateProvider` instead.

plumbing/msg/sender.go
```diff
@@ -2,160 +2,25 @@ package msg
...
-}
-
-// nextNonce returns the next expected nonce value for an account actor. This is the larger
-// of the actor's nonce value, or one greater than the largest nonce from the actor found in the message pool.
-func nextNonce(act *actor.Actor, outbox *core.MessageQueue, address address.Address) (uint64, error) {
-	actorNonce, err := actor.NextNonce(act)
-	if err != nil {
-		return 0, err
-	}
+	return s.outbox.Send(ctx, from, to, value, gasPrice, gasLimit, method, params...)
```
rosalinekarr: Let's just ditch this file completely and put the outbox on the plumbing API struct directly like the wallet.

node/node.go
rosalinekarr: Also, btw `chainFacade` was a terrible choice of name on my part, and I would totally be up for calling it something like `actorChainDecorator` or `chainStateProvider` instead.

anorth: I've replaced the actorProvider with a new method on the chain facade (which differs from the existing one in being parameterised by tipset key). I also renamed it to `ChainStateProvider`.

I agree that it should move moved from plumbing. Some of the methods are just compositions of lower-level bits, so it should probably go in porcelain (with a few primitives remaining behind).

I'd rather not combine the chain and actor interfaces from the outbox, since they are provided by different objects at the moment.

core/outbox.go
anorth: This is executed only when there *is* an error. It can't go at the end due to the unstructured early return statements throughout. (I went through the same thought process).

core/outbox.go
anorth: `nextNonce` inspects the message queue, and later in this method a message is added to that queue; the lock needs to at least span that distance. Since the actor state is also part of the calculation I think it's prudent that the lock cover it too (though there are still probably cases of a re-org where nonces become invalid).

I don't think there's much to be gained, so will leave this alone until profiles indicate contention here.

node/msg_publisher.go
anorth: This is a package-level dependency injection. The outbox doesn't know or care how messages are published to the network. This change removed the import of `net/pubsub` by the `core` package.

It's "wiring" which definitely feels miscellaneous but is what the node package alone should be doing.

plumbing/msg/sender.go
anorth: Done 🎉

core/outbox.go
rosalinekarr: ah, makes sense

core/outbox_test.go
```diff
@@ -0,0 +1,199 @@
...
+			nonces[uint64(message.Msg.Nonce)] = true
+		}
+
+		for i := 0; i < 60; i++ {
+			assert.True(t, nonces[uint64(actr.Nonce)+uint64(i)])
+
+		}
+	})
+
+	t.Run("nonce fails with non-account actor", func(t *testing.T) {
```
acruikshank: Should this read "send fails with non-account actor"? If it fails in nextNonce, it's opaque to the test.

PR #2798, opened 2019-05-20
Separate new tipset handling from message pool implementation, wrap in an inbox.

anorth:
With this change, the incoming and outbound message queue/pool follow a similar pattern.
They're not quite the same, since validation is performed inside the message pool but
prior to the outbound queue (which we could also align).

### DIFF
--- a/core/inbox.go
+++ b/core/inbox.go
@@ -1,0 +1,128 @@
+package core
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+)
+
+// InboxMaxAgeTipsets is maximum age (in non-empty tipsets) to permit messages to stay in the pool after reception.
+// It should be a little shorter than the outbox max age so that messages expire from mining
+// pools a little before the sender gives up on them.
+const InboxMaxAgeTipsets = 6
+
+// Inbox maintains a pool of received messages.
+type Inbox struct {
+	// The pool storing received messages.
+	pool *MessagePool
+	// Maximum age of a pool message.
+	maxAgeTipsets uint
+
+	chain InboxChainProvider
+}
+
+// InboxChainProvider provides chain access for updating the message pool in response to new heads.
+// Exported for testing.
+type InboxChainProvider interface {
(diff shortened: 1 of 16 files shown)


core/message_pool.go
```diff
@@ -250,10 +139,24 @@ func (pool *MessagePool) LargestNonce(address address.Address) (largest uint64,
...
+	pool.lk.RLock()
+	defer pool.lk.RUnlock()
+
+	var cids []cid.Cid
+	for c, msg := range pool.pending {
+		if msg.addedAt < minimumHeight {
+			cids = append(cids, c)
+		}
+	}
+	return cids
```
rosalinekarr: Not something we need to address in this PR, but we could get a big performance improvement here by sorting pending upon insertion and exiting early here. It might be a good idea to add an icebox task for optimizing this.

core/inbox.go
```diff
@@ -0,0 +1,128 @@
...
+func (ib *Inbox) Pool() *MessagePool {
+	return ib.pool
+}
+
+// HandleNewHead updates the message pool in response to a new head tipset.
+// This removes messages from the pool that are found in the newly adopted chain and adds back
+// those from the removed chain (if any) that do not appear in the new chain.
+// We think that the right model for keeping the message pool up to date is
+// to think about it like a garbage collector.
+func (ib *Inbox) HandleNewHead(ctx context.Context, oldHead, newHead types.TipSet) error {
```
anorth: This is copied verbatim from message_pool.go `UpdateMessagePool`

core/inbox_test.go
```diff
@@ -0,0 +1,472 @@
...
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateMessagePool(t *testing.T) {
```
anorth: These tests are copied from message_pool_test, with minor dependency changes.

anorth:
Ping @acruikshank

PR #2799, opened 2019-05-20
Feat/caught up nodes ignore far future blocks 2782

### LINKED ISSUE #2782: Chain sync rejects new heads farther than [finality] ahead of consensus head
### Description
As reported in #2755, the 0.2.x user devnets experienced a long mining delay when peers and blocks from the previous devnet leaked into the DHT/pubsub of the new one. This simulated an attack by a node proposing a long and heavy alternate chain to the consensus one. Filecoin nodes spent a long time fetching this alternate chain, only to eventually reject it when it didn't meet with the new network's genesis.

Unlike POW blockchains, filecoin nodes can't easily validate blocks without linking them back to their existing chain. This attack is mitigated by a caught-up (#1160) node limiting the distance ahead of its current head that it will accept a new block. A block that doesn't converge with the chain within [finality] blocks should be dropped, marked as bad, and not propagated (see #2783, although it may not be feasible to prevent pubsub validation).

https://github.com/filecoin-project/consensus/issues/67 gives us a working number of 600 rounds (~5 hours) for this finality.

Note: #2755 also identifies that nodes are blocked from mining while following the fake chain due to synchronous/blocking code fetching the chain. Changing that is not in scope for this story; adding asynchrony here will only make things harder to trace and understand at the moment.

### Acceptance criteria
- Nodes which are caught up and following a consensus chain ignore blocks claiming to be more than 600 rounds past their current head
- The [sync spec](https://github.com/filecoin-project/specs/blob/master/sync.md) clearly identifies this behaviour

### Risks + pitfalls
- This resolution depends on differentiating a caught-up mode #1160

rosalinekarr:
# Problem

Currently, a vulnerability exists in chain syncing when receiving a blocks for a chain extremely far ahead of the current consensus after entering `CaughtUp` mode which causes nodes to significantly slow down as they attempt to verify the new blocks.

# Solution

Once in `CaughtUp` mode, disregard any chains received more than 600 blocks ahead of the current consensus.

Resolves #2782

### DIFF
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -59,6 +59,7 @@ const (
 )
 
 type syncerChainReader interface {
+	BlockHeight() (uint64, error)
 	GetBlock(context.Context, cid.Cid) (*types.Block, error)
 	GetHead() types.SortedCidSet
 	GetTipSet(tsKey types.SortedCidSet) (*types.TipSet, error)
@@ -168,7 +169,7 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 	// Continue collecting the chain if we're either not yet caught up or the
 	// number of new input blocks is less than the FinalityLimit constant.
 	// Otherwise, halt assuming the new blocks come from an invalid chain.
-	for (syncer.syncMode == Syncing) || (len(chain) < FinalityLimit) {
+	for (syncer.syncMode == Syncing) || !syncer.exceedsFinalityLimit(chain) {
 		var blks []*types.Block
 		// check the cache for bad tipsets before doing anything
 		tsKey := tipsetCids.String()
@@ -451,4 +452,14 @@ func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids typ
 		parent = ts
 	}
 	return nil
+}
+
+func (syncer *DefaultSyncer) exceedsFinalityLimit(chain []types.TipSet) bool {
+	if len(chain) == 0 {
+		return false
+	}
+	blockHeight, _ := syncer.chainStore.BlockHeight()
+	finalityHeight := types.NewBlockHeight(blockHeight).Add(types.NewBlockHeight(uint64(FinalityLimit)))
+	chainHeight, _ := chain[0].Height()
+	return types.NewBlockHeight(chainHeight).LessThan(finalityHeight)
 }
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -37,7 +37,9 @@ var (
 
 var logSyncer = logging.Logger("chain.syncer")
 
-// SyncMode is the
+// SyncMode represents which behavior mode the chain syncer is currently in. By
+// default, the node starts in "Syncing" mode, and once it syncs the most
+// recently generated block, it switches to "Caught Up" mode.
 type SyncMode int
 
 const (
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -111,13 +111,13 @@ type DefaultSyncer struct {
 	// implementation in issue #1160.
 	//
 	// TODO: https://github.com/filecoin-project/go-filecoin/issues/1160
-	SyncMode SyncMode
+	syncMode SyncMode
 }
 
 var _ Syncer = (*DefaultSyncer)(nil)
 
 // NewDefaultSyncer constructs a DefaultSyncer ready for use.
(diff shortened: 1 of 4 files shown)


rosalinekarr:
Putting this work on ice at @anorth's [suggestion](https://github.com/filecoin-project/go-filecoin/issues/2782#issuecomment-493834198)

chain/default_syncer_test.go
```diff
@@ -560,6 +613,20 @@ func TestHeavierFork(t *testing.T) {
 	assertHead(t, chainStore, forklink3)
 }
 
+// Syncer errors if input blocks massively exceed the current block height in
+// caught up mode
```
anorth: Please test the behaviour when in syncing mode too, as a contrast.

chain/default_syncer.go
```diff
@@ -33,6 +37,25 @@ var (
...
+	// See the spec for more detail:
+	// https://github.com/filecoin-project/specs/blob/master/sync.md#syncing-mode
+	Syncing SyncMode = iota
+	// CaughtUp indicates that the node has caught up with consensus head and as a
+	// result can restrict which new blocks are accepted to mitigate consensus
+	// attacks.
+	//
+	// See the spec for more detail:
+	// https://github.com/filecoin-project/specs/blob/master/sync.md#caught-up-mode
+	CaughtUp
```
frrist: When do we enter this mode? I only see logic for `Syncing` mode. Is this to be done in a follow on or has it not been addressed because this is on ice?

anorth: Never, except tests, but that needn't stop this code from landing. However, it cannot close the underlying issue #2782

anorth:
@rosalinekarr please ensure this doesn't close #2782 when you land it, until the mode actually exists.

chain/default_syncer.go
```diff
@@ -133,7 +163,11 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 	fetchedHead := tipsetCids
 	defer logSyncer.Infof("chain fetch from network complete %v", fetchedHead)
 
-	for {
+	// Continue collecting the chain if we're either not yet caught up or the
+	// height of the input blocks has not yet exceeded the sum of the current
```
ZenGround0: Blocking: This comment gives a misleading picture to the reader about what is going on here.  It references the case when the new "too long" chain is built on top of the head of the heaviest chain.  In general the finality check applies to new chains that fork from far back in the past too.  This comment is also confusing because it references the height of the input blocks when the finality limit is based on number of blocks in the code.

chain/default_syncer.go
```diff
@@ -133,7 +163,11 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
...
+	// consensus height and the finalityLimit constant, otherwise ignore the input
+	// blocks as a likely invalid chain or denial of service attempt.
```
ZenGround0: We can just say "as an invalid chain" because our finality rule invalidates it.

chain/default_syncer.go
```diff
@@ -133,7 +163,11 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
...
+	for syncer.SyncMode == Syncing || len(chain) < FinalityLimit {
```
ZenGround0: Please note that this is number of blocks and not block height people.  I think this is the right thing to do right now, but this will trip us up if we don't internalize it.

chain/default_syncer_test.go
```diff
@@ -178,6 +178,40 @@ func requireSetTestChain(t *testing.T, con consensus.Protocol, mockStateRoots bo
 	}
 }
 
+// requireChainOfLength generates a fake chain of given length for testing
+// relating to chain length
+func requireChainOfLength(t *testing.T, length int, con consensus.Protocol) types.SortedCidSet {
```
ZenGround0: I would strongly prefer some way to avoid redundancy with [`requireGrowChain`](https://github.com/filecoin-project/go-filecoin/blob/master/chain/get_ancestors_test.go#L24).  I understand why you can't use that one here but a lot of the logic in both could be abstracted to something common.  Would you mind filing a follow-on / tech debt issue about building out a user-friendly chain growing framework for tests?

chain/default_syncer.go
anorth: Blocking: well spotted. The issues for this work identify number of rounds, not blocks. Please use block height arithmetic instead of `len()`.

PR #2801, opened 2019-05-20
Refactor/remove chain store interface 2800

### LINKED ISSUE #2800: Remove Java-style Chain Interfaces
### Description

As a follow up to #2552, we want to remove the `chain/store.go` file as it disagrees with our current methodology with regard to using interfaces in Golang. Early in the project, we mistakenly built several Java-style interfaces with strict, over-specified surfaces. Since then, we have adopted the Go-style of smaller interfaces specified at their invocation sites and implemented more fluidly.

rosalinekarr:
# Problem

`chain/store.go` is inconsistent with our current methodologies regarding interfaces on the project

# Solution

Kill it with fire

Resolves #2800


anorth:
Thanks! IMO there might be room for one or two narrow but shared interfaces, just for ease of re-use. For example, the `GetHead` and `GetTipset` methods often come together. A small set like those plus `HasBlock`, `GetBlock` could go together? No change required to this PR tho.

chain/default_syncer_test.go
```diff
@@ -317,11 +323,16 @@ func assertNoAdd(t *testing.T, chainStore chain.Store, cids types.SortedCidSet)
 	}
 }
 
-func requireHead(t *testing.T, chain chain.Store, head types.TipSet) {
+type headGetter interface {
```
ZenGround0: My intuition is that this should be used in the store test instead of redefining the same interface with the `requireHeadTipsetChainStore` name.

ZenGround0: It seems like redefining the same interface twice within the same package is too much, but please educated me otherwise. go-style interfaces are relatively new to me.

node/node.go
```diff
@@ -90,6 +90,8 @@ type nodeChainReader interface {
 	GetTipSetStateRoot(tsKey types.SortedCidSet) (cid.Cid, error)
 	HeadEvents() *ps.PubSub
 	Load(context.Context) error
+	PutTipSetAndState(context.Context, *chain.TipSetAndState) error
```
ZenGround0: Blocking: we still want read write separation.  Node functions should absolutely not write to the chain store and the interface should reflect that.  Only the syncer should have this capability in production code.

It should be no problem to keep casting chainForTest to a read-write interface, or doing other function decomposition and automatic casting tricks.

chain/default_syncer_test.go
rosalinekarr: No, you're right. I've combined the two into a single `HeadAndTipsetGetter` interface

node/node.go
rosalinekarr: So, the issue is that we pass `node.ChainReader` to `RequirePutTsas` in `TestUpdateMessagePool` which needs `PutTipSetAndState`. It's super ugly, and I'd rather remove it, but I can't find any way to rewrite the test without using `node.ChainReader` because of all the complexity in `GenNode`.

acruikshank: I'm in favor of removing `TestUpdateMessagePool`. It's way too complicated for a test that's essentially testing whether the node creates a message pool and updates it on new tip set. All the actual update logic is covered elsewhere. 

If you go that route, these methods should go away, `node.HeaviestTipSetHandled` should also go away.

Also, we need a test to properly test that the node subscribes things like the message pool to events like `NewHeaviestTipset`. This will probably require some refactoring, so it can be a new PR or just an issue.

chain/default_store_test.go
```diff
@@ -34,14 +34,14 @@ func initStoreTest(ctx context.Context, t *testing.T, dstP *DefaultSyncerTestPar
 	requireSetTestChain(t, con, true, dstP)
 }
 
-func newChainStore(dstP *DefaultSyncerTestParams) chain.Store {
+func newChainStore(dstP *DefaultSyncerTestParams) *chain.DefaultStore {
```
acruikshank: This should be renamed to `chain.Store`. It was named `DefaultStore` as an alternative to `StoreImpl`, which is the (bad) Java convention.

node/node.go
rosalinekarr: sounds good to me

node/node.go
```diff
@@ -91,6 +91,7 @@ type nodeChainReader interface {
 	GetTipSetStateRoot(tsKey types.SortedCidSet) (cid.Cid, error)
 	HeadEvents() *ps.PubSub
 	Load(context.Context) error
+	SetHead(context.Context, types.TipSet) error
```
acruikshank: We still need to figure out a way to get rid of this write method. Is this still necessary?

testhelpers/chain.go
```diff
@@ -171,9 +171,13 @@ func MustNewTipSet(blks ...*types.Block) types.TipSet {
...
 }
 
+type requirePutTsasChainStore interface {
+	PutTipSetAndState(context.Context, *chain.TipSetAndState) error
+}
+
 // RequirePutTsas ensures that the provided tipset and state is placed in the
 // input store.
-func RequirePutTsas(ctx context.Context, t *testing.T, chn chain.Store, tsas *chain.TipSetAndState) {
+func RequirePutTsas(ctx context.Context, t *testing.T, chn requirePutTsasChainStore, tsas *chain.TipSetAndState) {
```
acruikshank: Is this really necessary? It can be inlined as a one-liner, which would eliminate the need for the interface.

node/node.go
ZenGround0: Blocking: same comment as before, no node caller should ever be updating chain store state from node.  (Sorry if it wasn't clear that this applied to SetHead as well as PutTipSetAndState)

testhelpers/chain.go
rosalinekarr: I wasn't aware interface definitions could be inlined. How do you do that?

acruikshank: I mean inline `RequirePutTsas`. The name isn't great in the first place and it doesn't save us much.

Issue #2792, opened 2019-05-17
Compute CommP from piece data


acruikshank:
### Description

A client must be able to generate a data commitment from their piece data so that they can use it as the basis of piece inclusion conditions in their payments to the miner. 

The clients piece data commitment is the merkle root of the piece when it is bit aligned as it would be when staged into a sector. It's not reasonable to replicate the rust sector builder logic that bit pads the data in go. The ability to generate a data commitment should be exposed from rust fil proofs and used here.

### Acceptance criteria

1. A method that accepts a data reader and produces a data commitment is exposed to go-filecoin.
2. That method is integrated into the storage client so that the client can use the commitment to generate payments.
3. Tests still pass.
4. When we have PIPs, they work against the new CommP.

acruikshank:
Completed in #3066

PR #2802, opened 2019-05-21
client adds conditions to payments

### LINKED ISSUE #2705: Client adds `verifyPieceInclusion` conditions to payments
### Description

In order for a client to be able to prevent miners from collecting payment when they haven't been storing the client's file, the client needs to add piece inclusion conditions to payments. The the ability to specify and add a condition has been added in #2656 and #2573.

There are two parts to this work.

1. The client needs to compute the data commitment (aka CommP) for the piece they want to store.
2. When creating payments, the client should provide a piece inclusion condition for every payment except the last.

CommP can be fake (suggest using the data cid for now) until rust proofs comes up with a method for generating it.

The condition should be addressed to the miner actor of the storage miner. Its method should be 'verifyPieceInclusion' and it should include CommP as its one parameter.

The last payment should not have a condition. This allows the miner to collect payment without positing a PIP on chain so long as they convince the client that they are storing their piece (i.e. the client doesn't try to cancel payment), and they are willing to wait until the payment is complete to collect.

### Acceptance criteria

1. The first N-1 generated payment vouchers contain conditions constructed with the correct details, and the last one has no condition.
2. Proper construction of the payment conditions is fully tested.

acruikshank:
closes #2705

### Problem

Storage clients need to ensure they can recover their money once they discover that a miner is not storing their piece. They can do that by adding conditions to payments that ensure the miner is storing the piece at the time of payment.

#### Solution

Have the client add a 'verifyPieceInclusion' condition to each payment except the last. The last payment should not have a condition so the miner can recover funds without posting a piece inclusion proof at the end of the deal.

### Note

The `verifyPieceInclusion` method on miner requires a `CommP` as it's first parameter. We currently do not have the ability to generate `CommP` so we create a fake one in this PR (by using the CID of the piece) and leaving CommP integration as future work (#2792). 

The PR tests that the conditions are created correctly. It doesn't test that miners will be able to redeem payments if they have valid piece inclusion proofs. This round-trip testing should be performed soon after this PR (#2822).

### DIFF
--- a/porcelain/payments.go
+++ b/porcelain/payments.go
@@ -14,6 +14,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+const verifyPieceInclusionMethod = "verifyPieceInclusion"
+
 // cpPlumbing is the subset of the plumbing.API that CreatePayments uses.
 type cpPlumbing interface {
 	ChainBlockHeight() (*types.BlockHeight, error)
@@ -39,6 +41,13 @@ type CreatePaymentsParams struct {
 
 	// Duration is the amount of time (in block height) the payments will cover.
 	Duration uint64
+
+	// MinerAddress is the address of the miner actor representing the payment recipient.
+	// Conditions confirming that the miner is storing the client's piece will be directed towards this actor.
+	MinerAddress address.Address
+
+	// CommP is the clients data commitment. It will be the basis of piece inclusion conditions added to the payments.
+	CommP types.CommP
 
 	// PaymentInterval is the time between payments (in block height)
 	PaymentInterval uint64
@@ -72,7 +81,12 @@ type CreatePaymentsReturn struct {
 	Vouchers []*types.PaymentVoucher
 }
 
-// CreatePayments establishes a payment channel and create multiple payments against it
+// CreatePayments establishes a payment channel and create multiple payments against it.
+//
+// Each payment except the last will get a condition that calls verifyPieceInclusion on the recipient's miner
+// actor to ensure the storage miner is still storing the file at the time of redemption.
+// The last payment does not contain a condition so that the miner may collect payment without posting a
+// piece inclusion proof after the storage deal is complete.
 func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePaymentsParams) (*CreatePaymentsReturn, error) {
 	// validate
 	if config.From.Empty() {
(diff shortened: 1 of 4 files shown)


porcelain/payments.go
```diff
@@ -145,18 +166,17 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
...
 		}
 	}
 
-	if voucherAmount.LessThan(&config.Value) {
-		validAt := currentHeight.Add(types.NewBlockHeight(config.Duration))
-		err = createPayment(ctx, plumbing, response, &config.Value, validAt, nil)
-		if err != nil {
-			return response, err
-		}
+	// create last payment if we haven't already created a voucher for the full amount
```
anorth: The code no longer has an "if..."

porcelain/payments.go
```diff
@@ -40,6 +42,13 @@ type CreatePaymentsParams struct {
 	// Duration is the amount of time (in block height) the payments will cover.
 	Duration uint64
 
+	// MinerAddress is the address of the miner actor representing the payment recipient.
+	// Conditions confirming that the miner is storing the client's piece will be directed towards this actor.
+	MinerAddress address.Address
+
+	// CommP is the clients data commitment. It will be the basis of piece inclusion conditions added to the payments.
```
ZenGround0: nit picking at its extreme: client's

porcelain/payments.go
```diff
@@ -72,7 +81,12 @@ type CreatePaymentsReturn struct {
 	Vouchers []*types.PaymentVoucher
 }
 
-// CreatePayments establishes a payment channel and create multiple payments against it
+// CreatePayments establishes a payment channel and create multiple payments against it.
```
ZenGround0: more nitpicking: creates

porcelain/payments_test.go
```diff
@@ -140,17 +146,66 @@ func TestCreatePayments(t *testing.T) {
...
+			// assert all vouchers other than the last have a valid condition
+			assert.NotNil(t, voucher.Condition)
+		}
+
+		// last payment should be for the full amount and have no condition
+		assert.Equal(t, config.Value, paymentResponse.Vouchers[9].Amount)
+		assert.Nil(t, paymentResponse.Vouchers[9].Condition)
+	})
+
+	t.Run("Payments constructed correctly when duration < paymentInterval", func(t *testing.T) {
```
ZenGround0: Nice tests

protocol/storage/client.go
```diff
@@ -110,6 +110,10 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		return nil, errors.Wrap(err, "failed to determine the size of the data")
 	}
 
+	// TODO This is fake. CommP should be the merkle root of data, rather than its CID (issue #2792)
```
ZenGround0: thanks for linking issue 🥇

acruikshank:
I'm going to merge it so I can test it in nightly. Then well automate the tests in #2822 comes through.

PR #2819, opened 2019-05-22
pre-release tool to clean up old nightlies

ognots:
allows setting a limit of prereleases aka nightlies to keep.
requires a github api token to function.
tool is parameterized via flags otherwise

### DIFF
--- a/tools/prerelease-tool/main.go
+++ b/tools/prerelease-tool/main.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+type prereleaseTool struct {
+	Data        []*github.RepositoryRelease
+	Limit       int
+	Owner       string
+	Repo        string
+	client      *github.Client
+	Prereleases []github.RepositoryRelease
+	DryRun      bool
+}
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	token, ok := os.LookupEnv("GITHUB_TOKEN")
+	if !ok {
+		log.Fatal("Github token must be provided through GITHUB_TOKEN environment variable")
+	}
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	dryRun := flag.Bool("dry-run", false, "perform a dry run instead of executing create/delete/update actions")
+	limit := flag.Int("limit", 7, "limit of prereleases to keep")
+	owner := flag.String("owner", "filecoin-project", "github owner or organization")
+	repo := flag.String("repo", "go-filecoin", "github project repository")
+	flag.Parse()
+	r := prereleaseTool{
+		Limit:  *limit,
+		Owner:  *owner,
+		Repo:   *repo,
+		client: client,
+		DryRun: *dryRun,
+	}
+	if err := r.get(ctx); err != nil {
+		log.Fatalf("Could not find any releases: %+v", err)
+	}
+	r.getPrereleases()
+	r.sortReleasesByDate()
+	ok, err := r.trim(ctx, r.oldReleaseIDs())
+	if err != nil {
+		log.Fatalf("Problem attempting to delete releases: %+v", err)
+	}
+	if !ok {
+		log.Print("Remove --dry-run flag to apply changes")
+	}
+}
+
+func (r *prereleaseTool) get(ctx context.Context) error {
+	releases, _, err := r.client.Repositories.ListReleases(ctx, r.Owner, r.Repo, nil)
+	r.Data = releases
+	return err
+}
+
+func (r *prereleaseTool) getPrereleases() {
+	switch {
+	case len(r.Data) > 0:
+		for _, release := range r.Data {
+			if *release.Prerelease {
+				r.Prereleases = append(r.Prereleases, *release)
+			}
+		}
+	default:
+		log.Print("no releases found")
+	}
+
+}
+
+func (r *prereleaseTool) sortReleasesByDate() {
+	sort.Slice(r.Prereleases, func(i, j int) bool {
+		first := r.Prereleases[i].CreatedAt
+		second := r.Prereleases[j].CreatedAt
+		switch {
+		case first.Unix() < second.Unix():
+			return false
+		default:
+			return true
+		}
+	})
+}
+
+func (r *prereleaseTool) oldReleaseIDs() []int64 {
+	var idsToDelete []int64
+	switch {
+	case len(r.Prereleases) > r.Limit:
+		trimmedPrereleaseList := r.Prereleases[r.Limit:]
+		for _, release := range trimmedPrereleaseList {
+			idsToDelete = append(idsToDelete, *release.ID)
+		}
+	default:
+		log.Print("There are no outdated prereleases")
+	}
+	return idsToDelete
+}
+
+func (r *prereleaseTool) trim(ctx context.Context, ids []int64) (bool, error) {
+	var ok bool
+	switch {
+	case len(ids) > 0:
+		log.Print("Removing outdated Prereleases")
+		for _, id := range ids {
+			log.Printf("Prerelease ID %d selected for deletion", id)
+			if !r.DryRun {
+				ok = true
+				log.Printf("Deleting Prerelease %d", id)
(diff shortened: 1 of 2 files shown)


ognots:
Demo video https://drive.google.com/a/filecoin.org/file/d/1irO88e57cFgoGR-vz8QzW9nbwoLG56sv/view?usp=sharing

ognots:
Note: this still leaves the tags behind - only deletes release. Any opinions on whether we should be keeping or deleting tags?

pooja:
Great demo @ognots! I would prefer for us to delete tags. Is that an option?

pooja:
Awesome thanks!

frrist:
Looks good @ognots, couple questions:
- When will this tool run? Seems like we might be able to run it as part of nightly deployment?
- Does this need to live in the go-filecoin repo? Do we think the infra repo would be more appropriate place?

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+	Limit       int
+	Owner       string
+	Repo        string
+	client      *github.Client
+	Prereleases []github.RepositoryRelease
+	DryRun      bool
+}
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
frrist: Do we ever see a scenario when this will take more than 30 seconds? If so the tool should complain loudly that it failed to run.

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+	token, ok := os.LookupEnv("GITHUB_TOKEN")
+	if !ok {
+		log.Fatal("Github token must be provided through GITHUB_TOKEN environment variable")
+	}
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	dryRun := flag.Bool("dry-run", false, "perform a dry run instead of executing create/delete/update actions")
```
frrist: nit: can we move all of the flags and parsing to the top of main? Personal preference is to see all the flags a thing takes first.

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+				if err != nil {
+					return ok, err
+				}
+				if resp.StatusCode != 204 {
+					return ok, fmt.Errorf("Unexpected HTTP status code. Expected: 204 Got: %d", resp.StatusCode)
+				}
+			}
+		}
+	default:
+		log.Print("No preleases to delete")
```
frrist: typo: prereleases

ognots:
@frrist it would be run as part of nightly release. just wanting to vet it's working as expected and test if manually a couple times before adding to circleci config.

as far as it's home, I'm open to suggestions. it's not a bad fit in `go-filecoin` repo (especially if running as part of nightly build), but it could definitely be in it's own repo. the reason I didn't opt to put it in `go-filecoin-infra` is because there's currently no other Golang based tools there. I'm at least temporarily taking advantage of the existing build tooling in `go-filecoin`

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+	ok, err := r.trim(ctx, r.oldReleaseIDs())
+	if err != nil {
+		log.Fatalf("Problem attempting to delete releases: %+v", err)
+	}
+	if !ok {
+		log.Print("Remove --dry-run flag to apply changes")
+	}
+}
+
+func (r *prereleaseTool) get(ctx context.Context) error {
```
frrist: This gets all releases, correct? This means things outside of just the nightly's? If so can we call it `getReleases()`

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"sort"
+	"time"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+type prereleaseTool struct {
+	Data        []*github.RepositoryRelease
```
frrist: Can we call this `Releases`?

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+		Limit:  *limit,
+		Owner:  *owner,
+		Repo:   *repo,
+		client: client,
+		DryRun: *dryRun,
+	}
+	if err := r.get(ctx); err != nil {
+		log.Fatalf("Could not find any releases: %+v", err)
+	}
+	r.getPrereleases()
```
frrist: If this method doesn't find any releases do we want to continue?

tools/prerelease-tool/main.go
ognots: yeah, it might actually, I meant to set that higher, to 5 minutes. there's a good amount of releases to delete now. 
I'll set a receiver on the context cancel channel to make sure I log that. great suggestion!

tools/prerelease-tool/main.go
ognots: yes, I had already done that believe it or not during final cleanup, but obviously lost that change when I rebased from master :)

tools/prerelease-tool/main.go
ognots: it's safe to continue, but a waste of compute you're right. altering method to have a return value to check.

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+		Owner:  *owner,
+		Repo:   *repo,
+		client: client,
+		DryRun: *dryRun,
+	}
+	if err := r.getReleases(ctx); err != nil {
+		log.Fatalf("Could not find any releases: %+v", err)
+	}
+	r.getPrereleases()
+	r.sortReleasesByDate()
```
anorth: This pattern of an object having multiple state-mutating methods that must be called in order, and only one order that makes any sense, is a bit odd. It would be easier to use and understand with a single `DoIt()` method that can then call some functions/methods that take parameters and return values, rather than mutating shared state.

It's not a big deal if we don't expect the scope of this code to grow, I won't block on it.

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+func (r *prereleaseTool) sortReleasesByDate() {
+	sort.Slice(r.Prereleases, func(i, j int) bool {
+		first := r.Prereleases[i].CreatedAt
+		second := r.Prereleases[j].CreatedAt
+		switch {
+		case first.Unix() < second.Unix():
+			return false
+		default:
+			return true
+		}
```
anorth: `return r.Prereleases[i].CreatedAt.Before(r.Prereleases[j].CreatedAt)`

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+	})
+}
+
+func (r *prereleaseTool) oldReleaseIDs() []int64 {
+	var idsToDelete []int64
+	switch {
+	case len(r.Prereleases) > r.Limit:
```
anorth: Just use `if` where there are only two cases. Same at 114.

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, release := range trimmedPrereleaseList {
+			idsToDelete = append(idsToDelete, *release.ID)
+		}
+	default:
+		log.Print("There are no outdated prereleases")
+	}
+	return idsToDelete
+}
+
+func (r *prereleaseTool) trim(ctx context.Context, ids []int64) (bool, error) {
```
anorth: This is a very benign name for what it does. I suggest "deleteReleases".

tools/prerelease-tool/main.go
```diff
@@ -0,0 +1,134 @@
...
+				if err != nil {
+					return ok, err
+				}
+				if resp.StatusCode != 204 {
+					return ok, fmt.Errorf("Unexpected HTTP status code. Expected: 204 Got: %d", resp.StatusCode)
+				}
+			}
+		}
+	default:
+		log.Print("No prereleases to delete")
```
anorth: Instead of the logs at 106 and here, just log "Deleted %d releases" at the end of this function. That covers all cases.

tools/prerelease-tool/main.go
ognots: I did some method renaming and incorporated the sort method in to the getPreleases method. Right now the purpose of this is very tightly scoped but think it should be easy enough to adapt this if the scope grows.

ognots:
All change requests have been made. 
Updated demo: https://drive.google.com/a/filecoin.org/file/d/124qlBzqlW5D-pzn6v-NA_Zxp7LwPJFYO/view?usp=sharing

anorth:
FYI this code required changes to `go.mod` and `go.sum` which were not landed with this PR.

ognots:
@anorth thanks for resolving that. sorry for the miss

PR #2833, opened 2019-05-24
efficient reorg logging + log length

ZenGround0:


ZenGround0:
As part of this PR I'm removing the tipset pointer when it shows up.  The reason is that we don't get anything but slightly extra work for ourselves because [the golang map value is already a pointer](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it)

ZenGround0:
This PR
1) Introduces logging the amount of chain height reverted and added to the chain during a reorg
2) Makes reorg logging more efficient by only traversing back to common ancestor (instead of gathering all ancestors)
3) Removes tipset pointers from interfaces as explained above

chain/default_syncer.go
```diff
@@ -268,31 +268,44 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
 
 	return nil
 }
 
+func (syncer *DefaultSyncer) logReorg(ctx context.Context, curHead, newHead types.TipSet) {
+	curHeadIter := IterAncestors(ctx, syncer.chainStore, curHead)
+	newHeadIter := IterAncestors(ctx, syncer.chainStore, newHead)
+	commonAncestor, err := FindCommonAncestor(curHeadIter, newHeadIter)
+	if err != nil {
+		logSyncer.Debugf("unexpected error when running FindCommonAncestor for reorg log: %s", err.Error())
```
anorth: Debug is too quiet for a message that is very unexpected. I suggest warning (unless this is an expected path, in which case document why).

chain/default_syncer.go
```diff
@@ -268,31 +268,44 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
+		return
+	}
+
+	// Refresh iterators modified above
```
anorth: This comment doesn't make sense here

chain/reorg_test.go
```diff
@@ -1,71 +1,141 @@
...
+	requirePutBlocks(t, blockSource, block2)
+	superset := th.RequireNewTipSet(t, headBlock, block2)
+	supersetTsas := &chain.TipSetAndState{
+		TipSet:          superset,
+		TipSetStateRoot: dstP.genStateRoot,
+	}
+	th.RequirePutTsas(ctx, t, chainStore, supersetTsas)
+
+	return oldHead, superset, commonAncestor
+}
```
anorth: Move these helper functions to the bottom of the file so that the logic proceeds in a top-down manner.

chain/reorg_test.go
```diff
@@ -1,71 +1,141 @@
...
+
+	firstForkBlock := th.RequireMkFakeChild(t, fakeChildParams)
+	requirePutBlocks(t, blockSource, firstForkBlock)
+	firstForkTS := th.RequireNewTipSet(t, firstForkBlock)
+	firstForkTsas := &chain.TipSetAndState{
+		TipSet:          firstForkTS,
+		TipSetStateRoot: dstP.genStateRoot,
+	}
+	th.RequirePutTsas(ctx, t, chainStore, firstForkTsas)
+	err := chainStore.SetHead(ctx, firstForkTS)
```
frrist: Does setting the head here control which chain `requireGrowChain` grows?

chain/reorg_test.go
```diff
@@ -1,71 +1,141 @@
...
+	ctx, blockSource, chainStore := setupGetAncestorTests(t, dstP)
+	old, new, common := getSubsetOldNewCommon(ctx, t, chainStore, blockSource, dstP)
+
+	dropped, added, err := chain.ReorgDiff(old, new, common)
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(1), dropped)
+	assert.Equal(t, uint64(1), added)
+}
+
+// getForkOldNewCommon
```
frrist: Can we elaborate on this method more in the commnet? 

It looks like it grows the OG chain by 10 blocks, forks and grows the fork by 10, then grows the OG chain by 14 blocks. These numbers (10 & 14) seem like they might make good parameters to the method, what do you think? Also, is there anything significant about the numbers 10 and 14?

chain/reorg_test.go
```diff
@@ -1,71 +1,141 @@
...
+	// go back and complete the original chain
+	err = chainStore.SetHead(ctx, commonAncestor)
+	require.NoError(t, err)
+	requireGrowChain(ctx, t, blockSource, chainStore, 14, dstP)
+	newHead := requireHeadTipset(t, chainStore)
+
+	return oldHead, newHead, commonAncestor
+}
+
+func getPrefixOldNewCommon(ctx context.Context, t *testing.T, chainStore chain.Store, blockSource *th.TestFetcher, dstP *DefaultSyncerTestParams) (types.TipSet, types.TipSet, types.TipSet) {
```
frrist: Ditto the above comment about method description and numbers as parameters here.

chain/reorg_test.go
```diff
@@ -1,71 +1,141 @@
...
+	oldHead := requireHeadTipset(t, chainStore)
+
+	// Add 2 tipsets to the head
+	requireGrowChain(ctx, t, blockSource, chainStore, 2, dstP)
+	newHead := requireHeadTipset(t, chainStore)
+
+	return oldHead, newHead, oldHead
+}
+
+func getSubsetOldNewCommon(ctx context.Context, t *testing.T, chainStore chain.Store, blockSource *th.TestFetcher, dstP *DefaultSyncerTestParams) (types.TipSet, types.TipSet, types.TipSet) {
```
frrist: It looks like `getForkOldNewCommon`, `getPrefixOldNewCommon`, and `getSubsetOldNewCommon` all accept the same parameters. Consider creating a structure for this and defining the methods on it.

frrist: Also please comment the intent of this method.

chain/reorg_test.go
ZenGround0: Yes
General note: the testing framework that has grown organically in this package is in need of some intentional redesign.

Issue #2310, opened 2019-03-18
Extract tipset canonical block ordering and use in all tipset traversals


anorth:
### Description
The canonical ordering of blocks in a tipset is apparently by ticket bytes. Many places in the code iterate blocks in a tipset either in map range order, or by iterating keys of a  `SortedCidSet`.

Blocks could be considered to lose their identify for many purposes when combined in a tipset: the tipset is more primary. Some block-inspection functionality (e.g. iterating messages) should work with a tipset instead. Tipsets should only permit iteration of blocks in canonical (ticket) order.

See discussions [here](https://github.com/filecoin-project/go-filecoin/pull/2269#discussion_r265564100) and [here](https://github.com/filecoin-project/go-filecoin/pull/2272)

### Acceptance criteria
- [x] TipSet hides its internals and exposes blocks only in canonical order
- [x] Remove use of SortedCidSet for stable ordering of blocks in a tipset
- [ ] Iterate messages in canonical order in callers of `chains.CollectBlocksToCommonAncestor`
- [x] Spec is updated to match: https://github.com/filecoin-project/specs/issues/285
- [x] [bonus] Tipset is immutable

### Risks + pitfalls
- There are probably more than a few places that iterate blocks and/or sorted CIDs.

### Where to begin
- chain/traversal.go (iterators) and core/chains.go (traversal/collection)

anorth:
There could be a time-bomb buried here. The outbound message queue now checks that messages from an address are discovered in nonce order, and will error out of block handling if not. So messages from multiple blocks at the same height that happened to be mined in the right order from the vm's point of view also need to be discovered in the right order by the queue.

So we shouldn't let this one lie: either @ZenGround0 or myself should pick it up soon.

(If you're here because the time bomb went off, the quick fix is probably to remove the nonce sequence assertions in the outbound message queue).

shannonwells:
Looks like some of this may be addressed with Issue #1847, PR #2422

anorth:
@ZenGround0 I'm stealing this, I trust you haven't started on it.

anorth:
`runMessages` in `consensus/expected.go` iterates in CID order, with a TODO to process in ticket order. It calls `ProcessTipSet` in `processor.go`, which does a similar iteration (but the comments lie about it).

So iteration is in fact consistently in CID order at present. We should still fix it.

PR #2837, opened 2019-05-28
TipSet orders blocks by ticket, is immutable value type

anorth:
Block ticket is the canonical traversal order when a tipset is processed. This changes the critical traversal in consensus/expected.go to match that (previously iterated in sorted CID order). This is most of the resolution to #2310. See also https://github.com/filecoin-project/specs/issues/285

The representation is more memory efficient and can be iterated without allocating an intermediate slice. An empty/undefined tipset is now harder to construct and explicitly undefined; a few methods now always return nil error and I intend to remove the error return value from them in a follow-up.

SortedCidSet remains as the "key" type for a tipset. In a follow-up, I intend to also upgrade this to an immutable structure.

### DIFF
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -144,7 +144,7 @@ func (store *Store) Load(ctx context.Context) (err error) {
 		genesii = iterator.Value()
 	}
 	// Check genesis here.
-	if !genesii.IsSolo() {
+	if genesii.Len() != 1 {
 		return errors.Errorf("genesis tip set must be a single block, got %d blocks", genesii.Len())
 	}
 
--- a/chain/default_store.go
+++ b/chain/default_store.go
@@ -98,20 +98,20 @@ func (store *Store) Load(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
-	headTs := types.TipSet{}
+	var blocks []*types.Block
 	// traverse starting from head to begin loading the chain
-	var startHeight types.Uint64
 	for it := tipCids.Iter(); !it.Complete(); it.Next() {
 		blk, err := store.GetBlock(ctx, it.Value())
 		if err != nil {
 			return errors.Wrap(err, "failed to load block in head TipSet")
 		}
-		err = headTs.AddBlock(blk)
-		if err != nil {
-			return errors.Wrap(err, "failed to add validated block to TipSet")
-		}
-		startHeight = blk.Height
+		blocks = append(blocks, blk)
 	}
+	headTs, err := types.NewTipSet(blocks...)
+	if err != nil {
+		return errors.Wrap(err, "failed to add validated block to TipSet")
+	}
+	startHeight := headTs.At(0).Height
 	logStore.Infof("start loading chain at tipset: %s, height: %d", tipCids.String(), startHeight)
 	// esnures we only produce 10 log messages regardless of the chain height
 	logStatusEvery := startHeight / 10
@@ -144,11 +144,11 @@ func (store *Store) Load(ctx context.Context) (err error) {
 		genesii = iterator.Value()
 	}
 	// Check genesis here.
-	if len(genesii) != 1 {
-		return errors.Errorf("genesis tip set must be a single block, got %d blocks", len(genesii))
+	if !genesii.IsSolo() {
+		return errors.Errorf("genesis tip set must be a single block, got %d blocks", genesii.Len())
 	}
 
-	loadCid := genesii.ToSlice()[0].Cid()
+	loadCid := genesii.At(0).Cid()
 	if !loadCid.Equals(store.genesis) {
 		return errors.Errorf("expected genesis cid: %s, loaded genesis cid: %s", store.genesis, loadCid)
 	}
@@ -205,8 +205,8 @@ func (store *Store) putBlk(ctx context.Context, block *types.Block) error {
 // PutTipSetAndState persists the blocks of a tipset and the tipset index.
 func (store *Store) PutTipSetAndState(ctx context.Context, tsas *TipSetAndState) error {
 	// Persist blocks.
-	for _, blk := range tsas.TipSet {
-		if err := store.putBlk(ctx, blk); err != nil {
+	for i := 0; i < tsas.TipSet.Len(); i++ {
+		if err := store.putBlk(ctx, tsas.TipSet.At(i)); err != nil {
 			return err
 		}
 	}
@@ -312,7 +312,7 @@ func (store *Store) SetHead(ctx context.Context, ts types.TipSet) error {
 	logStore.Debugf("SetHead %s", ts.String())
 
 	// Add logging to debug sporadic test failure.
-	if len(ts) < 1 {
+	if !ts.Defined() {
 		logStore.Error("publishing empty tipset")
 		logStore.Error(debug.Stack())
 	}
@@ -374,7 +374,7 @@ func (store *Store) GetHead() types.SortedCidSet {
 	store.mu.RLock()
 	defer store.mu.RUnlock()
 
-	if store.head == nil {
+	if !store.head.Defined() {
 		return types.SortedCidSet{}
 	}
 
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -354,42 +354,51 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
 	// Lookup tipsets with the same parents from the store.
 	parentSet, err := ts.Parents()
 	if err != nil {
-		return nil, err
+		return types.NoTipSet, err
 	}
 	height, err := ts.Height()
 	if err != nil {
-		return nil, err
+		return types.NoTipSet, err
 	}
 	if !syncer.chainStore.HasTipSetAndStatesWithParentsAndHeight(parentSet.String(), height) {
-		return nil, nil
+		return types.NoTipSet, nil
 	}
 	candidates, err := syncer.chainStore.GetTipSetAndStatesByParentsAndHeight(parentSet.String(), height)
 	if err != nil {
-		return nil, err
+		return types.NoTipSet, err
 	}
 	if len(candidates) == 0 {
-		return nil, nil
+		return types.NoTipSet, nil
 	}
 
 	// Only take the tipset with the most blocks (this is EC specific logic)
-	max := candidates[0]
+	max := candidates[0].TipSet
 	for _, candidate := range candidates[0:] {
-		if len(candidate.TipSet) > len(max.TipSet) {
-			max = candidate
+		if candidate.TipSet.Len() > max.Len() {
+			max = candidate.TipSet
 		}
 	}
 
-	// Add blocks of the biggest tipset in the store to a copy of ts
-	wts := ts.Clone()
-	for _, blk := range max.TipSet {
-		if err = wts.AddBlock(blk); err != nil {
-			return nil, err
-		}
+	// Form a new tipset from the union of ts and the largest in the store, de-duped.
+	blocks := make(map[cid.Cid]*types.Block)
+	for i := 0; i < ts.Len(); i++ {
+		blocks[ts.At(i).Cid()] = ts.At(i)
+	}
+	for i := 0; i < max.Len(); i++ {
+		blocks[max.At(i).Cid()] = max.At(i)
+	}
+	var blockSlice []*types.Block
+	for _, b := range blocks {
+		blockSlice = append(blockSlice, b)
+	}
+	wts, err := types.NewTipSet(blockSlice...)
+	if err != nil {
+		return types.NoTipSet, err
 	}
 
 	// check that the tipset is distinct from the input and tipsets from the store.
-	if wts.String() == ts.String() || wts.String() == max.TipSet.String() {
-		return nil, nil
+	if wts.String() == ts.String() || wts.String() == max.String() {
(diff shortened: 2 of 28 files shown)


anorth:
Just FYI @whyrusleeping

types/tipset.go
```diff
@@ -2,155 +2,148 @@ package types
...
+	return ts.blocks[i]
 }
 
-// Equals returns true if the tipset contains the same blocks as another set.
-// Equality is not tested deeply.  If blocks of two tipsets are stored at
-// different memory addresses but have the same cids the tipsets will be equal.
-func (ts TipSet) Equals(ts2 TipSet) bool {
-	return ts.ToSortedCidSet().Equals(ts2.ToSortedCidSet())
+// IsSolo tests whether the tipset has a single block.
+func (ts TipSet) IsSolo() bool {
```
acruikshank: This is, at best, unnecessary. `ts.Len()==1` is exactly the same length as `ts.IsSolo()`, but the meaning of the former is unambiguous. It would take me a few guesses to get to IsSolo means contains one block.

types/tipset.go
```diff
@@ -2,155 +2,148 @@ package types
...
 }
 
-// Height returns the height of a tipset.
+// Height returns the height of a tipset, and nil error.
 func (ts TipSet) Height() (uint64, error) {
-	if len(ts) == 0 {
-		return uint64(0), ErrEmptyTipSet
-	}
-	return uint64(ts.ToSlice()[0].Height), nil
+	return uint64(ts.blocks[0].Height), nil
```
acruikshank: I like how making TipSet immutable simplifies the error handling here and elsewhere.

chain/default_syncer.go
```diff
@@ -302,42 +302,51 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
 	// Lookup tipsets with the same parents from the store.
 	parentSet, err := ts.Parents()
 	if err != nil {
-		return nil, err
+		return types.NoTipSet, err
```
acruikshank: We have a bit of a convention to use 'Undef' to for nil values of known types (`address.Undef`, `cid.Undef`). `types.UndefTipSet` would help distinguish this from a flag or enum. Not a big deal though.

chain/default_syncer.go
```diff
@@ -302,42 +302,51 @@ func (syncer *DefaultSyncer) widen(ctx context.Context, ts types.TipSet) (types.
...
+	for i := 0; i < ts.Len(); i++ {
+		blocks[ts.At(i).Cid()] = ts.At(i)
+	}
+	for i := 0; i < max.Len(); i++ {
+		blocks[max.At(i).Cid()] = max.At(i)
+	}
+	var blockSlice []*types.Block
+	for _, b := range blocks {
+		blockSlice = append(blockSlice, b)
+	}
```
acruikshank: I'd find this logic a little more straightforward if the first loop added to both the slice and the map, and the second loop added to the slice only if the cid wasn't in the map (and there was no 3rd loop). It's debatable.

Changing it would be more efficient, but that's not worth much here.

chain/traversal.go
```diff
@@ -16,21 +16,19 @@ type BlockProvider interface {
 // GetParentTipSet returns the parent tipset of a tipset.
 // The result is empty if the tipset has no parents (including if it is empty itself)
 func GetParentTipSet(ctx context.Context, store BlockProvider, ts types.TipSet) (types.TipSet, error) {
-	newTipSet := types.TipSet{}
 	parents, err := ts.Parents()
-	if err != nil {
-		return nil, err
+	if err != nil || parents.Len() == 0 {
```
acruikshank: breadcrumb: the second clause just short circuits the rest of the method. The results would be the same.

consensus/expected.go
```diff
@@ -247,19 +247,6 @@ func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ance
 		return nil, err
 	}
 
-	sl := ts.ToSlice()
```
acruikshank: breadcrumb: This appears to be removed because the TipSet constructor doesn't permit inconsistent blocks. Note there is a small inconsistency in that this compares duplicates blocks by block.String (cid + json) whereas NewTipSet only compares by cid. Comparing by CID seems better.

consensus/expected.go
```diff
@@ -360,9 +348,9 @@ func CreateChallengeSeed(parents types.TipSet, nullBlkCount uint64) (types.PoStC
 func (c *Expected) runMessages(ctx context.Context, st state.Tree, vms vm.StorageMap, ts types.TipSet, ancestors []types.TipSet) (state.Tree, error) {
 	var cpySt state.Tree
 
-	// TODO: order blocks in the tipset by ticket
```
acruikshank: 🎉

plumbing/msg/waiter.go
```diff
@@ -175,9 +177,8 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
 func (w *Waiter) receiptFromTipSet(ctx context.Context, msgCid cid.Cid, ts types.TipSet) (*types.MessageReceipt, error) {
 	// Receipts always match block if tipset has only 1 member.
 	var rcpt *types.MessageReceipt
-	blks := ts.ToSlice()
-	if len(ts) == 1 {
-		b := blks[0]
+	if ts.IsSolo() {
```
acruikshank: Perhaps now that we have a tipset struct, some future work would be cache receipts in tip sets. That would keep us from having to run a full state transition to fetch a single receipt when there's more than one block. This method is pretty ugly.

types/tipset_test.go
```diff
@@ -54,220 +49,138 @@ func block(t *testing.T, height int, parentCid cid.Cid, parentWeight uint64, msg
...
-	tf.UnitTest(t)
-
-	ts := RequireTestTipSet(t)
-	w, err := ts.ParentWeight()
-	assert.NoError(t, err)
-	assert.Equal(t, w, uint64(1337000))
-}
+	b1, b2, b3 := makeTestBlocks(t)
+
+	assert.False(t, NoTipSet.Defined())
```
acruikshank: This should have its own test case for documentation purposes.

types/tipset_test.go
```diff
@@ -54,220 +49,138 @@ func block(t *testing.T, height int, parentCid cid.Cid, parentWeight uint64, msg
...
+
+	t.Run("construction of singleton", func(t *testing.T) {
```
acruikshank: There's nothing inherently different about multiple block and single block tipsets (other than single block tipsets don't exercise some code). We have too many places where we treat these differently (only because we can optimize the 1 block case) already. This test should just test IsSolo (if you choose to keep it).

Len, ParentWeight, ToSortedCidSet, Parents, Height, and MinTicket should have separate test cases for documentation purposes and to be more explicit about what's under test.

types/tipset_test.go
```diff
@@ -54,220 +49,138 @@ func block(t *testing.T, height int, parentCid cid.Cid, parentWeight uint64, msg
...
+		assert.Equal(t, uint64(b1.Height), tsHeight)
+		tsTicket, _ := ts.MinTicket()
+		assert.Equal(t, b1.Ticket, tsTicket)
+
+		assert.Equal(t, "{ "+b1.Cid().String()+" }", ts.String())
+	})
 
-func TestTipSetToSortedCidSet(t *testing.T) {
-	tf.UnitTest(t)
+	t.Run("construction of multiblock", func(t *testing.T) {
```
acruikshank: See above about single block tipsset.

types/tipset_test.go
```diff
@@ -54,220 +49,138 @@ func block(t *testing.T, height int, parentCid cid.Cid, parentWeight uint64, msg
...
-func TestTipSetToSlice(t *testing.T) {
-	tf.UnitTest(t)
+		assert.NotEqual(t, slice, ts.ToSlice())
+		assert.Equal(t, []*Block{b1, b2, b3}, ts.ToSlice()) // tipset is immutable
+	})
 
-	ts := RequireTestTipSet(t)
-	b1, b2, b3 := RequireTestBlocks(t)
-	tips := []*Block{b1, b2, b3}
+	t.Run("empty", func(t *testing.T) {
```
acruikshank: This and the following test descriptions should contain `NewTipSet` to clarify exactly what's being tested.

types/tipset.go
```diff
@@ -2,155 +2,148 @@ package types
...
-	}
-	blks := ts.ToSlice()
-	min := blks[0].Ticket
-	for i := range blks[0:] {
-		if bytes.Compare(blks[i].Ticket, min) < 0 {
-			min = blks[i].Ticket
-		}
-	}
-	return min, nil
+	return ts.blocks[0].Ticket, nil
```
ZenGround0: Blocking: panics when called on undefined tipset

types/tipset.go
```diff
@@ -2,155 +2,148 @@ package types
...
 func (ts TipSet) Parents() (SortedCidSet, error) {
-	if len(ts) == 0 {
-		return SortedCidSet{}, ErrEmptyTipSet
-	}
-	return ts.ToSlice()[0].Parents, nil
+	return ts.blocks[0].Parents, nil
 }
 
-// ParentWeight returns the tipset's ParentWeight in fixed point form.
+// ParentWeight returns the tipset's ParentWeight in fixed point form, and nil error.
```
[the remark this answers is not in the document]
ZenGround0: Also if we do decide to go with this then there should be some comment explicitly saying that these errors are intended to go away soon.  As it is this code is the worst of both worlds: we are forced to write error handling code but that error handling code never gets used because this code panics instead of passing along the error

types/tipset.go
```diff
@@ -2,155 +2,148 @@ package types
...
-		return errors.Errorf("block parents %s don't match tipset parents %s", b.Parents.String(), p.String())
-	}
-	if uint64(b.ParentWeight) != weight {
-		return errors.Errorf("bBlock parent weight: %d doesn't match existing tipset parent weight: %d", uint64(b.ParentWeight), weight)
-	}
+	// Sort blocks by ticket.
+	sort.Slice(sorted, func(i, j int) bool {
+		cmp := bytes.Compare(sorted[i].Ticket, sorted[j].Ticket)
+		if cmp == 0 {
+			// Break ticket ties with the block CIDs, which are distinct.
```
ZenGround0: Lets make sure there is an issue somewhere (gfc or spec) tracking that this is recorded in the spec.

plumbing/msg/waiter.go
ZenGround0: This is probably the ugliest code in the entire project and it should be made better with or without caching receipts.  That said caching receipts is a great idea.  We should probably think about this alongside caching aggregate state cids / aggregate state trees, and limiting the in-memory tipset cache.

types/tipset.go
acruikshank: I am a little shocked at how often we do this check (this is a smell in itself). The problem is we seem to be doing it for different reasons in each case. So in that sense, this function obscures rather than adds semantic meaning to its code.

Categorizing usages:
* 2 are related to whether the tipset is a genesis block. This logic could be encapsulated within expected.
* 2 are optimizations that check to see if we can avoid reprocessing messages when the tipset contains only one block. I'd like to see this fix so all cases are optimized.
* The rest are used in tests, and are actually asserting the tipset has the right number of blocks, not that the tipset is in a special state (e.g. a state of Soloness)

anorth: Fair enough, reverted. I don't want to try to apply other optimizations or refactors here.

types/tipset.go
anorth: It's stated at the struct level. I've asked in a Go forum for advice.

chain/traversal.go
anorth: Not quite. `NewTipSet` returns an error if given no blocks, but this returns Undef and no error (for the genesis tipset). I added a comment.

chain/default_store.go
```diff
@@ -95,20 +95,20 @@ func (store *DefaultStore) Load(ctx context.Context) (err error) {
...
-			return errors.Wrap(err, "failed to add validated block to TipSet")
-		}
-		startHeight = blk.Height
+		blocks = append(blocks, blk)
+	}
+	headTs, err := types.NewTipSet(blocks...)
+	if err != nil {
+		return errors.Wrap(err, "failed to add validated block to TipSet")
 	}
+	startHeight := headTs.At(0).Height
```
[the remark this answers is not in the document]
anorth: Yeah but that returns an error that I know isn't, so nah.

chain/traversal.go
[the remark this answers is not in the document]
anorth: Parents returns a `SortedCidSet`, which I have avoided touching yet.

types/tipset.go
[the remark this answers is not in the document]
anorth: If this lands, I'll fast follow with removing the errors. Pending broader advice on the pattern.

types/tipset.go
anorth: I'll do this when fixing the spec generally to define this. https://github.com/filecoin-project/specs/pull/289

anorth:
I have pushed a commit with review changes, you may wish to look at just that one. Pending decision on suppressing the error from undefined tipset method calls.

anorth:
After more thought, I've decided to leave the error return value for empty tipset methods there. It's less work than removing it and not obviously better either way.

PR #2851, opened 2019-05-29
Add code cov config

travisperson:
This is the same config we removed a while back.


ognots:
Let me look how to resolve this error https://circleci.com/gh/filecoin-project/go-filecoin/18448

ognots:
@acruikshank @anorth @mishmosh can one of you please opt-in to using third party orbs in CircleCI Organization Security settings. Only an admin can do this.

Fixes https://circleci.com/gh/filecoin-project/go-filecoin/18448 error

mishmosh:
This should be done now.

anorth:
Thanks! Please link to the coverage report for this PR/commit when the hookup is working. Looks broken atm in https://codecov.io/gh/filecoin-project/go-filecoin/commits

I expect there will be some GitHub branch hook stuff to configure in order for results to show up in PRs automagically, so lmk when we're ready for that.

.codecov.yml
```diff
@@ -0,0 +1,32 @@
...
+        threshold: 2%
+        if_no_uploads: error
+        if_not_found: success
+        if_ci_failed: error
+
+    changes:
+      default:
+        enabled: no
+
+comment: false
```
ognots: I think we only need to set this to `true` to enable commenting - no special permissions are needed to allow commenting on PRs

travisperson: Ya, I'll update that. I can't even few the reports on codecov yet. I think the project is in a bit of a wonky state at the moment. We added codecov when the project was private, and then made it public. The hooks seem to still be in place though, but a few things don't work and give me errors or weird one about private scope.

I'll update this and see if the codecovs start to show up

Issue #2834, opened 2019-05-25
[bug] implement block delay validation (quick fix against the attack on 0.2.2)


steven004:
**Describe the bug**
Attack on version 0.2.2 devnet is happening, after some investigation, this attack is using the current consensus implementation, which is not fully functional. After discussed with @whyrusleeping and @ZenGround0. 
I would suggest we have a short-term solution before we fully implement VDF for consensus part,  since the current attack could be used to mislead investors. As @whyrusleeping suggested, this can be easily addressed, then, let's do it to make the devnet work.

See more discussion in [this thread](https://filecoinproject.slack.com/?redir=%2Fmessages%2FCEHHJNJS3%2Fconvo%2FCEHHJNJS3-1558602296.079500%2F). 

**To Reproduce**
This attack is most likely doing as: 
1) Shorten the block time to a few seconds or no wait;
2) Select messages to create a block, most times only pick the messages related to this miner, and discard all others;
3) When creating a block, to calculate the ticket to see if the miner could be the next winner, if yes, send it out, if not, just produce blank block, and repeat until the new ticket could make sure it will be the next winner.

**Expected behavior**
This kind of blocks could not pass in other honest miners validation process, so that the malicous node could get any benefit from this. 

**Version information**
go-filecoin-0.2.2

**Some of the email-exchange information**
```
Hi ZenGround and Why, 

This is a follow-up of the discussion in the filecoin discussion thread. The reason I'd like to discuss more here is: it's about attacking, not very sure if it should be exposed to all audiences. 

The attacking is happening right now, and after some investigation, I would think the attacker is doing like this: 
1) Shorten the block time to about 10s or less;
2) Select messages to create a block, most times only pick the messages related to this miner, and discard all others;
3) When creating a block, to calculate the ticket to see if the miner could be the next winner, if yes, send it out, if not, just produce blank block, and repeat until the new ticket could make sure it will be the next winner.

Since the height number could be changed due to current blank block mechanism, this is a way to pass the validation of winner, and the miner once get a chance to mine a block, it will guarantee it could get the right to mine next ones. 

From current implementation of view, please confirm if it is a possible way or most likely the thing happening right now in the devnet? 

If this is true, is there a short-term way to avoid this, instead of fully implementation of VDF? e.g. in the validation of mining winner process, let the validator check the height and compare with previous height number to see the difference, and check the time difference of the two? 

Best Regards,
Steven

```

Replied by @whyrusleeping : 
```
[...]

Issue #2412, opened 2019-03-25
Consensus protocol for long-running network


phritz:
This epic captures issues for completion of the consensus protocol ready for a long-running network.

ZenGround0:
Having a lot of trouble modifying this epic with zenhub.  

Both #2938 and #2946 should be included to land for 0.5.  Also #2224 is not scheduled to land for 0.5

anorth:
Fixed

Issue #2222, opened 2019-03-09
EC Upgrade 1: ticket arrays


ZenGround0:
### Description
Perhaps the biggest of several [recent changes](https://github.com/filecoin-project/specs/pull/138) to the EC spec is that blocks now must hold onto an array of tickets, one for each null block.  This issue tracks work to make this change.

### Acceptance criteria
* block format includes space for multiple tickets
* during mining new blocks are created with an array of tickets in the case of null block mining
* during validation the number of tickets is checked to make sure that it corresponds with the height field.

Don't worry about what we put in the ticket array so much.  Generating something that will actually be validated is tracked in part 4.  Because we don't have VDF bindings we need to mock some of this out anyway.

anorth:
Note that block timestamp validation must be changed to use ticket height.

ZenGround0:
This is tracking the syntactic change to the block header.
- [ ] use the spec compliant ticket type with the correct VDF and VRF subtypes
- [ ] use an array instead of a single ticket
- [ ] don't make any semantic changes to ticket generation, just put the same data we currently use in a subfield of a single ticket in the array.

PR #2855, opened 2019-05-29
Feat/soft block delay

frrist:
## What's this PR
- Add timestamp to the block structure as [outlined in the spec](https://github.com/filecoin-project/specs/blob/master/mining.md#block-creation)
- Verify block timestamps 

Details on why this is required: https://github.com/filecoin-project/go-filecoin/issues/2834

## What's left?
- [ ] Move blocktime to consensus
- [ ] Add block validation methods to consensus package
- [ ] remove DONOTMERGE comment about error message
- [ ] add testing around block timestamp verification
- [ ] rebase


chain/default_syncer_test.go
```diff
@@ -204,7 +208,7 @@ func loadSyncerFromRepo(t *testing.T, r repo.Repo, dstP *DefaultSyncerTestParams
 	chainStore := chain.NewDefaultStore(chainDS, calcGenBlk.Cid())
 
 	blockSource := th.NewTestFetcher()
-	syncer := chain.NewDefaultSyncer(cst, con, chainStore, blockSource) // note we use same cst for on and offline for tests
+	syncer := chain.NewDefaultSyncer(cst, con, chainStore, blockSource, 1) // note we use same cst for on and offline for tests
```
frrist: putting a `1` here is not a good permanent solution (imo). As part of the "What's left" section of the PR, this will be addressed.

It is my opinion that a refactor of this file is in order -- its over 1000 lines long and was not easy to extend. However I believe said refactor is outside the scope of this PR.

chain/default_syncer.go
```diff
@@ -83,12 +83,13 @@ type DefaultSyncer struct {
 	badTipSets *badTipSetCache
 	consensus  consensus.Protocol
 	chainStore syncerChainReader
+	blockTime  time.Duration
```
frrist: I am not sure its the chain syncers job to enforce a soft-block time, I could imagine this field going on the consensus interface as well. cc @ZenGround0 you have any thoughts on this?

chain/default_syncer.go
```diff
@@ -208,6 +210,31 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
+		if uint64(b.Timestamp) > uint64(time.Now().UTC().Unix()) {
+			return errors.Errorf("block: %s was generated by miner: %s too far in the future", b.Cid(), b.Miner.String())
+		}
+		limit := int64(pMinTs) + (int64(syncer.blockTime.Seconds()) * int64(len(tickets)))
+		// By setting this to strictly less than, we allow for a 1 second drift
+		// in clocks. This has an added benefit in that it keeps FAST tests
+		// working that mine at the sub-second level.
+		// The loss of precision in seconds need not matter.
+		if int64(b.Timestamp) < limit {
+			// DONOTMERGE clean up error message before merging, right now its nice to debug with
```
frrist: @frrist will remove this before mering.

chain/default_syncer_test.go
```diff
@@ -49,6 +49,8 @@ type DefaultSyncerTestParams struct {
 	minerAddress      address.Address
 	minerOwnerAddress address.Address
 	minerPeerID       peer.ID
+
+	blockTimeTicker func() uint64
```
frrist: Will likely need something more robust than this moving forward, what that looks like is fuzzy to me at the moment -- for now doing the simple thing.

chain/default_syncer.go
```diff
@@ -208,6 +210,31 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 		return nil
 	}
 
+	pMinTs, err := parent.MinTimestamp()
+	if err != nil {
+		return err
+	}
+
+	tickets, err := next.Tickets()
```
anorth: Blocking: this looks wrong. If a tipset has two blocks and each has two tickets (one null block) then `len(tickets)` below is 4 while I think the intention is 2. This should probably be using `Height` instead.

types/block.go
```diff
@@ -54,6 +54,9 @@ type Block struct {
 	// a challenge
 	Proof PoStProof `json:"proof"`
 
+	// A Unix Timestamp generated at block creation.
```
anorth: "The timestamp, in seconds since the Unix epoch, at which this block was created"

testhelpers/chain.go
```diff
@@ -251,3 +260,13 @@ func (bs *FakeBlockProvider) NewBlockWithMessages(nonce uint64, messages []*type
...
+
+// NewBlkTimeForTestGetter returns a closure that returns a uint64 unique to that invocation.
+// The uint64 is unique wrt the closure returned, not globally.
+func NewBlkTimeForTestGetter() func() uint64 {
+	var ticker uint64
+	return func() uint64 {
+		ticker++
+		return ticker
+	}
+}
```
anorth: I found the use of this really confusing, for probably a few reasons:
- there are no units specified anywhere. I guess it's intended to be seconds?
- neither a starting timestamp nor the block time are parameters
- the name / documentation doesn't describe intention

I *think* this is intended as a thing that generates a sequence of block timestamps, such that if sequential blocks use sequential values, their timestamps will be deemed valid. I think it's just a fluke that the implicit increment "1" matches a block time that we set for tests, but it's a huge smell to me that we can't (or don't) set a value of "30" (and still have tests run nearly instantly).

If that's right, then I think this should be called something like `NewBlockTimeTicker` and take at least  parameters, being a start timestamp and a tick duration. Possibly we want a type alias for `BlockTimestamp` so that the `uint64` makes sense out of context.

It would be more direct as struct with a method than a closure, too. The `func() uint64` when this type is referred to from elsewhere is pretty opaque. This would then let it be manipulated for tests, e.g. to generate an invalid timestamp.

chain/default_syncer_test.go
anorth: Messy code that's impeding some functionality, or testing thereof, that we wish to add is exactly the right signal to do a refactor. Please do such refactor in a preceding PR, though.

chain/default_syncer.go
anorth: I don't think it should go deeper. But rather than block time, this would be better as an abstract validator.

chain/default_syncer.go
ZenGround0: @frrist this *will* work but only after
(1) you move this check to be on a single block and access that single block's ticket array
(2) the implementation actually catches up with the spec and includes one ticket per preceding null block.  Right now this doesn't happen and all blocks have a single ticket.

For now as @anorth suggests comparing a block's height with its parent height is thing to do.

chain/default_syncer.go
```diff
@@ -208,6 +210,31 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
+	for _, b := range next.ToSlice() {
+		if uint64(b.Timestamp) > uint64(time.Now().UTC().Unix()) {
+			return errors.Errorf("block: %s was generated by miner: %s too far in the future", b.Cid(), b.Miner.String())
+		}
+		limit := int64(pMinTs) + (int64(syncer.blockTime.Seconds()) * int64(len(tickets)))
+		// By setting this to strictly less than, we allow for a 1 second drift
+		// in clocks. This has an added benefit in that it keeps FAST tests
+		// working that mine at the sub-second level.
+		// The loss of precision in seconds need not matter.
+		if int64(b.Timestamp) < limit {
```
ZenGround0: Please submit a PR to the spec with this change (can happen after you merge this one)

chain/default_syncer_test.go
ZenGround0: Moving this validation check to a method on `con` is "the right thing" IMO.  Mocking out block times then becomes part of mocking out a consensus module.  Also I don't disagree about this file needing a refactor.

types/tipset.go
```diff
@@ -131,6 +131,35 @@ func (ts TipSet) MinTicket() (Signature, error) {
 	return min, nil
 }
 
+// Tickets returns the tickets in the tipset.
+func (ts TipSet) Tickets() ([]Signature, error) {
```
ZenGround0: Blocking: This was introduced for an incorrect reason and should be removed.

ZenGround0:
@anorth just read through your comments now.  I generally agree: validation entrypoints should be separated into those that can run immediately and those that need the ancestral chain.  I also agree that separating out validation into components that can be swapped out/mocked during testing is needed.

Here is where we might differ:
(1) Validation cannot live in the chain package.  The current pattern is adding it as a function on the static consensus module.  The consensus package/module is basically just a validation package/module. Conceptually/Squishily I feel that having all validation rules organized under one roof is important.  Practically, this way functions can import this minimal dependency across the codebase (during mining or pubsub validation for example).  There are problems with this (have to mock out everything each time which makes testing onerous).  Having a separate validation package/module that consensus has as a dependency seems like a good way to improve this.

(2) This refactor should be a one shot thing that moves all validation operations around at once.  It probably should wait until after this PR.  Ideally it has a design doc.

anorth:
I'm fine with validation logic being centralised in the consensus package, so long as syntax validation can be injected to (not directly depended on by) code closer to the network boundary.

Either the refactor happens before this PR, or is pinky-promised to land immediately afterwards. Since there is currently no significant code to move, I prefer to avoid the structural regression.

frrist:
I see the trail forward as:
- Creating a BlockSyntaxValidator and BlockSemanticValidator interface in the consensus package.(as @anorth describes)
- Syncer uses the BlockSyntaxValidator close to network boundary (before fetching more information about it)
- Consensus uses the BlockSemanticValidator interface during RunStateTransition can replace/meld the `validateMining` method. 
- Validator impls can be tested independently. 
- Validator interfaces can be mocked out for testing -- ideally we can mock nearly all parts of the consensus package when testing the syncer.

I am going to need to tinker with the above decisions a bit in order to form a clearer picture of how they will fit together. This PR temporarily balloon in scope as these things are fleshed out at which point it will be broken into a set of smaller PRs.

types/tipset.go
frrist: Roger roger. Will remove and use the `Height` as @anorth describes.

From an above comment:

>  the implementation actually catches up with the spec and includes one ticket per preceding null block. Right now this doesn't happen and all blocks have a single ticket.

Do we have an issue to track this? Seems like a correctness issue.

chain/default_syncer.go
frrist: spec change being: `b.Timestamp <= limit` to strictly ` b.Timestamp < limit`?

chain/default_syncer.go
frrist: Will replace with an interface (`BlockSemanticValidator`) in the `consensus/` package. Validation refactor (that @ZenGround0 describes) can happen in a follow on.

testhelpers/chain.go
frrist: Agreed, this needs a better comment stating the intention.

> I think it's just a fluke that the implicit increment "1" matches a block time that we set for tests, but it's a huge smell to me that we can't (or don't) set a value of "30" (and still have tests run nearly instantly).

This is being used by unit tests, they don't perform any mining. This won't work for tests that do mining moving forward. I will sketch something out like the `NewBlockTimeTicker` you've described.

consensus/expected.go
```diff
@@ -100,50 +101,154 @@ type Expected struct {
...
+		if b.Parents.String() != one.Parents.String() {
+			return errors.Errorf("invalid parents for block: %s parents: %d & block: %s, parents: %d",
+				b.Cid().String(), b.Parents.String(), one.Cid().String(), one.Parents.String())
+		}
+
+		// validate same height
+		if b.Height != one.Height {
+			return errors.Errorf("invalid height for block: %s height: %d & block: %s, height: %d",
+				b.Cid().String(), b.Height, one.Cid().String(), one.Height)
+		}
```
anorth: When https://github.com/filecoin-project/go-filecoin/pull/2837 lands, the tipset consistency checks are redundant (I already copied from the same place you took these). I do think this should be limited to single block validation so it is usable before a block has been joined with others into a tipset.

consensus/expected.go
```diff
@@ -100,50 +101,154 @@ type Expected struct {
...
+type BlockSyntaxValidator interface {
+	ValidateSyntax(ctx context.Context, ts types.TipSet) error
+}
+
+type BlockSemanticValidator interface {
+	ValidateSemantics(ctx context.Context, cur, parent types.TipSet) error
+}
+
+type BlockStateValidator interface {
+	ValidateState(ctx context.Context, st state.Tree, cur, parent types.TipSet) error
```
anorth: If this is intended to be something like `RunStateTransition` then I don't think we should try to include it in this validation stuff (yet). In particular it's not clear that this doesn't have side-effects in the current implementation. Limit the scope of change as much as you can.

types/tipset.go
ZenGround0: Yes.  This change is tracked [here](https://github.com/filecoin-project/go-filecoin/issues/2222) as part of [this epic](https://github.com/filecoin-project/go-filecoin/issues/2412).

chain/default_syncer.go
```diff
@@ -208,6 +212,10 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 		return nil
 	}
 
+	if err := syncer.validator.ValidateSemantics(ctx, next, parent); err != nil {
```
ZenGround0: If the syncer has a `validator` field `RunStateTransition` should probably be a method on that field too for consistency.

chain/default_syncer_test.go
ZenGround0: Reiterating: this `1` should be moved into the constructor of the consensus validation module.

consensus/expected.go
```diff
@@ -243,30 +348,17 @@ func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, ance
 	span.AddAttributes(trace.StringAttribute("tipset", ts.String()))
 	defer tracing.AddErrorEndSpan(ctx, span, &err)
 
-	if err := c.validateMining(ctx, pSt, ts, ancestors[0]); err != nil {
+	if err := c.validateMining(ctx, st, ts, ancestors[0]); err != nil {
```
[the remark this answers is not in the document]
frrist: typo

frrist:
Closing as this ballooned in scope, https://github.com/filecoin-project/go-filecoin/pull/2882 is the outcome of some conversation here.

PR #2702, opened 2019-05-02
init creates a symlink to actual repo directory.

anorth:
The init command creates a symbolic link to a repo directory.
Updates lots of tests, especially the migration ones which necessarily inspected the filesystem.

PR #2865, opened 2019-05-31
feat/2841 - compute collateral as a function of sector size + track collateral as actor balance + collateral check before adding new commitments

### LINKED ISSUE #2841: The network computes collateral as a function of sector size
### Description

The miner owner transfers some funds to a miner actor when the miner actor is created. These funds act as the miner's collateral. The miner actor can lose some or all of this collateral if they (for example) fail to post a `submitPoSt` message to the blockchain for a proving period. 

Collateral should be computed as a function of miner's chosen sector size AND pledge; a miner should be penalized more for losing a large sector than a small one. This is consistent with the [spec](https://github.com/filecoin-project/specs/blob/master/actors.md#createstorageminer).

### Acceptance criteria

- implement `MinimumCollateral` function (which exists in spec pseudocode)
- implement `MinimumPledge` function (which exists in spec pseudocode)
- implement `CollateralForSectors` function (which exists in spec pseudocode)
- `CommitSector` method ensures that miner has sufficient remaining collateral to commit a new sector of the selected size

### Notes

- the crypto economics team hasn't yet come up with a function for computing collateral from sector size
- the collateral-calculating function can return, for now, the same value regardless of sector size used as input
- the minimum pledge for a miner equals the collateral for one sector of the miner's chosen sector size

laser:

Fixes #2841.

## Why does this PR exist?

The spec demonstrates that a miner's collateral is its balance, and the collateral which is has locked up for sectors which it is actively proving must be tracked via the `ActiveCollateral` field. This field's value must be incremented when a sector is committed, and commitments should be rejected if the miner doesn't have enough collateral to cover a new commitment.

The spec also demonstrates that per-sector collateral is a function of sector size. We believe that this will change soon (see _Pledge Collateral Design Doc (Baseline (Pro-rata) Construction)_), but this moves the codebase in the right direction.

## What's in this PR?

- fix the broken retrieval script (as of #2702, repo dir must not exist when `init` is called)
- track miner collateral via its balance
- add a method to VM context called `MyBalance` which does what you'd expect (this is a [specified function](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector))
- verify miner has enough collateral to cover commitments

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/abi"
 	"github.com/filecoin-project/go-filecoin/actor"
-	"github.com/filecoin-project/go-filecoin/actor/builtin/storagemarket"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/exec"
 	"github.com/filecoin-project/go-filecoin/proofs"
@@ -33,6 +32,9 @@ const MaximumPublicKeySize = 100
 // TODO: what is an actual workable value? currently set very high to avoid race conditions in test.
 // https://github.com/filecoin-project/go-filecoin/issues/966
 const ProvingPeriodBlocks = 20000
+
+// MinimumCollateralPerSector is the minimum amount of collateral required per sector
+var MinimumCollateralPerSector, _ = types.NewAttoFILFromFILString("0.001")
 
 // GracePeriodBlocks is the number of blocks after a proving period over
 // which a miner can still submit a post at a penalty.
@@ -165,6 +167,7 @@ func NewState(owner address.Address, key []byte, pledge *big.Int, pid peer.ID, s
 		Power:             types.NewBytesAmount(0),
 		NextAskID:         big.NewInt(0),
 		SectorSize:        sectorSize,
+		ActiveCollateral:  types.NewZeroAttoFIL(),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


actor/builtin/miner/miner.go
```diff
@@ -145,23 +151,23 @@ type State struct {
 	SectorSize *types.BytesAmount
 }
 
-// NewActor returns a new miner actor
-func NewActor() *actor.Actor {
-	return actor.NewActor(types.MinerActorCodeCid, types.NewZeroAttoFIL())
+// NewActor returns a new miner actor with the provided balance.
+func NewActor(balance *types.AttoFIL) *actor.Actor {
```
acruikshank: This is a little sketchy. This only gets used in tests an genesis, but it suggests you can create a miner actor in a state transition with a non-zero balance. This would be really bad, since it would be generating FIL from nowhere. For clarity's sake, I'd rather you not add balance to the constructor and set it manually after the actor is constructed.

actor/builtin/miner/miner.go
```diff
@@ -564,6 +578,14 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 	return 0, nil
 }
 
+// CollateralForSector returns the collateral required to commit a sector of the
+// given size.
+func CollateralForSector(sectorSize *types.BytesAmount) *types.AttoFIL {
+	// TODO: Replace this function with the baseline pro-rata construction.
+	// https://github.com/filecoin-project/go-filecoin/issues/2866
```
acruikshank: 🙏

actor/builtin/miner/miner.go
laser: > For clarity's sake, I'd rather you not add balance to the constructor and set it manually after the actor is constructed.

OK. This change is easy to roll back. I introduced it to [be consistent](https://github.com/filecoin-project/go-filecoin/blob/master/actor/builtin/account/account.go#L23) with the account actor-creation constructor. Do you think that the same logic applies to the account actor?

laser:
@ZenGround0 

> It's worth checking the wiki and potentially making an update explaining that there is a minimum collateral during the miner setup workflow (though it might not affect anything if users run with the default FIL values).

Sounds good. I'm about to delete the `pledge` concept from the storage miner (and corresponding `miner create` command). I'll update the wiki to reflect both changes simultaneously.

PR #2870, opened 2019-05-31
Clean gengen fixture data before generation

### LINKED ISSUE #2861: Remove contents fo `fixtures/[test/live]` during build prior to gengen
### Description

We load some data from disk in an `init` method in the fixtures package. This can result in a panic if the data-structure used to generate the data changes as we can no longer read it back.

This was discovered in #2859.

We should remove this data during build prior to running gengen so avoid this conflict in the future.

### Acceptance criteria

- Contents of `fixtures/[test/live]` are remove prior to a call to running gengen

### Where to begin

https://github.com/filecoin-project/go-filecoin/blob/master/build/main.go#L172-L189

travisperson:
Closes #2861

### DIFF
--- a/build/main.go
+++ b/build/main.go
@@ -171,7 +171,7 @@ func forceBuildFC() {
 	}...))
 }
 
-// cealnDirectory removes the child of a directly wihtout removing the directory itself, unlike `RemoveAll`.
+// cleanDirectory removes the child of a directly wihtout removing the directory itself, unlike `RemoveAll`.
 // There is also an additional parameter to ignore dot files which is important for directories which are normally
 // empty. Git has no concept of directories, so for a directory to automatically be created on checkout, a file must
 // exist in side of it. We use this pattern in a few places, so the need to keep the dot files around is impotant.
@@ -194,7 +194,7 @@ func cleanDirectory(dir string, ignoredots bool) error {
 		fpath := filepath.Join(dir, fname)
 
 		fmt.Println("Removing", fpath)
-		if err := os.Remove(fpath); err != nil {
+		if err := os.RemoveAll(fpath); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


build/main.go
```diff
@@ -169,15 +171,62 @@ func forceBuildFC() {
 	}...))
 }
 
+func cleanDirectory(dir string, ignoredots bool) error {
```
frrist: Can you add a small comment stating that we use this method (and not https://golang.org/pkg/os/#RemoveAll) to preserve the `.gitkeep` file? (or could rename the method to reflect that).

Is the `.gitkeep` necessary?

travisperson: Ya, the `.gitkeep` is necessary if we want the directory to exist on code checkout.

build/main.go
```diff
@@ -169,15 +171,66 @@ func forceBuildFC() {
 	}...))
 }
 
+// cealnDirectory removes the child of a directly wihtout removing the directory itself, unlike `RemoveAll`.
```
anorth: `cleanDirectory`

build/main.go
```diff
@@ -169,15 +171,66 @@ func forceBuildFC() {
...
+	for _, file := range files {
+		fname := file.Name()
+		if ignoredots && []rune(fname)[0] == '.' {
+			continue
+		}
+
+		fpath := filepath.Join(dir, fname)
+
+		fmt.Println("Removing", fpath)
+		if err := os.Remove(fpath); err != nil {
```
anorth: What if fpath is also a directory? This should probably RemoveAll.

build/main.go
```diff
@@ -169,15 +171,66 @@ func forceBuildFC() {
...
+		}
+	}
+
+	return nil
+}
+
 func generateGenesis() {
 	log.Println("Generating genesis...")
+
+	liveFixtures, err := filepath.Abs("./fixtures/live")
```
anorth: Factor out the duplicated path names (line 219)?

PR #2872, opened 2019-06-01
add timeout to daemon test Wait() calls

laser:
## Why is this PR needed?

The daemon tests create wait groups and schedule goroutines. If a panic happens in the goroutine (say, because `t.Require` was used), the wait group `Done()` method is never called. The test then hangs until the `go test` timeout expires.

## What's in this PR?

- add a helper function which waits a maximum amount of time for a wait group to complete
- use the helper function in the daemon tests

### DIFF
--- a/commands/message_daemon_test.go
+++ b/commands/message_daemon_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -104,7 +105,9 @@ func TestMessageWait(t *testing.T) {
 
 		d.RunSuccess("mining once")
 
-		wg.Wait()
+		if !th.WaitWithTimeout(&wg, time.Second*5) {
+			t.Fatal("timed out waiting for message wait command to return")
+		}
 	})
 }
 
--- a/commands/miner_daemon_test.go
+++ b/commands/miner_daemon_test.go
@@ -143,7 +143,10 @@ func TestMinerCreate(t *testing.T) {
 
 			// ensure mining runs after the command in our goroutine
 			d1.MineAndPropagate(time.Second, d)
-			wg.Wait()
+
+			if !th.WaitWithTimeout(&wg, time.Second*5) {
+				t.Fatal("timed out waiting for miner create command to return")
+			}
 
 			// expect address to have been written in config
 			config := d.RunSuccess("config mining.minerAddress")
@@ -208,7 +211,10 @@ func TestMinerCreate(t *testing.T) {
 
 		// ensure mining runs after the command in our goroutine
 		d1.MineAndPropagate(time.Second, d)
-		wg.Wait()
+
+		if !th.WaitWithTimeout(&wg, time.Second*5) {
+			t.Fatal("timed out waiting for miner create command to return")
+		}
 	})
 }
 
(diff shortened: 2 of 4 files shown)


anorth:
I recently removed a bunch of hardcoded and fairly arbitrary timeouts because they caused extreme test flakiness for me (in conjunction with a loaded machines and `-parallel` hardcoded too high). I don't think a bunch of timeouts scattered throughout the code is a good solution. People will add more, and with little more rationale behind the numbers than "worked for me". Test code can't make a good assessment about the environment in which it is executing and what kind of timeout would be appropriate.

Since I removed those other timeouts, and the hardcoded `30m` timeout to `go test`, I've had good success running with `-timeout 200s` locally (and could probably go lower).

Do I understand correctly that these timeouts *should* only be relevant in the case of tests failing due a legitimate error? In that uncommon case, I think a top-level timeout is more appropriate.

Another problem your change might be addressing is getting some diagnostic about what went wrong when a test times out. Is that part of the need? Can `go test` give us any hooks in this case?

If this is the only way, I think we should remove the constant `5` from call sites and have a single value that is waaaaay too large for the expected contexts, so that it could only possibly trigger in case of a real hang; maybe `60`.

commands/message_daemon_test.go
```diff
@@ -104,7 +105,9 @@ func TestMessageWait(t *testing.T) {
 
 		d.RunSuccess("mining once")
 
-		wg.Wait()
+		if !th.WaitWithTimeout(&wg, time.Second*5) {
```
anorth: Would it make sense to push these deeper so that *all* the call sites don't need this? Like if every method on `d` carried a timeout, configured at construction?

laser:
@anorth 

> Do I understand correctly that these timeouts should only be relevant in the case of tests failing due a legitimate error?

There are two scenarios which this PR addresses:

#### Early Exit in Scheduled Goroutines

We oftentimes create a wait group in a test, increment its counter, and then schedule a goroutine. We decrement the wait group's counter in the goroutine. If something causes the goroutine to exit early (e.g. a call to `TestingT#Equal`, which delegates to `TestingT#FailNow`), the wait group counter doesn't decrement and the parent goroutine hangs on `WaitGroup#Wait`.

If this behavior manifests in a test (after this PR has been merged), the programmer would see that both a timeout occurred (waiting for the child goroutine to exit) and that a failure occurred in the goroutine (e.g. `require.Equal` was not satisfied).

Instead of introducing a test timeout to address this issue, we could audit the test code to ensure that we `defer wg.Done()` in descendant goroutines which use methods on values returned from `require.New()`.

#### Waiting Forever for Output

A test can also hang if, for example, [a CLI command never returns](https://github.com/filecoin-project/go-filecoin/blob/master/commands/message_daemon_test.go#L93). We have at least two blocking CLI commands (`message wait` and `miner create`). We don't want tests to hang indefinitely if either of these two CLI commands do not return.

shannonwells:
> Instead of introducing a test timeout to address this issue, we could audit the test code to ensure that we `defer wg.Done()` in descendant goroutines which use methods on values returned from `require.New()`.

It seems on the surface of it that a helper function like you first proposed discourages good go practice, i.e. the 'defer' call that you refer to above. My initial reaction is it's better to do this audit you suggest, rather than introducing a black box helper that gets called.  I realize that this idea could apply to nearly anything (i.e. why abstract anything at all) but it seems for go routines, there should be explicit choices about these wait groups and defer calls in each test. Also if there is other cleanup to do it's not added to the defer stack but done all in one place, in a guaranteed and obvious order.

mishmosh:
This has been sitting for a while. @laser @anorth can you discuss & resolve when you meet on video today?

laser:
@mishmosh - Thanks for reminding me. This PR is defunct; I'll close it now.

PR #2873, opened 2019-06-02
[part 2] parameterize CreateStorageMiner with a sector size

laser:
This PR makes incremental progress on #2530.

Note that this PR is branched off of #2871. Once the upstream PR is merged, this PR will need to be rebased onto `master`. 

## Why does this PR exist?

The specification shows that the `CreateStorageMiner` method should accept a sector size, and that the created storage miner actor must be forever pinned to that sector size.

## What's in this PR?

This PR modifies the implementation to select, based on the storage market actor's proofs mode, a sector size to be passed in the `createStorageMiner` message.

## What's next?

I need to modify the `miner create` command (and corresponding wiki docs) to accept a sector size as input.

### DIFF
--- a/commands/miner.go
+++ b/commands/miner.go
@@ -62,6 +62,25 @@ additional sectors.`,
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
 		var err error
 
+		pp, err := GetPorcelainAPI(env).ProtocolParameters(env.Context())
+		if err != nil {
+			return err
+		}
+
+		// TODO: Modify this command to accept a sector size as an argument,
+		// ensuring that the provided sector size is supported by the network.
+		// https://github.com/filecoin-project/go-filecoin/issues/2530
+		//
+		// TODO: It may become the case that the protocol does not specify an
+		// enumeration of supported sector sizes, but rather that any sector
+		// size for which a miner has Groth parameters and a verifying key is
+		// supported.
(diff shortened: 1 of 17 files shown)


commands/miner.go
```diff
@@ -62,6 +62,25 @@ additional sectors.`,
...
+		pp, err := GetPorcelainAPI(env).ProtocolParameters(env.Context())
+		if err != nil {
+			return err
+		}
+
+		// TODO: Modify this command to accept a sector size as an argument,
+		// ensuring that the provided sector size is supported by the network.
+		// https://github.com/filecoin-project/go-filecoin/issues/2530
+		//
+		// TODO: It may become the case that the protocol does not specify an
```
acruikshank: This is contentious and at any rate, it's not really a TODO. I'd remove this todo.

gengen/util/gengen.go
```diff
@@ -30,8 +30,11 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Miner is
-type Miner struct {
+// CreateStorageMinerConfig holds configuration options used to create a storage
+// miner in the genesis block. Note: Instances of this struct can be created
+// from the contents of fixtures/setup.json, which means that a JSON
+// encoder/decoder must exist for any of the struct's fields' types.
```
acruikshank: thank you

laser: no prob bob

PR #2874, opened 2019-06-02
[part 3] "miner create" command must accept "sectorsize" flag

### LINKED ISSUE #2530: A miner selects a sector size when pledging

laser:
Fixes #2530.

Note that this PR is based off of #2873. Once that PR is merged, this PR will need to be rebased onto `master`.

## Why does this PR exist?

We need to give an operator a command option which lets them pick the size of the sectors which their newly-created storage miner actor will commit.

## What's in this PR?

- modify `miner create` to accept an optional `sectorsize` flag
- default to whatever sector size is supported by the proofs mode

### Invalid `sectorsize`

```
$ ./go-filecoin miner create 1 --sectorsize=woot
Error: invalid sector size: woot
```

### Unsupported `sectorsize`

```
$ ./go-filecoin miner create 1 --sectorsize=12
Error: unsupported sector size: 12 (supported sizes: 268435456)
```

### DIFF
--- a/commands/miner.go
+++ b/commands/miner.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"math/big"
 	"strconv"
+	"strings"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit"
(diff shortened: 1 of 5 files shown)


commands/miner.go
```diff
@@ -50,10 +51,11 @@ miner's collateral drops below 0.001FIL, the miner will not be able to commit
 additional sectors.`,
 	},
 	Arguments: []cmdkit.Argument{
-		cmdkit.StringArg("collateral", true, false, "The amount of collateral, in FIL"),
+		cmdkit.StringArg("collateral", true, false, "The amount of collateral, in FIL."),
```
anorth: FWIW I think we could make this an option (rather than argument) too

PR #2875, opened 2019-06-02
compute proving period as a function of sector size

### LINKED ISSUE #2839: The network computes proving period blocks as a function of sector size
### Description

A miner actor's proving period is a function of the miner's selected sector size. Small sectors take less time to prove than larger sectors do, so proving period blocks should be a function of sector size. This is consistent with the [spec](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector).

### Acceptance criteria

- proving period becomes a function of sector size

### Notes

- the crypto economics team hasn't yet come up with a function for computing proving period from sector size
- the proving period-calculating function can return, for now, the same value (20000 blocks) regardless of sector size used as input

laser:
Fixes #2839 

## Why is this PR required?

A miner actor's proving period is a function of the miner's selected sector size. Small sectors take less time to prove than larger sectors do, so proving period blocks should be a function of sector size. This is consistent with the [spec](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector).

## What's in this PR?

Make proving period blocks a function of sector size. This is currently a constant function. Make `AncestorRoundsNeeded` rely on the maximum value which `ProvingPeriodDuration` can return.

### DIFF
--- a/consensus/expected.go
+++ b/consensus/expected.go
@@ -69,7 +69,11 @@ const ECPrM uint64 = 100
 
 // AncestorRoundsNeeded is the number of rounds of the ancestor chain needed
 // to process all state transitions.
-const AncestorRoundsNeeded = miner.ProvingPeriodBlocks + miner.GracePeriodBlocks
+//
+// TODO: If the following PR is merged - and the network doesn't define a
+// largest sector size - this constant will need to be reconsidered.
+// https://github.com/filecoin-project/specs/pull/318
+const AncestorRoundsNeeded = miner.LargestSectorSizeProvingPeriodBlocks + miner.GracePeriodBlocks
 
 // A Processor processes all the messages in a block or tip set.
 type Processor interface {
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -727,9 +727,17 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) {
 		return
 	}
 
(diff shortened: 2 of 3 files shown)


protocol/storage/miner.go
```diff
@@ -728,7 +728,7 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) {
 	}
 
 	h := types.NewBlockHeight(height)
-	provingPeriodHeight := types.NewBlockHeight(miner.ProvingPeriodBlocks)
+	provingPeriodHeight := types.NewBlockHeight(miner.LargestSectorSizeProvingPeriodBlocks)
```
anorth: Blocking: I think this should be *this miner's* period, not the max.

consensus/expected.go
```diff
@@ -69,7 +69,11 @@ const ECPrM uint64 = 100
 
 // AncestorRoundsNeeded is the number of rounds of the ancestor chain needed
 // to process all state transitions.
-const AncestorRoundsNeeded = miner.ProvingPeriodBlocks + miner.GracePeriodBlocks
+//
+// TODO: If the following PR is merged - and the network doesn't define a
+// largest sector size - this constant will need to be reconsidered.
+// https://github.com/filecoin-project/specs/pull/318
+const AncestorRoundsNeeded = miner.LargestSectorSizeProvingPeriodBlocks + miner.GracePeriodBlocks
```
ZenGround0: Good to know about.

PR #2876, opened 2019-06-02
generation attack threshold is a function of sector size

### LINKED ISSUE #2840: The network computes generation attack time as a function of sector size
### Description

A miner is allowed to post a `submitPoSt` message for a proving period to the blockchain after the proving period has ended, but they will lose (burn) some FIL for doing so. The message must be included in a block whose height <= the proving period end PLUS generation attack time (in blocks). Small sectors have a shorter proving period than large sectors do, and small sectors have a smaller generation attack threshold than large sectors do.

### Acceptance criteria

- implement `GenerationAttackTime` function (which exists in spec pseudocode)

### Notes

- the crypto economics team hasn't yet come up with a function for computing generation attack time from sector size
- the generation attack time-calculating function can return, for now, the same value regardless of sector size used as input (developer gets to pick a value)

laser:
Fixes #2840 

## Why is this PR needed?

The specification shows, in pseudocode, that generate attack threshold is a function of sector size.

## What's in this PR?

Introduce `GenerationAttackTime` (constant) function which returns, for the largest sector size, the number of blocks after a proving period P ends after which a delinquent miner (i.e. they haven't yet posted a `submitPoSt` message for P) is subject to storage fault slashing.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -38,14 +38,18 @@ const MaximumPublicKeySize = 100
 // https://github.com/filecoin-project/specs/pull/318
 const LargestSectorSizeProvingPeriodBlocks = 20000
 
+// LargestSectorGenerationAttackThresholdBlocks defines the number of blocks
+// after a proving period ends after which a miner using the largest sector size
+// supported by the network is subject to storage fault slashing.
+//
+// TODO: If the following PR is merged - and the network doesn't define a
+// largest sector size - this constant and consensus.AncestorRoundsNeeded will
+// need to be reconsidered.
+// https://github.com/filecoin-project/specs/pull/318
+const LargestSectorGenerationAttackThresholdBlocks = 100
+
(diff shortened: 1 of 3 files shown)


actor/builtin/miner/miner.go
```diff
@@ -33,15 +33,19 @@ const MaximumPublicKeySize = 100
 // https://github.com/filecoin-project/go-filecoin/issues/966
 const ProvingPeriodBlocks = 20000
 
+// LargestSectorGenerationAttackThresholdBlocks defines the number of blocks
+// after a proving period ends after which a miner using the largest sector size
+// supported by the network is subject to storage fault slashing.
+//
+// TODO: If the following PR is merged - and the network doesn't define a
+// largest sector size - this constant and consensus.AncestorRoundsNeeded will
```
ZenGround0: nit: Since you already have a comment in the other PR over consensus.AncestorRoundsNeeded you don't need to mention it here.

PR #2879, opened 2019-06-03
Change chain traversal to hit tipset cache

anorth:
@ZenGround0 suggested that the chain traversals should use a TipsetProvider rather than BlockProvider so they could hit the store's tip index.

This allows most traversals to hit the in-memory tipset cache and bypass a heap of disk access and decoding. For the traversals that can't use an index, added an adaptor that builds tipsets on the fly from a block provider (i.e. the old way).

### DIFF
--- a/chain/traversal.go
+++ b/chain/traversal.go
@@ -8,40 +8,21 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// BlockProvider provides blocks. This is a subset of the ReadStore interface.
-type BlockProvider interface {
-	GetBlock(ctx context.Context, cid cid.Cid) (*types.Block, error)
-}
-
-// GetParentTipSet returns the parent tipset of a tipset.
-// The result is empty if the tipset has no parents, but an error if the tipset is undefined.
-func GetParentTipSet(ctx context.Context, store BlockProvider, ts types.TipSet) (types.TipSet, error) {
-	parents, err := ts.Parents()
-	// Parents is empty (without error) for the genesis tipset, and does't produce an error here either.
-	if err != nil || parents.Len() == 0 {
-		return types.UndefTipSet, err
-	}
-	var newBlocks []*types.Block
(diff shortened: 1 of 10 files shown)


anorth:
@ZenGround0 note the failing tests. I can change GetTipSet to fall back to loading the blocks (which it won't be able to cache in the current tipindex) but that will require plumbing a `Context` through all the existing callers. I think that's probably the right thing to do: lack of a context is exposing an implementation detail that the tipset is (until now) served from memory.

anorth:
Updated and ready for review. Rather than plumb context through GetTipSet I added an adaptor that builds tipsets on the fly from blocks. This has a slightly funky smell because it needs to store a context for passing to GetBlock, but I think that this is pragmatic for now. It'll probably be quite a while before we need to limit the tip index size and hence pass contexts around.

chain/traversal.go
```diff
@@ -57,12 +38,52 @@ func (it *TipsetIterator) Complete() bool {
...
+		return err
+	}
+}
+
+// BlockProvider provides blocks.
+type BlockProvider interface {
+	GetBlock(ctx context.Context, cid cid.Cid) (*types.Block, error)
+}
+
+// LoadTipsetBlocks loads all the block for a tipset from the store.
```
ZenGround0: nit: blocks

core/inbox_test.go
```diff
@@ -408,6 +409,19 @@ func (p *fakeChainProvider) GetBlock(ctx context.Context, cid cid.Cid) (*types.B
 	return &blk, nil
 }
 
+func (p *fakeChainProvider) GetTipSet(tsKey types.SortedCidSet) (types.TipSet, error) {
+	ctx := context.TODO() // Should GetTipSet require a context everywhere?
+	var blocks []*types.Block
```
ZenGround0: I think we can reuse `chain.LoadTipsetBlocks` here without any issue.

anorth: Thanks, I intended that and forgot :-)

PR #2880, opened 2019-06-03
Feat/mig cbor meta 2595

shannonwells:
Use @ZenGround0 's [chainstore metadata conversion spike](https://github.com/filecoin-project/go-filecoin/tree/spike/cbor-meta) to create a first repo migration.

A fixture tarball is included for testing purposes; the same validation that is run in test is run on a migrated repo.

Validation steps:
After conversion, the old and new chain stores are loaded, then we compare old and new head tipsets, and iterate over the chainstores and ensure that the state roots are the same at each step, as well as verifying that they are both the same length.

### WARNING
This updates the repo version to 2; after this is merged, unless you reset your chain, **go-filecoin will exit until migration is run**.

#### Reviewers please help by checking out this branch and verify against a test repo:
* I used localnet to generate a small test repo, and I made a tarball of it so I could keep checking against it.
* Build and run go-filecoin against your repo and verify that it exits saying you need to migrate.
* You will want to run `go-filecoin-migrate` with the `--verbose` option, otherwise all log output goes to  `~/.filecoin-migration-logs`
* After migration, be sure to run go-filecoin daemon built from this branch to make sure that it will run with the newly migrated repo.  
* If you change back to master, build go-filecoin and run the daemon, it ought to exit because it will see that the version is too high.


tools/migration/main.go
```diff
@@ -78,17 +78,39 @@ func main() { // nolint: deadcode
 	switch command {
 	case "-h", "--help":
 		showUsageAndExit(0)
-	case "describe", "buildonly", "migrate", "install":
-		logFile, err := openLogFile()
+	case "describe":
```
shannonwells: Even though there is some duplicated code here, I separated this out because it doesn't need to do nearly as much as the others.

tools/migration/migrations/repo-1-2/migration_test.go
```diff
@@ -0,0 +1,72 @@
...
+	assert.Equal(t, expected, mig.Describe())
+}
+
+func TestMigrateSomeRepo(t *testing.T) {
+	tf.UnitTest(t)
+
+	mig := migration12.MetadataFormatJSONtoCBOR{}
+	_, newVer := mig.Versions()
+
+	curDir, err := syscall.Getwd()
```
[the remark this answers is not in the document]
shannonwells: I don't know, but I figure if it's passing CI it should be fine? 🤷‍♀ The test cleans up after itself.

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,373 @@
...
+		}
+		blocks = append(blocks, blk)
+	}
+
+	headTs, err := types.NewTipSet(blocks...)
+	if err != nil {
+		return errors.Wrap(err, "failed to add validated block to TipSet")
+	}
+
+	for iterator := chain.IterAncestors(ctx, m.chainStore, headTs); !iterator.Complete(); err = iterator.Next() {
```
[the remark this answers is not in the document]
shannonwells: yikes, ok thanks for the heads up just in case

tools/migration/main.go
```diff
@@ -78,32 +78,36 @@ func main() { // nolint: deadcode
...
+		logger, err := newLoggerWithVerbose(true)
```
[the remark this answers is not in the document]
shannonwells: I made this call because I tried to run describe and didn't like the behavior. I couldn't imagine why you'd ever have a script run it and care about the log content.  Since the instantiation of the migration runner requires a logger, it's created but it's ignored if describe is called. We could always log it if we decide to log describe output later.

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,376 @@
...
+		}
+
+		oldStateRoot, err := loadStateRootAsJSON(oldIt.Value(), oldChainStore)
+		if err != nil {
+			return err
+		}
+		if !newStateRoot.Equals(oldStateRoot) {
+			return errors.New("current state root not equal for block")
+		}
+		if err = oldIt.Next(); err != nil {
```
ZenGround0: Because of the way [this iterator works](https://github.com/filecoin-project/go-filecoin/blob/master/chain/traversal.go#L17) if the old chain is somehow shorter than the new one then this loop will end up operating on an undefined tipset and raise a more cryptic error at line 329.  It would be a bit nicer to check `oldIt.Complete` at the top of the loop and explicitly return "old chain store is shorter than new chain store" there.

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,376 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !newHeadTs.Equals(oldHeadTs) {
+		return errors.New("new and old head tipsets not equal")
+	}
+
+	oldIt := chain.IterAncestors(ctx, newChainStore, newHeadTs)
+	for newIt := chain.IterAncestors(ctx, oldChainStore, oldHeadTs); !newIt.Complete(); err = newIt.Next() {
```
ZenGround0: Blocking: looks like these names are reversed (`oldIt` paired with `newChainStore` and vice versa)

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,376 @@
...
+			return errors.Wrap(err, "old chain store Next failed")
+		}
+	}
+	if !oldIt.Complete() {
+		return errors.New("old chain store is longer than new chain store")
+	}
+	return nil
+}
+
+func loadBlocksToChainStore(ctx context.Context, cidSet types.SortedCidSet, chainStore *migrationChainStore) (headTs types.TipSet, err error) {
```
ZenGround0: This name confused me a bit because loading the blocks into memory isn't the thing the caller cares about (though I do get that it is a necessary part of what is happening), the caller cares about getting out the referenced tipset.  Maybe `loadTipSet`, or `getTipSet`?

tools/migration/migrations/repo-1-2/migration.go
ZenGround0: How is validation passing with this?

tools/migration/main.go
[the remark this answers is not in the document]
ZenGround0: I understand that and believe it was the right call.  I was trying to ask whether it makes sense to remove the logger altogether from this case in main.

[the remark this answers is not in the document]
ZenGround0: i.e. make a noop logger by wiring up the logger to /dev/null or something when running describe.

tools/migration/migrations/repo-1-2/migration.go
ZenGround0: Can CBOR/JSON dec/encoders read/write both formats? 👻

shannonwells: whups, thanks

tools/migration/migrations/repo-1-2/migration.go
shannonwells: ok, good idea

tools/migration/migrations/repo-1-2/migration.go
shannonwells: It passed either because it's comparing something with itself or the swap in variables was a complete swap.

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,376 @@
...
+		return errors.New("new and old head tipsets not equal")
+	}
+
+	oldIt := chain.IterAncestors(ctx, newChainStore, newHeadTs)
+	for newIt := chain.IterAncestors(ctx, oldChainStore, oldHeadTs); !newIt.Complete(); err = newIt.Next() {
+		if err != nil {
+			return err
+		}
+
+		newStateRoot, err := loadStateRoot(newIt.Value(), true, newChainStore)
```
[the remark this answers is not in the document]
shannonwells: I had it but it became unused so I deleted it.

tools/migration/main.go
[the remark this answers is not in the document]
shannonwells: ah, right. Yes I considered this too but chose expedience.

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,380 @@
...
+// Describe describes the steps this migration will take.
+func (m *MetadataFormatJSONtoCBOR) Describe() string {
+	return `MetadataFormatJSONtoCBOR migrates the storage repo from version 1 to 2.
+
+    This migration changes the chain store metadata serialization from JSON to CBOR.
+    The chain store metadata will be read in as JSON and rewritten as CBOR. 
+	Chain store metadata consists of associations between tipset keys and state 
+	root cids and the tipset key of the head of the chain.
+	No other repo data is changed.  Migrations are performed on a copy of the
+	chain store.
```
anorth: This last statement is not under the control of this migration, so I suggest dropping it

tools/migration/migrations/repo-1-2/migration.go
```diff
@@ -1 +1,380 @@
...
+		return cid.Undef, errors.Wrapf(err, "failed to cast state root of tipset %s", ts.String())
+	}
+	return stateRoot, nil
+}
+
+// compareChainStores loads each chain store and iterates through each, comparing heads
+// then state root at each iteration, returning error if:
+//     * at any point state roots are not equal
+//     * the old and new chain store iterators are not complete at the same time
+// 	   * at the end the two heads are not equal
```
anorth: Thanks, this is just the kind of comment I think is needed for an internal function like this.

PR #2881, opened 2019-06-03
Add a timeout option to "message wait" command

### LINKED ISSUE #1552: the message wait command should accept a timeout
Timeout should be a string in golang duration units. Add a deadline with it to the context in command.

ingar:
## Summary
Add a `timeout` argument to `message wait` that accepts duration strings compatible with [`time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration).

Resolves #1552

### DIFF
--- a/commands/message.go
+++ b/commands/message.go
@@ -1,10 +1,12 @@
 package commands
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"strconv"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit"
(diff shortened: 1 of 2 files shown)


commands/message.go
```diff
@@ -156,6 +158,7 @@ var msgWaitCmd = &cmds.Command{
 		cmdkit.BoolOption("message", "Print the whole message").WithDefault(true),
 		cmdkit.BoolOption("receipt", "Print the whole message receipt").WithDefault(true),
 		cmdkit.BoolOption("return", "Print the return value from the receipt").WithDefault(false),
+		cmdkit.StringOption("timeout", "Maximum time to wait for message. e.g., 300ms, 1.5h, 2h45m.").WithDefault("1m"),
```
anorth: I think the default should be very long, so that this is an opt-in to giving up early. I suggest 10m

ingar: Done.

Issue #2783, opened 2019-05-15
Prevent propagation of invalid blocks with block pubsub topic validator


anorth:
### Description
#2755 reported a network progress issue that ultimately came down to propagation of invalid blocks (which were far ahead of the consensus head #2782).

The spec describes [block validation logic](https://filecoin-project.github.io/specs/docs/validation/) which must execute before a block is accepted or propagated to peers. Validation includes basic well-formedness checks as well as checking that the block can be linked to the nodes existing chain and the contained messages and receipts describe valid state transitions. See @whyrusleeping's [comment on 2755]( https://github.com/filecoin-project/go-filecoin/issues/2755#issuecomment-491590106)

Implement a [libp2p pubsub topic validator](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L930) which validates blocks before accepting them locally or propagating them to peers. 

It's unclear whether "propagation" in the spec refers to libp2p pubsub validation, or the higher level availability of a block to peers that request it directly. Comprehensive block validation is expensive and requires access to all parent blocks and the preceding state (i.e. fully synced);  doing this in the topic validator may slow block propagation. Alternatively we could split validation into two parts, only one of which applies to the pubsub. We need @whyrusleeping to help us make a call here and, if necessary, update the spec to clarify.

### Acceptance criteria
- Blocks are validated before nodes perform any other processing
- Invalid blocks are not propagated

### Risks + pitfalls
- Tradeoff between block propagation time and validation
- If a block references a parent not already held by the node, it can't validate, or must block validation, until the parent is received.
- The node currently subscribes to the block pubsub before syncing the chain; it won't be able to validate blocks before that so should probably join pubsub only after catching up (#1160)
- The pubsub validator will prevent a node receiving an invalid block on `Subscription.Next()` which will also prevent it storing an explicit list of "bad" blocks, which the spec refers to (but is not yet implemented). Is refusing to accept or propagate a block sufficient? Is caching some validation results in the topic validator sufficient?

vyzo:
Note that we just merged https://github.com/libp2p/go-libp2p-pubsub/pull/176 which reworked the validation pipeline, and there is new pubsub release tag (https://github.com/libp2p/go-libp2p-pubsub/releases/tag/v0.0.3) that includes it.

Another issue you should be aware of is validation time and the interprlay with the timecache that prevents cycles in pubsub.
[...]

PR #2882, opened 2019-06-04
Feat/block validation interface

frrist:
This PR was generated from the conversation had in: https://github.com/filecoin-project/go-filecoin/pull/2855/files

### What's new?
- Add Block syntax and semantic validation
- `consensus.NewExpected` accepts a `BlockVlaidator` interface.
- `consensus.Protocol` implements `BlockValidator` interface.

### DIFF
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -189,19 +189,6 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 			return nil, err
 		}
 
-		// validate blocks we maybe got from the network
-		// TODO: if we didn't get these blocks from the network and instead got
-		// them from disk, validation feels redundant. This requires
-		// looser coupling between the store and the fetcher.
-		// De-tangle syner and store: #2128
-		// Prevent Bad block propagation: #2783
-		for _, b := range blks {
-			if err := syncer.consensus.ValidateSyntax(ctx, b); err != nil {
-				syncer.badTipSets.Add(tsKey)
-				syncer.badTipSets.AddChain(chain)
-				return nil, err
-			}
-		}
 		ts, err := types.NewTipSet(blks...)
 		if err != nil {
 			return nil, err
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -170,7 +170,6 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 	// number of new input blocks is less than the FinalityLimit constant.
 	// Otherwise, halt assuming the new blocks come from an invalid chain.
 	for (syncer.syncMode == Syncing) || !syncer.exceedsFinalityLimit(chain) {
-		var blks []*types.Block
 		// check the cache for bad tipsets before doing anything
 		tsKey := tipsetCids.String()
 
@@ -190,10 +189,19 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 			return nil, err
 		}
 
-		ts, err := syncer.consensus.NewValidTipSet(ctx, blks)
+		// validate blocks we maybe got from the network
+		// TODO: if we didn't get these blocks from the network and instead got
+		// them from disk, validation feels redundant. This requires
+		// looser coupling between the store and the fetcher.
+		for _, b := range blks {
+			if err := syncer.consensus.ValidateSyntax(ctx, b); err != nil {
+				syncer.badTipSets.Add(tsKey)
+				syncer.badTipSets.AddChain(chain)
+				return nil, err
+			}
+		}
+		ts, err := types.NewTipSet(blks...)
 		if err != nil {
-			syncer.badTipSets.Add(tsKey)
-			syncer.badTipSets.AddChain(chain)
 			return nil, err
 		}
 
@@ -244,6 +252,12 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 	// if tipset is already head, we've been here before. do nothing.
 	if head.Equals(next.ToSortedCidSet()) {
 		return nil
+	}
+
+	for _, b := range next.ToSlice() {
+		if err := syncer.consensus.ValidateSemantic(ctx, parent.At(0), b); err != nil {
+			return err
+		}
 	}
 
 	// Lookup parent state. It is guaranteed by the syncer that it is in
@@ -431,6 +445,7 @@ func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids typ
 	// Walk the chain given by the input blocks back to a known tipset in
 	// the store. This is the only code that may go to the network to
 	// resolve cids to blocks.
+	// collectChain performs block syntax validation
 	chain, err := syncer.collectChain(ctx, tipsetCids)
 	if err != nil {
 		return err
--- a/chain/default_syncer.go
+++ b/chain/default_syncer.go
@@ -193,6 +193,8 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 		// TODO: if we didn't get these blocks from the network and instead got
 		// them from disk, validation feels redundant. This requires
 		// looser coupling between the store and the fetcher.
+		// De-tangle syner and store: #2128
+		// Prevent Bad block propagation: #2783
 		for _, b := range blks {
 			if err := syncer.consensus.ValidateSyntax(ctx, b); err != nil {
 				syncer.badTipSets.Add(tsKey)
@@ -252,12 +254,6 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 	// if tipset is already head, we've been here before. do nothing.
 	if head.Equals(next.ToSortedCidSet()) {
 		return nil
-	}
-
-	for _, b := range next.ToSlice() {
-		if err := syncer.consensus.ValidateSemantic(ctx, parent.At(0), b); err != nil {
-			return err
-		}
 	}
 
 	// Lookup parent state. It is guaranteed by the syncer that it is in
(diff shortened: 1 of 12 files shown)


chain/default_syncer.go
```diff
@@ -190,10 +189,19 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
 			return nil, err
 		}
 
-		ts, err := syncer.consensus.NewValidTipSet(ctx, blks)
+		// validate blocks we maybe got from the network
+		// TODO: if we didn't get these blocks from the network and instead got
+		// them from disk, validation feels redundant. This requires
+		// looser coupling between the store and the fetcher.
```
anorth: Agreed, this should happen further out. If this is the pragmatic place to put it right now, then ok.

When pushed out to the network, a syntactically invalid block won't even make it as far as badTipSets.

Don't add a TODO without an issue, but here are two prepared earlier: #2128, #2783

chain/default_syncer.go
```diff
@@ -246,6 +254,12 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
 		return nil
 	}
 
+	for _, b := range next.ToSlice() {
```
anorth: Prefer indexed iteration with `Len()/At()` to avoid an unnecessary slice copy.

chain/default_syncer.go
```diff
@@ -246,6 +254,12 @@ func (syncer *DefaultSyncer) syncOne(ctx context.Context, parent, next types.Tip
...
+		if err := syncer.consensus.ValidateSemantic(ctx, parent.At(0), b); err != nil {
```
anorth: I think @ZenGround0 was suggesting this check actually move into RunStateTransition, so this call becomes unnecessary. I have a slight preference for that too. When the syntax check moves out to the boundary, the net will be zero changes to syncer.

chain/default_syncer.go
```diff
@@ -431,6 +445,7 @@ func (syncer *DefaultSyncer) HandleNewTipset(ctx context.Context, tipsetCids typ
 	// Walk the chain given by the input blocks back to a known tipset in
 	// the store. This is the only code that may go to the network to
 	// resolve cids to blocks.
+	// collectChain performs block syntax validation
```
anorth: Remove, I think this comment will rot as we clean things up.

consensus/block_validation.go
```diff
@@ -0,0 +1,69 @@
...
+	ValidateSemantic(ctx context.Context, child, parent *types.Block) error
+}
+
+type BlockSyntaxValidator interface {
+	ValidateSyntax(ctx context.Context, blk *types.Block) error
+}
+
+type BlockValidationClock interface {
+	Now() int64
+	BlockTime() time.Duration
```
anorth: This is unused right now and boils down to a constant. I suggest removing it from the interface and making it a direct member of the DefaultBlockValidator.

The clock interface will then be more widely useable (later): perhaps rename `Now` to `EpochSeconds` for extensibility.

consensus/block_validation.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func NewDefaultBlockValidator(c BlockValidationClock) *DefaultBlockValidator {
+	return &DefaultBlockValidator{
+		clock: c,
+	}
+}
+
+func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child, parent *types.Block) error {
+	// TODO validate timestamp
```
anorth: File an issue for semantic validation and link to it. The spec is https://github.com/filecoin-project/specs/blob/master/validation.md (with clarification of syntax/semantics landing shortly).

Ditto for syntactic validation below.

consensus/expected_test.go
```diff
@@ -148,10 +92,10 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
 		minerPower := types.NewBytesAmount(1)
 		totalPower := types.NewBytesAmount(1)
 
-		ptv := testhelpers.NewTestPowerTableView(minerPower, totalPower)
-		exp := consensus.NewExpected(cistore, bstore, testhelpers.NewTestProcessor(), ptv, genesisBlock.Cid(), verifier)
+		ptv := th.NewTestPowerTableView(minerPower, totalPower)
+		exp := consensus.NewExpected(cistore, bstore, th.NewTestProcessor(), th.NewTestBlockValidator(), ptv, genesisBlock.Cid(), verifier)
 
-		pTipSet, err := exp.NewValidTipSet(ctx, []*types.Block{genesisBlock})
+		pTipSet, err := types.NewTipSet(genesisBlock)
```
anorth: FYI there exists RequireNewTipSet that checks the error

testhelpers/consensus.go
```diff
@@ -127,6 +127,21 @@ func (tbr *TestBlockRewarder) GasReward(ctx context.Context, st state.Tree, mine
 	return nil
 }
 
+// TestBlockValidator passes everything as valid
+type TestBlockValidator struct{}
```
anorth: I think we should be a bit more clear that this doesn't validate anything. I suggest `FakeBlockValidator` or `NilBlockValidator`. The former offers a bit of flexibility for later parameterising it by whether it will accept or reject.

consensus/block_validation.go
frrist: Done: #2886

testhelpers/consensus.go
frrist: Agreed, was following the pattern that's being used elsewhere in the file -- will go with FakeBlockValidator

consensus/block_validation.go
```diff
@@ -0,0 +1,87 @@
...
+
+// BlockSemanticValidator defines an interface used to validate a blocks
+// semantics.
+type BlockSemanticValidator interface {
+	ValidateSemantic(ctx context.Context, child, parent *types.Block) error
+}
+
+// BlockSyntaxValidator defines an interface used to validate a blocks
+// syntax.
+type BlockSyntaxValidator interface {
```
ZenGround0: nit: I don't see the value in splitting these interfaces up here instead of defining a single `BlockValidator` interface with the two functions.  But not a strong opinion.

consensus/block_validation.go
```diff
@@ -0,0 +1,87 @@
...
+// to validate blocks and uses the DefaultBlockValidationClock.
+func NewDefaultBlockValidator(blkTime time.Duration) *DefaultBlockValidator {
+	return &DefaultBlockValidator{
+		clock:     NewDefaultBlockValidationClock(),
+		blockTime: blkTime,
+	}
+}
+
+// ValidateSemantic validates a block is correctly derived from its parent.
+func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child, parent *types.Block) error {
```
ZenGround0: Blocking: `parent` cannot be a block.  When we fill this in with all of the needed validation we are going to need to operate on the whole parent which is a tipset.  For example we will need the min ticket.  As a more relevant example, we will need the min timestamp of the parent tipset to do the soft block time validation.

consensus/expected_test.go
```diff
@@ -32,68 +32,12 @@ func TestNewExpected(t *testing.T) {
...
-		ptv := testhelpers.NewTestPowerTableView(types.NewBytesAmount(1), types.NewBytesAmount(5))
-		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), ptv, types.SomeCid(), verifier)
+		ptv := th.NewTestPowerTableView(types.NewBytesAmount(1), types.NewBytesAmount(5))
+		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), th.NewFakeBlockValidator(), ptv, types.SomeCid(), verifier)
 		assert.NotNil(t, exp)
 	})
 }
 
-// TestExpected_NewValidTipSet also tests validateBlockStructure.
-func TestExpected_NewValidTipSet(t *testing.T) {
```
ZenGround0: 🔴 🎉

consensus/expected_test.go
```diff
@@ -110,22 +54,22 @@ func requireMakeBlocks(ctx context.Context, t *testing.T, pTipSet types.TipSet,
 
 		ownerPubKeys[i] = kis[i].PublicKey()
 
-		ownerActor := testhelpers.RequireNewAccountActor(t, types.NewZeroAttoFIL())
+		ownerActor := th.RequireNewAccountActor(t, types.NewZeroAttoFIL())
```
ZenGround0: Nice rename, it's the little things sometimes

chain/default_syncer.go
frrist: Awesome, thank you for that link -- will move this validation check to the fetcher.

chain/default_syncer.go
```diff
@@ -190,10 +189,21 @@ func (syncer *DefaultSyncer) collectChain(ctx context.Context, tipsetCids types.
...
+		for _, b := range blks {
+			if err := syncer.consensus.ValidateSyntax(ctx, b); err != nil {
+				syncer.badTipSets.Add(tsKey)
+				syncer.badTipSets.AddChain(chain)
+				return nil, err
+			}
+		}
+		ts, err := types.NewTipSet(blks...)
 		if err != nil {
-			syncer.badTipSets.Add(tsKey)
```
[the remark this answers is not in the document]
frrist: this file will be unchanged once the syntax check is moved to fetcher (in a coming commit on this PR)

consensus/block_validation.go
frrist: One advantage is that we can pass just the syntax validator to methods that need it, e.g:
```go
NewFetcher(ctx context.Context, bsrv bserv.BlockService, bv consensus.BlockSyntaxValidator) *Fetcher
```
There are other ways to accomplish this -- like defining a sub-interface in the `fetcher.go` file. As work in this areas continues I am opening to changing these interfaces as needed, for now will leave as is.

chain/default_syncer.go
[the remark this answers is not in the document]
frrist: flagging that the check on a blocks state root existing is now done in `net/fetcher.go`, this is why we no longer add blocks to the badTipSet cache.

consensus/block_validation.go
anorth: I think splitting is right. The interface definitions should move to the consumers as this matures.

PR #2894, opened 2019-06-05
move BlockTime to plumbing add BlockClock interface

frrist:
### What
This PR move BlockTime to a method on the plumbing api and adds an interface called `BlockClock`. In coming PR's this interface can be extended to return the current time (`time.Now()`) so that mining may timestamp blocks. This makes progress towards implementing a soft-block dealy.

### DIFF
--- a/consensus/block_validation.go
+++ b/consensus/block_validation.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/filecoin-project/go-filecoin/plumbing/clock"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -49,16 +50,14 @@ func (ebc *DefaultBlockValidationClock) EpochSeconds() uint64 {
 
 // DefaultBlockValidator implements the BlockValidator interface.
 type DefaultBlockValidator struct {
-	clock     BlockValidationClock
-	blockTime time.Duration
+	clock clock.BlockClock
 }
 
 // NewDefaultBlockValidator returns a new DefaultBlockValidator. It uses `blkTime`
 // to validate blocks and uses the DefaultBlockValidationClock.
-func NewDefaultBlockValidator(blkTime time.Duration) *DefaultBlockValidator {
+func NewDefaultBlockValidator(c clock.BlockClock) *DefaultBlockValidator {
 	return &DefaultBlockValidator{
-		clock:     NewDefaultBlockValidationClock(),
-		blockTime: blkTime,
+		clock: c,
 	}
 }
 
@@ -83,5 +82,5 @@ func (dv *DefaultBlockValidator) ValidateSyntax(ctx context.Context, blk *types.
 // BlockTime returns the block time the DefaultBlockValidator uses to validate
 /// blocks against.
 func (dv *DefaultBlockValidator) BlockTime() time.Duration {
-	return dv.blockTime
+	return dv.clock.BlockTime()
 }
--- a/mining/worker.go
+++ b/mining/worker.go
@@ -23,10 +23,6 @@ import (
 
 var log = logging.Logger("mining")
 
-// DefaultBlockTime is the estimated proving period time.
-// We define this so that we can fake mining in the current incomplete system.
-const DefaultBlockTime = 30 * time.Second
-
 // Output is the result of a single mining run. It has either a new
 // block or an error, mimicing the golang (retVal, error) pattern.
 // If a mining run's context is canceled there is no output.
--- a/mining/worker.go
+++ b/mining/worker.go
@@ -72,6 +72,10 @@ type MessageApplier interface {
 	ApplyMessagesAndPayRewards(ctx context.Context, st state.Tree, vms vm.StorageMap, messages []*types.SignedMessage, minerOwnerAddr address.Address, bh *types.BlockHeight, ancestors []types.TipSet) (consensus.ApplyMessagesResponse, error)
 }
 
+type workerPorcelainAPI interface {
+	BlockTime(ctx context.Context) time.Duration
+}
+
 // DefaultWorker runs a mining job.
 type DefaultWorker struct {
 	createPoSTFunc DoSomeWorkFunc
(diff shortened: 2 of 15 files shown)


consensus/block_validation.go
```diff
@@ -49,16 +50,14 @@ func (ebc *DefaultBlockValidationClock) EpochSeconds() uint64 {
 
 // DefaultBlockValidator implements the BlockValidator interface.
 type DefaultBlockValidator struct {
-	clock     BlockValidationClock
```
frrist: I believe we can replace the `BlockValidationClock` with the `BlockClock` interface once it matures a bit more (in the coming block validation PR)

mining/worker.go
```diff
@@ -72,6 +72,10 @@ type MessageApplier interface {
 	ApplyMessagesAndPayRewards(ctx context.Context, st state.Tree, vms vm.StorageMap, messages []*types.SignedMessage, minerOwnerAddr address.Address, bh *types.BlockHeight, ancestors []types.TipSet) (consensus.ApplyMessagesResponse, error)
 }
 
+type workerPorcelainAPI interface {
```
frrist: This _could_ be extended to replace other fields on the `DefaultWorker`, however that may be outside the scope of the PR.

node/node.go
```diff
@@ -378,8 +378,14 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 	// set up pinger
 	pingService := ping.NewPingService(peerHost)
 
+	// DO NOT MERGE this code smells like poo, this smell exists elsewhere,
+	// should have defaults set on it _before_ we get here?
+	if nc.Clock == nil {
```
frrist: I'm not a fan of this pattern, however since `--block-time` is a command line option I am unsure if there is a way around it. If there are no suggestions during review I will remove this smelly comment.

plumbing/clock/clock.go
```diff
@@ -0,0 +1,29 @@
+package clock
+
+import (
+	"time"
+)
+
+// BlockClock defines an interface for fetching the block time.
+type BlockClock interface {
```
frrist: This is the interface we can extend, adding a method like `EpochSeconds`. Miners will then call this method when adding a timestamp to the block.

consensus/block_validation.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/filecoin-project/go-filecoin/plumbing/clock"
```
ZenGround0: I think this is the first time plumbing has made its way into the consensus package.  I am interested in the long term implications of this change.  Does it mean consensus can't be part of plumbing itself?  Does it mean consensus will have a stronger pull towards being put on plumbing?

It seems like there is lack of intentional design around the interactions between "core components" and the plumbing injection things.  These are just general observations, no action required in this PR.

mining/worker.go
ZenGround0: I agree it could.  Furthermore I think the mining worker should eventually be code in the block miner protocol's `miner.go` and most of the dependencies it shares with the node should be accessed through porcelain.

I also agree all of this is out of scope of your PR.

node/node.go
ZenGround0: What about passing an integer parsed from the CLI (or the default) on the node config and passing that to a clock constructor during the `porcelain.New` call below?

node/node.go
```diff
@@ -1026,11 +1022,11 @@ func (node *Node) setupProtocols() error {
 	node.BlockMiningAPI = &blockMiningAPI
 
 	// set up retrieval client and api
-	retapi := retrieval.NewAPI(retrieval.NewClient(node.host, node.blockTime, node.PorcelainAPI))
```
ZenGround0: 🎉

ZenGround0: Note: `node.host` should be accessed on porcelain too in a perfect world. (just observation not relevant to this PR)

frrist:
>  I don't immediately see how a function that always returns the same value is better than a constant here, but no real objection.

Initially BlockTime had a [setter and a getter](https://github.com/filecoin-project/go-filecoin/pull/2894/files#diff-380e79adccbaf1276bf6c91cc551cd7fL760). Since the setter was never referenced anywhere I opted to remove it rather than add it to plumbing. In the event that we want a setter for this value, the code change to add it will be smaller if this is left as a getter.

Moving this to plumbing also allows the BlockTime value to be referenced in the `consensus/` package. Previously there existed a circular dep between `mining/` and `consensus/` that prevented this.

> I object to calling this a clock. A method that returns the same value every time you call it is just what a clock isn't. This returns a Duration while a clock will return something like a Time. This is just a configuration value getter. 

I agree that naming this thing `*Clock` is a bit confusing. In a follow on PR I add a method [`EpochSeconds`](https://github.com/filecoin-project/go-filecoin/pull/2899/files#diff-c5c90458f89d14085f0c637ff03d7f71R14) to this interface, it returns `time.Now()` and also has a [mock for tests](https://github.com/filecoin-project/go-filecoin/pull/2897/files#diff-c67c02ad9a5f2978a87f472ad45456aaR10). Together, both of these methods are used by the [`BlockValidator`](https://github.com/filecoin-project/go-filecoin/pull/2899/files#diff-66c31b19e7bd7764c065d832890ef781R44) to validate blocks. 

I can split this interface up and modify `BlockValidator` to accept them both as parameters.

anorth:
Depending on existing structure, it may be better to have the block validator take the duration configuration at construction, but the current time as a parameter to the `Validate()` method (if the caller needs a clock anyway, or it's easier to plumb there). Or, take both the configuration and clock at construction. Either is fine.

node/node.go
ingar: You leave the concrete implementation of `BlockClock` configurable here for testing?  Like for a FAST test or something?  If not I kind of agree with @ZenGround0.. you can just instantiate the `BlockClock` given the `block-time` option.

frrist:
Closing in favor of https://github.com/filecoin-project/go-filecoin/pull/2914

PR #2896, opened 2019-06-06
Make show deal test use FAST.  Increase default FAST block time to 50 ms

shannonwells:
1. converts a daemon tests to FAST tests for the `show deal` command.  
1. changes the interface of the FAST `ShowDeal` function to take a deal instead of a CID
1. increases the default block time from 1ms to 50ms because this was needed to get the tests passing; otherwise there wasn't enough time to accept the deal.

The reason for changing the default time was that  I ran into a bug which, when you pass a DaemonOpt to the NewEnvironment function, for some reason the repo/api file is never created. I spent just a little time trying to figure out why but then I figured out a workaround, so for this PR I am using the workaround, as it's basically harmless anyway.  I filed an issue against this here: #2900

### DIFF
--- a/commands/show_test.go
+++ b/commands/show_test.go
@@ -1,14 +1,27 @@
 package commands_test
 
 import (
+	"bytes"
+	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"github.com/filecoin-project/go-filecoin/fixtures"
+	"github.com/filecoin-project/go-filecoin/proofs"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
 	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+	files "github.com/ipfs/go-ipfs-files"
+	"github.com/multiformats/go-multihash"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"io"
+	"math/big"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
+	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fastenvOpts)
 
-	minerDaemon.RunSuccess("mining", "start")
-	minerDaemon.UpdatePeerID()
+	clientNode := env.GenesisMiner
+	require.NoError(t, clientNode.MiningStart(ctx))
 
-	minerDaemon.ConnectSuccess(clientDaemon)
+	minerNode := env.RequireNewNodeWithFunds(1000)
+	defer func() {
```
frrist: BLOCKING: This should be called directly after the `NewTestEnvironment` call otherwise if `require.NoError(t, clientNode.MiningStart(ctx))` errors the defer function will never be called.

commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
+		require.NoError(t, env.Teardown(ctx))
+	}()
 
-	addAskCid := minerDaemon.MinerSetPrice(fixtures.TestMiners[0], fixtures.TestAddresses[0], "20", "10")
+	// Connect the clientNode and the minerNode
+	err := series.Connect(ctx, clientNode, minerNode)
```
frrist: nit: you could wrap this in the `NoError` call -- up to you.

commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
+	price := big.NewFloat(0.000000001)      // price per byte/block
+	expiry := big.NewInt(24 * 60 * 60 / 30) // ~24 hours
 
-		proposeDealOutput := clientDaemon.RunSuccess("client", "propose-storage-deal", fixtures.TestMiners[0], dataCid, "0", "5").ReadStdoutTrimNewlines()
+	ask, err := series.CreateStorageMinerWithAsk(ctx, minerNode, collateral, price, expiry)
+	require.NoError(t, err)
 
-		splitOnSpace := strings.Split(proposeDealOutput, " ")
+	// Create some data that is the full sector size and make it autoseal asap
+	var data bytes.Buffer
```
frrist: do we end up using this for anything, at first glance this looks like its not required.

commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
-		th.KeyFile(fixtures.KeyFilePaths()[1]),
-		th.DefaultAddress(fixtures.TestAddresses[1]),
-	).Start()
-	defer clientDaemon.ShutdownSuccess()
+	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fastenvOpts)
 
-	minerDaemon.RunSuccess("mining", "start")
-	minerDaemon.UpdatePeerID()
+	clientNode := env.GenesisMiner
+	require.NoError(t, clientNode.MiningStart(ctx))
```
travisperson: We need to figure out what we want to do about registering the mining protocol without running MiningStart. It'd be nice if we could avoid starting it.

tools/fast/action_show.go
```diff
@@ -23,11 +23,11 @@ func (f *Filecoin) ShowBlock(ctx context.Context, ref cid.Cid) (*types.Block, er
 }
 
 // ShowDeal runs the `show deal` command against the filecoin process
-func (f *Filecoin) ShowDeal(ctx context.Context, ref cid.Cid) (*storagedeal.Deal, error) {
+func (f *Filecoin) ShowDeal(ctx context.Context, dealResp *storagedeal.Response) (*storagedeal.Deal, error) {
```
travisperson: Do we want to pass in the entire `storagedeal.Response` here?

commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
-		assert.Contains(t, showdeal, "Duration: 5 blocks")
-		assert.Contains(t, showdeal, "State: staged")
-		assert.Contains(t, showdeal, "Size: 12 bytes")
-		assert.Contains(t, showdeal, "Total Price: 1200 FIL")
+	t.Run("showDeal outputs correct information", func(t *testing.T) {
+		showDeal, err := clientNode.ShowDeal(ctx, deal)
+		require.NoError(t, err)
+
+		assert.Equal(t, ask.Miner.String(), showDeal.Miner.String())
+		assert.Equal(t, "accepted", showDeal.Response.State.String())
```
travisperson: These both can be checked against there value and don't need to be converted to strings first.

commands/show_test.go
shannonwells: I got it from another test -- I think it was to cause autoseal to happen.

commands/show_test.go
```diff
@@ -59,48 +71,69 @@ func TestBlockDaemon(t *testing.T) {
...
+
+		prFIL := types.NewAttoFIL(big.NewInt(10000000000))
+		totalPrice := prFIL.MulBigInt(big.NewInt(maxBytesi64))
```
travisperson: Can be calculate this value? I'm not sure where the `10000000000` comes from at the moment.

Currently `ImportAndStore` has a hardcoded `10` for the expiry, but we can add that as an argument. With the expiry and the ask we should be able to determine the price.

tools/fast/action_show.go
shannonwells: I initially had it take the CID and that would match the other function. I'm fine with changing it back.

commands/show_test.go
shannonwells: good catch, thanks

commands/show_test.go
travisperson: I believe it was there just to tee the data off into so we could keep a copy of the random data around to assert on later. I don't believe you need to to call `io.TeeReader` here as I don't believe the test actually cares about the contents of the piece being stored.

PR #2897, opened 2019-06-06
Feat/timestamp block

frrist:
### What
This PR adds the timestamp field to block. This PR must be merged _after_ #2914

### DIFF
--- a/commands/schema/filecoin_block.schema.json
+++ b/commands/schema/filecoin_block.schema.json
@@ -64,6 +64,9 @@
             "string",
             "null"
           ]
+        },
+        "timestamp": {
+          "type": "string"
         }
       },
       "required": [
@@ -74,7 +77,8 @@
         "parents",
         "proof",
         "stateRoot",
-        "ticket"
+        "ticket",
+        "timestamp"
       ],
       "type": "object"
     },
--- a/types/block.go
+++ b/types/block.go
@@ -52,6 +52,9 @@ type Block struct {
 	// a challenge
 	Proof PoStProof `json:"proof"`
 
+	// The timestamp, in seconds since the Unix epoch, at which this block was created.
+	Timestamp Uint64 `json:"timestamp"`
+
 	cachedCid cid.Cid
 
 	cachedBytes []byte
--- a/types/block_test.go
+++ b/types/block_test.go
@@ -72,11 +72,12 @@ func TestTriangleEncoding(t *testing.T) {
 			ParentWeight:    Uint64(1000),
 			Proof:           NewTestPoSt(),
 			StateRoot:       SomeCid(),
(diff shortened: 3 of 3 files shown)


types/tipset_test.go
```diff
@@ -153,7 +154,12 @@ func TestTipSet(t *testing.T) {
 		// String shouldn't really need testing, but some existing code uses the string as a
 		// datastore key and depends on the format exactly.
 		assert.Equal(t, "{ "+b1.Cid().String()+" }", RequireNewTipSet(t, b1).String())
-		assert.Equal(t, "{ "+b1.Cid().String()+" "+b2.Cid().String()+" "+b3.Cid().String()+" }",
+
+		// DONOTMETGE the below assertion needed to be fixed after adding a timestamp to block.
```
frrist: @anorth could you confirm my suspicion here, blame shows you as the author. Does the order of these matter?

plumbing/clock/testing.go
```diff
@@ -0,0 +1,52 @@
+package clock
+
+import (
+	"time"
+
+	"github.com/benbjohnson/clock"
```
ZenGround0: I noticed from looking that the issues that this isn't actively maintained and has some open issues about races (they might actually be addressed but issues are open).  Probably fine but something to keep in mind if we start seeing races in tests using it.

ZenGround0: Looks like the de-racing PR got closed down without merging and the author recommended this project: https://github.com/jonboulle/clockwork.

frrist: Ohh nice find! I remember looking at that issue but didn't see the recommendation, will address and bubble up the changes to the validation branch.

types/tipset_test.go
anorth: Yes, order matters (from SortedCidSet)

anorth: You can robustify the test by actually using a sortedcidset to order them.

ingar: This is fixed in #2908

plumbing/clock/testing.go
```diff
@@ -0,0 +1,52 @@
...
+// MockBlockClock defines a testing structure used to mock a BlockClock.
+type MockBlockClock struct {
+	blockClock *clock.Mock
+	blockTime  time.Duration
+}
+
+// NewMockBlockClock returns a new MockBlockClock with the provided blockTime.
+// Calls to MockBlockClock.EpochSeconds() will return 0 until a time is set using
+// the method SetClock().
+func NewMockBlockClock(blkTime time.Duration) *MockBlockClock {
```
anorth: I suggest taking the mock time value in this constructor, saving most users a call to `SetClock`

PR #2908, opened 2019-06-10
Normalize `types.AttoFIL`

ingar:
## Problem
`AttoFIL` is a struct wrapping a pointer to a [`big.Int`](https://golang.org/pkg/math/big/#Int).  Its usage in the code base is confusing and inconsistent.  In most cases, it is passed as as a pointer to its methods, and the methods are defined on a pointer receiver, even though the receiver is never modified and usually the parameters aren't either.

Additionally, an `AttoFIL` with a value of `0` has 3 representations:
1. `AttoFIL{}` (The zero value of an `AttoFIL`)
1. `AttoFIL{val: &big.Int{}}` (With a pointer to the zero value of a `big.Int`)
1. `AttoFIL{val: big.NewInt(0)}` (With a pointer to a zero-valued `big.Int`)

This can cause some serialization/deserialization asymmetry.

## Solution
Pass `AttoFIL`s by value always; they can be treated as immutable values.

Also, embed a `big.Int` in the `AttoFIL` instead of wrapping just a `*big.Int`.  This is the most questionable part of this PR.  It increases the size of the struct from just a pointer to a pointer, a bool, and a slice.  However, this has some nice benefits:

1. We don't have [check for `nil`](https://github.com/filecoin-project/go-filecoin/blob/742c9836f43479de8a14bdd5c5b5e27ca34b0870/types/atto_fil.go#L173) in every method that does calculations.
1. Serialization "symmetry" when serializing `AttoFIL{}`.

### DIFF
--- a/types/atto_fil.go
+++ b/types/atto_fil.go
@@ -58,6 +58,10 @@ func (z AttoFIL) MarshalJSON() ([]byte, error) {
 // AttoFIL represents a signed multi-precision integer quantity of
 // attofilecoin (atto is metric for 10**-18). The zero value for
 // AttoFIL represents the value 0.
+//
+// Reasons for embedding a big.Int instead of *big.Int:
+//   - We don't have check for nil in every method that does calculations.
+//   - Serialization "symmetry" when serializing AttoFIL{}.
 type AttoFIL struct{ val big.Int }
 
 // NewAttoFIL allocates and returns a new AttoFIL set to x.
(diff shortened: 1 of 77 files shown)


ingar:
I just remembered there is an unresolved issue with `types/tipset_test.go`, in which a test is failing.  I believe this is a faulty test, but I would like to talk to @anorth about this.  I have a minimal failing example up in 714156086b107ac770eb5fdbbcafd07762261724.  Here are the [changes](https://github.com/filecoin-project/go-filecoin/compare/af634cbb...tipset-block-order-doesnt-match-sorted-cid-order).  This should probably be a separate issue.

types/atto_fil.go
```diff
@@ -123,43 +102,39 @@ func NewAttoFILFromFILString(s string) (*AttoFIL, bool) {
...
 
 // Sub sets z to the difference x-y and returns z.
-func (z *AttoFIL) Sub(y *AttoFIL) *AttoFIL {
-	ensureZeroAmounts(&z, &y)
-	newVal := big.NewInt(0)
-	newVal.Sub(z.val, y.val)
-	newZ := AttoFIL{val: newVal}
-	return &newZ
+func (z AttoFIL) Sub(y AttoFIL) AttoFIL {
+	newZ := AttoFIL{}
```
ZenGround0: so much clearer

types/atto_fil.go
```diff
@@ -169,76 +144,65 @@ func (z *AttoFIL) DivCeil(y *AttoFIL) *AttoFIL {
...
 }
 
 // CalculatePrice treats z as a price in AttoFIL/Byte and applies it to numBytes to calculate a total price.
-func (z *AttoFIL) CalculatePrice(numBytes *BytesAmount) *AttoFIL {
-	ensureZeroAmounts(&z)
+func (z AttoFIL) CalculatePrice(numBytes *BytesAmount) AttoFIL {
 	ensureBytesAmounts(&numBytes)
 	unitPrice := z
 
 	newVal := big.NewInt(0)
```
ZenGround0: nit: For consistency with other operations consider `newVal := AttoFIL{}`

types/tipset_test.go
```diff
@@ -149,13 +149,13 @@ func TestTipSet(t *testing.T) {
...
 	})
 
-	t.Run("string", func(t *testing.T) {
-		// String shouldn't really need testing, but some existing code uses the string as a
-		// datastore key and depends on the format exactly.
-		assert.Equal(t, "{ "+b1.Cid().String()+" }", RequireNewTipSet(t, b1).String())
-		assert.Equal(t, "{ "+b1.Cid().String()+" "+b2.Cid().String()+" "+b3.Cid().String()+" }",
-			RequireNewTipSet(t, b3, b2, b1).String())
-	})
+	//t.Run("string", func(t *testing.T) {
```
ZenGround0: FYI: @frrist's related comments in PR [here](https://github.com/filecoin-project/go-filecoin/pull/2897#discussion_r291278340)

vm/vm.go
```diff
@@ -73,9 +73,9 @@ func Transfer(fromActor, toActor *actor.Actor, value *types.AttoFIL) error {
 		return errors.Errors[errors.ErrInsufficientBalance]
 	}
 
-	if toActor.Balance == nil {
-		toActor.Balance = types.NewZeroAttoFIL()
-	}
+	//if toActor.Balance == nil {
```
ZenGround0: This can probably be deleted all the way

types/atto_fil.go
```diff
@@ -71,37 +58,29 @@ func (z AttoFIL) MarshalJSON() ([]byte, error) {
 // AttoFIL represents a signed multi-precision integer quantity of
 // attofilecoin (atto is metric for 10**-18). The zero value for
 // AttoFIL represents the value 0.
-type AttoFIL struct{ val *big.Int }
+type AttoFIL struct{ val big.Int }
```
anorth: Copy your words describing the tradeoff between embedding the bigint directly here, vs as a pointer, from the PR description. Save the next person the effort of retracing your steps.

types/atto_fil.go
anorth: I don't think it would be "too clever" to use a named return value to reduce these to on-liners.

types/tipset_test.go
anorth: We do need this test. It can be made more robust by using a SortedCidSet explicitly to construct the "expected" string.

PR #2910, opened 2019-06-11
Refactor/list deals command

rosalinekarr:
# Problem

The `go-filecoin client list-deals` and `go-filecoin miner list-deals` commands are largely redundant, and we ultimately intend to reorganize several commands under a new `go-filecoin deals` command.

# Solution

Combine `go-filecoin client list-deals` and `go-filecoin miner list-deals` into a new `go-filecoin deals list` command.


commands/deals.go
```diff
@@ -0,0 +1,108 @@
...
+	},
+	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
+		minerAddress, _ := GetPorcelainAPI(env).ConfigGet("mining.minerAddress")
+		dealsCh, err := GetPorcelainAPI(env).DealsLs(req.Context)
+		if err != nil {
+			return err
+		}
+
+		if filterForMiner, _ := req.Options[minerOnly].(bool); filterForMiner {
+			dealsCh = filterDealChannel(dealsCh, func(deal *storagedeal.Deal) bool {
```
acruikshank: Creating a goroutine and channel here is over-complicated. The `filterDealChannel` function can be replaced with an if statement in the dealsCh loop using the filter function defined here or the one defined for the client.

commands/deals_daemon_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"/": "%s"
+	},
+	"state": 7
+}`, fixtures.TestMiners[0], dealCid)
+
+	t.Run("with no filters", func(t *testing.T) {
+		listClientDealsOutput := clientDaemon.RunSuccess("deals", "list").ReadStdoutTrimNewlines()
+		assert.Equal(t, expectedClientOutput, listClientDealsOutput)
+
+		listMinerDealsOutput := minerDaemon.RunSuccess("deals", "list").ReadStdoutTrimNewlines()
```
acruikshank: These tests are difficult to parse (although they are correct). We also have no examples that demonstrate that the filters can filter some deals while leaving others.

It would be better if the setup were to have daemon A make a deal with daemon B, and then have daemon B make a deal with daemon B. Then run all the tests against just daemon A.

commands/deals.go
```diff
@@ -1,24 +1,97 @@
...
+	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
+		minerAddress, _ := GetPorcelainAPI(env).ConfigGet("mining.minerAddress")
+		dealsCh, err := GetPorcelainAPI(env).DealsLs(req.Context)
+		if err != nil {
+			return err
+		}
+
+		filterForMiner, _ := req.Options[minerOnly].(bool)
+		filterForClient, _ := req.Options[clientOnly].(bool)
+
```
shannonwells: does it break anything if you provide both "c" and "m"?

rosalinekarr: it doesn't error, but it returns no deals

commands/deals.go
```diff
@@ -1,24 +1,97 @@
...
+			if err = re.Emit(out); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	},
+	Type: DealsListResult{},
+	Encoders: cmds.EncoderMap{
+		cmds.Text: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, res *DealsListResult) error {
```
shannonwells: Also the output looks like:
```
{
        "minerAddress": "t2zcjccxr647t453wxsqr5yak7qin75ho2wxds54y",
        "pieceCid": {
                "/": "QmP2n11wtNtGEQzkDwLMoacRptB6uH7zhvPfqgheWqAWnY"
        },
        "proposalCid": {
                "/": "zDPWYqFCvAHa76rdfVX1MtXicsgDhd3EvhUMcFpYbEeCwh1tC7rT"
        },
        "state": 2
}
```
State = 2 doesn't really tell a human much, but I don't know what we're doing with enumerations like this. I kind of want the state to be converted to "Accepted" or whatever.

Issue #2850, opened 2019-05-29
Create `redeem-payment` command.


acruikshank:
### Description

We currently have no reasonable way for a miner to redeem the payment vouchers contained within a deal. When a miner invokes:

``` bash
go-filecoin deals redeem <deal id>
```

The command will

1. Find the deal
2. Return error if no deal is found or deal contains no vouchers
3. Find the first voucher with the greatest value whose validAt is not less than the current block height
4. Return if no voucher is found
5. Return if the vouchers value is not greater than its payment channel's `AmmountRedeemed`
6. Send a `paych redeem` message with the voucher. If a `verifyPieceInclusion` condition is present, it will also submit the sector id and piece inclusion proof. 
7. Returns the CID of the submitted message.

### Acceptance criteria

1. The command exists and works.
2. Most of the command is implemented in porcelain.
3. The porcelain is well tested
4. The command has daemon tests written in FAST

PR #2911, opened 2019-06-11
Moved show deal command to deals show

rosalinekarr:
# Problem

We want to reorganize several commands under a new `go-filecoin deals` subcommand.

# Solution

Move `go-filecoin show deal` to `go-filecoin deals show`.


rosalinekarr:
@shannonwells, no, this came out of work on #2850. I can add an issue, if you think that makes sense.

Issue #2903, opened 2019-06-07
[bug]All nodes are offline, whether Filecoin devnet has crashed ？


zhengboowen:
All nodes are offline, whether Filecoin devnet has crashed ？

The website of Filecoin devnet shows that there are many nodes’ block height reached more than 1.5 million. I want to know if this is normal.

<img width="1001" alt="All nodes are offline, whether Filecoin devnet has crashed" src="https://user-images.githubusercontent.com/51158073/59104129-4a21af80-8963-11e9-9aaa-e4a6a24715a4.png">

rosalinekarr:
@zhengboowen, we're investigating this right now.

@mslipper @frrist, can you update here when we have more answers?

frrist:
Hey @zhengboowen, this behaviour is not normal and will be addressed in the next major release (0.3.*).

This was more than likely caused by a node encountering a chain that was greater than finality limit ahead, the fix for this [exists here](https://github.com/filecoin-project/go-filecoin/pull/2799).

PR #2912, opened 2019-06-11
add metric to count reorgs

frrist:
### What
This PR was created to help debug problems like #2903. It adds a metric called  `chain/reorg_count` to the filecoin prometheus endpoint that may be used to determine when and how frequently reorgs occur.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -13,11 +13,18 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/actor/builtin"
 	"github.com/filecoin-project/go-filecoin/consensus"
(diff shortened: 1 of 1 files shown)


chain/default_syncer.go
```diff
@@ -334,6 +341,7 @@ func (syncer *DefaultSyncer) logReorg(ctx context.Context, curHead, newHead type
 	if reorg {
 		dropped, added, err := ReorgDiff(curHead, newHead, commonAncestor)
 		if err == nil {
+			reorgCnt.Inc(ctx, 1)
```
travisperson: I think we should move this up a level. It looks like the reorg still occurs, and this check is only for calculating the diff.

PR #2913, opened 2019-06-12
Add IntSet implementation

ingar:
## Summary
`IntSet` implementation, primarily intended for manipulating and passing around groups of Sector Ids.

Supports the basic set operations: Union, Difference, and Intersection, which should take care of most of our uses cases.

This implementation pretty much follows the [design document](https://docs.google.com/document/d/1aE5a-QZojprMkig6IyYkwV1SrMDDLOnNBqNdnwpu9tk), the relevant part quoted here:
> Proposed data structure: use a golang-collections bitarray.  Don’t worry too much about efficient construction and set difference.  Most operations are fairly trivial.  We will KISS and not avoid iterating over members of the set.  We won’t do any clever run or range based operations.  This will of course change if this is seen to be a state machine processing bottleneck.

### DIFF
--- a/types/intset.go
+++ b/types/intset.go
@@ -11,31 +11,30 @@ type IntSet struct {
 
 // NewIntSet returns a new IntSet, optionally initialized with integers
 func NewIntSet(ints ...uint64) IntSet {
+	// We are ignoring errors from SetBit, GetBit since SparseBitArrays never return errors for those methods
 	out := IntSet{ba: bitarray.NewSparseBitArray()}
 	for _, i := range ints {
-		out.Add(i)
+		_ = out.ba.SetBit(i)
 	}
 	return out
 }
 
-// Add adds an integer to this IntSet
-func (is IntSet) Add(i uint64) {
-	// Ignoring errors as we are using SparseBitArrays, which never return errors
-	_ = is.ba.SetBit(i)
-}
-
-// Contains returns whether an integer exists in this IntSet
-func (is IntSet) Contains(i uint64) bool {
-	// Ignoring errors as we are using SparseBitArrays, which never return errors
+// Has returns whether an integer exists in this IntSet
+func (is IntSet) Has(i uint64) bool {
 	isSet, _ := is.ba.GetBit(i)
 	return isSet
 }
 
-// IsSubset returns true if every element in other is also in the receiver
-func (is IntSet) IsSubset(other IntSet) bool {
+// HasSubset returns true if every element in other is also in the receiver
+func (is IntSet) HasSubset(other IntSet) bool {
 	// The method we're calling here seems weird, but in bitarray (https://github.com/Workiva/go-datastructures/blob/f07cbe3f82ca2fd6e5ab94afce65fe43319f675f/bitarray/block.go#L97)
 	// "intersect" means, "wholly contained within"
 	return is.ba.Intersects(other.ba)
+}
+
+// Add returns a new IntSet, the result of adding i to the receiver
+func (is IntSet) Add(i uint64) IntSet {
+	return is.Union(NewIntSet(i))
 }
 
 // Union returns a new IntSet, the result of a set union of the receiver and other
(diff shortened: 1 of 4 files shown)


types/intset.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Add adds an integer to this IntSet
+func (is IntSet) Add(i uint64) {
+	// Ignoring errors as we are using SparseBitArrays, which never return errors
+	_ = is.ba.SetBit(i)
+}
+
+// Contains returns whether an integer exists in this IntSet
+func (is IntSet) Contains(i uint64) bool {
```
ZenGround0: `Contains` should operate on another IntSet (this wasn't clear from the design doc).  If you are interested in the motivation check out `submitPoSt` and the `ValidateFaultSet` specs here: https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost

types/intset.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/golang-collections/go-datastructures/bitarray"
+)
+
+// IntSet is a space-efficient set of uint64
+type IntSet struct {
+	ba bitarray.BitArray
+}
+
+// NewIntSet returns an new IntSet, optionally initialized with integers
```
ZenGround0: nit: "a new"

ingar:
@ZenGround0 I was talking to @acruikshank about this; he mentioned that the [specs](https://github.com/filecoin-project/specs/blob/6841949ccf6ecb50902128233464ce116357d006/data-structures.md#basic-type-encodings) specify a `BitField` as the data type, not an `IntSet`.  Do you think this is worth mentioning in the specs repo?

types/intset_test.go
```diff
@@ -0,0 +1,90 @@
...
+		assert.True(t, is.Contains(0xffffffffffff))
+		assert.False(t, is.Contains(2))
+	})
+
+	t.Run("IsSubset", func(t *testing.T) {
+		is0 := types.NewIntSet(1, 2, 3)
+		is1 := types.NewIntSet()
+
+		// {} ⊆ {1, 2, 3}
+		assert.True(t, is0.IsSubset(is1))
```
anorth: The phrasing given by this method name is confusing, suggesting the converse relationship "is subset of". It would be less ambiguous as `is0.HasSubset(is1) == true`

types/intset.go
```diff
@@ -0,0 +1,67 @@
...
+func (is IntSet) IsSubset(other IntSet) bool {
+	// The method we're calling here seems weird, but in bitarray (https://github.com/Workiva/go-datastructures/blob/f07cbe3f82ca2fd6e5ab94afce65fe43319f675f/bitarray/block.go#L97)
+	// "intersect" means, "wholly contained within"
+	return is.ba.Intersects(other.ba)
+}
+
+// Union returns a new IntSet, the result of a set union of the receiver and other
+func (is IntSet) Union(other IntSet) IntSet {
+	return IntSet{ba: is.ba.Or(other.ba)}
+}
```
anorth: For conceptual consistency, I think the API offered here would be better if it were all in a mutable style, or all immutable. If our expected usage requires a mutable structure for performance, we can go that way. This would mean `AddAll`, `RetainAll` and `RemoveAll` instead of union, intersection and difference (and probably need a `Copy()` too). Else an immutable structure with methods like these.

This is not blocking feedback, use your discretion.

ZenGround0:
> the specs specify a BitField as the data type, not an IntSet. Do you think this is worth mentioning in the specs repo?

@ingar good question.  I don't think it is too important right now.  I've been imagining that `Bitfield` is the name we give the abi type that wraps a byte slice containing a serialization of an intset.  If you'd rather call this type a `Bitfield` and name that parameter type something else (or just keep it as bytes) that is fine with me too.

This question gets close to the fuzzy boundary between what is implementation specific and what is in the spec.  Specifically right now the spec conflates the `Bitfield` input parameter and the state machine's in memory representation of it so go-filecoin needs two names for something with one name in the spec.  Being more precise about this throughout the spec would be good but probably isn't the right thing to focus on right now.

types/intset_test.go
ingar: Oh yeah, that reads much better.

types/intset.go
ingar: Done, lets try out an immutable style

PR #2914, opened 2019-06-12
move blocktime onto consensus and plumb

frrist:
- This commit moves the blocktime default to the consensus package and
adds plumbing and porcelain cals to access it.

### DIFF
--- a/consensus/expected.go
+++ b/consensus/expected.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-hamt-ipld"
@@ -56,6 +57,11 @@ type TicketSigner interface {
 	GetAddressForPubKey(pk []byte) (address.Address, error)
 	SignBytes(data []byte, signerAddr address.Address) (types.Signature, error)
 }
+
+// DefaultBlockTime is the estimated proving period time.
+// We define this so that we can fake mining in the current incomplete system.
+// We also use this to enforce a soft block validation.
+const DefaultBlockTime = 30 * time.Second
 
 // TODO none of these parameters are chosen correctly
 // with respect to analysis under a security model:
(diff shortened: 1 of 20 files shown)


consensus/expected.go
```diff
@@ -57,6 +58,10 @@ type TicketSigner interface {
 	SignBytes(data []byte, signerAddr address.Address) (types.Signature, error)
 }
 
+// DefaultBlockTime is the estimated proving period time.
+// We define this so that we can fake mining in the current incomplete system.
```
ingar: This comment is a little confusing to me.  `DefaultBlockTime` is defined for more than just faking the VDF, right?  Also, is the terminology "proving period" correct?  I always associate that with a storage miner.

porcelain/protocol.go
```diff
@@ -10,16 +11,18 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// ProtocolParams TODO(rosa)
+// ProtocolParams contains paramethers that modify the filecoin nodes protocol
```
ingar: spelling of `paramethers`

consensus/protocol.go
```diff
@@ -38,4 +39,7 @@ type Protocol interface {
 
 	// ValidateSemantic validates a block is correctly derived from its parent.
 	ValidateSemantic(ctx context.Context, child *types.Block, parents *types.TipSet) error
+
+	// BlockTime returns the block time used by the consensus protocol.
+	BlockTime(ctx context.Context) time.Duration
```
ingar: I am unclear on where we draw the line on what methods require a `Context` and which don't, seems like all of them?

frrist: Good point, the method returns a constant, I should not have added context as a parameter here, will remove.

consensus/expected.go
frrist: I think when the comment was written we assumed that this constant would go away, since then the spec has change s.t. we need it. Will update.

consensus/protocol.go
acruikshank: There is a case to be made that all interface methods that retrieve state should take a context. The context is primarily used to cleanly cancel IO calls and other code that's not entirely synchronous. So a method failing to take a context leaks implementation details. The same logic would require all methods to return an error. The general pattern on this project is thread through contexts when needed and leave them when the need goes away. So this context is ok but not strictly necessary.

consensus/expected.go
anorth: I'm pretty sure it's nothing to do with PoSt proving period.

PR #2915, opened 2019-06-12
update repo directory to latest path

ognots:
the devnets have changed repo directory structure. currently if the node gets restarted, the restart check will never evaluate to true


ognots:
see https://github.com/filecoin-project/go-filecoin-infra/issues/254

PR #2917, opened 2019-06-12
Refactor/remove unnecessary syncer interface 2916

### LINKED ISSUE #2916: Remove Java-style Syncer interface
### Description

Just like we removed the Store interface in #2800, we should remove `chain/syncer.go` as it disagrees with our current style regarding defining interfaces.

rosalinekarr:
# Problem

The `chain/syncer.go` file disagrees with our current style regarding defining interfaces.

# Solution

Kill it.

Resolves #2916


rosalinekarr:
This is just a quick refactor that I've been meaning to knock out. Not a high priority or anything.

PR #2919, opened 2019-06-12
create fast environment package

frrist:
### Wat
This PR repackages the way FAST is laid out to make it more friendly to work with deps
This PR adds `GetFunds` to the environment interface and modifies DevNet and MemoryGenesis to impl it.

### DIFF
--- a/tools/fast/environment/environment.go
+++ b/tools/fast/environment/environment.go
@@ -1,4 +1,4 @@
-package fast
+package environment
 
 import (
 	"context"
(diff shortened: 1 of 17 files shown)


tools/fast/environment/environment.go
```diff
@@ -53,17 +48,20 @@ type Environment interface {
...
 	// Teardown runs anything that the environment may need to do to
 	// be nice to the the execution area of this code.
 	Teardown(context.Context) error
 
 	// TeardownProcess runs anything that the environment may need to do
 	// to remove a process from the environment in a clean way.
-	TeardownProcess(context.Context, *Filecoin) error
+	TeardownProcess(context.Context, *fast.Filecoin) error
+
+	// GetFunds retrieves a fixed amount of tokens from the environment.
```
anorth: This needs more explanation. The method only returns an error, so what does "get" mean? From where are they got? Where is the fixed amount specified? Are they guaranteed to have landed when this method returns?

Edit: having now read implementation, it would probably be better to be more explicit and call this `RequestFaucetFunds` or similar. The questions above still stand.

PR #2924, opened 2019-06-13
Show more information by '--messages' flag when show block

waynewyang:
I think we should show messages in block when run 'go-filecoin show block', then we can know the history messages, including transaction and other messages.

 
```
waynewyang:go-filecoin waynewyang$ ./go-filecoin show block zDPWYqFCwxAV66BT3jnYZMVWNZvayYf95EjjK6f8n46n9AVsReSD
Block Details
Miner:  t2ej7i72xtkglfpbo7nc56pqg2fsdzlspqe63vnoq
Weight: 0.000
Height: 1
Nonce:  0
Timestamp:  0
Message:  [SignedMessage cid=[zDPWYqFCz3f9sTvKp5W1u6Vo57u2xH4T7Nicv2Xx8jz1zoQ1UL1v]: {
  "meteredMessage": {
    "message": {
      "to": "t2ej7i72xtkglfpbo7nc56pqg2fsdzlspqe63vnoq",
      "from": "t1strybge3p4daa6qgtz5jsfxalzg4mr6bix5s2ti",
      "nonce": "2",
      "value": "0",
      "method": "updatePeerID",
      "params": "gVgiEiDbXbBmGs1XgNWvOhNU42yekJfMBSr7U4fZYPsDbHxqFQ=="
    },
    "gasPrice": "0.000000001",
    "gasLimit": "300"
  },
  "signature": "mpnLn2DQ2kbObOrG2z0nfihYHUUnEUl8E85NjZ6Ok6RCgucH2cUSOppN2f96uYc3By+L0S/rSrVwyxeKgZofcAE="
} SignedMessage cid=[zDPWYqFCyU6B4qiJCBQ1k8pDP16KJDueJUMm1jtM7QAGEEkvYc3u]: {
  "meteredMessage": {
    "message": {
      "to": "t1cgp7xhkvsarve2mm32jidxnlbxuxgqjqcvya2qa",
      "from": "t1strybge3p4daa6qgtz5jsfxalzg4mr6bix5s2ti",
      "nonce": "3",
      "value": "100",
      "method": "",
      "params": null
    },
    "gasPrice": "0.000000001",
    "gasLimit": "300"
  },
  "signature": "Djb6b8K8tmwZAPBrlaLqkC0aaLfCoXLqyYV+YafAy09R/skyGm0tFIzOTLflt0ppmHi5cQWB3l64f0IEDkI9TgE="
}]
```

### DIFF
--- a/commands/show.go
+++ b/commands/show.go
@@ -31,6 +31,9 @@ all other block properties will be included as well.`,
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("cid", true, false, "CID of block to show"),
 	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("messages", "m", "show messages in block"),
+	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
 		cid, err := cid.Decode(req.Arguments[0])
 		if err != nil {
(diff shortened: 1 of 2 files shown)


anorth:
Blocks may have many, many messages and this could make the output hard to digest for someone only after the header. Would you be open to adding a flag to switch this on? E.g. `--messages` (timestamp is good to include always).

Note that in the future, messages will not be directly inlined in blocks, and must be fetched separately (referenced by CID of a merkle tree), so this option would prepare for that.

waynewyang:
@anorth Thank you, I've added a flag named 'messages' .

waynewyang:
@shannonwells I had added the test case, thanks.

PR #2925, opened 2019-06-13
Changed the deals redeem command test to be less flaky

rosalinekarr:
This is a follow up to #2850. I didn't notice the math bug until after I merged to master.

### DIFF
--- a/commands/deals_daemon_test.go
+++ b/commands/deals_daemon_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/tools/fast"
 	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
 	"github.com/filecoin-project/go-filecoin/tools/fast/series"
+	"github.com/filecoin-project/go-filecoin/types"
 )
 
 func TestDealsRedeem(t *testing.T) {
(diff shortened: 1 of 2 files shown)


commands/deals_daemon_test.go
```diff
@@ -70,6 +71,9 @@ func TestDealsRedeem(t *testing.T) {
 	newWalletBalance, err := minerDaemon.WalletBalance(ctx, minerOwnerAddress)
 	require.NoError(t, err)
 
+	// InDelta is necessary because sometimes the big.Int math can be off by 2 attoFIL
```
travisperson: This sounds like a bug someplace. Why can it sometimes be off by 2 attoFIL?

rosalinekarr: not sure. at first i thought it was something in `math/big`, but now i'm beginning to suspect is may have something to do with gas costs.

travisperson: Hmm okay. I think we should merge this if it's causing flakes, but we should follow up with an issue to root cause the issue. The 2 attoFIL is just a wearly small number.

acruikshank: At this point, gas cost should be fully deterministic (and never less than 100 AttoFIL). The value of the deal is determined by the file size, the price in the ask, and the duration of the deal. All of these should be constant for this test and wouldn't account for this small of a delta. We're probably doing some double arithmetic and getting some round off error, but I can't explain why it would be happening inconsistently.

rosalinekarr: Maybe it has something to do with a race condition around `MiningStop`? I stop mining before getting the initial wallet balance to prevent the miner from getting a block reward which can change the difference in it's wallet balance, but unfortunately, `MiningStop` is an async command with no way of checking when it's done, so there could be some kind of race condition here.

rosalinekarr:
Closing since this doesn't seem to resolve the flakiness

PR #2929, opened 2019-06-13
Refactor/extract config from plumbing 2928

### LINKED ISSUE #2928: Remove config package from plumbing
### Description

Plumbing is intended to be a thin wrapper around core services, so the logic present in `plumbing/cfg` disagrees with our stated goals around architecture in the project. We need to extract logic from this package into `config` and `repo` and remove it from `plumbing`.

rosalinekarr:
# Problem

We don't want config logic in plumbing. It should be thin wrappers.

# Solution

Extract `plumbing/cfg` logic to `config` and `repo` and remove the package.

Resolves #2928

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -292,6 +295,9 @@ func ReadFile(file string) (*Config, error) {
 // Set sets the config sub-struct referenced by `key`, e.g. 'api.address'
 // or 'datastore' to the json key value pair encoded in jsonVal.
 func (cfg *Config) Set(dottedKey string, jsonString string) error {
+	cfg.lock.Lock()
+	defer cfg.lock.Unlock()
```
anorth: Moving the lock here is good. `Get()` should be protected too.

rosalinekarr:
abandoning this...

PR #2931, opened 2019-06-13
Bug/deals redeem flaky test

rosalinekarr:
# Problem

I accidentally introduced flakiness in #2850 

# Solution

Fix it

### DIFF
--- a/commands/deals_daemon_test.go
+++ b/commands/deals_daemon_test.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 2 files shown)


rosalinekarr:
I've been running tests on this branch in a loop all morning, and it's passed every time, so I'm confident that this resolves the issue.

travisperson:
Sleeping is fixing the the 2 attoFIL issue?

rosalinekarr:
@travisperson, it seems so. I have no idea why, but I ran tests 10 times this morning without a failure.

PR #2934, opened 2019-06-14
Refactor/remove message send with default address

rosalinekarr:
# Problem

We want to eventually remove the message package from plumbing, but the `MessageSendWithDefaultAddress` porcelain currently depends on it. Additionally, we've stated that `MessageSendWithDefaultAddress` shouldn't exist because we should check for non-provided parameters in the commands layer rather than accepting nil values to porcelain methods.

# Solution

Extract default addresses to the commands layer and remove `MessageSendWithDefaultAddress`


commands/from_address.go
```diff
@@ -1,17 +1,18 @@
...
 	"github.com/filecoin-project/go-filecoin/address"
 )
 
-func optionalAddr(o interface{}) (ret address.Address, err error) {
-	if o != nil {
-		ret, err = address.NewFromString(o.(string))
-		if err != nil {
-			err = errors.Wrap(err, "invalid from address")
-		}
+func fromAddr(req *cmds.Request, env cmds.Environment) (address.Address, error) {
```
acruikshank: How about "fromAddrOrDefault" to make the result of this method clear.

commands/optional_address.go
```diff
@@ -0,0 +1,17 @@
+package commands
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/filecoin-project/go-filecoin/address"
+)
+
+func optionalAddr(o interface{}) (ret address.Address, err error) {
```
acruikshank: Can this be inlined? Do we call it from outside from_address? I don't like having a file per helper function.

rosalinekarr: This is used on lines 374 & 413 of `commands/miner.go` as well as line 150 of `commands/payment_channel.go` for addresses other than the `fromAddr`.

PR #2944, opened 2019-06-17
Iterate messages in canonical tipset order when removing from outbox

### LINKED ISSUE #2310: Extract tipset canonical block ordering and use in all tipset traversals

anorth:
This is the last item in #2310, ensuring that messages are removed from the outbox in the order they were mined (i.e. correct nonce order).

Closes #2310

### DIFF
--- a/core/inbox.go
+++ b/core/inbox.go
@@ -3,9 +3,10 @@ package core
 import (
 	"context"
 
+	"github.com/ipfs/go-cid"
+
 	"github.com/filecoin-project/go-filecoin/chain"
 	"github.com/filecoin-project/go-filecoin/types"
-	"github.com/ipfs/go-cid"
 )
 
 // InboxMaxAgeTipsets is maximum age (in non-empty tipsets) to permit messages to stay in the pool after reception.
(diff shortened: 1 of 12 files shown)


core/inbox.go
```diff
@@ -58,31 +59,37 @@ func (ib *Inbox) Pool() *MessagePool {
...
+			for _, msg := range block.Messages {
+				_, err = ib.pool.Add(ctx, msg, uint64(block.Height))
+				if err != nil {
+					log.Info(err)
+				}
 			}
 		}
 	}
 
 	// Remove all messages in the new blocks from the pool, now mined.
```
shannonwells: Shouldn't this comment "Remove all messages in the new blocks" be changed since it's operating on the TipSet now?

core/policy.go
```diff
@@ -50,28 +50,28 @@ func NewMessageQueuePolicy(store chain.BlockProvider, maxAge uint64) *DefaultQue
...
-	_, newBlocks, err := CollectBlocksToCommonAncestor(ctx, p.store, oldHead, newHead)
+	_, newTips, err := CollectTipsToCommonAncestor(ctx, p.store, oldHead, newHead)
 	if err != nil {
 		return err
 	}
 
 	// Remove from the queue all messages that have now been mined in new blocks.
 
-	// Rearrange the blocks in increasing height order so messages are discovered in order.
-	// Note: this is imperfect until CollectBlocksToCommonAncestor is updated to return blocks
```
shannonwells: What about this note?  Is it still relevant, has it been fixed, or is there a story?

shannonwells: I didn't realize the "Note" part was part of it, ok.

core/policy_test.go
```diff
@@ -150,6 +152,45 @@ func TestMessageQueuePolicy(t *testing.T) {
...
+		b1.Ticket = []byte{0}
+		b2.Ticket = []byte{1}
+		b2.Timestamp = 0 // Tweak to force CID ordering to be opposite to ticket ordering.
+		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) > 0)
+
+		err := policy.HandleNewHead(ctx, q, requireTipset(t, root), requireTipset(t, b1, b2))
+		require.NoError(t, err)
+		assert.Empty(t, q.List(alice))
+
+		// If the blocks were ordered the other way, this would fail.
```
ZenGround0: This comment is confusing.  I think that this test would fail at the require on line 181 if the blocks were ordered in the other way.  `HandleNewHead` would have called `RemoveNext` on `b2` first and `RemoveNext` would return [this error](https://github.com/filecoin-project/go-filecoin/blob/master/core/message_queue.go#L91).  Putting this comment here makes it look like the test would fail on line 186 which it would never get to.

ZenGround0: Oh I understand what you mean now.  One way to make this clearer might be to add a comment above line 181 like "If the blocks are ordered `b1, b2` then this passes" and then have this comment.

You could also change this comment to say "*When* the blocks are ordered in reverse, this *should* fail" to make it clear that the "failure" you are talking about is captured in the code block below.

PR #2952, opened 2019-06-17
rename some deal states

### LINKED ISSUE #2806: Reconcile storagedeal.State with deal states in the specification
### Description

Deal states in `protocol/storage/state` don't match the deal states in the spec. They need to.

### Acceptance Criteria

1. Delete any state in state.go that aren't used and aren't in the spec.
2. Rename states to match spec.
3. If any states still exist that don't exist in the spec, fix the spec.

acruikshank:
closes #2806

### Problem

The names we use for deal states are inconsistent with [the spec](https://github.com/filecoin-project/specs/blob/ed6f64914edc4fef47fc287a68305b7cd1e7afb8/network-protocols.md#deal-state-values)

### Solution

1. Add `Unset` state to distinguish unknown deals from deals with their state unset
2. Rename `Posted` to `Complete`.
3. Sync comments.

### DIFF
--- a/protocol/storage/storagedeal/state.go
+++ b/protocol/storage/storagedeal/state.go
@@ -8,7 +8,7 @@ import (
 type State int
 
 const (
-	// This implies a programmer error and should never appear in an actual message.
+	// Unset indicates a programmer error and should never appear in an actual message.
 	Unset = State(iota)
 
 	// Unknown signifies an unknown negotiation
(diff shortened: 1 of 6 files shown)


protocol/storage/storagedeal/state.go
```diff
@@ -46,12 +47,10 @@ func (s State) String() string {
...
 	case Failed:
 		return "failed"
-	case Posted:
-		return "posted"
-	case Complete:
-		return "complete"
 	case Staged:
 		return "staged"
+	case Complete:
+		return "posted"
```
anorth: blocking: `"complete"`

acruikshank: whoops

PR #2953, opened 2019-06-17
Use ProvingPeriodEnd instead of ProvingPeriodStart

### LINKED ISSUE #2806: Reconcile storagedeal.State with deal states in the specification

acruikshank:
closes #2806

# Problem

`ProvingPeriodStart` is referred to as `ProvingPeriodEnd` in [the spec](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector). We need to store the ProvingPeriodEnd. We also need to consolidate the computation of the proving period into the state machine as much as possible.

# Solution

1. Rename `ProvingPeriodStart` to `ProvingPeriodEnd`
2. Compute `ProvingPeriodEnd` as the block height + the proving period in `CommitSector`
3. Change `GetProvingPeriodEnd` to `GetProvingPeriod` and return the start and end of the proving period.
4. Modify protocol/storage/miner to use start and end from GetProvingPeriodEnd

This PR also adds unit testing for `storage/miner.OnNewHeaviestTipSet`. Making storagee/miner more testable, adding more utilities for testing and adding the tests themselves account for most of the bulk of the PR.

I made an attempt early in the PR to move the chain randomness retrieval in storage miner into actor. There are early commits with this code. Since this value must be computed the same way during proof generation as in verification, I feel pretty strongly that this is the right thing to do. Unfortunately, when ancestors were wired through to message processing, they weren't also wired into query messaging. Doing that work is well outside the scope of this PR, so I backed out.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -146,8 +146,8 @@ type State struct {
 
 	LastUsedSectorID uint64
 
-	ProvingPeriodStart *types.BlockHeight
-	LastPoSt           *types.BlockHeight
+	ProvingPeriodEnd *types.BlockHeight
+	LastPoSt         *types.BlockHeight
 
 	// The amount of space committed to the network by this miner.
 	Power *types.BytesAmount
@@ -557,7 +557,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		state.ActiveCollateral = state.ActiveCollateral.Add(collateral)
 
 		if state.Power.Equal(types.NewBytesAmount(0)) {
-			state.ProvingPeriodStart = ctx.BlockHeight()
+			state.ProvingPeriodEnd = ctx.BlockHeight()
 		}
 		inc := state.SectorSize
 		state.Power = state.Power.Add(inc)
@@ -737,9 +737,8 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 		}
 
 		// Calcuate any penalties for late submission
-		provingPeriodEnd := state.ProvingPeriodStart.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
 		generationAttackGracePeriod := GenerationAttackTime(state.SectorSize)
-		if chainHeight.GreaterThan(provingPeriodEnd.Add(generationAttackGracePeriod)) {
+		if chainHeight.GreaterThan(state.ProvingPeriodEnd.Add(generationAttackGracePeriod)) {
 			// The PoSt has been submitted after the generation attack time.
 			// The miner can expect to be slashed, and so for now the PoSt is rejected.
 			// An alternative would be to apply the penalties here, duplicating the behaviour
@@ -748,7 +747,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 				generationAttackGracePeriod)
 		}
 
-		feeRequired := LatePoStFee(state.ActiveCollateral, provingPeriodEnd, chainHeight, generationAttackGracePeriod)
+		feeRequired := LatePoStFee(state.ActiveCollateral, state.ProvingPeriodEnd, chainHeight, generationAttackGracePeriod)
 
 		// The message value has been added to the actor's balance.
 		// Ensure this value fully covers the fee which will be charged to this balance so that the resulting
@@ -809,7 +808,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 		}
 
 		// transition to the next proving period
-		state.ProvingPeriodStart = provingPeriodEnd
+		state.ProvingPeriodEnd = state.ProvingPeriodEnd.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
 		state.LastPoSt = chainHeight
 
 		return nil, nil
@@ -821,7 +820,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 	return 0, nil
 }
 
-// GetProvingPeriodStart returns the current ProvingPeriodStart value.
+// GetProvingPeriodStart returns the current ProvingPeriodEnd value.
 func (ma *Actor) GetProvingPeriodStart(ctx exec.VMContext) (*types.BlockHeight, uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
@@ -837,7 +836,7 @@ func (ma *Actor) GetProvingPeriodStart(ctx exec.VMContext) (*types.BlockHeight, 
 		return nil, errors.CodeError(err), err
 	}
 
-	return state.ProvingPeriodStart, 0, nil
+	return state.ProvingPeriodEnd, 0, nil
 }
 
 //
@@ -917,7 +916,7 @@ func LatePoStFee(pledgeCollateral types.AttoFIL, provingPeriodEnd *types.BlockHe
 //
 
 func currentProvingPeriodPoStChallengeSeed(ctx exec.VMContext, state State) (types.PoStChallengeSeed, error) {
-	bytes, err := ctx.SampleChainRandomness(state.ProvingPeriodStart)
+	bytes, err := ctx.SampleChainRandomness(state.ProvingPeriodEnd)
 	if err != nil {
 		return types.PoStChallengeSeed{}, err
 	}
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -266,10 +266,6 @@ var minerExports = exec.Exports{
 		Params: nil,
 		Return: []abi.Type{abi.BytesAmount},
 	},
-	"getPoStChallengeSeed": &exec.FunctionSignature{
-		Params: nil,
-		Return: []abi.Type{abi.PoStChallengeSeed},
-	},
 	"getProvingPeriod": &exec.FunctionSignature{
 		Params: []abi.Type{},
 		Return: []abi.Type{abi.BlockHeight, abi.BlockHeight},
@@ -783,7 +779,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 		//
 		// This switching will be removed when issue #2270 is completed.
 		if !ma.Bootstrap {
-			seed, _, err := GetPoStChallengeSeed(ctx)
+			seed, err := getPoStChallengeSeed(ctx, state)
 			if err != nil {
 				return nil, errors.RevertErrorWrap(err, "failed to sample chain for challenge seed")
 			}
@@ -868,27 +864,17 @@ func (ma *Actor) GetProvingPeriod(ctx exec.VMContext) (*types.BlockHeight, *type
 	return provingPeriodStart(state), state.ProvingPeriodEnd, 0, nil
 }
 
-// GetPoStChallengeSeed(ctx ex
-func GetPoStChallengeSeed(ctx exec.VMContext) (types.PoStChallengeSeed, uint8, error) {
-	chunk, err := ctx.ReadStorage()
-	if err != nil {
-		return types.PoStChallengeSeed{}, errors.CodeError(err), err
-	}
-
-	var state State
-	if err := actor.UnmarshalStorage(chunk, &state); err != nil {
-		return types.PoStChallengeSeed{}, errors.CodeError(err), err
-	}
-
+// getPoStChallengeSeed returns some chain randomness
+func getPoStChallengeSeed(ctx exec.VMContext, state State) (types.PoStChallengeSeed, error) {
 	randomness, err := ctx.SampleChainRandomness(provingPeriodStart(state))
 	if err != nil {
-		return types.PoStChallengeSeed{}, 1, err
+		return types.PoStChallengeSeed{}, err
 	}
 
 	seed := types.PoStChallengeSeed{}
 	copy(seed[:], randomness)
 
-	return seed, 0, nil
+	return seed, nil
 }
 
 //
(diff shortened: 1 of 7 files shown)


actor/builtin/miner/miner.go
```diff
@@ -793,7 +793,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 	return 0, nil
 }
 
-// GetProvingPeriodStart returns the current ProvingPeriodStart value.
+// GetProvingPeriodStart returns the current ProvingPeriodEnd value.
 func (ma *Actor) GetProvingPeriodStart(ctx exec.VMContext) (*types.BlockHeight, uint8, error) {
```
laser: BLOCKING: This method should be called `GetProvingPeriodEnd`.

actor/builtin/miner/miner.go
```diff
@@ -553,7 +553,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		state.ActiveCollateral = state.ActiveCollateral.Add(collateral)
 
 		if state.Power.Equal(types.NewBytesAmount(0)) {
-			state.ProvingPeriodStart = ctx.BlockHeight()
+			state.ProvingPeriodEnd = ctx.BlockHeight()
```
laser: BLOCKING: If you're going to replace `ProvingPeriodStart` with `ProvingPeriodEnd`, you should set  `ProvingPeriodEnd` to be equal to `ctx.BlockHeight() + ProvingPeriodDuration(miner.SectorSize)` as per the spec.

actor/builtin/miner/miner.go
```diff
@@ -739,7 +739,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 		}
 
 		// Check if we submitted it in time
-		provingPeriodEnd := state.ProvingPeriodStart.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
+		provingPeriodEnd := state.ProvingPeriodEnd.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
 		if ctx.BlockHeight().GreaterThan(provingPeriodEnd.Add(GenerationAttackTime(state.SectorSize))) {
```
laser: BLOCKING: This should be modified to be:

```
if ctx.BlockHeight().GreaterThan(state.ProvingPeriodEnd.Add(GenerationAttackTime(state.SectorSize))) {
```

actor/builtin/miner/miner.go
anorth: Must add proving period duration to block height to compute the end (copy from 742)

actor/builtin/miner/miner.go
```diff
@@ -809,11 +809,11 @@ func (ma *Actor) GetProvingPeriodStart(ctx exec.VMContext) (*types.BlockHeight,
 		return nil, errors.CodeError(err), err
 	}
 
-	return state.ProvingPeriodStart, 0, nil
+	return state.ProvingPeriodEnd, 0, nil
 }
 
 func currentProvingPeriodPoStChallengeSeed(ctx exec.VMContext, state State) (types.PoStChallengeSeed, error) {
-	bytes, err := ctx.SampleChainRandomness(state.ProvingPeriodStart)
+	bytes, err := ctx.SampleChainRandomness(state.ProvingPeriodEnd)
```
laser: BLOCKING: Suppose a miner completes the PoSt generation routine some number of blocks N before the end of their proving period. If N is greater than `sampling.LookbackParameter`, then there will exist no block with a matching height from which to sample randomness from.

actor/builtin/miner/miner.go
```diff
@@ -146,8 +146,8 @@ type State struct {
 
 	LastUsedSectorID uint64
 
-	ProvingPeriodStart *types.BlockHeight
-	LastPoSt           *types.BlockHeight
+	ProvingPeriodEnd *types.BlockHeight
```
anorth: Please comment the semantics. I assume that with the name changes comes the semantic meaning "the end of the current proving period" (as opposed to the end of the prior proving period).

actor/builtin/miner/miner.go
```diff
@@ -739,7 +739,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (u
 		}
 
 		// Check if we submitted it in time
-		provingPeriodEnd := state.ProvingPeriodStart.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
+		provingPeriodEnd := state.ProvingPeriodEnd.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
```
anorth: Remove this calculation here, it goes at 784 instead

actor/builtin/miner/miner.go
anorth: Rename the method

actor/builtin/miner/miner.go
anorth: Probably necessary to subtract the proving period duration here to retain semantics.

actor/builtin/miner/miner.go
```diff
@@ -811,17 +819,42 @@ func (ma *Actor) GetProvingPeriodEnd(ctx exec.VMContext) (*types.BlockHeight, ui
...
+
+	var state State
+	if err := actor.UnmarshalStorage(chunk, &state); err != nil {
+		return nil, nil, errors.CodeError(err), err
+	}
+
+	return provingPeriodStart(state), state.ProvingPeriodEnd, 0, nil
+}
+
+// GetPoStChallengeSeed(ctx ex
```
anorth: borked comment

actor/builtin/miner/miner.go
```diff
@@ -269,6 +269,14 @@ var minerExports = exec.Exports{
 		Params: nil,
 		Return: []abi.Type{abi.BytesAmount},
 	},
+	"getPoStChallengeSeed": &exec.FunctionSignature{
+		Params: nil,
+		Return: []abi.Type{abi.PoStChallengeSeed},
+	},
+	"getProvingPeriod": &exec.FunctionSignature{
```
anorth: If adding this, delete `getProvingPeriodEnd` above.

protocol/storage/miner.go
```diff
@@ -748,18 +748,18 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) {
...
 }
 
-func (sm *Miner) getProvingPeriodEnd() (*types.BlockHeight, error) {
+func (sm *Miner) getProvingPeriod() (*types.BlockHeight, *types.BlockHeight, error) {
 	res, err := sm.porcelainAPI.MessageQuery(
 		context.Background(),
 		address.Undef,
 		sm.minerAddr,
-		"getProvingPeriodEnd",
+		"isInProvingPeriod",
```
anorth: Blocking: This looks like something from an interim attempt. LGTM if changed to `getProvingPeriod`.

protocol/storage/miner.go
```diff
@@ -706,70 +700,61 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) {
...
 	}
 
 	// the block height of the new heaviest tipset
 	h := types.NewBlockHeight(height)
 
-	// compute the block height at which the miner's current proving period ends
-	provingPeriodEnd := provingPeriodStart.Add(types.NewBlockHeight(miner.ProvingPeriodDuration(sectorSize)))
-
 	if h.GreaterEqual(provingPeriodStart) {
 		if h.LessThan(provingPeriodEnd) {
```
laser: BLOCKING: There is a bug here - not one which you introduced. I think that if the miner submits their PoSt when the block height _equals_ the end of the proving period that their submission should not be penalized. Do you disagree?

protocol/storage/miner.go
```diff
@@ -580,17 +579,12 @@ func (sm *Miner) onCommitFail(ctx context.Context, dealCid cid.Cid, message stri
 		resp.Message = message
 		resp.State = storagedeal.Failed
 	})
-	log.Errorf("commit failure but could not update to deal 'Failed' state: %s", err)
+	sm.log.Errorf("commit failure but could not update to deal 'Failed' state: %s", err)
 }
 
 // currentProvingPeriodPoStChallengeSeed produces a PoSt challenge seed for
 // the miner actor's current proving period.
```
laser: Given that you added the `currentProvingPeriodStart` parameter, this comment is no longer true. Rather, it should be rewritten to say:

```
// currentProvingPeriodPoStChallengeSeed produces a PoSt challenge 
// using randomness from the chain at the provided block height
```

protocol/storage/miner.go
```diff
@@ -580,17 +579,12 @@ func (sm *Miner) onCommitFail(ctx context.Context, dealCid cid.Cid, message stri
...
-func (sm *Miner) currentProvingPeriodPoStChallengeSeed(ctx context.Context) (types.PoStChallengeSeed, error) {
-	currentProvingPeriodStart, err := sm.getProvingPeriodStart()
-	if err != nil {
-		return types.PoStChallengeSeed{}, errors.Wrap(err, "error obtaining current proving period")
-	}
-
+func (sm *Miner) currentProvingPeriodPoStChallengeSeed(ctx context.Context, currentProvingPeriodStart *types.BlockHeight) (types.PoStChallengeSeed, error) {
```
laser: STRONGLY-WORDED SUGGESTION: This method's name should be changed to `poStChallengeSeedAtBlockHeight`, or you should drop the `currentProvingPeriodStart` parameter and instead, in the body of the method, make a call to `getProvingPeriod`.

protocol/storage/miner.go
```diff
@@ -706,70 +700,61 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) {
...
+				sm.log.Errorf("error obtaining challenge seed: %s", err)
 				return
 			}
 
 			go sm.submitPoSt(provingPeriodStart, provingPeriodEnd, seed, inputs)
 		} else {
 			// we are too late
 			// TODO: figure out faults and payments here
-			log.Errorf("too late start=%s  end=%s current=%s", provingPeriodStart, provingPeriodEnd, h)
+			sm.log.Errorf("too late start=%s  end=%s current=%s", provingPeriodEnd, provingPeriodEnd, h)
```
laser: BLOCKING: The first non-format-string argument should be `provingPeriodStart`, not `provingPeriodEnd`.

protocol/storage/miner_test.go
```diff
@@ -297,18 +304,319 @@ func TestOnCommitmentAddedToChain(t *testing.T) {
...
+
+		// return true, indicating this miner is a bootstrap miner
+		api.messageHandlers["isBootstrapMiner"] = func(a address.Address, v types.AttoFIL, p ...interface{}) ([][]byte, error) {
+			return [][]byte{}, errors.New("test error")
+		}
+
+		miner.OnNewHeaviestTipSet(types.TipSet{})
+
+		require.Equal(t, 1, len(logger.ErrorMessages))
+		assert.Contains(t, logger.ErrorMessages[0], "bootstrapping")
```
laser: SUGGESTION: If your intent is to exercise the invariant-checking code, I suggest that you modify `OnNewHeaviestTipSet` to return `error` (it's a void method now). Then, you can make assertions on the existence and message contained within any error returned from said method.

protocol/storage/miner_test.go
```diff
@@ -297,18 +304,319 @@ func TestOnCommitmentAddedToChain(t *testing.T) {
...
+	return handlers
+}
+
+type testNode struct{}
+
+func (tn *testNode) BlockService() bserv.BlockService           { return nil }
+func (tn *testNode) Host() host.Host                            { return nil }
+func (tn *testNode) SectorBuilder() sectorbuilder.SectorBuilder { return &testSectorBuilder{} }
+
+type testSectorBuilder struct{}
```
laser: OBSERVATION: This is a lot of test code, which is unfortunate because only two of the methods are consumed by the storage miner (`AddPiece` and `ReadPieceFromSealedSector`). 

I suggest that you modify the `storage.node` interface such that

```
type node interface {
	BlockService() bserv.BlockService
	Host() host.Host
	SectorBuilder() sectorbuilder.SectorBuilder
}
```

becomes

```
type node interface {
	BlockService() bserv.BlockService
	Host() host.Host
	SectorBuilder() someNewInterfaceWithJustWhatStorageMinerNeeds
}

type someNewInterfaceWithJustWhatStorageMinerNeeds interface {
    AddPiece(...)
    ReadPieceFromSealedSector(...)
}
```

You'd need to do the same thing in the retrieval miner code.

protocol/storage/miner_test.go
```diff
@@ -297,18 +304,319 @@ func TestOnCommitmentAddedToChain(t *testing.T) {
...
+
+	for i, value := range values {
+		encoded, err := value.Serialize()
+		require.NoError(t, err)
+		out[i] = encoded
+	}
+	return out
+}
+
+type testLogger struct {
```
laser: OBSERVATION: I think that all of this test code goes away if you change `OnNewHeaviestTipSet` to return an `error`.

protocol/storage/miner.go
laser: A PoSt is going to take a long time to generate, so I guess this doesn't matter.

acruikshank: I think the ProvingPeriodEnd should be exclusive so proving periods don't overlap. The spec seems to disagree with me. As you mention here, it's mostly irrelevant, because it's very unlikely that PoSt generation could start on the last block of the proving period and make it on chain in the same block. I'll change it anyway.

protocol/storage/miner_test.go
acruikshank: Good suggestion. Done.

protocol/storage/miner_test.go
acruikshank: This is a good idea, but out of scope for this PR.

protocol/storage/miner_test.go
acruikshank: It did.

Issue #2900, opened 2019-06-06
[bug] Providing DaemonOpts to fastesting.NewTestEnvironment somehow causing repo/api not found


shannonwells:
## Describe the bug
When creating a FAST test, I attempted to pass a DaemonOpt for increasing the block time to `fastesting.NewTestEnvironment`.  This consistently will cause the following error:

```Could not get PeerID, non-zero exit code```

but when you turn on logging with `log.SetDebugLogging()`, you'll see why:
```
13:16:45.796 WARNI  util: open  /var/folders/j0/foo...bar/T/TestReproduceFASTFailure613116831/0/repo/api: no such file or directory util.go:27
```
One of these messages occurs with the default options, i.e. passing in `fast.FilecoinOpts{}`, but adding an option results in this file never being found at all, so it looks like the api is for some reason never created, or maybe it's looking in the wrong directory or ?

## To Reproduce
Use the following code to reproduce the behavior. By commenting out the `DaemonOpts` line (and removing the "time" import), the test will pass, leaving it will fail.
```
func TestReproduceFASTFailure(t *testing.T){
	tf.IntegrationTest(t)
	log.SetDebugLogging()

	fastOpts := fast.FilecoinOpts{
		DaemonOpts: []fast.ProcessDaemonOption{fast.POBlockTime(time.Second)},
	}

	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fastOpts)

	clientNode := env.GenesisMiner
	require.NoError(t, clientNode.MiningStart(ctx))
}
```

## Expected
Opts passed to NewTestEnvironment should be successfully merged with default settings, and also should not prevent the test from running.

### Reproduces: 100%

PR #2955, opened 2019-06-17
Feat/add payment voucher show deal 2844

### LINKED ISSUE #2844: Make `show deal` output include deals payment vouchers
### Description
request from @acruikshank via @rosalinekarr : include a deals payment vouchers in the output from  show deals.

### Acceptance criteria
Add deals payment vouchers info to the show deal output

### Where to begin
commands/show.go

shannonwells:
Closes #2844 Make `deals show` command show payment voucher info, and for each voucher includes the encoded voucher string.  Because there was extra info requested for vouchers, I created output structs.

Other changes:
* improved checking for payment_channel_daemon_test
* change output struct of deals show
* doubled the block time for FAST test environment to help with flaky Deals Show tests

Example output:
```
$ ./go-filecoin deals show zDPWYqFCx3AS8hVNJGyUaSh1wEYguQbnuRxurLM1mncFPYNdeYuT --repodir=~/.filecoin-localnet/0/repo
{
        "deal_cid": {
                "/": "zDPWYqFCx3AS8hVNJGyUaSh1wEYguQbnuRxurLM1mncFPYNdeYuT"
        },
        "state": 2,
        "miner_address": "t2f42g3ug5krmlzd2rb2ojsd5u26rs6l4r77qfefi",
        "duration_blocks": 10,
        "deal_size": "1016",
        "total_price": "0.00001016",
        "payment_vouchers": [
                {
                        "index": 0,
                        "amount": "0.00001016",
                        "channel_id": 4,
                        "condition": null,
                        "payer": "t1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy",
                        "valid_at_block": 13,
                        "encoded_as": "z2P729JDGqP3ATb7MAWa16kF8o4byGLy4yYTX4N4837RWxUSnfDyTC677EcTzF8rFXXYHB9xtwsquUZDhDj6o7W5J4G1B7GGk6ZVF1TMCAU8CvZ8Nubvwy6zADjmocUL3Se4kedykEkrMnSmYJqUH8TrAkCKzUKSLisNimAvKY33RtGFSCXEV7rGzHoPNvMCL9i8fMXebfuoBoB2EwxfCGGKCvZgap41Jk8h9E2QdSMEZwPbY2iHgXiwa"
                }
        ]
}
```

### DIFF
--- a/commands/deals_daemon_test.go
+++ b/commands/deals_daemon_test.go
@@ -3,13 +3,10 @@ package commands_test
 import (
 	"context"
 	"crypto/rand"
-	"fmt"
 	"io"
 	"math/big"
-	"strconv"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-files"
@@ -245,7 +242,7 @@ func TestDealsShowPaymentVouchers(t *testing.T) {
 	clientNode := env.GenesisMiner
 	require.NoError(t, clientNode.MiningStart(ctx))
 	defer func() {
-		assert.NoError(t, clientNode.MiningStop(ctx))
+		require.NoError(t, clientNode.MiningStop(ctx))
 	}()
 
 	minerNode := env.RequireNewNodeWithFunds(1000)
@@ -262,7 +259,7 @@ func TestDealsShowPaymentVouchers(t *testing.T) {
 	ask, err := series.CreateStorageMinerWithAsk(ctx, minerNode, collateral, price, expiry)
 	require.NoError(t, err)
 	defer func() {
-		assert.NoError(t, minerNode.MiningStop(ctx))
+		require.NoError(t, minerNode.MiningStop(ctx))
 	}()
 
 	// Create some data that is the full sector size and make it autoseal asap
@@ -278,9 +275,6 @@ func TestDealsShowPaymentVouchers(t *testing.T) {
 
 	_, deal, err := series.ImportAndStoreWithDuration(ctx, clientNode, ask, durationui64, files.NewReaderFile(dataReader))
 	require.NoError(t, err)
-
-	require.NoError(t, minerNode.MiningStop(ctx))
-	require.NoError(t, clientNode.MiningStop(ctx))
 
 	t.Run("Vouchers output as JSON have the correct info", func(t *testing.T) {
 		res, err := clientNode.DealsShow(ctx, deal.ProposalCid)
--- a/commands/deals_daemon_test.go
+++ b/commands/deals_daemon_test.go
@@ -128,12 +128,6 @@ func TestDealsList(t *testing.T) {
 	splitOnSpace := strings.Split(proposeDealOutput, " ")
 	dealCid := splitOnSpace[len(splitOnSpace)-1]
 
-	t.Run("with --help", func(t *testing.T) {
-		clientOutput := clientDaemon.RunSuccess("deals", "list", "--help").ReadStdoutTrimNewlines()
-		assert.Contains(t, clientOutput, "only return deals made as a client")
-		assert.Contains(t, clientOutput, "only return deals made as a miner")
-	})
-
 	t.Run("with no filters", func(t *testing.T) {
 		// Client sees the deal
 		clientOutput := clientDaemon.RunSuccess("deals", "list").ReadStdoutTrimNewlines()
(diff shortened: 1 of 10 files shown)


commands/payment_channel_daemon_test.go
```diff
@@ -150,14 +150,17 @@ func TestPaymentChannelVoucherSuccess(t *testing.T) {
 	chanid, _ := rsrc.requirePaymentChannel(ctx, t, channelAmount, channelExpiry)
 
 	voucherAmount := types.NewAttoFILFromFIL(10)
-	voucherValidAt := types.NewBlockHeight(0)
+	voucherValidAt := types.NewBlockHeight(1)
```
shannonwells: I noticed that not all the info was being checked as part of debugging the deals show test, and added more asserts.

tools/fast/series/import_and_store.go
```diff
@@ -13,16 +14,23 @@ import (
 
 // ImportAndStore imports the `data` to the `client`, and proposes a storage
 // deal using the provided `ask`, returning the cid of the import and the
-// created deal.
+// created deal. It uses a duration of 10 blocks
 func ImportAndStore(ctx context.Context, client *fast.Filecoin, ask porcelain.Ask, data files.File) (cid.Cid, *storagedeal.Response, error) {
+	return ImportAndStoreWithDuration(ctx, client, ask, 10, data)
+}
+
+// ImportAndStoreWithDuration imports the `data` to the `client`, and proposes a storage
```
shannonwells: This is to support the need in the deals show test to provide a duration so >1 payment voucher would be created.  I figured this would probably be useful so I added it here rather than copying almost all of the function and making it private to the test package.

types/payment_voucher.go
```diff
@@ -74,3 +75,12 @@ func (voucher *PaymentVoucher) Encode() (string, error) {
 
 	return multibase.Encode(multibase.Base58BTC, cborVoucher)
 }
+
+// SortVouchersByValidAt takes a list of payment vouchers and sorts them
+// in increasing value of ValidAt (block height)
+func SortVouchersByValidAt(sorted []*PaymentVoucher) []*PaymentVoucher {
```
shannonwells: This is part of the ask for the issue:  that the vouchers be sorted in increasing ValidAt block heights, and then indexed, and the index should be output as well.

commands/deals_daemon_test.go
```diff
@@ -180,27 +190,33 @@ func TestShowDeal(t *testing.T) {
...
+	defer func() {
+		require.NoError(t, minerNode.MiningStop(ctx))
+	}()
 
 	// Create some data that is the full sector size and make it autoseal asap
 
 	maxBytesi64 := int64(getMaxUserBytesPerStagedSector())
 	dataReader := io.LimitReader(rand.Reader, maxBytesi64)
+
+	// Duration is 10
```
acruikshank: It's not clear what this comment is referring to or why it's needed.

shannonwells: That was a note to myself when I was trying to do some debugging. I'll take it out.  It's the default duration for ImportAndStore.

commands/deals_daemon_test.go
```diff
@@ -84,8 +90,11 @@ func TestDealsRedeem(t *testing.T) {
 	newWalletBalance, err := minerDaemon.WalletBalance(ctx, minerOwnerAddress)
 	require.NoError(t, err)
 
+	// this is to fix flaky test failures due to the amount being 11.8999999
```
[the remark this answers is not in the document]
shannonwells: Yes

tools/fast/fastesting/basic.go
```diff
@@ -61,7 +61,7 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.File
 
 	fastenvOpts = fast.FilecoinOpts{
 		InitOpts:   append([]fast.ProcessInitOption{fast.POGenesisFile(genesisURI)}, fastenvOpts.InitOpts...),
-		DaemonOpts: append([]fast.ProcessDaemonOption{fast.POBlockTime(50 * time.Millisecond)}, fastenvOpts.DaemonOpts...),
+		DaemonOpts: append([]fast.ProcessDaemonOption{fast.POBlockTime(100 * time.Millisecond)}, fastenvOpts.DaemonOpts...),
```
anorth: If this actually makes a difference then we are in a very bad world that I thought we had left. Any dependence on wall clock time is **fundamentally flaky**. Tests that pass or fail depending on the load on the machine they are running on are broken. Tests that rely on sequencing of events must use actual synchronisation primitives to ensure that such events have happened. Otherwise we will continually increase this time in a futile attempt to have stable tests, forever paying more and more dev time waiting for them to run and discouraging their use.

The `time.Sleep` in the test that we've disabled is an absolute no-go. That test needs to be rewritten for multiple reasons, of which that is one.

Which tests are sensitive to this block time? If it was only the deal redemption one, please revert this. If you feel it's necessary to commit this change, please file an issue that those tests depend on simulated block time and are a liability.

cc @acruikshank @travisperson

shannonwells: @anorth The story behind the change you have notated is: the original block time set across all FAST tests was 1 ms. The original deals show test was consistently failing when creating deals. The message wait was bailing out immediately. It appeared that the context had ended/disappeared before the deal was accepted on chain.  When I passed in a longer block time for the `fast.FilecoinOpts{}`, it failed for an entirely different reason, and an issue was filed for this (#2900).  To get the original test to pass I felt it was harmless to temporarily increase it to 50 ms, which passes locally but is flaky on CI.

These two integration tests for deals show are the only ones I know of sensitive to block time.

Given your comment above, and that the FAST environment setup is designed to allow you to pass in a block time, are you also saying that block time should not be configurable?

anorth: As discussed, only the `deals show` test has this dependency, so let's make it a parameter so all the other tests can be fast, and the problematic ones easy to identify.

shannonwells: Also it will come in a separate PR.

PR #2956, opened 2019-06-18
Add built-in burnt funds account actor to genesis

anorth:
As [specified](https://github.com/filecoin-project/specs/blob/master/actors.md#built-in-actors)

Also updated the other built-in actors to have ID-style addresses.

### DIFF
--- a/consensus/genesis.go
+++ b/consensus/genesis.go
@@ -27,9 +27,10 @@ var (
 
 func init() {
 	defaultAccounts = map[address.Address]types.AttoFIL{
-		address.NetworkAddress: types.NewAttoFILFromFIL(10000000000),
-		address.TestAddress:    types.NewAttoFILFromFIL(50000),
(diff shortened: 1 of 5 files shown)


consensus/genesis.go
```diff
@@ -210,26 +211,24 @@ func SetupDefaultActors(ctx context.Context, st state.Tree, storageMap vm.Storag
...
 		return err
 	}
 
-	pbAct := actor.NewActor(types.PaymentBrokerActorCodeCid, types.ZeroAttoFIL)
+	pbAct := paymentbroker.NewActor()
 	err = (&paymentbroker.Actor{}).InitializeState(storageMap.NewStorage(address.PaymentBrokerAddress, pbAct), nil)
 	if err != nil {
 		return err
 	}
+	err = st.SetActor(ctx, address.PaymentBrokerAddress, pbAct)
```
ZenGround0: nit: I think this could just be `return st.SetActor`

Issue #2960, opened 2019-06-18
[bug] TestDealsRedeem fails due to rounding error non-deterministically 


frrist:
**Describe the bug**
`TestDealsRedeem` is failing in CI due to what appears to be a rounding error.

**To Reproduce**
Run the `TestDealsRedeem` test. It is a non-deterministic failure, meaning you nay need to run this more than once.

**Expected behavior**
The test passes everytime.

**Screenshots**
```
Failed
deals_daemon_test.go:88: 
Error Trace:	deals_daemon_test.go:88
Error:      	Not equal: 
            	expected: "11.9"
            	actual  : "11.8999999999999998"
            	
            	Diff:
            	--- Expected
            	+++ Actual
            	@@ -1 +1 @@
            	-11.9
            	+11.8999999999999998
Test:       	TestDealsRedeem
log_writer.go:44: >>>> start-dump
log_writer.go:44: ---- command     go-filecoin client query-storage-deal zDPWYqFCvC1dkmvRfGkT7cikBPPyvgWu4yXjD98qt864Y8eDpQ8a --enc=json
log_writer.go:44: ---- exit-code   0
log_writer.go:44: ---- error       nil
log_writer.go:44: ---- stdout
log_writer.go:44: {"State":5,"Message":"","ProposalCid":{"/":"zDPWYqFCvC1dkmvRfGkT7cikBPPyvgWu4yXjD98qt864Y8eDpQ8a"},"ProofInfo":{"SectorID":1,"CommitmentMessage":{"/":"zDPWYqFD6txFXsghC6uHvTj6N1UcGVBy8dhMZhT2FfkLBJfELQDE"},"PieceInclusionProof":"EiDAatZGvn+eyPm7KNRJwqivm0N7cEiXXa3b7cVidIjOj7LnXHSl3rilpTB0y3iPbjl8G8URqk4deoNhVfyyDw=="},"Signature":"c2lnbmF0dXJycmVlZQ=="}
log_writer.go:44: ---- stderr
log_writer.go:44: <<<< end-dump
log_writer.go:44: >>>> start-dump
log_writer.go:44: ---- command     go-filecoin wallet balance t1kfxxwlc7jr2i4qrnuy53ecbzaqwrshcslcaatey --enc=json
log_writer.go:44: ---- exit-code   0
log_writer.go:44: ---- error       nil
log_writer.go:44: ---- stdout
log_writer.go:44: "9810.8999999999999998"
[...]

PR #2964, opened 2019-06-19
Disable flaky assertion in TestDealsRedeem FAST test

anorth:
We don't know the cause of this yet. It's a significant problem (#2960) but meanwhile this flaky test prevents us from merging PRs. Leaving it disabled is a louder warning that we have a problem than relaxing it to accept a rounding error.


anorth:
I am invoking moderator privilege to merge this based on in-person discussion earlier today, to unblock landing of other PRs.

Issue #2942, opened 2019-06-16
Miner pays fees for late PoSt submission


anorth:
### Description
A miner may submit a PoSt after the proving period ends, but before the generation attack time. In that case, it must pay a fee from pledge collateral. The spec currently also requires that that fee is covered by the value of the SubmitPoSt message.

See the [faults and slashing design doc](https://docs.google.com/document/d/1U3b9GVNVOLoS_-q9kXU3a9dqHGvcrHprwDFzNYgLI0M/edit?pli=1#heading=h.9vj78uxo22yu) and [SubmitPost spec](https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost)

The fee calculation is pending https://github.com/filecoin-project/specs/issues/328 and https://github.com/filecoin-project/specs/issues/329. If those are not yet resolved when implementing, calculate the fee as a proportion of pledge collateral equal to amount of lateness as a fraction of the maximum allowable lateness (the generation attack period).

See also fees paid in #2940.

### Acceptance criteria
- SubmitPost penalises late PoSt submissions
- A miner attaches message value to cover the late fee
- Unit tests for fee calculations

### Risks + pitfalls
The actor and miner worker must both calculate the same fee. Be careful finding a pattern here that doesn't introduce unwanted dependencies between actor and worker code.

PR #2966, opened 2019-06-19
StorageMiner.SubmitPoSt actor method charges late fees.

anorth:
The fee is a proportion of pledge collateral equal to the lateness as a fraction of the maximum allowable lateness. Proofs after the generation attack threshold are rejected (the miner actor can expect to be slashed).

The worker doesn't yet compute and attach any required fee, so late proofs will continue to be rejected. Part of the work for #2942.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -276,6 +276,10 @@ func (ma *Actor) Exports() exec.Exports {
 	return minerExports
 }
 
+//
+// Exported actor methods
+//
+
 // AddAsk adds an ask to this miners ask list
 func (ma *Actor) AddAsk(ctx exec.VMContext, price types.AttoFIL, expiry *big.Int) (*big.Int, uint8,
 	error) {
@@ -583,14 +587,6 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 	return 0, nil
 }
 
-// CollateralForSector returns the collateral required to commit a sector of the
-// given size.
-func CollateralForSector(sectorSize *types.BytesAmount) types.AttoFIL {
-	// TODO: Replace this function with the baseline pro-rata construction.
-	// https://github.com/filecoin-project/go-filecoin/issues/2866
-	return MinimumCollateralPerSector
-}
-
 // VerifyPieceInclusion verifies that proof proves that the data represented by commP is included in the sector.
 // This method returns nothing if the verification succeeds and returns a revert error if verification fails.
 func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, sectorID uint64, proof []byte) (uint8, error) {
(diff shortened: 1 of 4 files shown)


actor/builtin/miner/miner_test.go
```diff
@@ -514,6 +548,40 @@ func TestGetProofsMode(t *testing.T) {
...
+		assert.True(t, af(1).Equal(LatePoStFee(pledgeCollateral, bh(1000), bh(1010), bh(10000))))
+	})
+
+	t.Run("fee capped at total pledge", func(t *testing.T) {
+		assert.True(t, pledgeCollateral.Equal(LatePoStFee(pledgeCollateral, bh(1000), bh(1100), bh(100))))
+		assert.True(t, pledgeCollateral.Equal(LatePoStFee(pledgeCollateral, bh(1000), bh(2000), bh(100))))
+	})
+}
+
+func af(h int64) types.AttoFIL {
```
ZenGround0: I like these.  Wonder if they belong somewhere more generally accessible?

actor/builtin/miner/miner_test.go
```diff
@@ -216,7 +215,7 @@ func TestMinerGetPower(t *testing.T) {
 
 		st, vms := core.CreateStorages(ctx, t)
 
-		minerAddr := createTestMinerWith(120, types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress,
+		minerAddr := createTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress,
```
shannonwells: Why the choice of these numbers for the pledge (240 and 100 down below) and why 240 here (twice as much as before).  A comment saying that the number is enough to cover fees or not enough to cover fees or whatever the reason for that amount would be helpful.

actor/builtin/miner/miner_test.go
```diff
@@ -326,40 +325,75 @@ func TestMinerSubmitPoSt(t *testing.T) {
...
+	proof := th.MakeRandomPoStProofForTest()
+
+	miner := state.MustGetActor(st, minerAddr)
+	minerBalance := miner.Balance
+	owner := state.MustGetActor(st, address.TestAddress)
+	ownerBalance := owner.Balance
+
+	firstCommitBlockHeight := uint64(3)
+	secondProvingPeriodStart := LargestSectorSizeProvingPeriodBlocks + firstCommitBlockHeight
+	lastSubmission := secondProvingPeriodStart + LargestSectorSizeProvingPeriodBlocks + LargestSectorGenerationAttackThresholdBlocks
```
shannonwells: Did you intend to add `LargestSectorSizeProvingPeriodBlocks` twice to obtain lastSubmission?

actor/builtin/miner/miner.go
```diff
@@ -867,3 +890,52 @@ func GenerationAttackTime(sectorSize *types.BytesAmount) *types.BlockHeight {
 func ProvingPeriodDuration(sectorSize *types.BytesAmount) uint64 {
 	return LargestSectorSizeProvingPeriodBlocks
 }
+
+// LatePostFee calculates the fee from pledge collateral that a miner must pay for submitting a PoSt
```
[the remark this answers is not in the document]
anorth: This and CollateralForSector will have other callers, e.g. the miner worker computing the fee to attach. An alternative is to move these methods to a different place, callable from both. I opted for this so we can try to keep the actor dependencies tightly constrained (they're currently much too loose).

actor/builtin/miner/miner_test.go
anorth: I don't think so. If widely available, would need to have better names and descriptions which would defeat the purpose. It's totally acceptable for this to be duplicated a few times.

actor/builtin/miner/miner_test.go
anorth: I didn't change anything here except removing an unused first parameter "pledge".

actor/builtin/miner/miner_test.go
anorth: No, I intended to pay the maximum fee before the submission would be rejected entirely.

PR #2971, opened 2019-06-20
Codecov requires a base report before posting a GH comment.

anorth:
This improves the quality of the coverage diffs, though I'm not 100% sure it's right yet.

We do have a small amount of nondeterministic coverage in node.go and hello.go, but on the order of <10 lines. 
The coverage report can be confusing while the Circle build is still running: it shows partial results when only some of the jobs have uploaded their coverage reports. It's necessary to wait until the build is finished and then reload the PR page to see a sensible report.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,10 +9,10 @@ coverage:
     changes: off
 
 comment:
-  layout: "reach, diff, flags, files"
+  layout: "diff, flags, files"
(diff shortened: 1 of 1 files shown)


anorth:
I have pushed into this PR a few times and eventually it's settled on a diff with only node/hello coverage changes.

Issue #2814, opened 2019-05-22
Storage protocol – slashing and faults


anorth:
This epic covers issues related to storage fault reporting and slashing required for completion of the storage protocol.

anorth:
I've moved remaining work from here into #2813

PR #2972, opened 2019-06-20
Factor proof calculation out of storage miner for testability.

anorth:
Interesting tests will come later when some of the outstanding TODOs are addressed.

I think the boundary that I've drawn could actually be pushed a bit further, incorporating much of the logic in `Miner.OnNewHeaviestTipSet`. But right now I need to test late fee calculations.

### DIFF
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -42,7 +42,6 @@ const queryDealProtocol = protocol.ID("/fil/storage/qry/1.0.0")
 
 // TODO: replace this with a queries to pick reasonable gas price and limits.
 const submitPostGasPrice = 1
-const submitPostGasLimit = 300
 
 const waitForPaymentChannelDuration = 2 * time.Minute
 
@@ -94,14 +93,9 @@ type node interface {
 	SectorBuilder() sectorbuilder.SectorBuilder
 }
 
-// generatePostInput is a struct containing sector id and related commitments
-// used to generate a proof-of-spacetime
-type generatePostInput struct {
-	commD     types.CommD
-	commR     types.CommR
-	commRStar types.CommRStar
-	sectorID  uint64
-}
+// Static check that miner implements the needs of the Prover
+var _ ProofReader = (*Miner)(nil)
+var _ ProofCalculator = (*Miner)(nil)
 
 // NewMiner is
 func NewMiner(minerAddr, minerOwnerAddr address.Address, nd node, dealsDs repo.Datastore, porcelainAPI minerPorcelain) (*Miner, error) {
@@ -582,19 +576,6 @@ func (sm *Miner) onCommitFail(ctx context.Context, dealCid cid.Cid, message stri
 	log.Errorf("commit failure but could not update to deal 'Failed' state: %s", err)
 }
 
-// poStChallengeSeedAtBlockHeight produces a PoSt challenge seed for the proving period starting at the given height.
-func (sm *Miner) poStChallengeSeedAtBlockHeight(ctx context.Context, currentProvingPeriodStart *types.BlockHeight) (types.PoStChallengeSeed, error) {
-	bytes, err := sm.porcelainAPI.ChainSampleRandomness(ctx, currentProvingPeriodStart)
-	if err != nil {
-		return types.PoStChallengeSeed{}, errors.Wrap(err, "error sampling chain for randomness")
-	}
-
-	seed := types.PoStChallengeSeed{}
-	copy(seed[:], bytes)
-
-	return seed, nil
-}
-
 // isBootstrapMinerActor is a convenience method used to determine if the miner
 // actor was created when bootstrapping the network. If it was,
 func (sm *Miner) isBootstrapMinerActor(ctx context.Context) (bool, error) {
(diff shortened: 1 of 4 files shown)


protocol/storage/deals_awaiting_seal.go
```diff
@@ -13,7 +13,7 @@ func init() {
 	cbor.RegisterCborType(dealsAwaitingSeal{})
 }
 
-// sectorInfo combines sector Metadata from rust proofs with go-filecoin specific data
+// sectorInfo combines sector Metadata from rust Proofs with go-filecoin specific data
```
laser: NIT: Accidental capitalization, perhaps.

protocol/storage/prover.go
```diff
@@ -0,0 +1,109 @@
...
+
+// PoStSubmission is the information to be submitted on-chain for a proof.
+type PoStSubmission struct {
+	Proofs   []types.PoStProof
+	Fee      types.AttoFIL
+	GasLimit types.GasUnits
+}
+
+// NewProver constructs a new Prover.
+func NewProver(actor address.Address, owner address.Address, reader ProofReader, caculator ProofCalculator) *Prover {
```
laser: Nice. I like the `ProofReader` and `ProofCalculator` interfaces.

protocol/storage/prover.go
```diff
@@ -0,0 +1,109 @@
...
+	return &Prover{
+		actorAddress: actor,
+		ownerAddress: owner,
+		chain:        reader,
+		calculator:   caculator,
+	}
+}
+
+// CalculatePoSt computes and returns a proof-of-spacetime ready for posting on chain.
+func (sm *Prover) CalculatePoSt(ctx context.Context, start, end *types.BlockHeight, inputs []PoStInputs) (*PoStSubmission, error) {
```
laser: I like `PoStSubmission`.

protocol/storage/miner_test.go
```diff
@@ -454,7 +454,7 @@ func minerWithAcceptedDealTestSetup(t *testing.T, proposalCid cid.Cid, sectorID
 	}
 
 	// Simulates miner.acceptProposal without going to the network to fetch the data by storing the deal.
-	// Mapping the proposalCid to a sectorID simulates staging the sector.
+	// Mapping the proposalCid to a SectorID simulates staging the sector.
```
laser: NIT: Inconsistent capitalization on this line. Do you object to the following?

```
// Mapping the proposal CID to a sector ID simulates staging the sector.
```

protocol/storage/miner.go
```diff
@@ -95,14 +94,9 @@ type node interface {
...
 
-// generatePostInput is a struct containing sector id and related commitments
-// used to generate a proof-of-spacetime
-type generatePostInput struct {
-	commD     types.CommD
-	commR     types.CommR
-	commRStar types.CommRStar
-	sectorID  uint64
-}
+// Static check that miner implements the needs of the Prover
```
shannonwells: 👍 thanks for that

protocol/storage/miner.go
```diff
@@ -95,14 +94,9 @@ type node interface {
...
+var _ ProofReader = (*Miner)(nil)
+var _ ProofCalculator = (*Miner)(nil)
```
shannonwells: 🏅 we don't do enough of this

protocol/storage/prover.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	proofs, faults, err := sm.calculator.CalculatePost(proofs.NewSortedCommRs(commRs...), seed)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate PoSt")
+	}
+
+	// Compute fees.
+	if len(faults) != 0 {
+		log.Warningf("some faults when generating PoSt: %v", faults)
+		// TODO: proper fault handling
```
shannonwells: I know these TODOs are old but I think the "proper fault handling" one can be connected to some part of #2814 epic.  The other ones ... ?

protocol/storage/miner.go
```diff
@@ -772,102 +772,59 @@ func (sm *Miner) getProvingPeriodStart() (*types.BlockHeight, error) {
...
+	req := sectorbuilder.GeneratePoStRequest{
+		SortedCommRs:  sortedCommRs,
+		ChallengeSeed: seed,
 	}
-
-	resp := sm.Query(ctx, q.Cid)
-
-	if err := cbu.NewMsgWriter(s).WriteMsg(resp); err != nil {
-		log.Errorf("failed to write query response: %s", err)
+	res, err := sm.node.SectorBuilder().GeneratePoSt(req)
```
shannonwells: Miner only needs SectorBuilder and BlockService from node.  I wonder if there is a good way to do this where miner doesn't have to access node.  I'm saying this b/c I'm not sure how much I like having Miner be the default ProofCalculator and ProofReader rather than their being separate (either separate structs or in a separate package), since the SectorBuilder is the one that generates the PoSt. It feels a little like mixing concerns, OTOH what else is concerned with coming up with proofs but a Miner....

PR #2974, opened 2019-06-20
RLE+ encoding

### LINKED ISSUE #2868: RLE+ <---> IntSet
### Description

Implement encoding and decoding to and from rle+ from and to the IntSet built in #2867. 
 How to do this is roughly specified here: https://docs.google.com/document/d/1aE5a-QZojprMkig6IyYkwV1SrMDDLOnNBqNdnwpu9tk/edit

If we went with an existing package for the IntSet we will want some `IntSet` wrapper around it for holding onto the encode and decode functions.

### Acceptance criteria
rle+ encode and decode vetted against rust reference implementation

### Risks + pitfalls
Keep an eye out for errors in reference implementation.  Shouldn't be there but it is very new.
Reference implementation is in rust so get our your crab hammers.

### Where to begin
Read the linked doc and its links to rle+ spec and reference impl

ingar:
## Problem
We need a compact representation of `IntSet` for network transmission and storage.

## Solution
Implementation of [RLE+](https://github.com/filecoin-project/specs/blob/master/data-structures.md#rle-bitset-encoding) encoding.

The encoding happens in [`rleplus.go`](https://github.com/filecoin-project/go-filecoin/blob/c526b748d24c062a9caa922b533db5512d76acbc/rleplus/rleplus.go) via the `Encode()` and `Decode()` functions, with an internal helper `BitVector` class.

Resolves #2868

## TODO
- [x] Determine where this package finally lives
- [x] Find out if the version string is real or not
- [x] Determine where byte-packing order (and network transmission) is part of this story

### DIFF
--- a/rleplus/internal/bitvector.go
+++ b/rleplus/internal/bitvector.go
@@ -36,8 +36,7 @@ type BitVector struct {
 
 // NewBitVector constructs a new BitVector from a slice of bytes.
 //
-// The bytePacking parameter is required to know how to interpret the
-// bit ordering within the bytes.
+// The bytePacking parameter is required to know how to interpret the bit ordering within the bytes.
 func NewBitVector(buf []byte, bytePacking BitNumbering) *BitVector {
 	return &BitVector{
 		BytePacking: bytePacking,
@@ -48,8 +47,7 @@ func NewBitVector(buf []byte, bytePacking BitNumbering) *BitVector {
 
 // Push adds a single bit to the BitVector.
 //
-// Although it takes a byte, only the low-order
-// bit is used, so just use 0 or 1.
+// Although it takes a byte, only the low-order bit is used, so just use 0 or 1.
 func (v *BitVector) Push(val byte) {
 	if v.Len%8 == 0 {
 		v.Buf = append(v.Buf, 0)
@@ -66,7 +64,7 @@ func (v *BitVector) Push(val byte) {
 	v.Len++
 }
 
-// Get returns byte either 0, 1
+// Get returns a single bit as a byte -- either 0 or 1
 func (v *BitVector) Get(idx int) (byte, error) {
 	if idx >= v.Len {
 		return 0, ErrOutOfRange
@@ -84,8 +82,8 @@ func (v *BitVector) Get(idx int) (byte, error) {
 // Extend adds up to 8 bits to the receiver
 //
 // Given a byte b == 0b11010101
-// v.Extend(b, 4, LSB0) would add <1, 0, 1, 0>
-// v.Extend(b, 4, MSB0) would add <1, 1, 0, 1>
+// v.Extend(b, 4, LSB0) would add < 1, 0, 1, 0 >
+// v.Extend(b, 4, MSB0) would add < 1, 1, 0, 1 >
 func (v *BitVector) Extend(val byte, count uint, order BitNumbering) {
 	if count > 8 {
 		count = 8
--- a/rleplus/internal/bitvector.go
+++ b/rleplus/internal/bitvector.go
@@ -29,9 +29,14 @@ const (
 
 // BitVector is used to manipulate ordered collections of bits
 type BitVector struct {
+	Buf []byte
+
+	// BytePacking is the bit ordering within bytes
 	BytePacking BitNumbering
-	Buf         []byte
-	Len         int
+
+	// Len is the logical number of bits in the vector.
+	// The last byte in Buf may have undefined bits if Len is not a multiple of 8
+	Len int
 }
 
 // NewBitVector constructs a new BitVector from a slice of bytes.
--- a/rleplus/internal/bitvector.go
+++ b/rleplus/internal/bitvector.go
@@ -1,0 +1,187 @@
+package bitvector
+
+import "errors"
+
+var (
+	// ErrOutOfRange - the index passed is out of range for the BitVector
+	ErrOutOfRange = errors.New("index out of range")
+)
+
+// BitNumbering indicates the ordering of bits, either
+// least-significant bit in position 0, or most-significant bit
+// in position 0.
+//
+// It it used in 3 ways with BitVector:
+// 1. Ordering of bits within the Buf []byte structure
+// 2. What order to add bits when using Extend()
+// 3. What order to read bits when using Take()
+//
+// https://en.wikipedia.org/wiki/Bit_numbering
+type BitNumbering int
+
+const (
+	// LSB0 - bit ordering starts with the low-order bit
+	LSB0 BitNumbering = iota
+
+	// MSB0 - bit ordering starts with the high-order bit
+	MSB0
+)
+
+// BitVector is used to manipulate ordered collections of bits
+type BitVector struct {
+	BytePacking BitNumbering
+	Buf         []byte
+	Len         int
+}
+
+// NewBitVector constructs a new BitVector from a slice of bytes.
+//
+// The bytePacking parameter is required to know how to interpret the
+// bit ordering within the bytes.
+func NewBitVector(buf []byte, bytePacking BitNumbering) *BitVector {
+	return &BitVector{
+		BytePacking: bytePacking,
+		Buf:         buf,
+		Len:         len(buf) * 8,
+	}
+}
+
+// Push adds a single bit to the BitVector.
+//
+// Although it takes a byte, only the low-order
+// bit is used, so just use 0 or 1.
+func (v *BitVector) Push(val byte) {
+	if v.Len%8 == 0 {
+		v.Buf = append(v.Buf, 0)
+	}
+
+	switch v.BytePacking {
+	case LSB0:
+		v.pushLSB0(val)
+	default:
+		v.pushMSB0(val)
+	}
+
+	v.Len++
+}
+
+// Get returns byte either 0, 1
+func (v *BitVector) Get(idx int) (byte, error) {
+	if idx >= v.Len {
+		return 0, ErrOutOfRange
+	}
+
+	switch v.BytePacking {
+	case LSB0:
+		return v.getLSB0(idx), nil
+	default:
+		return v.getMSB0(idx), nil
+	}
+}
+
+// Extend adds up to 8 bits to the receiver, in the order indicated
+//
+// Given a byte b == 0b11010101
+// v.Extend(b, 4, LSB0) would add <1, 0, 1, 0>
+// v.Extend(b, 4, MSB0) would add <1, 1, 0, 1>
+func (v *BitVector) Extend(val byte, count uint, order BitNumbering) {
+	if count > 8 {
+		count = 8
+	}
+
+	switch order {
+	case LSB0:
+		v.extendLSB0(val, count)
+	default:
+		v.extendMSB0(val, count)
+	}
+}
+
+// Take reads up to 8 bits at the given index, and returns them in a byte.
+// The byte packing is determined by the order parameter.
+//
+// Given a BitVector < 1, 1, 0, 1, 0, 1, 0, 1 >
+// v.Take(1, 4, LSB0) would return 0b00001011
+// v.Take(1, 4, MSB0) would return 0b00001101
+func (v *BitVector) Take(index int, count int, order BitNumbering) byte {
+	if count > 8 {
+		count = 8
+	}
+
+	switch order {
+	case LSB0:
+		return v.takeLSB0(index, count)
+	default:
+		return v.takeMSB0(index, count)
+	}
+}
+
+// Iterator returns a function, which when invokes, returns the number
+// of bits passed, and increments an internal cursor.
+//
+// When the end of the BitVector is reached, it returns zeroes indefinitely
+func (v *BitVector) Iterator(order BitNumbering) func(int) byte {
+	cursor := 0
+	return func(count int) (out byte) {
+		out = v.Take(cursor, count, order)
+		cursor += count
+		return
+	}
+}
+
+func (v *BitVector) getLSB0(idx int) byte {
+	blockIdx := idx / 8
+	return v.Buf[blockIdx] >> uint(idx%8) & 1
+}
+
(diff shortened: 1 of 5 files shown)


dignifiedquire:
> Find out if the version string is real or not

There are no strings quotes (`"`) mean just the literal bytes. Not sure if this solves the question.

dignifiedquire:
> Determine where byte-packing order (and network transmission) is part of this story

Ups, missed sth there in the spec :) Technically the format doesn't care about it, but we need to fix it for filecoin, so I would suggest just going with Little Endian now, as the rust implementation does.

ingar:
Ok thanks for the clarification @dignifiedquire on the version.  I was wondering because it wasn't in the rust impl, but that makes sense since since it was just for comparison purposes.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,187 @@
...
+		v.extendMSB0(val, count)
+	}
+}
+
+// Take reads up to 8 bits at the given index, and returns them in a byte.
+// The byte packing is determined by the order parameter.
+//
+// Given a BitVector < 1, 1, 0, 1, 0, 1, 0, 1 >
+// v.Take(1, 4, LSB0) would return 0b00001011
+// v.Take(1, 4, MSB0) would return 0b00001101
```
ZenGround0: There is something weird about this method's interpretation of `MSB0`.  To preserve symmetry across the package `v.Take(1, 4, MSB0)` would be `0b11010000`.  If the output of `Take` in `MSB0` mode is actually interpreted as an `MSB0` byte then it is multiplied by something like `2^(8-count)` when compared with the `LSB0` output of Take.

I haven't gotten to consumption of `Take` and I'm sure it makes sense when doing RLE+, but the difference and or motivation is probably worth explaining in the comment here.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,187 @@
...
+
+// extendMSB0 pushes up to 8 bits onto the BitVector, high-order bit first
+// e.g. v.extendMSB0(171, 4) // adds <...., 1, 0, 1, 0> since 171 == 0b10101011
+func (v *BitVector) extendMSB0(val byte, count uint) {
+	for i := uint(0); i < count; i++ {
+		v.Push((val >> (7 - i)) & 1)
+	}
+}
+
+// takeMSB0 fetches up to 8 bits and returns it as a byte, high-order bit first
```
ZenGround0: Related to comment about `Take`'s asymmetry above: maybe say "zero padded, high order bit first"?

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,187 @@
...
+
+	switch order {
+	case LSB0:
+		return v.takeLSB0(index, count)
+	default:
+		return v.takeMSB0(index, count)
+	}
+}
+
+// Iterator returns a function, which when invokes, returns the number
```
ZenGround0: nit:  invoked

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/filecoin-project/go-filecoin/rleplus/internal"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"gotest.tools/assert"
+)
+
+func assertBitVector(t *testing.T, expected []byte, actual bitvector.BitVector) {
```
ZenGround0: might be worth mentioning in comment that `expected`'s elements should only ever be `0` and `1` bytes.  Perhaps you could call it `expectedBits` too.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,187 @@
...
+// extendLSB0 pushes bits onto the BitVector, low-order bit first
+// e.g. v.extendLSB0(171, 4) // adds <...., 1, 1, 0, 1> since 171 == 0b10101011
+func (v *BitVector) extendLSB0(val byte, count uint) {
+	for i := uint(0); i < count; i++ {
+		v.Push((val >> i) & 1)
+	}
+}
+
+// takeLSB0 fetches up to 8 bits and returns it as a byte, low-order bit first
+// e.g. BitVector <1, 1, 0, 1, 0, 1, 0, 1 >, v.takeLSB0(1, 4) returns 0b00001011
```
ZenGround0: Blocking: I think this comment should say `v.takeLSB0(0, 4)` as `Get` is 0 indexed.  Same for the comment on `takeMSB0` below.

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,124 @@
...
+		bits := []byte{1, 0, 1, 0, 1, 0, 1, 1}
+		for _, bit := range bits {
+			v.Push(bit)
+		}
+
+		assert.Equal(t, byte(11), v.Take(4, 4, bitvector.MSB0))
+		assert.Equal(t, byte(13), v.Take(4, 4, bitvector.LSB0))
+	})
+
+	t.Run("Iterator", func(t *testing.T) {
```
ZenGround0: Very cool abstraction

rleplus/internal/bitvector.go
ingar: 👍 Yeah, the semantics of `Take` and `Extend` are confusing.  Will rationalize.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,187 @@
...
+func (v *BitVector) takeLSB0(index int, count int) (out byte) {
+	for i := 0; i < count; i++ {
+		val, _ := v.Get(index + i)
+		out |= val << uint(i)
+	}
+	return
+}
+
+// extendMSB0 pushes up to 8 bits onto the BitVector, high-order bit first
```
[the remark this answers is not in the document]
ingar: I just inlined this function; didn't really make too much sense on its own.

rleplus/internal/bitvector.go
ingar: 👍 I removed these internal methods; take a look at the new comments for [`Take` and `Extend`](https://github.com/filecoin-project/go-filecoin/blob/1a177e62fe225c753235072ab1097c27fb26b5ab/rleplus/internal/bitvector.go#L84-L125)

rleplus/rleplus.go
```diff
@@ -0,0 +1,174 @@
...
+			v.Push(1)
+		case run < 16:
+			v.Push(0)
+			v.Push(1)
+			v.Extend(byte(run), 4, bitvector.LSB0)
+		case run >= 16:
+			v.Push(0)
+			v.Push(0)
+			buf := make([]byte, 10)
+			numBytes := binary.PutUvarint(buf, run)
```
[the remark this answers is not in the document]
ZenGround0: Reading into this closer I now realize this 9 bytes requirement is pedantic on the part of multiformats spec and won't impact correctness because the multiformats spec is wrong about the golang varint implementation.

Still a good idea to impose a smaller max varint for RLE+ / total number of sector ids per miner actor.  I think limiting the max varint should happen later in a separate PR.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,136 @@
...
+			v.Push((val >> (7 - i)) & 1)
+		}
+	}
+}
+
+// Take reads up to 8 bits at the given index.
+//
+// Given a BitVector < 1, 1, 0, 1, 0, 1, 0, 1 >
+// v.Take(0, 4, LSB0) would return 0b00001011
+// v.Take(0, 4, MSB0) would return 0b11010000
```
ZenGround0: 👍

rleplus/rleplus.go
```diff
@@ -0,0 +1,174 @@
...
+
+			if val == 1 {
+				// short block
+				runLength = int(take(4))
+			} else {
+				// long block
+				var buf []byte
+				for {
+					b := take(8)
+					buf = append(buf, b)
```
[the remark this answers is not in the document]
ZenGround0: Brainstorming: the max varint should the max sector ID allowable + 1 or else we need to modify RLE+

rleplus/rleplus.go
```diff
@@ -0,0 +1,174 @@
...
+//
+// A set of unsigned integers { 0, 2, 4, 5, 6 } can be thought of as
+// indices into a bitset { 1, 0, 1, 0, 1, 1, 1 } where bitset[index] == 1.
+//
+// The bit set run lengths of this set would then be { 1, 1, 1, 1, 3 },
+// representing lengths of runs alternating between 1 and 0, starting
+// with a first bit of 1.
+//
+// This is a helper function for Encode()
+func RunLengths(ints []uint64) (firstBit byte, runs []uint64) {
```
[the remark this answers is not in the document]
ZenGround0: Looking closer it looks like duplicates are safely ignored.  Making a note of that would be nice.

rleplus/rleplus.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(ints) == 0 {
+		return
+	}
+
+	// Sort our incoming numbers
+	sort.Slice(ints, func(i, j int) bool { return ints[i] < ints[j] })
+
+	last := ints[0]
```
ZenGround0: nit: Feel free to ignore, but maybe `prev` instead? To me `last` carries around connotations of pointing at some "final" thing.

rleplus/rleplus_test.go
```diff
@@ -0,0 +1,157 @@
...
+			1, 0, 0, 0, 1, 0, 0, 0,
+		}
+
+		v := bitvector.BitVector{}
+		for _, bit := range expectedBits {
+			v.Push(bit)
+		}
+		actualBytes, _ := rleplus.Encode(ints)
+
+		for idx, expected := range v.Buf {
```
ZenGround0: nit: maybe check lengths too to make sure there is nothing extra in `actualBytes`'s `Buf`

rleplus/rleplus.go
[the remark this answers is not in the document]
ingar: The current thinking is that the max # sectors is 16384; I can bound it at 2 bytes.

rleplus/internal/bitvector.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// MSB0 - bit ordering starts with the high-order bit
+	MSB0
+)
+
+// BitVector is used to manipulate ordered collections of bits
+type BitVector struct {
+	BytePacking BitNumbering
+	Buf         []byte
+	Len         int
```
anorth: Please document these:
// BytePacking is the bit ordering within bytes
// Len is the logical number of bits in the vector. The last byte in Buf may have undefined bits if Len is not a multiple of 8.

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,125 @@
...
+// Note: When using this helper assertion, expectedBits should *only* be 0s and 1s.
+func assertBitVector(t *testing.T, expectedBits []byte, actual bitvector.BitVector) {
+	assert.Equal(t, len(expectedBits), actual.Len)
+
+	for idx, bit := range expectedBits {
+		actualBit, err := actual.Get(idx)
+		assert.NilError(t, err)
+		assert.Equal(t, bit, actualBit)
+	}
+}
```
anorth: Nit: move helpers to the bottom when reasonable.

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+func TestBitVector(t *testing.T) {
+	tf.UnitTest(t)
+
+	t.Run("Zero Value", func(t *testing.T) {
```
anorth: Nit: there are some inconsistencies, but we mostly use lower case for test run names.

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,125 @@
...
+		var v bitvector.BitVector
+
+		assert.Equal(t, bitvector.LSB0, v.BytePacking)
```
anorth: Test the iterator too

rleplus/internal/bitvector_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		assert.Equal(t, byte(176), v.Take(4, 4, bitvector.MSB0))
+		assert.Equal(t, byte(13), v.Take(4, 4, bitvector.LSB0))
+	})
+
+	t.Run("Iterator", func(t *testing.T) {
+		var v bitvector.BitVector
+
+		for i := 0; i < 256; i++ {
+			v.Push(byte(rand.Int() & 1))
```
anorth: I don't think `rand` belongs in tests. To be sure of through coverage, how about initialising Buf with all 256 unique values? Or a fixed set that exercise a bunch of cases.

rleplus/rleplus.go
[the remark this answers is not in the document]
acruikshank: This is a good point @ZenGround0. It sounds like it's not really a problem because the size is low. If it were, I think we could figure out other ways to deal with it (e.g. predict the size from the RLE without decoding it, and charging gas on the size).

rleplus/rleplus.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(buf) == 0 {
+		return
+	}
+
+	v := bitvector.NewBitVector(buf, bitvector.LSB0)
+	take := v.Iterator(bitvector.LSB0)
+
+	// Read version and discard
+	take(1)
+	take(1)
```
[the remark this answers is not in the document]
ingar: I added a proper version check, thanks.  I wasn't sure if the version was real or not since it wasn't in the ref implementation.

rleplus/rleplus.go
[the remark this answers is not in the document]
steven004: Sorry, just find the discussion here while I encountered an issue while running my new code which uses different sectorId segments for different disks, so that I could take advantage of RLE+ to save bits when some disks lost. Unfortunately, the I defined segment size as 2^32, then it is not allowed since we limit the runlength < 2^14.

Issue #2959, opened 2019-06-18
Implement command and porcelain to set miner.Worker


acruikshank:
### Description

Miner operators will likely be multi-person operations and will need to have the ability to run their accounts from multiple wallets. As of #2954, we will have a mechanism to alter the address from which miner operational messages are sent. We need a command to alter the address, and also a command to show the current value.

We also need to ensure we use the worker address for all automated message that perform mining operations. This means we need to look up the worker address and retrieve and use it though out the code base.

#### Covered by PR #3261 
- [x]  Implement `SetWorkerAddress` in porcelain/miner

    - [x]  Call SetWorker on the miner actor (just return the message CID)
- [x]  Implement `miner worker` command, which displays the current worker address.
- [x]  Implement `miner set-worker [new address]`, which sets the worker address to the new address.
- [x]  storage.FaultSlasher's messages must be signed by the mining worker

#### Covered by PR #3286 
- [x]  Every automated message sent to a miner address, instead of using the default wallet address, should look up the worker address and use it as the from
- [x]  Block tickets must be signed from the worker address and verified against it

anorth:
@acruikshank done?

acruikshank:
The worker address exists, and the miner actor checks that messages come from the worker address. This currently relies on the fact that the worker address is always the same as the miner owner.

This story is about implementing the ability to change the worker address and using it whenever we send messages to the miner actor.

Issue #2954, opened 2019-06-17
Implement miner.Actor.ChangeWorker


acruikshank:
### Description

As of #2936, miner actors can be managed with messages signed from two different private keys. Messages signed by the owner create the miner and can set the worker address. The worker address sends all the other message. This story is about creating the [ChangeWorker](https://github.com/filecoin-project/specs/blob/master/actors.md#changeworker) command to change the worker address.

### Acceptance criteria

1. Implement SetWorker to ensure message comes from miner owner
2. Implement SetWorker to update miner.Worker field

PR #2976, opened 2019-06-20
Replace miner public key with miner worker address

### LINKED ISSUE #2936: Replace `miner.Actor.PublicKey` with `miner.Actor.Worker` address
### Description

We would like to separate the accounts used to create workers and receive mining rewards from the account used for storage mining operations. To begin, we will introduce a `Worker` address to miner state. For now, this address can be the same as the owners and initialized in the same way. 

The miner's PublicKey is actually redundant with the address (the address is just a hashed public key). We should also remove the PublicKey and compare signatures against the new worker address instead.

### Acceptance criteria

1. The miner has a `Worker address.Address` field.
1. The `Worker` field is initialized to the same thing as the `Owner` when the miner is initialized with the same value.
1. CommitSector and SubmitPoSt validate that the message is sent from the worker instead of the owner.
1. The `PublicKey` field is removed from miner state.
1. Every place we had used the public key, we now extract the public key from the signature, hash it, and compare it to the worker address.

acruikshank:
closes #2936

### Problem

The [Filecoin Spec](https://github.com/filecoin-project/specs/blob/master/actors.md#storage-miner-actor) requests that the miner actor store a `Worker` address rather than a `PublicKey`. This address is also the address that will be used to interact with the miner.

### Solution

This PR:
1. Replaces the `PublicKey` field in miner state with `Worker`.
2. Modifies `storagemarket.Actor.CreateStorageMiner` to stop taking public key as a parameter. The worker address is initialized to be the miner owner by default. #2954 and #2959 define how the address will be updated.
3. Modifies mining worker to take the worker address rather than a public key for ticket signing.
4. Creates tickets using the worker address for signing.
5. Changes lots of tests to use the wallet address rather than the wallet public key to create a worker.


acruikshank:
@shannonwells Yes it has changed in the spec. That change is the motivation for this PR.

actor/builtin/miner/miner.go
```diff
@@ -121,14 +117,20 @@ type Ask struct {
 
 // State is the miner actors storage.
 type State struct {
+	// Owner is the address of the account that owns this miner. Income and returned
+	// collateral are paid to this address. This address is also allowed to change the
+	// worker address for the miner.
 	Owner address.Address
 
+	// Worker is the address of the worker account for this miner.
+	// This will be the key that is used to sign blocks created by this miner, and
```
laser: > This will be the key that is used to sign blocks created by this miner,

Should this read: 

> This actor will own the key that is used to sign blocks created by this miner,

?

acruikshank: This is  not the best wording. But an address is a hashed public key and we use them as such.

laser: Ah, disregard. I misunderstood what an `Address` was.

actor/builtin/miner/miner.go
```diff
@@ -164,11 +166,11 @@ func NewActor() *actor.Actor {
 }
 
 // NewState creates a miner state struct
-func NewState(owner address.Address, key []byte, pid peer.ID, sectorSize *types.BytesAmount) *State {
+func NewState(owner address.Address, pid peer.ID, sectorSize *types.BytesAmount) *State {
```
anorth: Could we push this up one more level, accepting two addresses here and passing the same parameter value at call sites? Then we can test behaviour of interactions with the two addresses properly. It's ok for most unit tests to use the same key for both, but it should be possible to test that, e.g., block rewards are paid to the correct address.

Leave a comment at those call sites pointing to an issue.

actor/builtin/miner/miner_test.go
```diff
@@ -118,26 +117,54 @@ func TestAskFunctions(t *testing.T) {
...
 	st, vms := core.CreateStorages(ctx, t)
 
-	signature := []byte("my public key")
-	minerAddr := createTestMiner(t, st, vms, address.TestAddress, signature, th.RequireRandomPeerID(t))
+	minerAddr := createTestMiner(t, st, vms, address.TestAddress, th.RequireRandomPeerID(t))
 
 	// retrieve key
-	result := callQueryMethodSuccess("getKey", ctx, t, st, vms, address.TestAddress, minerAddr)
-	assert.Equal(t, result[0], signature)
+	result := callQueryMethodSuccess("getWorker", ctx, t, st, vms, address.TestAddress, minerAddr)
```
anorth: This isn't testing that it retrieves the right address since it's impossible to distinguish between the owner and worker address at creation.

chain/get_ancestors_test.go
```diff
@@ -32,14 +32,15 @@ func requireGrowChain(ctx context.Context, t *testing.T, blockSource *th.TestFet
 	link := requireHeadTipset(t, chainStore)
 
 	signer, ki := types.NewMockSignersAndKeyInfo(1)
-	mockSignerPubKey := ki[0].PublicKey()
+	minerOwner, err := ki[0].Address()
```
anorth: Name this `minerWorker` for consistency with the same lines below (or change the others to match this).

actor/builtin/miner/miner_test.go
acruikshank: This is true. The next PR adds the ability to change the address. I've already written that test.

Issue #2965, opened 2019-06-19
Integration tests for payment voucher redemption


anorth:
### Description
The TestDealsRedeem daemon test, apart from being flaky (#2960) takes many minutes because it waits for a sector to be sealed. This is not acceptable duration for an integration test that developers are expected to run with any frequency.

Voucher redemption after a successful deal must be tested without waiting for real sector sealing. This probably means an in-memory integration test.

It's probably still appropriate to leave a stable version of the existing test as a functional test, to be run occasionally on CI and to validate real networks.

### Acceptance criteria
- Payment voucher redemption is integration tested but only takes a few seconds

### Risks + pitfalls
- We don't have good patterns for this yet, this is an opportunity to define them.

### Where to begin
See also #2822, #2194.

ZenGround0:
Something I've noticed from fixing the flakiness: it is actually unnecessary and poor separation of concerns to wait until sealing completes.  Vouchers are valid after the deal duration is complete, which is well before sealing finishes.

Of course we will eventually need something that can handle waiting for a complete seal in order to test the edge cases of deal arbitration (specifically testing that when the sector is committed and the client erroneously arbitrates the miner can correctly end arbitration in their favor).

anorth:
Ancient history

PR #2977, opened 2019-06-20
remove flake from redeem test

ZenGround0:
After investigating #2960 I found a few things.  First going off a helpful tip from @travisperson there is solid evidence that the nondeterminism is not caused by AttoFil math at all, but rather a race condition in the existing test.  Currently the test assumes that the `commitSector` message gets into the chain and its gas cost is deducted *before* the old balance is recorded for comparison.  Investigating the chain I found that the gas price of `commitSector` is `0.0000000000000002`, exactly the amount missing during flakes.  <details><summary>(Click here for the block json)</summary>
<p>

```json
Block cid=[zDPWYqFCt75CwJyReU4u6f6ZtFcy6umYX7kBkQ1AiRheCR3t2VcH]: {
  "miner": "t2udhmwzokf4hq7oe4zj4g3faah7oz4dnbxtxaqoy",
  "ticket": "IM8DZo6YyvGMWWawTlnMX/aILVqmmjcWWiSYBtCdwQcGmmQSIONle2+pSzSXbog6PUHRt6MiAnGrQPqsMS90FwE=",
  "parents": [
    {
      "/": "zDPWYqFCsYX7uHw7ibjnF8TjxgQeADwXXn7YHBxuEUJRue1tNo55"
    }
  ],
  "parentWeight": "321310000",
  "height": "2922",
  "nonce": "0",
  "messages": [
    {
      "meteredMessage": {
        "message": {
          "to": "t22ttnc7b3fnrj72sksamjmymwknagad4p2bmhnha",
          "from": "t13irzerbpsr3jepcg3edbmgfaczlas5sxo5ufi7q",
          "nonce": "2",
          "value": "0",
          "method": "commitSector",
          "params": "hUEBWCDOj7LnXHSl3rilpTB0y3iPbjl8G8URqk4deoNhVfyyD1ggfBmGGWrR9vQoSJLTfScFQdPldRU7uZgHu4Aaot42fQpYIGhKLSHm09qnqeTtVnN9ey9za8eCmFRDqN5mDb7deBk1WQGAoeuZaG0t+66xMyqGhthzLweLDVlKkrOcqA45Yzi8Z/Hxk4ejHVUbl3nZ6UvFrjiDuEQ0pkejmvFe3cGhIdejMXL38//hiusbCayvDufgVLjKPMHLq1j7QdDU8hnHeuq4F0HKOfE9KDO2GBM1NHXtdsl6gEOkYnwoy60gjcxnaEs8pRgZSPVxSd/9I+hrAE/8p7hNkwwrL7tAcbf8Re8ARslel+6uCu+ezxlZ31IolbHuZkR+dS8Z8OY7ILiF56zBgxhcp72v7QBm22uS6kI7UFs8nODsZgGDJNhakX4hJs+9y3EU/6Qk/0OYhIP2Re+7jgI/GLLG5yEAoNcN6GjcWtb9VaaYajg/VsMQwJjamradfXLCSYJj17oRnh+0VgkpFh+E18clcPDi6rXlYqDbCdk5o7CiA60L8RTnUb9rWr3B9IjAD0S4DguaBHYcJfWIowmkg2cPuiSZHABJ6lDd9MJf6u5WYnCiWaoLkq0C7l8mC8K9hn3dUriP1YA2pdRN"
        },
        "gasPrice": "0.000000000000000001",
        "gasLimit": "300"
      },
      "signature": "sK3BB7Di/9X0wTAV0129LaY5v4BfJdBKZ9E6aeZxzIkUIuINAuRlp6coImbgvZOEiyCgnDVCxoTb2zApViT4MAA="
    }
  ],
  "stateRoot": {
    "/": "zdpuAqVgF9RPuMeThpMH6GvJWp54htQyq4UJT2pMusaAgroa3"
  },
  "messageReceipts": [
    {
      "exitCode": 0,
      "return": null,
      "gasAttoFIL": "0.0000000000000002"
    }
  ],
  "proof": null,
  "timestamp": "1561059969"
}

```
</p>
</details>

I did not directly correlate that this gas price gets paid after recording the first balance and before recording the final balance in the case of flakes but am confident in this hypothesis.

A related problem I ran into is that there is another race in the current test between time to complete sealing and number of blocks before the channel expired.  On my machine I consistently could not redeem because the channel had expired long before sealing finished.

A final related issue recorded in #2965 is that this test waits around to seal before completing which is too slow.  I would argue waiting around for sealing to complete is a poor testing strategy.  The conditions on the deal are that the miner gets paid after a certain number of blocks from channel creation, not from committing the sector.  As written the deal is only for 1 block, so the miner should be able to get paid well before sealing finishes.

I've changed the test to redeem after the deal is accepted without waiting for sealing.  This speeds up the test time to something manageable and avoids the expiry time out (though technically the race still exists if the redeem message got sent out 1000 blocks after `DealsRedeem` were called).  Conveniently it also avoids calculating balances in a window when the `commitSector` message is sent so it does not flake.  The race still exists and but this test will now only flake if sealing time goes down to a few seconds.

### DIFF
--- a/commands/deals_daemon_test.go
+++ b/commands/deals_daemon_test.go
@@ -3,11 +3,11 @@ package commands_test
 import (
 	"context"
 	"crypto/rand"
+	"fmt"
 	"io"
 	"math/big"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-files"
@@ -29,12 +29,6 @@ import (
 )
 
 func TestDealsRedeem(t *testing.T) {
-	// DISABLED: this test has nondeterministic rounding errors
-	// https://github.com/filecoin-project/go-filecoin/issues/2960
-	// It also takes many minutes due to waiting for real sector sealing. This is unacceptable
-	// for an integration test (possibly ok for a functional test).
-	// https://github.com/filecoin-project/go-filecoin/issues/2965
-	t.Skipf("Flaky and slow: #2960, #2965")
 	tf.IntegrationTest(t)
 
 	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fast.FilecoinOpts{})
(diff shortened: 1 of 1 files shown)


ingar:
Wow, nice job!  I attempted to look at this but was looking in totally the wrong place

commands/deals_daemon_test.go
```diff
@@ -7,7 +7,8 @@ import (
 	"math/big"
 	"strings"
 	"testing"
-	"time"
+
```
anorth: blank line

commands/deals_daemon_test.go
```diff
@@ -94,8 +98,12 @@ func TestDealsRedeem(t *testing.T) {
 	newWalletBalance, err := minerDaemon.WalletBalance(ctx, minerOwnerAddress)
 	require.NoError(t, err)
 
+	expectedGasCost := 0.1
+	expectedBalanceDiff := float64(dealDuration*dataPriceOneBlock) - expectedGasCost
+	expectedBalanceStr := fmt.Sprintf("%.1f", expectedBalanceDiff)
```
acruikshank: Is this rounding still necessary?

ZenGround0: yes.  I actually had to add it in because otherwise the trailing 0s you get by default from float64 cause this to fail.

ZenGround0: And note this will still fail if actual balanceDiff is off by a very small fraction because we don't round actualBalanceDiff

Issue #2973, opened 2019-06-20
Hide non-consensus actor methods from processing


acruikshank:
### Description

We have two slightly conflicting needs:

1. Keep actor methods binary compatible with spec so that we achieve consensus.
2. Allow the implementation to access chain state using the same logic as the actor methods.

The second need (and a few other needs) argue strongly for using actor methods to query chain state in the go-filecoin implmentation. Adding these methods to the spec, however, requires changes to the spec that don't actually affect the network implementation.

We can't simply add an actor method to the implementation without adding it to the spec, because attempts to call those methods from across the network would work on some implementations and fail with a method not found on others. This would cause consensus failures.

We need a way to flag actor methods that are needed by the implementation but not the networrk. This may be all methods that do not alter state. It's certainly hard to imagine why someone would want to pay for a message to go on chain that does not alter state when they can query their own chain for free.

When running state transitions, messages directed at query methods should behave exactly as if they do not exist.

### Acceptance criteria

1. Add a flag to `exec.FunctionSignature` that signals whether the method is part of the spec (consensus? query? specified? mutable?)
2. Add a parameter to `exec.VMContext` to signal whether the method is running as a query or part of message processing.
3. in `vm.Send` return a `ErrMissingExport` if we are in message processing and the method is only used for querying

### Protocol Changes

Once this functionality is in place, we should remove all read only methods from the spec.

anorth:
At least some read-only methods are needed for the actors to interoperate with each other. As a (not great) example, the miner actor calls `getProofsMode` on the storage market. As another, the storage market's total storage power will be an input to miner collateral calculations. Query methods are generally useful, and the spec might not reflect all the utility.

Nevertheless, I agree in general that we need ways for non-actor code to inspect actor state and probably want to share code with actor method implementations.

If methods are omitted from the `Exports` list then I assume they are not invokable as messages and so not part of ABI compatibility. Presence in that list is the flag we desire. I think the question then becomes how to invoke the non-exported methods without going through the message-passing paraphanalia. Perhaps we need a code path to load an actor/state with a read-only VM context? Something similar to the current query-message, but that just returns an actor object instead of sending it a message. This would conveniently skip all the parameter encoding/decoding too.

acruikshank:
[...]

PR #2980, opened 2019-06-21
Miner worker calculates and pays late fees when submitting PoSt

### LINKED ISSUE #2942: Miner pays fees for late PoSt submission

anorth:
This change is in two commits. The first does the real work but introduced some messy dependency plumbing. The second re-arranges dependencies to avoid passing prover dependencies through the miner, but touches quite a few files on the way. The `CalculatePoSt` method is added to plumbing.

cc @acruikshank as this touches a few things you've edited recently.

Closes #2942

### DIFF
--- a/protocol/storage/prover.go
+++ b/protocol/storage/prover.go
@@ -5,13 +5,21 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/proofs"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// This will likely depend on the sector size and proving period.
-const submitPostGasLimit = 300
+const (
+	// Maximum number of rounds delay to allow for when submitting a PoSt for computing any
+	// fee necessary due to late submission. The miner expects the PoSt message to be mined
+	// into a block at most `buffer` rounds in the future.
+	postSubmissionDelayBufferRounds = 10
+
+	// This will likely depend on the sector size and proving period.
+	submitPostGasLimit = 300
+)
 
 // ProofReader provides information about the blockchain to the proving process.
 type ProofReader interface {
@@ -19,6 +27,10 @@ type ProofReader interface {
 	ChainHeight() (*types.BlockHeight, error)
 	// ChallengeSeed returns the PoSt challenge seed for a proving period.
 	ChallengeSeed(ctx context.Context, periodStart *types.BlockHeight) (types.PoStChallengeSeed, error)
+	// PledgeCollateralRequirement returns the pledge collateral required to be posted by a miner.
+	PledgeCollateralRequirement(ctx context.Context, addr address.Address) (types.AttoFIL, error)
+	// WalletBalance returns the balance for an actor.
+	WalletBalance(ctx context.Context, addr address.Address) (types.AttoFIL, error)
 }
 
 // ProofCalculator creates the proof-of-spacetime bytes.
@@ -30,10 +42,11 @@ type ProofCalculator interface {
 
 // Prover orchestrates the calculation and submission of a proof-of-spacetime.
 type Prover struct {
-	actorAddress address.Address
-	ownerAddress address.Address
-	chain        ProofReader
-	calculator   ProofCalculator
+	actorAddress  address.Address
+	workerAddress address.Address
+	sectorSize    *types.BytesAmount
+	chain         ProofReader
+	calculator    ProofCalculator
 }
 
 // PoStInputs contains the sector id and related commitments used to generate a proof-of-spacetime.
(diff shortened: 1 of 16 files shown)


protocol/storage/prover.go
```diff
@@ -74,36 +88,79 @@ func (sm *Prover) CalculatePoSt(ctx context.Context, start, end *types.BlockHeig
...
+		return types.PoStChallengeSeed{}, err
+	}
+
+	seed := types.PoStChallengeSeed{}
+	copy(seed[:], bytes)
+	return seed, nil
+}
+
+// calculateFee calculates any fees due with a proof submission due to faults or lateness.
+func (p *Prover) calculateFee(ctx context.Context, height *types.BlockHeight, end *types.BlockHeight) (types.AttoFIL, error) {
```
acruikshank: BLOCKING: This method is collecting various elements of miner actor state (ProvingPeriodEnd, ActiveCollateral, SectorSize) and then calling static methods on the miner actor (GenerationAttackTime, LatePoStFee) to compute the late fee. This really should be a single query method on the miner.Actor. The miner actor already has this logic implemented in a similar fashion.

Consolidation of consensus logic in the actors will prevent potential consensus bugs and also simplify the model for change as these algorithms change.

plumbing/proofs/proofs.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"github.com/filecoin-project/go-filecoin/proofs"
+	"github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// Provides the sector builder for proofs.
+type sectorBuilderer interface {
+	SectorBuilder() sectorbuilder.SectorBuilder
```
shannonwells: Since it provides the sector builder for proofs, can we call it sectorBuilderProvider instead?

protocol/storage/prover.go
anorth: I see what you're saying, but I think there's a reasonable separation to be made somewhere here. Firstly, the *actual* fee calculation code is in the actor (factored out as a static function) and shared. This method is invoking the same calculation but not with the same parameters: the height/lateness is a hypothetical, and a comment indicates that we might also want the collateral to be a different value too.

For reasons you cited in #2973, I don't want to add an exported actor method to do this. The decision about how much buffer to add to lateness and/or collateral is not part of consensus.

plumbing/proofs/proofs.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Proofs is plumbing for proof calculation and verification.
+type Proofs struct {
```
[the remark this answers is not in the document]
anorth: I do want to avoid exposing the whole sector builder interface. I was tempted to push the request/response marshalling down (essentially removing the structs from the sector builder interface) in which case this would be a vanishingly thin wrapper. I don't really follow why exposing the parts separately in plumbing is undesirable, but I'll try pushing it up to porcelain.

protocol/storage/prover.go
```diff
@@ -5,20 +5,32 @@ import (
...
 
 // ProofReader provides information about the blockchain to the proving process.
 type ProofReader interface {
-	// ChainHeight returns the current height of the best chain.
-	ChainHeight() (*types.BlockHeight, error)
-	// ChallengeSeed returns the PoSt challenge seed for a proving period.
-	ChallengeSeed(ctx context.Context, periodStart *types.BlockHeight) (types.PoStChallengeSeed, error)
+	// ChainBlockHeight returns the current height of the best chain.
+	ChainBlockHeight() (*types.BlockHeight, error)
+	// ChainSampleRandomness returns bytes derived from the blockchain before `sampleHeight`.
```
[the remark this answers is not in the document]
anorth: Understood, but I'll leave this comment for simplicity here and expect the implementation to handle it.

protocol/storage/prover.go
anorth: We discussed this in person. I'm intending to move the `LatePoStFee` function from a free function in the miner actor to be a method instead. This turns out to be a bit of a pain though; if I follow existing patterns it requires adding it to porcelain and probably changing the parameters, since it will necessarily be a function of a specific miner instance. I might try to find a less verbose pattern than adding every actor query method to porcelain.

acruikshank: It would be my preference __not__ to add all query methods to porcelain. I believe we have several internal query method calls that don't have porcelain wrappers. The requirement for adding a porcelain method really should be that it's user facing and there is a command that calls it. That goal has gotten muddy in the porcelain implementation.

It appears that the new `getPledgeCollateralRequirement` is currently only used to provide a parameter to feed a parameter back to the late fee calculation. I would replace this function with an actor method named something like 'computeLateFeeAtBlockHeight' and have it pull the collateral requirement from state. This leaks fewer details about the fee calculation.

plumbing/proofs/proofs.go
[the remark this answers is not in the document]
anorth: I have exposed SectorBuilder on plumbing and moved `CalculatePoSt` wrapper to porcelain.

anorth:
@acruikshank sorry this is a bit large now. I've attempted to reflect the ideas about actor method that we've been talking about. There is a lot of boilerplate involved in invoking actor methods, and they also make testing awkward.

On particular worry I have is that while the prover is unit tested, I don't have an integration test involving an actual actor and VM. I can kinda see the path to adding one, but it would involve re-implementing the bits of miner (and other) porcelain on which the prover depends just as test wiring: all that `queryAndDeserialize` junk. But that wiring is precisely the thing an integration test would want to exercise. I can't see a good way to do this without lots of work. I think this is an instance of a general problem, that accessing actor state externally is currently very hard.

Issue #2968, opened 2019-06-19
Add ability to propose storage deal with miner without running the mining scheduler


travisperson:
### Description

Currently to propose a storage deal, a user must have called `mining start`. This has presented a few issues with regard to the way we are currently aiming to write our integration tests.

In our integration tests we would like to have complete control over the mining cycle. This enables us to have great control over our block height while maintaining quick execution of our tests. However, currently any test that requires proposing a storage deal does not work in this manner because we must call `mining start` to register the storage miner protocol with the libp2p host. This is due to the fact that we [register the libp2p protocols during `NewMiner`](https://github.com/filecoin-project/go-filecoin/blob/103b4521af55ff7154d552b2067918be5be7914e/protocol/storage/miner.go#L125) of the storage protocol which is constructed during `mining start`.

The `mining start` command does a few thing

1) Creates a sector builder, if not already built
2) Create a mining worker, if not already built
3) Create a mining scheduler, if not already built
4) Create a storage miner
5) Listen for SectorSealResults and send `commitSector` messages to miner.

I would like to propose that we provide a way to set everything up without the schedule, and allow the use of `mining once`. I think ideally the scheduler should be as simple as issuing the `mining once` command on the desired interval. This would allow the mining once API to continue to exist even on the TestNet / MainNet.

### Acceptance criteria

- Storage miner protocol can be setup without running the mining scheduler
- Mining once can be used in place of the mining scheduler

### Risks + pitfalls

- Possibly quite a bit of work

PR #2984, opened 2019-06-21
Fix for setting block time in FilecoinOpts

### LINKED ISSUE #2900: [bug] Providing DaemonOpts to fastesting.NewTestEnvironment somehow causing repo/api not found

shannonwells:
Closes #2900
This is a workaround for the bug found in the above issue. 
I considered a few approaches before settling on this and I'm not thrilled with it, but it seems to be the best of not great solutions.

Bonus: this ought to reduce CI time since FAST is back down to a 1ms default block time.

# Considered alternatives
*  Fixing the source of the bug.  Drilling down to figure out why having two --block-time options sent to RunCmd would cause a failure to find the api file was taking a lot of time.
*  DefaultFilecoinOpts struct, or a CreateDefaultFilecoinOpts function that returns all the defaults, both InitOpts and DaemonOpts. The problem with this is setting up the genesis, which depends on the genesis environment you want to run. These need to be declared together because the URI has to come from that environment instance.
* The struct or function creating the defaults has only DaemonOpts, not InitOpts.  Not ideal because it's not really all the defaults.  Also this and the above would require changing 9 other files and even more tests.

# Chosen solution
* I landed on splitting up the treatment of InitOpts and DaemonOpts in NewTestEnvironment. 
    - InitOpts must append the genesis URL regardless, so this line stays the same.
    - Check DaemonOpts, and if it is empty, we'll set a 1ms block time.  There is no easy and accurate way to match opt functions to determine if there is already a block time option set.  Doing advanced checking with reflections seems like overkill. I did not want to go deeper into FAST code to reject duplicate option strings.  Commandkit already rejects duplicate options on the command line.
- I added a comment that if you do pass in non-empty DaemonOpts you should set your own block time.

### DIFF
--- a/tools/fast/fastesting/basic.go
+++ b/tools/fast/fastesting/basic.go
@@ -35,8 +35,6 @@ type TestEnvironment struct {
 }
 
 // NewTestEnvironment creates a TestEnvironment with a basic setup for writing tests using the FAST library.
-// NOTE: If you provide non-empty DaemonOpts, it should include a block time option,
-// or the go-filecoin default will be used.
 func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.FilecoinOpts) (context.Context, *TestEnvironment) {
 
 	// Create a directory for the test using the test name (mostly for FAST)
@@ -64,10 +62,8 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.File
 
 	fastenvOpts.InitOpts = append([]fast.ProcessInitOption{fast.POGenesisFile(genesisURI)}, fastenvOpts.InitOpts...)
 
-	// The old method appended a default blocktime and this caused a bug when callers set blocktime
-	// as well. See Issue
-	if len(fastenvOpts.DaemonOpts) == 0 {
-		fastenvOpts.DaemonOpts = []fast.ProcessDaemonOption{fast.POBlockTime(time.Millisecond)}
+	if isMissingBlockTimeOpt(fastenvOpts) {
+		fastenvOpts.DaemonOpts = append([]fast.ProcessDaemonOption{fast.POBlockTime(time.Millisecond)}, fastenvOpts.DaemonOpts...)
 	}
 
 	// Setup the first node which is used to help coordinate the other nodes by providing
@@ -162,4 +158,14 @@ func dumpEnvOutputOnFail(t *testing.T, procs []*fast.Filecoin) {
 		}
 		require.NoError(t, w.Close())
 	}
+}
+
+func isMissingBlockTimeOpt(opts fast.FilecoinOpts) bool {
+	for _, fn := range opts.DaemonOpts {
+		s := fn()
+		if len(s) > 0 && s[0] == "--block-time" {
+			return false
+		}
+	}
+	return true
 }
--- a/tools/fast/fastesting/basic.go
+++ b/tools/fast/fastesting/basic.go
@@ -36,6 +36,7 @@ type TestEnvironment struct {
 
 // NewTestEnvironment creates a TestEnvironment with a basic setup for writing tests using the FAST library.
 func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.FilecoinOpts) (context.Context, *TestEnvironment) {
+
 	// Create a directory for the test using the test name (mostly for FAST)
 	// Replace the forward slash as tempdir can't handle them
 	dir, err := ioutil.TempDir("", strings.Replace(t.Name(), "/", ".", -1))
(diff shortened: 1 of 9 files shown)


anorth:
@shannonwells please assign reviewers if you want to land this

shannonwells:
> @shannonwells please assign reviewers if you want to land this

I need to get it passing tests and then got busy & forgot about it... I'll mark it as WIP until I can get back to it

tools/fast/fastesting/basic_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestSetFilecoinOpts(t *testing.T) {
+	tf.IntegrationTest(t)
+	log.SetDebugLogging()
+
+	fastOpts := fast.FilecoinOpts{
+		DaemonOpts: []fast.ProcessDaemonOption{fast.POBlockTime(10 * time.Millisecond)},
+	}
+
+	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fastOpts)
```
frrist: We need a call like the below to avoid leaking processes since this method starts one:
```go
	// Teardown after test ends
	defer func() {
		err := env.Teardown(ctx)
		require.NoError(t, err)
	}()
```

tools/fast/fastesting/basic.go
```diff
@@ -59,9 +62,12 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.File
 	genesisMiner, err := env.GenesisMiner()
 	require.NoError(t, err)
 
-	fastenvOpts = fast.FilecoinOpts{
-		InitOpts:   append([]fast.ProcessInitOption{fast.POGenesisFile(genesisURI)}, fastenvOpts.InitOpts...),
-		DaemonOpts: append([]fast.ProcessDaemonOption{fast.POBlockTime(100 * time.Millisecond)}, fastenvOpts.DaemonOpts...),
+	fastenvOpts.InitOpts = append([]fast.ProcessInitOption{fast.POGenesisFile(genesisURI)}, fastenvOpts.InitOpts...)
+
+	// The old method appended a default blocktime and this caused a bug when callers set blocktime
+	// as well. See Issue
```
frrist: Se should link to the issue or remove this comment

tools/fast/fastesting/basic.go
```diff
@@ -59,9 +62,12 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.File
...
+	if len(fastenvOpts.DaemonOpts) == 0 {
```
frrist: Should we instead explicitly check if the `POBlockTime` opt was passed here? There could be daemon opts passed that are unrelated to block time.

shannonwells: do you know how to do that? I couldn't figure out how

travisperson: I don't think there is a super clean way to actually check if the blocktime was passed. I guess the best thing to do would be to iterator over all `DaemonOpts` invoking them and doing a check for `--block-time`.

This might be the best way to move forward at the moment.

```
isBlockTimeSet := false
for _, fn := range fastenvOpts.DaemonOpts {
	s := fn()
	if len(s) > 0 && s[0] == "--block-time" {
		isBlockTimeSet = true
		break
	}
}
```

Currently the requirement to change the blocktime stems from an issue of having to use `MiningStart` to get around an issue where the storage protocol and sector builder is tied to mining. I have an issue to track this in #2968.

So I'm not sure how common it's going to be to actually want to change the blocktime after we get a fix in. The issue we face is that the actual default of 30s is not what we want to default to in these tests most of the time. So I think regardless we are going to want to be able to detect block-time being passed, and add the daemon opt if it's not provided.

shannonwells: Ok so should this change be isolated to detecting if a block time was passed in and then adding one if it's not?

tools/fast/fastesting/basic_test.go
shannonwells: good catch

tools/fast/fastesting/basic.go
travisperson: Ya, I think everything else in this PR looks good.

PR #2990, opened 2019-06-24
UpdateStorage accepts negative values

ZenGround0:
This PR includes tests demonstrating that UpdateStorage already accepts negative values.

It also refactors state machine test helpers so that we can effectively test UpdateStorage which previously was not tested.

Feedback welcome if leveraging possibility of negative values in `BytesAmount` is too hacky and negative values should be more explicitly created in that type.

### DIFF
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -15,7 +15,6 @@ import (
 	. "github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/consensus"
-	"github.com/filecoin-project/go-filecoin/core"
 	"github.com/filecoin-project/go-filecoin/state"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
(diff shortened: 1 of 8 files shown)


actor/builtin/miner/miner_test.go
```diff
@@ -15,47 +16,22 @@ import (
...
 	"github.com/filecoin-project/go-filecoin/consensus"
-	"github.com/filecoin-project/go-filecoin/core"
 	"github.com/filecoin-project/go-filecoin/state"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/vm"
 )
 
-func createTestMiner(t *testing.T, st state.Tree, vms vm.StorageMap, minerOwnerAddr address.Address, pid peer.ID) address.Address {
```
ZenGround0: Moved these to testhelpers so we can use them from within storagemarket_test

core/testing.go
```diff
@@ -2,39 +2,14 @@ package core
...
-	"github.com/filecoin-project/go-filecoin/actor"
-	"github.com/filecoin-project/go-filecoin/actor/builtin"
-	"github.com/filecoin-project/go-filecoin/address"
-	"github.com/filecoin-project/go-filecoin/consensus"
-	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
-	"github.com/filecoin-project/go-filecoin/vm"
 )
 
-// MustGetNonce returns the next nonce for an actor at an address or panics.
```
ZenGround0: These changes pay down some technical debt. For more discussion see #2992

testhelpers/consensus.go
```diff
@@ -185,15 +185,21 @@ func newMessageApplier(smsg *types.SignedMessage, processor *consensus.DefaultPr
 	return nil, err
 }
 
-// CreateAndApplyTestMessage wraps the given parameters in a message and calls ApplyTestMessage
-func CreateAndApplyTestMessage(t *testing.T, st state.Tree, vms vm.StorageMap, to address.Address, val, bh uint64, method string, ancestors []types.TipSet, params ...interface{}) (*consensus.ApplicationResult, error) {
+// CreateAndApplyTestMessageFrom wraps the given parameters in a message and calls ApplyTestMessage.
```
ZenGround0: With new storagemarket tests for the first time the `From` address of a message has become semantically important so this refactor allows tests to set this field.

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -156,10 +156,10 @@ func (sma *Actor) CreateStorageMiner(vmctx exec.VMContext, sectorSize *types.Byt
 	return ret.(address.Address), 0, nil
 }
 
-// UpdatePower is called to reflect a change in the overall power of the network.
+// UpdateStorage is called to reflect a change in the overall power of the network.
```
laser: [Very good.](https://github.com/filecoin-project/specs/blob/master/actors.md#updatestorage).

ZenGround0:
Hey @acruikshank I couldn't find any other tests for insufficient gas to draw inspiration from.  It turned out to be non-trivial for me to add this in for UpdateStorage given the existing testing framework, especially because the test messages need to come from the miner actor.  I decided to not block on this and have added an issue here for testing actors against queries with insufficient gas here: #2995.

Issue #2889, opened 2019-06-04
Add fault and recovered sets to SubmitPost


ZenGround0:
### Description
FaultSets are a simple data structure wrapping a Bitfield and a BlockHeight.  

### Acceptance criteria
Implement this wrapper.  
Add it as a parameter to the submitPoSt state machine function signature. 
Pass in dummy value of no faults from the storage miner protocol code.

### Risks + pitfalls
Representing the 0 value might be a bit weird.
This depends on on Bitfield encodings being implemented: #2868 

### Where to begin
See [this relevant design doc](https://docs.google.com/document/d/1aE5a-QZojprMkig6IyYkwV1SrMDDLOnNBqNdnwpu9tk/edit)

ZenGround0:
Note, there are some todos introduced in #2999 referencing this issue as the place to change IntSets to be serialized as RLE+.  That might be better split into a preceding PR.

ingar:
Question about implementing the [`FaultSet`](https://github.com/filecoin-project/specs/blob/master/data-structures.md#faultset) as described in the spec:
```
type FaultSet struct {

        // offset from start of proving period
	index UInt

        // sector ids that had faults
	bitField BitField
}
```
Currently, the rust sector builder [doesn't return any notion of a block height](https://github.com/filecoin-project/go-filecoin/blob/707e7c31406ca1127ee380fee2598221cfbdb063/proofs/sectorbuilder/rustsectorbuilder.go#L390-L393).  Is having the block height needed? Will we at some point be getting a slice of faults back from rust-fil-proofs?  Should we ignore it for now?

@laser @ZenGround0 @dignifiedquire

laser:
Now that I look at it, FaultSet is an interesting type. What does the “index” field indicate? The block height at which a storage fault is detected? Is it expected that a single Rational PoSt generate call produce more than one FaultSet? What about Beacon? or Surprise?

Of all the PoSt constructions I know about, Beacon seems like the only one which would be able to produce more than one FaultSet (and, at maximum, one FaultSet per “step”). The other constructions either learn of the chain height once, when PoSt generation begins or, as is the case with Surprise, at some single point during the proving period.

ZenGround0:
As @laser makes clear in the comment above this is likely to swirl as the PoSt construction develops.  We should feel free to choose something for now and expect it to change later.

PR #2999, opened 2019-06-25
Add NextDoneSet and ProvingSet abstractions to miner actor

ZenGround0:
This PR adds the NextDoneSet and ProvingSet abstractions to the miner actor.  It is best read commit by commit.  The first commit introduces a done parameter into submitPoSt.  The second adds the NextDoneSet abstraction.  The third introduces a SectorSet abstraction to keep the next commit clean.  The fourth introduces the ProvingSet.

The edge case of the ProvingSet being updated during the first commitSector message is not handled in this PR.  I'm working on this in an upcoming followup.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -3,7 +3,6 @@ package miner
 import (
 	"bytes"
 	"math/big"
-	"strconv"
 
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
@@ -58,8 +57,8 @@ const PieceInclusionGracePeriodBlocks = 10000
 const (
 	// ErrInvalidSector indicates and invalid sector id.
 	ErrInvalidSector = 34
-	// ErrSectorCommitted indicates the sector has already been committed.
-	ErrSectorCommitted = 35
+	// ErrSectorIDInUse indicates a sector has already been committed at this ID.
+	ErrSectorIDInUse = 35
 	// ErrStoragemarketCallFailed indicates the call to commit the deal failed.
 	ErrStoragemarketCallFailed = 36
 	// ErrCallerUnauthorized signals an unauthorized caller.
@@ -81,7 +80,7 @@ const (
 // Errors map error codes to revert errors this actor may return.
 var Errors = map[uint8]error{
 	ErrInvalidSector:           errors.NewCodedRevertErrorf(ErrInvalidSector, "sectorID out of range"),
-	ErrSectorCommitted:         errors.NewCodedRevertErrorf(ErrSectorCommitted, "sector already committed"),
+	ErrSectorIDInUse:           errors.NewCodedRevertErrorf(ErrSectorIDInUse, "sector already committed at this ID"),
 	ErrStoragemarketCallFailed: errors.NewCodedRevertErrorf(ErrStoragemarketCallFailed, "call to StorageMarket failed"),
 	ErrCallerUnauthorized:      errors.NewCodedRevertErrorf(ErrCallerUnauthorized, "not authorized to call the method"),
 	ErrInsufficientPledge:      errors.NewCodedRevertErrorf(ErrInsufficientPledge, "not enough pledged"),
@@ -142,7 +141,17 @@ type State struct {
 	// stringified.
 	//
 	// See also: https://github.com/polydawn/refmt/issues/35
-	SectorCommitments map[string]types.Commitments
+	SectorCommitments SectorSet
+
+	// NextDoneSet is a set of sectorIDs reported during the last PoSt
+	// submission as being 'done'.  The collateral for them is still being
+	// held until the next PoSt submission in case early sector removal
+	// penalization is needed.
+	NextDoneSet types.IntSet
+
+	// ProvingSet is the set of sectorIDs this miner is currently proving.
+	// It is only updated when a PoSt is submitted.
+	ProvingSet types.IntSet
 
 	LastUsedSectorID uint64
 
@@ -169,7 +178,9 @@ func NewState(owner, worker address.Address, pid peer.ID, sectorSize *types.Byte
 		Owner:             owner,
 		Worker:            worker,
 		PeerID:            pid,
-		SectorCommitments: make(map[string]types.Commitments),
+		SectorCommitments: NewSectorSet(),
+		NextDoneSet:       types.EmptyIntSet(),
+		ProvingSet:        types.EmptyIntSet(),
 		Power:             types.NewBytesAmount(0),
 		NextAskID:         big.NewInt(0),
 		SectorSize:        sectorSize,
@@ -241,7 +252,7 @@ var minerExports = exec.Exports{
 		Return: []abi.Type{abi.BytesAmount},
 	},
 	"submitPoSt": &exec.FunctionSignature{
-		Params: []abi.Type{abi.PoStProofs},
+		Params: []abi.Type{abi.PoStProofs, abi.IntSet},
 		Return: []abi.Type{},
 	},
 	"changeWorker": &exec.FunctionSignature{
@@ -455,7 +466,7 @@ func (ma *Actor) GetSectorCommitments(ctx exec.VMContext) (map[string]types.Comm
 
 	var state State
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		return state.SectorCommitments, nil
+		return (map[string]types.Commitments)(state.SectorCommitments), nil
 	})
 	if err != nil {
 		return map[string]types.Commitments{}, errors.CodeError(err), err
@@ -538,13 +549,8 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 			return nil, Errors[ErrCallerUnauthorized]
 		}
 
-		// TODO: use uint64 instead of this abomination, once refmt is fixed
-		// https://github.com/polydawn/refmt/issues/35
-		sectorIDstr := strconv.FormatUint(sectorID, 10)
-
-		_, ok := state.SectorCommitments[sectorIDstr]
-		if ok {
-			return nil, Errors[ErrSectorCommitted]
+		if state.SectorCommitments.Has(sectorID) {
+			return nil, Errors[ErrSectorIDInUse]
 		}
 
 		// make sure the miner has enough collateral to add more storage
@@ -568,8 +574,9 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 		copy(comms.CommD[:], commD)
 		copy(comms.CommR[:], commR)
 		copy(comms.CommRStar[:], commRStar)
+
 		state.LastUsedSectorID = sectorID
-		state.SectorCommitments[sectorIDstr] = comms
+		state.SectorCommitments.Add(sectorID, comms)
 		_, ret, err := ctx.Send(address.StorageMarketAddress, "updateStorage", types.ZeroAttoFIL, []interface{}{inc})
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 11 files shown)


actor/builtin/miner/miner_test.go
```diff
@@ -384,11 +384,58 @@ func TestMinerCommitSector(t *testing.T) {
...
+
+	// submit post
+	res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, firstCommitBlockHeight+5, "submitPoSt", ancestors, []types.PoStProof{proof}, done)
+	assert.NoError(t, err)
+	assert.NoError(t, res.ExecutionError)
+	assert.Equal(t, uint8(0), res.Receipt.ExitCode)
+
+	// Check that the state NextDoneSet is updated
+	miner := state.MustGetActor(st, minerAddr)
+	storage := vms.NewStorage(minerAddr, miner)
```
ZenGround0: Please let me know if there is a better way to look at actor state.  This seems sloppy

actor/builtin/miner/miner_test.go
```diff
@@ -384,11 +384,58 @@ func TestMinerCommitSector(t *testing.T) {
 		// fail because commR already exists
 		res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, 4, "commitSector", nil, uint64(1), commD, commR, commRStar, th.MakeRandomBytes(types.TwoPoRepProofPartitions.ProofLen()))
 		require.NoError(t, err)
-		require.EqualError(t, res.ExecutionError, "sector already committed")
+		require.EqualError(t, res.ExecutionError, "sector already committed at this ID")
 		require.Equal(t, uint8(0x23), res.Receipt.ExitCode)
 	})
 }
 
+func TestMinerSubmitPoStNextDoneChanges(t *testing.T) {
```
ZenGround0: I made a new test instead of a new `t.Run` subfunction to try to prevent further complicating the `TestMinerSubmitPoSt` setup.  This seemed worth it to me but I'd appreciate feedback on best practices for actor tests if anything about this seems ugly.

types/intset.go
```diff
@@ -1,9 +1,35 @@
...
 )
 
+func init() {
+	cbor.RegisterCborType(intSetAtlasEntry)
+}
+
+var intSetAtlasEntry = atlas.BuildEntry(IntSet{}).Transform().
+	TransformMarshal(atlas.MakeMarshalTransformFunc(
+		func(is IntSet) ([]byte, error) {
+			// TODO #2889: we should be using RLE+ to serialize intsets when it lands,
```
ZenGround0: cc @Ingar when taking #2889 these calls should be updated with `is.ToRLE()` and `IntSetFromRLE()` (or whatever interface you go with)

actor/builtin/miner/sector_set.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// Note that this type's interface is a bit of a placeholder.  We expect to
+// move to something like actor.Lookup eventually which will manage its own
+// hash linked storage.  See #2887.  This will require some interface changes.
+
+// SectorSet is a collection of sector commitments indexed by their integer
+// sectorID.  Due to a bug in refmt, the sector id-keys need to be stringified.
+// See also: https://github.com/polydawn/refmt/issues/35.
+type SectorSet map[string]types.Commitments
```
ZenGround0: This type decouples the SectorSet's implementation from its interface a bit better than before which helps with updating the miner actor's commitments and proving sets.  This type doesn't capture the interface we can expect in the near future which will need to pass around the miner's storage and return cids.

actor/builtin/miner/miner.go
```diff
@@ -719,7 +726,7 @@ func (ma *Actor) GetPower(ctx exec.VMContext) (*types.BytesAmount, uint8, error)
 
 // SubmitPoSt is used to submit a coalesced PoST to the chain to convince the chain
 // that you have been actually storing the files you claim to be.
-func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof) (uint8, error) {
+func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, done types.IntSet) (uint8, error) {
```
laser: QUESTION: I observe that [the spec uses the `BitField` type](https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost) (an RLE+ encoded bit set) where you've used an `IntSet` (a slice of u64). Is this a temporary divergence from the spec (perhaps because we don't yet have a `BitField` type with corresponding set operations)? If not, would you upstream this change to the spec?

ZenGround0: This is not temporary and I have considered upstreaming changes to the spec.  The issue is that the spec doesn't really distinguish between encodings of VM data and internal VM datastructures and the Bitfield is just the encoding of an IntSet.  I've been told that trying to put specifics of VM internals (like how the IntSet is implemented) into the spec is out of scope because that gets into implementation choices.  So there is a fine line I don't have a good feel for.  I don't think I am the right person to start introducing this subtle distinction into the spec or that it would even be a welcome addition.

laser: > I've been told that trying to put specifics of VM internals (like how the IntSet is implemented) into the spec is out of scope because that gets into implementation choices.

I guess all that really matters is that go-filecoin can send and receive `submitPoSt` messages whose bytes are conformant with the specification (i.e. go-filecoin knows how to send/receive RLE+ bit sets, in this context). If we choose to transform the decoded bit set into a slice of uint64s (in Go), then that's an implementation detail of go-filecoin that doesn't need upstreaming. Does this overlap with what you're thinking, @ZenGround0?

ZenGround0: Yes this lines up.  And once the RLE+ PR lands all IntSet parameters will be RLE+ encoded.

actor/builtin/miner/miner.go
```diff
@@ -536,13 +547,9 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 			return nil, Errors[ErrCallerUnauthorized]
 		}
 
-		// TODO: use uint64 instead of this abomination, once refmt is fixed
-		// https://github.com/polydawn/refmt/issues/35
-		sectorIDstr := strconv.FormatUint(sectorID, 10)
-
-		_, ok := state.SectorCommitments[sectorIDstr]
+		ok := state.SectorCommitments.Has(sectorID)
```
anorth: "ok" is a bad name for this. How about "has" or "committed"?

actor/builtin/miner/miner_test.go
```diff
@@ -384,11 +384,108 @@ func TestMinerCommitSector(t *testing.T) {
...
+	res, err := th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, firstCommitBlockHeight, "commitSector", ancestors, uint64(1), th.MakeCommitment(), th.MakeCommitment(), th.MakeCommitment(), th.MakeRandomBytes(types.TwoPoRepProofPartitions.ProofLen()))
+	require.NoError(t, err)
+	require.NoError(t, res.ExecutionError)
+	require.Equal(t, uint8(0), res.Receipt.ExitCode)
+
+	// add another sector
+	res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, firstCommitBlockHeight+1, "commitSector", ancestors, uint64(2), th.MakeCommitment(), th.MakeCommitment(), th.MakeCommitment(), th.MakeRandomBytes(types.TwoPoRepProofPartitions.ProofLen()))
+	require.NoError(t, err)
+	require.NoError(t, res.ExecutionError)
+	require.Equal(t, uint8(0), res.Receipt.ExitCode)
```
anorth: Check ProvingSet is empty

actor/builtin/miner/sector_set.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil
+}
+
+// IDs returns all of the sectorIDs in the sectorset.  They are not sorted.
+func (ss SectorSet) IDs() ([]uint64, error) {
+	var ids []uint64
+	for idStr := range ss {
+		id, err := str2ID(idStr)
+		if err != nil {
+			return nil, errors.RevertErrorWrap(err, "corrupt sectorset id")
```
anorth: I would expect this wrapping to be done at the VM boundary: code inside the actor can either succeed or fail, and any failure will revert the transaction but not be a "fault" error. The actor code in general should not depend on these error abstractions.

actor/builtin/miner/sector_set.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		ids = append(ids, id)
+	}
+	return ids, nil
+}
+
+// TODO: use uint64 as map keys instead of this abomination, once refmt is fixed.
+// https://github.com/polydawn/refmt/issues/35
```
anorth: Please file a tracking issue in go-filecoin too

anorth:
Sorry for the delay

actor/builtin/miner/miner.go
```diff
@@ -144,7 +143,17 @@ type State struct {
...
+
+	// NextDoneSet is a set of sectorIDs reported during the last PoSt
+	// submission as being 'done'.  The collateral for them is still being
+	// held until the next PoSt submission in case early sector removal
+	// penalization is needed.
+	NextDoneSet types.IntSet
+
+	// ProvingSet is the set of sectorIDs this miner is currently proving.
+	// It is only updated when a PoSt is submitted.
+	ProvingSet types.IntSet
```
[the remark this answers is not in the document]
ZenGround0: We have different ideas about what the proving set entails.  The way I read the spec the ProvingSet is the set of `sectorID`s the miner has committed and must prove to the network during this proving period.  `sectorID`s of sectors without any PoSts enter into the proving set all the time after a miner first commits sectors.  This is the way this code is written too.  I don't know where the "at least one PoSt" requirement is coming from.

actor/builtin/miner/sector_set.go
ZenGround0: Am I correctly reading this as a suggestion that we not call `errors.NewRevertError` or `errors.NewFaultError` from within actor code?  If so this is a big technical debt concern because we do this very commonly in actor code.

Am I misreading this?  If so could you please clarify the concrete change you'd like to see?

actor/builtin/miner/sector_set.go
ZenGround0: #3007

actor/builtin/miner/miner.go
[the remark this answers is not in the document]
acruikshank: > sectorIDs of sectors without any PoSts enter into the proving set all the time after a miner first commits sectors. 

My reading of the spec is that a sector enters the proving set:

1. When the sector is committed if the ProvingSet is empty. In this case the proving period is also reset.
2. At the next SubmitPoSt if proving is already underway (the ProvingSet is non-empty).

This logic means a sector will never be added to a proving set in the middle of a proving period and that proving periods will start as soon as possible. I think it's a bit too much detail to include when commenting the declaration of of the variable. My preference would be to simply remove the "It is only updated when a PoSt is submitted" and leave "ProvingSet is the set of scctorIDs this miner is currently proving", which is at least accurate.

actor/builtin/miner/sector_set.go
acruikshank: The way the VM is set up, it requires actors to declare errors to be reverts or faults. This makes some sense because this determination can't be made externally, but it would be nice if faults had to be wrapped and any non-fault error was treated as a revert. I think this system was set up for safety (forcing the developer to choose).

anorth: !! @ZenGround0 you interpreted correctly but I misunderstood the current convention and am now very alarmed by it.

> set up for safety

But it's not! Nothing forces the developer to choose because the methods are declared to return simply `error`. A quick grep shows at least five calls to methods like `errors.{New|Wrap|Errorf}` in actor code. I would bet large amounts of money that there are errors returned from code called by actors that are not wrapped. 

Also, this is supposed to be a VM, how can anything from inside here be intentionally causing a node to shutdown? If such a case is part of message execution, a node could only crash-loop until the message expired from all pools. That's a horrible DOS vector.

anorth: It looks like "faults" are not handled as harshly as they're documented to be, so a node won't shutdown. I've filed #3009 to clear this all up (later).

PR #3002, opened 2019-06-27
Update power based on previous proving set

ZenGround0:
## What's in this PR?
- move power and total storage updates to submitPoSt
- add power update tests
- proving set initialized to first sector on first PoSt submit
- address failures from out-dated power setting assumptions

## FYI
This is based against feat/2948 in PR at #2999 and you'll need to green light it again that one merges and I change the base to master.  Might be worth waiting for #2999 to merge until you review this.

Initializing the proving set to the first committed sector during commitSector is something some might consider in scope for #2999 but I added it here.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -137,20 +137,22 @@ type State struct {
 	NextAskID *big.Int
 
 	// SectorCommitments maps sector id to commitments, for all sectors this
-	// miner has committed. Due to a bug in refmt, the sector id-keys need to be
+	// miner has committed.  Sector ids are removed from this collection
+	// when they are included in the done or fault parameters of submitPoSt.
+	// Due to a bug in refmt, the sector id-keys need to be
 	// stringified.
 	//
 	// See also: https://github.com/polydawn/refmt/issues/35
 	SectorCommitments SectorSet
 
-	// NextDoneSet is a set of sectorIDs reported during the last PoSt
+	// NextDoneSet is a set of sector ids reported during the last PoSt
 	// submission as being 'done'.  The collateral for them is still being
 	// held until the next PoSt submission in case early sector removal
 	// penalization is needed.
 	NextDoneSet types.IntSet
 
-	// ProvingSet is the set of sectorIDs this miner is currently proving.
-	// It is only updated when a PoSt is submitted.
+	// ProvingSet is the set of sector ids of sectors this miner is 
+	// currently required to prove.
 	ProvingSet types.IntSet
 
 	LastUsedSectorID uint64
@@ -159,7 +161,8 @@ type State struct {
 	ProvingPeriodEnd *types.BlockHeight
 	LastPoSt         *types.BlockHeight
 
-	// The amount of space committed to the network by this miner.
+	// The amount of space proven to the network by this miner in the
+	// latest proving period.
 	Power *types.BytesAmount
 
 	// SectorSize is the amount of space in each sector committed to the network
@@ -561,11 +564,18 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 
 		state.ActiveCollateral = state.ActiveCollateral.Add(collateral)
 
-		if state.Power.Equal(types.NewBytesAmount(0)) {
+		// Case 1: If the miner is not currently proving any sectors,
+		// start proving immediately on this sector.
+		//
+		// Case 2: If the miner is adding sectors during genesis
+		// construction all committed sectors accumulate in their
+		// proving set.  This  allows us to add power immediately in
+		// genesis with commitSector and submitPoSt calls without 
+		// adding special casing for bootstrappers.
+		if state.ProvingSet.Size() == 0 || ctx.BlockHeight().Equal(types.NewBlockHeight(0)) {
+			state.ProvingSet = state.ProvingSet.Add(sectorID)
 			state.ProvingPeriodEnd = ctx.BlockHeight().Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
 		}
-		inc := state.SectorSize
-		state.Power = state.Power.Add(inc)
 		comms := types.Commitments{
 			CommD:     types.CommD{},
 			CommR:     types.CommR{},
@@ -577,13 +587,6 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 
 		state.LastUsedSectorID = sectorID
 		state.SectorCommitments.Add(sectorID, comms)
-		_, ret, err := ctx.Send(address.StorageMarketAddress, "updateStorage", types.ZeroAttoFIL, []interface{}{inc})
-		if err != nil {
-			return nil, err
-		}
-		if ret != 0 {
-			return nil, Errors[ErrStoragemarketCallFailed]
-		}
 		return nil, nil
 	})
 	if err != nil {
@@ -839,18 +842,33 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 		state.ProvingPeriodEnd = state.ProvingPeriodEnd.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
 		state.LastPoSt = chainHeight
 
+		// Update miner power to the amount of data actually proved
+		// during the last proving period.
+		oldPower := state.Power
+		// TODO subtract total faulted size from ProvingSet size #2889
+		newPower := types.NewBytesAmount(uint64(state.ProvingSet.Size())).Mul(state.SectorSize)
+		state.Power = newPower
+		delta := newPower.Sub(oldPower)
+		_, ret, err := ctx.Send(address.StorageMarketAddress, "updateStorage", types.ZeroAttoFIL, []interface{}{delta})
+		if err != nil {
+			return nil, err
+		}
+		if ret != 0 {
+			return nil, Errors[ErrStoragemarketCallFailed]
+		}
+
 		// Update SectorSet, DoneSet and ProvingSet
 		if err = state.SectorCommitments.Drop(done.Values()); err != nil {
 			return nil, err
 		}
-		
+
 		sectorIDsToProve, err := state.SectorCommitments.IDs()
 		if err != nil {
 			return nil, err
 		}
 		state.ProvingSet = types.NewIntSet(sectorIDsToProve...)
 		state.NextDoneSet = done
-		
+
 		return nil, nil
 	})
 	if err != nil {
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -461,103 +461,347 @@ func TestMinerCommitSector(t *testing.T) {
 	})
 }
 
-func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
-	tf.UnitTest(t)
+// minerActorLiason provides a set of test friendly calls for setting up, reading
+// internals, and transitioning the portion of the filecoin state machine
+// related to a particular miner actor.
+type minerActorLiason struct {
+	st state.Tree
+	vms vm.StorageMap
+	ancestors []types.TipSet
+	minerAddr address.Address
+	t *testing.T
+	currentHeight uint64
+}
 
+func (mal *minerActorLiason) requireHeightNotPast(blockHeight uint64) {
+	require.True(mal.t, blockHeight >= mal.currentHeight)
+	mal.currentHeight = blockHeight
+}
+
+func (mal *minerActorLiason) requireCommit(blockHeight, sectorID uint64) {
+	mal.requireHeightNotPast(blockHeight)
+	res, err := th.CreateAndApplyTestMessage(mal.t, mal.st, mal.vms, mal.minerAddr, 0, blockHeight, "commitSector", mal.ancestors, sectorID, th.MakeCommitment(), th.MakeCommitment(), th.MakeCommitment(), th.MakeRandomBytes(types.TwoPoRepProofPartitions.ProofLen()))
+	require.NoError(mal.t, err)
+	require.NoError(mal.t, res.ExecutionError)
+	require.Equal(mal.t, uint8(0), res.Receipt.ExitCode)
+}
+
+func (mal *minerActorLiason) requirePoSt(blockHeight uint64, done types.IntSet) {
+	mal.requireHeightNotPast(blockHeight)
+	res, err := th.CreateAndApplyTestMessage(mal.t, mal.st, mal.vms, mal.minerAddr, 0, blockHeight, "submitPoSt", mal.ancestors, []types.PoStProof{th.MakeRandomPoStProofForTest()}, done)
+	assert.NoError(mal.t, err)
+	assert.NoError(mal.t, res.ExecutionError)
+	assert.Equal(mal.t, uint8(0), res.Receipt.ExitCode)
+}
+
+func (mal *minerActorLiason) requireReadState() State {
+	miner := state.MustGetActor(mal.st, mal.minerAddr)
+	storage := mal.vms.NewStorage(mal.minerAddr, miner)
+	stateBytes, err := storage.Get(storage.Head())
+	require.NoError(mal.t, err)
+	var minerState State
+	err = actor.UnmarshalStorage(stateBytes, &minerState)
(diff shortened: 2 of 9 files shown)


actor/builtin/miner/miner.go
```diff
@@ -560,11 +560,20 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
+		// ther proving period now.  Every miner's first PoSt will be
+		// over a single sector.
+		//
+		// Case 2: if the miner is adding sectors during genesis
+		// construction all committed sectors accumulate in their
+		// proving set.  This is a bit of a hack but it is fairly
+		// clean and allows us to add power immediately in genesis
+		// with commitSector and submitPoSt calls without adding
+		// additional special casing for bootstrappers.
+		if state.SectorCommitments.Size() == 0 || ctx.BlockHeight().Equal(types.NewBlockHeight(0)) {
```
ZenGround0: This is likely the most controversial thing in this PR.  I was pleased with the simplicity and lack of special casing involved in this solution.  Please take a look and tell me if there is something flawed about it.

actor/builtin/miner/miner_test.go
```diff
@@ -418,6 +421,21 @@ func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
 	require.NoError(t, res.ExecutionError)
 	require.Equal(t, uint8(0), res.Receipt.ExitCode)
 
+	t.Run("proving-set contains only the first sector committed before submitPoSt", func(t *testing.T) {
```
ZenGround0: Is it better for me to break off the same `commitSector` and `submitPoSt` flow into another test that only tests power?

gengen/util/gengen.go
```diff
@@ -292,6 +292,19 @@ func setupMiners(st state.Tree, sm vm.StorageMap, keys []*types.KeyInfo, miners
 				return nil, err
 			}
 		}
+		if m.NumCommittedSectors > 0 {
```
ZenGround0: Because power is now added during `submitPoSt` this is needed for setting power in the genesis block.  Again let me know if this bootstrapping solution is flawed.

acruikshank: This works for me.

actor/builtin/miner/miner.go
```diff
@@ -813,20 +815,40 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 
 		// transition to the next proving period
 		state.ProvingPeriodEnd = state.ProvingPeriodEnd.Add(types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize)))
+		if ctx.BlockHeight().Equal(types.NewBlockHeight(0)) {
```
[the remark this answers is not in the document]
ZenGround0: @acruikshank Yeah that sounds good, I also don't like this.  I pushed this up to try to fix a test failure and didn't expect anyone to look at it today :)

Are you ok with just removing this and not doing the other special casing, or would you prefer that we specially set the ProvingPeriodEnd low during genesis commitSectors as described above?

actor/builtin/miner/miner_test.go
anorth: Probably yes. Adding more to this chain of update and assertions will get hard to maintain

actor/builtin/miner/miner_test.go
```diff
@@ -418,6 +421,21 @@ func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
...
+		// Check that state.ProvingSet is now [1, 2] and NextDoneSet is []
```
anorth: Misleading comment

actor/builtin/miner/miner.go
```diff
@@ -560,11 +560,20 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
...
+		// over a single sector.
+		//
+		// Case 2: if the miner is adding sectors during genesis
+		// construction all committed sectors accumulate in their
+		// proving set.  This is a bit of a hack but it is fairly
+		// clean and allows us to add power immediately in genesis
+		// with commitSector and submitPoSt calls without adding
+		// additional special casing for bootstrappers.
+		if state.SectorCommitments.Size() == 0 || ctx.BlockHeight().Equal(types.NewBlockHeight(0)) {
+			state.ProvingSet = state.ProvingSet.Add(sectorID)
```
anorth: Blocking: this muddies the semantics of ProvingSet. Before this change, it identified the set of sectors a miner has submitted at least one proof for (and not yet stopped/failed), and so was a proxy for power. Now it can be non-empty when a miner has no power.

I'm ok with bootstrap miners having a special case, but not all miners. Shouldn't boostrap miners have a `state.Power = xxx` here?

actor/builtin/miner/miner.go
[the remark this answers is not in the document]
acruikshank: I think we'll panic when the bootstrap miner submits a post in that case. In fact, I think we have a bug where we'll panic if a miner submits a post before committing a sector in any case. Initializing LastProvingPeriod to zero when we initialize the state should be fine to fix both problems..

actor/builtin/miner/miner.go
ZenGround0: > Before this change, it identified the set of sectors a miner has submitted at least one proof for

This is not true.  Before this change the proving set identifies the set of sectors the miner is required to submit a PoSt for over the current proving period.  The fact that it is non-empty when a miner has no power is expected during the edge case when an actor first adds power.

This non-bootstrap case is coming from [the spec](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector).  If we want to improve these ideas I think filing a spec issue or spec PR is the right way forward.

> Shouldn't boostrap miners have a state.Power = xxx here?

This is probably a good idea eventually.  I wrote it this way because the change required very little special casing.  I have a feeling we should eventually be crafting genesis blocks without calling actor code at all (and setting `state.Power = xxx` in gengen code) in order to avoid lugging around genesis special casing during live execution.

acruikshank: @ZenGround0's reading of the spec is consistent with my understanding. 

It's probably worth noting that checking `SectorCommitments` size instead of `ProvingSet` is inconsistent with the spec. Given the logic implemented here and the fact that SectorCommitments are a superset of ProvingSet, `SectorCommitments.Size() == 0` is equivalent statement to `ProvingSet.Size() == 0`, but it'd be nice if developers didn't have to work that out. If it's unreasonable to get the `ProvingSet` size, the equivalence should be justified in a comment.

ZenGround0: @acruikshank that's a good point.  It was just a mistake and will be easy to fix to match the spec.

commands/deals_daemon_test.go
```diff
@@ -260,12 +260,10 @@ func TestDealsShowPaymentVouchers(t *testing.T) {
 	price := big.NewFloat(0.000000001)      // price per byte/block
 	expiry := big.NewInt(24 * 60 * 60 / 30) // ~24 hours
 
-	// This also starts the Miner
```
ZenGround0: Thanks @shannonwells for this patch!

actor/builtin/miner/miner_test.go
```diff
@@ -490,6 +493,120 @@ func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
 	require.NoError(t, res.ExecutionError)
 	require.Equal(t, uint8(0), res.Receipt.ExitCode)
 
+	t.Run("power is 0 until first submitPoSt", func(t *testing.T) {
```
acruikshank: It's a little weird to include a bunch of subtests in a test that depend on each other. I usually expect subtests to be independent. I think writing tests that assert a sequence of things is fine (though some disagree). I would:

1. Remove the subtests and replace the descriptions with comments.
2. Create a test struct with st, vms, and minerAddress.
3. Give it a methods that can call submitPoSt or commitSector so that it accept the message parameters, runs the message and then return the miner's power and the total storage power.
4. Replace the 7 calls in this class with that method.

That should make this test pretty short and easy to understand.

actor/builtin/miner/miner_test.go
```diff
@@ -490,6 +493,120 @@ func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NoError(t, res.ExecutionError)
+	require.Equal(t, uint8(0), res.Receipt.ExitCode)
+
+	// add another sector
+	res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, firstCommitBlockHeight+1, "commitSector", ancestors, uint64(2), th.MakeCommitment(), th.MakeCommitment(), th.MakeCommitment(), th.MakeRandomBytes(types.TwoPoRepProofPartitions.ProofLen()))
+	require.NoError(t, err)
+	require.NoError(t, res.ExecutionError)
+
+	t.Run("proving-set contains only the first sector committed before submitPoSt", func(t *testing.T) {
```
acruikshank: See last comment. If the method described above could also return the proving set from state, you could use it throughout this test as well. There are fewer dependencies between subtests here, but I still don't like them.

actor/builtin/miner/miner_test.go
ZenGround0: Thanks, this is all great advice.  Here are some meta-notes

I copied the "sequential dependent subtest" pattern from TestMinerSubmitPoSt.  I assumed this was the way things were done in miner actor testing land and leaned on this structure to get the new tests written.  The tests I wrote are a good example of how bad habits multiply when new people jump into half-baked testing packages without realizing they are incomplete.

Your PR comment is a good example of what we need to do to prevent these bad habits from ballooning.

In general this project suffers from a lack of reusable quality test frameworks.  1) because the emergent test frameworks are low quality (like the one I just copied) they are burdensome to use to write new tests if you are new to the code. 2) the burden to improve these emergent frameworks falls on people implementing features which slows us all down.

actor/builtin/miner/miner_test.go
```diff
@@ -490,72 +493,195 @@ func TestMinerSubmitPoStSectorSetUpdates(t *testing.T) {
...
+		t:         t,
+		st:        st,
+		vms:       vms,
+		ancestors: ancestors,
+		minerAddr: minerAddr,
+	}
+}
+
+func TestMinerSubmitPoStSectorDoneSetUpdates(t *testing.T) {
+	tf.UnitTest(t)
```
ZenGround0: @acruikshank What do you think of this refactoring?  I separated out submitting and committing from reading the chain state (here I had already made a separate helper, and I'd do something similar for power).

It's almost more confusing to read without the subtests.  It feels like this test is doing too much, but at the same time we need to test longer flows like this.  Any further suggestions?

ZenGround0: Feel free to advise when you get here @anorht

actor/builtin/miner/miner.go
anorth: Before this change, `ProvingSet` was only updated in SubmitPost so could not possibly have referenced a sector not yet proven. I am reading what the code does, not what the spec suggests it should do.

Please update the comment on the state field `ProvingSet` to clearly state the "required to prove" semantics you are now implementing. Delete the sentence "It is only updated when a PoSt is submitted".

Also please update the comment for `SectorCommitments` to account for the fact that sectors are dropped from this collection when "done".

While you're there, update the comment on `Power` to also not be misleading: it's proven, not committed, storage.

actor/builtin/miner/miner_test.go
anorth: The general problem here is that we're attempting to test a range of different conditions by establishing them in sequence and checking those states. This is a brittle pattern and leads to long, hard-to-understand test methods. Better would be to break this into multiple tests (or subtests) that do not share state. We sometimes end up in this place when the state is tedious to establish, and I'm ok with a pragmatic approach there sometimes so long as it doesn't get out of hand. 

This is pretty close to out of hand. I suggest breaking it up.

> we need to test longer flows like this

I don't think that's a goal. We need to test lots of state transitions, but not chaining them together.

acruikshank: Yeah. I agree with @anorth. This is the pattern that works for me:

1. Extract setup to a multi-return function. This can actually be defined at the start of the test.
2. Create a sub-test for each section of this test that contains an assertion (i.e. every comment in here starting with "Check").
3. Call setup in each sub-test (so the sub-tests are independent)
4. Run the minimum number of steps in each sub-test to get to the point where you can test the thing you're trying to test. Ensure that each of these steps is tested elsewhere and don't make any assertions on the intermediate steps.
5. Extract some setup steps into functions if they are repeated enough times (I usually want the total number of lines to half before i decide this is worth it).

The `posterCommitter` is still a nice refactor though.

ZenGround0:
Hey @anorth feel free to take a look and provide any more feedback on testing code.

actor/builtin/miner/miner_test.go
```diff
@@ -461,103 +461,344 @@ func TestMinerCommitSector(t *testing.T) {
...
+	})
+	
+	t.Run("power accumulates over multiple proving periods", func(t *testing.T) {
+		mal := setupMinerActorLiason(t)
+		
+		// Period 1 commit and prove
+		mal.requireCommit(firstCommitBlockHeight, uint64(1))
+		mal.requireCommit(firstCommitBlockHeight+1, uint64(2))
+		done := types.EmptyIntSet()
+		mal.requirePoSt(firstCommitBlockHeight+5, done)
```
anorth: Check the power here (which will be just one sector?)

actor/builtin/miner/miner_test.go
```diff
@@ -461,103 +461,344 @@ func TestMinerCommitSector(t *testing.T) {
...
-	t.Run("proving-set updates to newly added sectors", func(t *testing.T) {
-		// submit post
-		res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, firstCommitBlockHeight+5, "submitPoSt", ancestors, []types.PoStProof{proof}, done)
-		assert.NoError(t, err)
-		assert.NoError(t, res.ExecutionError)
-		assert.Equal(t, uint8(0), res.Receipt.ExitCode)
+		// Period 2 commit and prove
+		mal.requireCommit(secondProvingPeriodStart+1, uint64(16))
+		mal.requireCommit(secondProvingPeriodStart+2, uint64(17))
+		mal.requirePoSt(secondProvingPeriodStart+5, done)
```
anorth: Check power here too

actor/builtin/miner/miner_test.go
```diff
@@ -461,103 +461,344 @@ func TestMinerCommitSector(t *testing.T) {
...
+		// Period 3 prove and drop 1 and 2
+		done = types.NewIntSet(1,2)
+		mal.requirePoSt(thirdProvingPeriodStart+5, done)
 		
-		// submit post
-		res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, secondProvingPeriodStart + 2, "submitPoSt", ancestors, []types.PoStProof{proof}, done)
-		assert.NoError(t, err)
-		assert.NoError(t, res.ExecutionError)
-		assert.Equal(t, uint8(0), res.Receipt.ExitCode)
+		// power lags removal by a proving period
```
anorth: Fascinating...

actor/builtin/miner/miner_test.go
```diff
@@ -461,103 +461,344 @@ func TestMinerCommitSector(t *testing.T) {
...
-		done = done.Add(5)
-		res, err = th.CreateAndApplyTestMessage(t, st, vms, minerAddr, 0, thirdProvingPeriodStart + 2, "submitPoSt", ancestors, []types.PoStProof{proof}, done)		
-		assert.NoError(t, err)
-		assert.NoError(t, res.ExecutionError)
-		assert.Equal(t, uint8(0), res.Receipt.ExitCode)
+	t.Run("only one sector added to proving set during first period", func(t *testing.T) {
+		mal := setupMinerActorLiason(t)
+		// commit several sectors
+		mal.requireCommit(firstCommitBlockHeight, uint64(1))
+		mal.requireCommit(firstCommitBlockHeight+1, uint64(2))
```
anorth: Test what happens when two commits arrive in the firstCommitBlockHeight.

actor/builtin/miner/miner_test.go
ZenGround0: following your advice from before (don't chain together lots of assertions instead separately set up different state) I separated this check out into "power is 1 after first post"

Issue #2509, opened 2019-04-08
Separate devnet DHTs


ZenGround0:
### Description
Currently different devnets have the same peer ids which could cause problems.  One such problem is that DHT records could mistakenly cross networks and cause problems.  One solution for this and other such problems is to namespace filecoin DHTs to reference a specific network.

### Acceptance criteria
We separate DHT namespaces.

### Where to begin
The code change should be a small modification to the protocol id string used for the DHT.  This should be set depending on the network being joined.

anorth:
There are lots of issues about related proposals. Whoever does some work on this, please gather these together and propose a set of changes to resolve them as a group: #1700, #2357, #2421, #2426 

One of these existing issues is probably the right tracking issue. #1700 proposes extracting the network protocol string to a configuration parameter, which could make some of these easy to resolve. But don't get in the way of upcoming protocol versioning and upgrading, which will require more nuance than strictly segregating networks of nodes running different code.

anorth:
Another concern is to ensure that developers can connect their node to any network they want to if explicitly opting in.

anorth:
@acruikshank I think this is ready to implement with the network name.

PR #3012, opened 2019-07-03
Move topics to net package and modify topic string

frrist:
### What
This PR changes the Topic string used for publishing messages and blocks to the network, this will help prevent old invalid chains from entering the network slowing consensus down until the `chain/` package is updated to address this.

This PR also changes the protocol string used for the dht. This address #2509, however there is still room for improvment.


net/protocols.go
```diff
@@ -0,0 +1,8 @@
+package net
+
+import (
+	"github.com/libp2p/go-libp2p-protocol"
+)
+
+// FilecoinDHT is the protocol of the filecoin DHT.
+const FilecoinDHT protocol.ID = "/fil/kad/user-3"
```
ZenGround0: nit: `user` isn't my favorite name, especially since this protocol id will be deployed on nightly.  How about `devnet` or `dev`?

frrist: ` devnet` sounds good to me, addressed

PR #3020, opened 2019-07-04
Reduce message validation failure after re-org to debug level

anorth:
This quiets the "nonce too low" validation failure that is extremely common whenever the chain head switches and the same message/s were mined in both chains.

`08:40:52.287  INFO     mqueue: validation error adding message to pool: nonce too low inbox.go:75`

@ZenGround0 if we cherry-pick anything into 0.3, I propose including this too.

### DIFF
--- a/core/policy.go
+++ b/core/policy.go
@@ -15,7 +15,7 @@ import (
 // pools a little before the sending node gives up on them.
 const OutboxMaxAgeRounds = 10
 
-var log = logging.Logger("mqueue")
+var log = logging.Logger("core")
 
 // QueuePolicy manages a message queue state in response to changes on the blockchain.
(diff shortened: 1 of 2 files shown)


core/policy.go
```diff
@@ -15,7 +15,7 @@ import (
 // pools a little before the sending node gives up on them.
 const OutboxMaxAgeRounds = 10
 
-var log = logging.Logger("mqueue")
+var log = logging.Logger("core")
```
anorth: This logger is shared by both inbound and outbound message handling, so the previous name was misleading and its location in `policy.go` arbitrary. We can do better when the `core` package is reorganized but I don't want to shave that yak today.

PR #3023, opened 2019-07-06
Fault Set Type [#2889]

### LINKED ISSUE #2889: Add fault and recovered sets to SubmitPost

ingar:
## Summary
Add a `FaultSet` type and ABI type registration.  Add a fault set as a parameter to the `SubmitPoSt` method.  Use RLE+ serialization for IntSets.

## TODO
- [x] Verify if [power adjustment](https://github.com/filecoin-project/go-filecoin/blob/14945fe813ba8f74083b295c6f1da07c3594af21/actor/builtin/miner/miner.go#L848) from faults should be done in this PR @ZenGround0 
- [x] If ^^ is true, test power adjustment with faults
- [x] Verify with @laser that rust-fil-proofs can return and take sector ids during PoSt processing.  It may be dealing with indices into the SortedCommR slice.

Resolves #2889

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -966,6 +966,10 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, fa
 			return nil, err
 		}
 
+		if err = state.SectorCommitments.Drop(faults.SectorIds.Values()); err != nil {
+			return nil, err
+		}
+
 		sectorIDsToProve, err := state.SectorCommitments.IDs()
 		if err != nil {
 			return nil, err
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -271,7 +271,7 @@ var minerExports = exec.Exports{
 		Return: []abi.Type{abi.BytesAmount},
 	},
 	"submitPoSt": &exec.FunctionSignature{
-		Params: []abi.Type{abi.PoStProofs, abi.IntSet},
+		Params: []abi.Type{abi.PoStProofs, abi.FaultSet, abi.IntSet},
 		Return: []abi.Type{},
 	},
 	"slashStorageFault": &exec.FunctionSignature{
@@ -854,7 +854,7 @@ func (ma *Actor) GetActiveCollateral(ctx exec.VMContext) (types.AttoFIL, uint8, 
 
 // SubmitPoSt is used to submit a coalesced PoST to the chain to convince the chain
 // that you have been actually storing the files you claim to be.
-func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, done types.IntSet) (uint8, error) {
+func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, faults types.FaultSet, done types.IntSet) (uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -929,7 +929,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 			req := verification.VerifyPoStRequest{
 				ChallengeSeed: seed,
 				SortedCommRs:  sortedCommRs,
-				Faults:        []uint64{},
+				Faults:        faults.SectorIds.Values(),
 				Proofs:        poStProofs,
 				SectorSize:    state.SectorSize,
 			}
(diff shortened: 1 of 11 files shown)


types/fault_set.go
```diff
@@ -0,0 +1,37 @@
+package types
+
+import (
+	"fmt"
+
+	cbor "github.com/ipfs/go-ipld-cbor"
+)
+
+func init() {
+	cbor.RegisterCborType(FaultSet{})
```
ZenGround0: not 100% but I think this is unnecessary because you're not defining a custom atlas for this type

ingar: Yeah, apparently you do, you get an error, unless we enable auto-atlasing.
```
panic: missing an atlas entry describing how to marshal type types.FaultSet (and auto-atlasing for structs is not enabled)
```
Looking at the source it is [unclear](https://github.com/polydawn/refmt/blob/10eebc3c830751b23b9236fc4df06a2d95e0f894/obj/marshalSlab.go#L146) whether this would even work..

laser:
@ingar 

> Verify with @laser that rust-fil-proofs can return and take sector ids during PoSt processing.

We did not discuss modifying the `generate_post` operation, exposed in `libsectorbuilder_ffi.a` to _accept_ sector ids. We did discuss modifying `generate_post` to produce sector ids in its faults output.

That said, I will investigate modifying `generate_post` to accept sector ids (and transform to replica commitments in sector builder code) this afternoon.

actor/builtin/miner/miner.go
```diff
@@ -950,8 +950,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 		// Update miner power to the amount of data actually proved
 		// during the last proving period.
 		oldPower := state.Power
-		// TODO subtract total faulted size from ProvingSet size #2889
-		newPower := types.NewBytesAmount(uint64(state.ProvingSet.Size())).Mul(state.SectorSize)
+		newPower := types.NewBytesAmount(uint64(state.ProvingSet.Size() - faults.SectorIds.Size())).Mul(state.SectorSize)
```
acruikshank: Blocking: faults need to be subtracted from the sector commitments. Otherwise power will be restored on the next submission. See
``` go
	// update sector sets and proving set
	miner.Sectors.Subtract(done)
	miner.Sectors.Subtract(permLostSet)
```
in [the spec](https://github.com/filecoin-project/specs/blob/master/actors.md).

actor/builtin/miner/miner.go
```diff
@@ -950,8 +950,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
...
 		state.Power = newPower
 		delta := newPower.Sub(oldPower)
 		_, ret, err := ctx.Send(address.StorageMarketAddress, "updateStorage", types.ZeroAttoFIL, []interface{}{delta})
```
acruikshank: I know this wasn't your doing, but we really should skip this call if the delta is zero. Would you mind adding that?

ZenGround0: Feel free to add a follow on "good first" issue alternatively

actor/builtin/miner/miner_test.go
```diff
@@ -677,27 +680,55 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
+		// commit several sectors
+		mal.requireCommit(firstCommitBlockHeight, uint64(1))
+		mal.requireCommit(firstCommitBlockHeight, uint64(2))
+		mal.requireCommit(firstCommitBlockHeight, uint64(3))
+
+		done := types.EmptyIntSet()
+
+		// PoSt twice to get power for my sectors
+		mal.requirePoSt(firstCommitBlockHeight+1, done, faults)
+		mal.requirePoSt(firstCommitBlockHeight+1, done, faults)
```
ZenGround0: Oh this is weird.  Your post should be too early for it to give you power.  cc @acruikshank did we get rid of stopping early PoSt submissions (or did these never make it into the implementation in the first place?).  I am pretty sure we don't want early posts like this because then slight gains can accumulate drastically and trivially allow generation attacks given enough proving periods.  Let's check the spec and file an issue about this.

@ingar this shouldn't get in the way of your PR.  Could you change this test to run the PoSt only once per proving period in the same fashion as the other tests so that once the above correction happens this test is still valid?

ZenGround0: @acruikshank of course the PoSt construction might change enough in the next month for this to be irrelevant.  I'll leave it up to you to decide if you think this is stable enough to be worth fixing.

PR #3026, opened 2019-07-08
Chain sampling for challenge seed skips over null blocks.

### LINKED ISSUE #3017: [bug] Challenge seed fails to generate if proving period lands on null block
**Describe the bug**

If a miners proving period start happens to turn out to be a null block, we fail to sample the chain as we can't find the block.

filecoin-3: `failed to calculate PoSt: failed to fetch PoSt challenge seed: error sampling chain for randomness: sample height out of range: 2136`

https://explorer.test.kittyhawk.wtf/blocks/zDPWYqFD5RcMoJ1LDuemGXu9jdeZd1dbDfSSAG8x1SzvnBVX9Gx3

filecoin-2: `failed to calculate PoSt: failed to fetch PoSt challenge seed: error sampling chain for randomness: sample height out of range: 2125`

https://explorer.test.kittyhawk.wtf/blocks/zDPWYqFCtuSB3Z6z3Po2sshrQMpgGmoo4cTBqqRH8GaVvKamPVms

https://github.com/filecoin-project/go-filecoin/blob/81e75fd7f9722f9a100b8d03852e4e823a4b209d/sampling/chain_randomness.go#L31-L41

**Expected behavior**
Miners should be able to generate a PoSt seed if their proving period starts on a null block.

**Additional context**

anorth:
Fixes #3017.

Removed tests for VM context that are duplicates of tests for the actual chain sampling.

Note that we're actually supposed to be sampling tickets, which form a contiguous chain, so this will all go away when we implement ticket chains properly.

### DIFF
--- a/sampling/chain_randomness.go
+++ b/sampling/chain_randomness.go
@@ -6,62 +6,57 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// LookbackParameter is the protocol parameter defining how many blocks in the
-// past to look back to sample randomness values.
+// LookbackParameter defines how many non-empty tiptsets (not rounds) earlier than any sample
+// height (in rounds) from which to sample the chain for randomness.
+// This constant is a protocol (actor) parameter and should be defined in actor code.
 const LookbackParameter = 3
 
-// SampleChainRandomness produces a slice of random bytes sampled from a tip set
-// in the provided slice of tip sets at a given height (minus lookback). This
-// function assumes that the tip set slice is sorted by block height in
-// descending order.
-//
-// SampleChainRandomness is useful for things like PoSt challenge seed
-// generation.
-func SampleChainRandomness(sampleHeight *types.BlockHeight, tipSetsSortedByBlockHeightDescending []types.TipSet) ([]byte, error) {
+// SampleChainRandomness produces a slice of bytes (a ticket) sampled from the tipset `LookbackParameter`
+// tipsets (not rounds) prior to the highest tipset with height less than or equal to `sampleHeight`.
+// The tipset slice must be sorted by descending block height.
+func SampleChainRandomness(sampleHeight *types.BlockHeight, tipSetsDescending []types.TipSet) ([]byte, error) {
+	// Find the first (highest) tipset with height less than or equal to sampleHeight.
+	// This is more complex than necessary: https://github.com/filecoin-project/go-filecoin/issues/3025
 	sampleIndex := -1
-	tipSetsLen := len(tipSetsSortedByBlockHeightDescending)
-	lastIdxInTipSets := tipSetsLen - 1
-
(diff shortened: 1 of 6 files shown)


sampling/chain_randomness.go
```diff
@@ -6,62 +6,57 @@ import (
...
-	for i := 0; i < tipSetsLen; i++ {
-		height, err := tipSetsSortedByBlockHeightDescending[i].Height()
+	for i, tip := range tipSetsDescending {
+		height, err := tip.Height()
 		if err != nil {
 			return nil, errors.Wrap(err, "error obtaining tip set height")
 		}
 
-		if types.NewBlockHeight(height).Equal(sampleHeight) {
+		if types.NewBlockHeight(height).LessEqual(sampleHeight) {
```
ZenGround0: I don't think we should do this.  While it will probably work for some time this is not compatible with a finished PoSt generate/verify system.  We need to sample an exact seed from an exact block height so that proving and verifying interoperate correctly and we should be able to provide it every time.  I believe the root of this bug is that the miner actor is getting into a state where its `provingPeriodStart` is not 100% for sure associated with a tipset blockheight that exists.  We should address this directly at the root.

I'm looking into this.  cc @acruikshank I can't see any issue with the change [here](https://github.com/filecoin-project/go-filecoin/commit/331e559db35cc2c6c98b419c0464cc1ee2d2d73e) but it is my primary suspect.

ZenGround0: Looking closer at the commit mentioned above I see that we *do* have a consistent reference seed.  In light of that this change is ok, especially since all seed sampling code is going to be seeing some major improvements in the near future.

ZenGround0:
Ok [block removed](https://github.com/filecoin-project/go-filecoin/pull/3026#discussion_r301123454) from previous comment.  

New block because now `GetRecentAncestors` will not correctly gather all the ancestors needed by the new implementation.  When `GetRecentAncestors` calls `CollectTipSetsOfHeightAtLeast` [here](https://github.com/filecoin-project/go-filecoin/blob/master/chain/get_ancestors.go#L76) it does not take the first ancestor less than the proving period start.  So in the (currently rare) case where a post is submitted at the very last moment, the ancestors loaded up by consensus processing will miss the last ancestor tipset needed.

Working on this now.  The approach is to change the [`CollectTipSetsOfHeightAtLeast` method](https://github.com/filecoin-project/go-filecoin/blob/master/chain/get_ancestors.go#L104) to include one tipset less than the target.

sampling/chain_randomness.go
acruikshank: I agree with the second comment. This will generate a consistent seed and is the correct approach.

ZenGround0:
After discussing with @acruikshank I'm removing the second block too.  The failure mode is harmless enough that it doesn't need to make it into the latest release and can be PRed during the next cycle. Work for that fix is now tracked in #3027.

PR #3035, opened 2019-07-09
Poll daemon to determine if it's ready

travisperson:
When a node dies from OOM, the API file will not be cleaned up properly. This resulted in the api file check to pass with a false positive.

### DIFF
--- a/bin/node_restart
+++ b/bin/node_restart
@@ -3,10 +3,17 @@ set -e
 filecoin_repo="/var/local/filecoin/repo"
 filecoin_exec="go-filecoin --repodir=${filecoin_repo}"
 
-
+# Number of time to check before giving up
+limit=3600
+count=0
 while ! ${filecoin_exec} id >&2 2> /dev/null
 do
   echo "Waiting for daemon to start..." && sleep 1
+  ((i++))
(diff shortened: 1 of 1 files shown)


bin/node_restart
```diff
@@ -4,7 +4,7 @@ filecoin_repo="/var/local/filecoin/repo"
 filecoin_exec="go-filecoin --repodir=${filecoin_repo}"
 
 
-while [ ! -f "${filecoin_repo}/api" ]
+while ! ${filecoin_exec} id >&2 2> /dev/null
```
frrist: ditto lasers comment about the number of times this can run, otherwise seems reasonable to me.

bin/node_restart
```diff
@@ -3,10 +3,17 @@ set -e
...
 
-
-while [ ! -f "${filecoin_repo}/api" ]
+# Number of time to check before giving up
+limit=3600
+count=0
+while ! ${filecoin_exec} id >&2 2> /dev/null
 do
   echo "Waiting for daemon to start..." && sleep 1
+  ((i++))
```
laser: You want to increment `count`, not `i`.

PR #3036, opened 2019-07-09
split CGO bindings to libsectorbuilder_ffi.a into extractable package

laser:
## Why does this PR exist?

@whyrusleeping needs to use the CGO bindings for `libsectorbuilder.a` in a separate Go project. As it currently stands, those CGO bindings were intermixed with go-filecoin specific types and concerns (the seal status poller).

## What's in this PR?

This changeset moves all CGO-related code into `cgo_bindings.go` in the `libsectorbuilder` package. The `RustVerifier` and `RustSectorBuilder` and corresponding interfaces continue to expose fancy go-filecoin types to the system which they create after calling down to `libsectorbuilder.cgo_bindings` functions. This changeset does not change any of the behavior of `RustSectorBuilder` or `RustVerifier`, just some implementation details.

## Next Steps

We need to split the code in the `libsectorbuilder` package into a separate Git repository.


proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -192,104 +138,83 @@ func (sb *RustSectorBuilder) AddPiece(ctx context.Context, pieceRef cid.Cid, pie
...
-		copy(commRStar[:], commRStarSlice)
-
-		proof := goBytes(resPtr.proof_ptr, resPtr.proof_len)
-
-		ps, err := goPieceInfos(resPtr.pieces_ptr, resPtr.pieces_len)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to marshal from string to cid")
-		}
-
+	if status.SealStatusCode == 0 {
```
shannonwells: I'd prefer it if these numbers were descriptive constants, but at least the comment is there, which serves the same purpose.

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -192,104 +138,83 @@ func (sb *RustSectorBuilder) AddPiece(ctx context.Context, pieceRef cid.Cid, pie
...
-			Proof:     proof,
-			SectorID:  sectorID,
+			CommD:     status.CommD,
+			CommR:     status.CommR,
+			CommRStar: status.CommRStar,
+			Pieces:    info,
+			Proof:     status.Proof,
+			SectorID:  status.SectorID,
 		}, nil
+	} else if status.SealStatusCode == 1 || status.SealStatusCode == 3 {
```
shannonwells: same

proofs/verification/rustverifier.go
```diff
@@ -0,0 +1,45 @@
+package verification
+
+import "github.com/filecoin-project/go-filecoin/proofs/libsectorbuilder"
+
+// RustVerifier provides proof-verification methods.
+type RustVerifier struct{}
```
shannonwells: Is this struct just for namespacing within the package?  Trying to understand the choice to not make these free functions.

laser: Totally legitimate question. There exist other structs which satisfy the Verifier interface (for DI purposes), and some functions accept a Verifier. We could modify the functions which accept a Verifier to instead accept two functions of the appropriate types, but the interface and struct approach is consistent with the rest o’ the codebase.

proofs/sectorbuilder/rustsectorbuilder.go
laser: Cool. I’ll factor these out into some enum-like thingy ASAP.

Issue #2939, opened 2019-06-16
Monitor for storage slashing condition while block mining


anorth:
### Description
A block-producing node should monitor for storage faults (failure to submit PoSt) and attempt to locally mine a message invoking `SlashStorageFault` on behalf of the miner owner account. The message should not be broadcast to other nodes.

The implementation is outlined in the [faults and slashing design doc](https://docs.google.com/document/d/1U3b9GVNVOLoS_-q9kXU3a9dqHGvcrHprwDFzNYgLI0M/edit?pli=1#heading=h.k011apgz1kwa)

See also #2938 which monitors for consensus faults.

### Acceptance criteria
- A miner detects and penalises storage faults
- Unit tests for detection and message construction
- An integration test demonstrating successful slashing

### Where to begin
A listener for new heaviest tipset state transitions.

PR #3043, opened 2019-07-10
add GetPoStState for lateness determination

shannonwells:
Support for #2939 
Storage fault monitor will ask a minor actor for its lateness status ("PoStState") so as to determine whether there is a fault and what type.  This leaves all the logic for determining lateness in one place -- since this is already calculated during an actual PoSt submission -- and will reduce the number of message queries to actors in order to tally unreported faults.

This PR implements the GetPoStState export in the miner actor.  Better name suggestions welcome.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -477,23 +477,21 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
 	return ma.Bootstrap, 0, nil
 }
 
-// GetPoStState returns whether the miner's last submitPoSt within the proving period,
+// GetPoStState returns whether the miner's last submitPoSt is within the proving period,
 // late or after the generation attack threshold.
 func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
-
 	var state State
-	bh := ctx.BlockHeight()
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		generationAttackGracePeriod := GenerationAttackTime(state.SectorSize)
 		// Don't check lateness unless there is storage to prove
-		if len(state.ProvingSet.Values()) == 0 {
+		if state.ProvingSet.Size() == 0 {
 			return int64(PoStStateNoStorage), nil
 		}
-		lateState, _ := lateState(state.ProvingPeriodEnd, bh, generationAttackGracePeriod)
+		lateState, _ := lateState(state.ProvingPeriodEnd, ctx.BlockHeight(), GenerationAttackTime(state.SectorSize))
 		return lateState, nil
 	})
 
 	if err != nil {
+		return nil, errors.CodeError(err), err
 		return nil, errors.CodeError(err), err
 	}
 
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -92,6 +92,13 @@ var Errors = map[uint8]error{
 	ErrInsufficientCollateral:  errors.NewCodedRevertErrorf(ErrInsufficientCollateral, "insufficient collateral"),
 }
 
+const (
+	PoStStateNoStorage = iota
+	PoStStateWithinProvingPeriod
+	PoStStateAfterProvingPeriod
+	PoStStateAfterGenerationAttackThreshold
+)
+
 // Actor is the miner actor.
 //
 // If `Bootstrap` is `true`, the miner will not verify seal proofs. This is
@@ -470,12 +477,6 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
 	return ma.Bootstrap, 0, nil
 }
 
-const (
-	WithinProvingPeriod = iota
-	AfterProvingPeriod
-	AfterGenerationAttackThreshold
-)
-
 // GetPoStState returns whether the miner's last submitPoSt within the proving period,
 // late or after the generation attack threshold.
 func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
@@ -484,6 +485,10 @@ func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
 	bh := ctx.BlockHeight()
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
 		generationAttackGracePeriod := GenerationAttackTime(state.SectorSize)
+		// Don't check lateness unless there is storage to prove
+		if len(state.ProvingSet.Values()) == 0 {
+			return int64(PoStStateNoStorage), nil
+		}
 		lateState, _ := lateState(state.ProvingPeriodEnd, bh, generationAttackGracePeriod)
 		return lateState, nil
 	})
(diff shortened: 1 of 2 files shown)


actor/builtin/miner/miner.go
```diff
@@ -461,6 +465,36 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
 	return ma.Bootstrap, 0, nil
 }
 
+const (
+	WithinProvingPeriod = iota
+	AfterProvingPeriod
+	AfterGenerationAttackThreshold
```
acruikshank: This enumeration is a good idea. A few suggestions:

1. Please move this definition up with the other constants.
2. This struct already defines several constants. Please prefix with 'PoStState', e.g. `PoStStateWithinProvingPeriod`.
3. There should be a separate state for the state when the miner is not proving any storage. `PoStStateNoStorage`?

actor/builtin/miner/miner.go
```diff
@@ -461,6 +465,36 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
...
+)
+
+// GetPoStState returns whether the miner's last submitPoSt within the proving period,
+// late or after the generation attack threshold.
```
acruikshank: This method should only check whether the PoSt is late if the miner is actively proving storage (the ProvingSet is non-empty). Otherwise this method will suggest the miner is late when they are not.

actor/builtin/miner/miner.go
```diff
@@ -973,23 +1007,46 @@ func ProvingPeriodDuration(sectorSize *types.BytesAmount) uint64 {
...
 		return pledgeCollateral
-	} else if roundsLate.GreaterThan(types.NewBlockHeight(0)) {
+	} else if lateState == AfterProvingPeriod {
 		// fee = collateral * (roundsLate / maxRoundsLate)
 		var fee big.Int
 		fee.Mul(pledgeCollateral.AsBigInt(), roundsLate.AsBigInt())
 		fee.Div(&fee, maxRoundsLate.AsBigInt()) // Integer division in AttoFIL, rounds towards zero.
 		return types.NewAttoFIL(&fee)
 	}
+	//roundsLate := chainHeight.Sub(provingPeriodEnd)
```
acruikshank: delete

actor/builtin/miner/miner_test.go
```diff
@@ -1063,6 +1074,42 @@ func TestLatePoStFee(t *testing.T) {
 	})
 }
 
+func TestActor_GetPoStState(t *testing.T) {
+	tf.UnitTest(t)
+
+	firstCommitBlockHeight := uint64(3)
+	//secondProvingPeriodStart := LargestSectorSizeProvingPeriodBlocks + firstCommitBlockHeight
+	//thirdProvingPeriodStart := 2*LargestSectorSizeProvingPeriodBlocks + firstCommitBlockHeight
```
acruikshank: delete

actor/builtin/miner/miner_test.go
```diff
@@ -1063,6 +1074,42 @@ func TestLatePoStFee(t *testing.T) {
 	})
 }
 
+func TestActor_GetPoStState(t *testing.T) {
```
acruikshank: Add test for miner with no sectors.

actor/builtin/miner/miner.go
```diff
@@ -461,6 +472,34 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
 	return ma.Bootstrap, 0, nil
 }
 
+// GetPoStState returns whether the miner's last submitPoSt within the proving period,
```
ZenGround0: nit "was within"

actor/builtin/miner/miner.go
```diff
@@ -461,6 +472,34 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
...
+// late or after the generation attack threshold.
+func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
+
```
ZenGround0: nit extra newline

actor/builtin/miner/miner.go
```diff
@@ -461,6 +472,34 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
...
+	var state State
+	bh := ctx.BlockHeight()
+	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
+		generationAttackGracePeriod := GenerationAttackTime(state.SectorSize)
+		// Don't check lateness unless there is storage to prove
+		if len(state.ProvingSet.Values()) == 0 {
```
ZenGround0: There's a fancy `ProvingSet.Size()` function you can use now

actor/builtin/miner/miner.go
```diff
@@ -461,6 +472,34 @@ func (ma *Actor) IsBootstrapMiner(ctx exec.VMContext) (bool, uint8, error) {
...
+			return int64(PoStStateNoStorage), nil
+		}
+		lateState, _ := lateState(state.ProvingPeriodEnd, bh, generationAttackGracePeriod)
```
ZenGround0: minor style preference: you could avoid the variable declarations and pass in `ctx.BlockHeight()` and `GenerationAttackTime(state.SectorSize)` as the last two arguments.

actor/builtin/miner/miner_test.go
```diff
@@ -533,7 +533,18 @@ func (mal *minerActorLiason) assertPoStFail(blockHeight uint64, done types.IntSe
 	assert.Equal(mal.t, exitCode, res.Receipt.ExitCode)
 }
 
-func newMinerActorLiason(t *testing.T, st state.Tree, vms vm.StorageMap, ancestors []types.TipSet, minerAddr address.Address ) *minerActorLiason {
+func (mal *minerActorLiason) assertPoStStateAtHeight(expected int64, queryHeight uint64) {
```
ZenGround0: I'm glad someone else is using the liason!  If you can think of a better name don't hesitate to rename it.

actor/builtin/miner/miner_test.go
```diff
@@ -1063,6 +1074,48 @@ func TestLatePoStFee(t *testing.T) {
...
+		mal.assertPoStStateAtHeight(PoStStateWithinProvingPeriod, firstCommitBlockHeight)
+		mal.assertPoStStateAtHeight(PoStStateWithinProvingPeriod, firstCommitBlockHeight+1)
+		mal.assertPoStStateAtHeight(PoStStateWithinProvingPeriod, firstCommitBlockHeight+6)
+	})
+
+	t.Run("is reported as PoStStateAfterProvingPeriod after the proving period", func(t *testing.T) {
+		mal := setupMinerActorLiason(t)
+		mal.requireCommit(firstCommitBlockHeight, uint64(1))
+
+		mal.assertPoStStateAtHeight(PoStStateAfterProvingPeriod, firstCommitBlockHeight+LargestSectorSizeProvingPeriodBlocks+1)
```
ZenGround0: something I found helpful for readability from other tests in this package is defining `firstCommitBlockHeight+LargestSectorSizeProvingPeriodBlocks` as a constant `lastHeightOfFirstPeriod` or something next to `firstCommitBlockHeight`.

ZenGround0: on line 1106 you could use `lastHeightOfSecondPeriod` etc

actor/builtin/miner/miner.go
shannonwells: I wasn't sure about it, now I know!

actor/builtin/miner/miner.go
shannonwells: I not no stinkin verbs

PR #3045, opened 2019-07-10
Turn SortedCidSet into an immutable value type and rename it to TipSetKey.

anorth:
This parallels the transformation of TipSet into a value type.

FYI @acruikshank @frrist


types/sorted_cid_set.go
```diff
@@ -5,188 +5,173 @@ import (
...
 	return true
 }
 
-// Contains checks if s2 is a sub-tipset of s
-func (s *SortedCidSet) Contains(s2 *SortedCidSet) bool {
-	for it := s2.Iter(); !it.Complete(); it.Next() {
-		if !s.Has(it.Value()) {
-			return false
+// ContainsAll checks if another set is a subset of this one.
+func (s *TipSetKey) ContainsAll(other TipSetKey) bool {
```
ZenGround0: style nit: I don't think the `All` makes the name clearer.

types/sorted_cid_set.go
```diff
@@ -5,188 +5,173 @@ import (
...
+	// Since the slices are sorted we can perform one pass over this set, advancing
+	// the other index index whenever the values match.
```
ZenGround0: nit: duplicate word index

types/sorted_cid_set.go
```diff
@@ -5,188 +5,173 @@ import (
...
+		// Uniq-ify the sorted array.
+		this, next := 0, 1
+		for next < len(cids) {
+			if cids[next] != cids[this] {
+				this++
+				cids[this] = cids[next]
+			}
+			next++
+		}
+		return TipSetKey{cids[:this+1]}
```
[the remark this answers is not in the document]
ZenGround0: If you keep it a descriptive comment (maybe a link describing this technique if it's standard) will help

types/sorted_cid_set.go
```diff
@@ -5,188 +5,173 @@ import (
...
-// TODO: This should probably go into go-cid package - see https://github.com/ipfs/go-cid/issues/45.
-type SortedCidSet struct {
-	s []cid.Cid // should be maintained sorted
+// TipSetKey is an immutable set of CIDs forming a unique key for a TipSet.
+// Equal keys will have equivalent iteration order, but note that the CIDs are *not* maintained in
+// the same order as the canonical iteration order of blocks in a tipset (which is by ticket).
+// TipSetKey is a lightweight value type; passing by pointer is usually unnecessary.
+type TipSetKey struct {
+	// The slice is wrapped in a struct to enforce immutability.
+	cids []cid.Cid
```
ZenGround0: way better than `s`

types/sorted_cid_set.go
```diff
@@ -5,188 +5,173 @@ import (
...
 // UnmarshalJSON parses JSON into the set.
-func (s *SortedCidSet) UnmarshalJSON(b []byte) error {
-	var ts []cid.Cid
-	if err := json.Unmarshal(b, &ts); err != nil {
+func (s *TipSetKey) UnmarshalJSON(b []byte) error {
+	var cids []cid.Cid
+	if err := json.Unmarshal(b, &cids); err != nil {
 		return err
 	}
-	for i := 0; i < len(ts)-1; i++ {
```
ZenGround0: Blocking: don't we still need this check?  I'm not sure if we're even using JSON serialization anymore but I don't think we're checking the ordering anywhere without this check.

ZenGround0: Consider deleting these if they are unused.

consensus/processor.go
```diff
@@ -50,8 +51,8 @@ type ApplicationResult struct {
 // of message conflicts.
 type ProcessTipSetResponse struct {
 	Results   []*ApplicationResult
-	Successes types.SortedCidSet
-	Failures  types.SortedCidSet
+	Successes map[cid.Cid]struct{}
```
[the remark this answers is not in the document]
anorth: My original plan was only to make the set immutable, but then I discovered that every case that actually needed that was using it as a key, and "key" is a very concise concept.

We can make the change you suggest pretty easily in the future, though.

types/sorted_cid_set.go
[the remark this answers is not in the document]
anorth: I wouldn't be convinced by any pencil and paper work, so I wrote tests :-)

It's not a particularly standard/common technique (or problem) AFAIK. You're right, if I were doing it again using a second slice would also have been faster to write. I didn't expect it to come out this "clever". Since it's now here and tested, I'm deciding to leave it in. I added an explanatory comment.

types/sorted_cid_set.go
anorth: Oops, this was an oversight, thanks. We certainly still use the serialisation, and there are tests for deserialisation, so I fixed it.

PR #3046, opened 2019-07-10
add Fetcher interface, impl with BitswapFetcher

frrist:
### What
This PR defines a Fetcher interface and renames the original Fetcher to BitswapFetcher. The goal here is to define an interface for other types of fetchers. Some examples include a blocksync fetcher and a graphsync fetcher.


frrist:
will reopen after rebasing onto branch with TipSetKey type #3045

PR #3047, opened 2019-07-10
add Fetcher interface, impl with BitswapFetcher

frrist:
### What
This PR defines a Fetcher interface and renames the original Fetcher to BitswapFetcher. The goal here is to define an interface for other types of fetchers. Some examples include a blocksync fetcher and a graphsync fetcher.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/metrics"
 	"github.com/filecoin-project/go-filecoin/metrics/tracing"
+	"github.com/filecoin-project/go-filecoin/net"
 	"github.com/filecoin-project/go-filecoin/sampling"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
@@ -85,11 +86,6 @@ type syncerChainReader interface {
 	HasAllBlocks(ctx context.Context, cs []cid.Cid) bool
 }
 
-type syncFetcher interface {
-	GetBlocks(context.Context, []cid.Cid) ([]*types.Block, error)
-	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, recur int) ([]types.TipSet, error)
-}
-
 // Syncer updates its chain.Store according to the methods of its
 // consensus.Protocol.  It uses a bad tipset cache and a limit on new
 // blocks to traverse during chain collection.  The Syncer can query the
@@ -115,7 +111,7 @@ type Syncer struct {
 	mu sync.Mutex
 	// fetcher is the networked block fetching service for fetching blocks
 	// and messages.
-	fetcher syncFetcher
+	fetcher net.Fetcher
 	// stateStore is the cborStore used for reading and writing state root
 	// to ipld object mappings.
 	stateStore *hamt.CborIpldStore
@@ -132,7 +128,7 @@ type Syncer struct {
 }
 
 // NewSyncer constructs a Syncer ready for use.
-func NewSyncer(cst *hamt.CborIpldStore, c consensus.Protocol, s syncerChainReader, f syncFetcher, syncMode SyncMode) *Syncer {
+func NewSyncer(cst *hamt.CborIpldStore, c consensus.Protocol, s syncerChainReader, f net.Fetcher, syncMode SyncMode) *Syncer {
 	return &Syncer{
 		fetcher:    f,
 		stateStore: cst,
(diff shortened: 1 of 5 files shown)


chain/syncer.go
```diff
@@ -151,11 +152,20 @@ func NewSyncer(cst *hamt.CborIpldStore, c consensus.Protocol, s syncerChainReade
 // otherwise resolved over the network.  This method will timeout if blocks
 // are unavailable.  This method is all or nothing, it will error if any of the
 // blocks cannot be resolved.
-func (syncer *Syncer) getBlksMaybeFromNet(ctx context.Context, blkCids []cid.Cid) ([]*types.Block, error) {
+func (syncer *Syncer) getBlksMaybeFromNet(ctx context.Context, blkCids []cid.Cid) ([]types.TipSet, error) {
 	ctx, cancel := context.WithTimeout(ctx, blkWaitTime)
 	defer cancel()
 
-	return syncer.fetcher.GetBlocks(ctx, blkCids)
+	tss, err := syncer.fetcher.FetchTipSets(ctx, types.NewTipSetKey(blkCids...), 1)
```
frrist: Since we don't have "Initial sync"/"caught up" modes this will just behave like it did before.

chain/syncer.go
```diff
@@ -151,11 +152,20 @@ func NewSyncer(cst *hamt.CborIpldStore, c consensus.Protocol, s syncerChainReade
 // otherwise resolved over the network.  This method will timeout if blocks
 // are unavailable.  This method is all or nothing, it will error if any of the
 // blocks cannot be resolved.
-func (syncer *Syncer) getBlksMaybeFromNet(ctx context.Context, blkCids []cid.Cid) ([]*types.Block, error) {
+func (syncer *Syncer) getBlksMaybeFromNet(ctx context.Context, blkCids []cid.Cid) ([]types.TipSet, error) {
```
ZenGround0: nit: name should change to reflect semantic changes.

chain/syncer.go
```diff
@@ -151,11 +152,20 @@ func NewSyncer(cst *hamt.CborIpldStore, c consensus.Protocol, s syncerChainReade
...
 	ctx, cancel := context.WithTimeout(ctx, blkWaitTime)
 	defer cancel()
 
-	return syncer.fetcher.GetBlocks(ctx, blkCids)
+	tss, err := syncer.fetcher.FetchTipSets(ctx, types.NewTipSetKey(blkCids...), 1)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(tss) > 1 {
```
ZenGround0: I doubt we'll forget to nuke this but maybe leave a "// TODO this will go away once improved fetching lands" to keep other people wandering through this code from worrying.

chain/syncer.go
```diff
@@ -195,12 +205,9 @@ func (syncer *Syncer) collectChain(ctx context.Context, tipsetCids types.TipSetK
...
 
-		blks, err := syncer.getBlksMaybeFromNet(ctx, tipsetCids.ToSlice())
-		if err != nil {
-			return nil, err
-		}
-
-		ts, err := types.NewTipSet(blks...)
+		// This will return a single tipset, error if not found and panic
+		// if more than 1 tipset is returned
+		ts, err := syncer.getBlksMaybeFromNet(ctx, tipsetCids.ToSlice())
```
ZenGround0: nit: I got a bit confused going off of the comment and var name.  I think it should say "This will return a slice containing a single tipset" and `ts` should probably be `tss` or some name indicating multiple values.

net/fetcher.go
```diff
@@ -13,28 +13,60 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// Fetcher is used to fetch data over the network.  It is implemented with
+// Fetcher defines an interface that may be used to fetch data from the network.
+type Fetcher interface {
```
ZenGround0: The syncFetcher interface is defined in the consumer as I would expect.  What does this guy do?  It appears extraneous.

chain/syncer.go
frrist: I'll just make it return a single tipset like it did before.

net/fetcher.go
frrist: Its used by the node: https://github.com/filecoin-project/go-filecoin/pull/3047/files#diff-380e79adccbaf1276bf6c91cc551cd7fR161 and I think I might remove the `syncFetcher` interface as it is no different than the `Fetcher` interface.

frrist: I think the consumer side interface definition is helpful when asking for a subset of functions from something big, like porcelain and plumbing. But I don't think its a pattern we should use everywhere. As an example I am having some mixed feelings about the number of `*ChainReader`/`*ChainProvider` interfaces we have defined (at least 10) -- feels like we might want to rethink this strategy and create 2  or 3 "cancolical" chain interface (read, write, query) that we use everywhere with corresponding Mocks/Fakes used in testing. My 2 cents here at least.

ZenGround0: I agree with that proposal.

PR #3057, opened 2019-07-11
storagemarket.GetLateMiners

shannonwells:
Support for #2939 

The fault monitor needs a list of storage miners that are late. Storage miner has the list of miners and can ask each actor for its lateness. Create an export in storage market that returns a pointer to a map of address strings with PoStStates, but only if that state is late.

### DIFF
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -335,7 +335,7 @@ func TestMinerGetPower(t *testing.T) {
 
 		st, vms := th.RequireCreateStorages(ctx, t)
 
-		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t))
+		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t), 0)
 
 		// retrieve power (trivial result for no proven sectors)
 		result := callQueryMethodSuccess("getPower", ctx, t, st, vms, address.TestAddress, minerAddr)
@@ -352,7 +352,7 @@ func TestMinerGetProvingPeriod(t *testing.T) {
 
 		st, vms := th.RequireCreateStorages(ctx, t)
 
-		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t))
+		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t), 0)
 
 		// retrieve proving period
 		result := callQueryMethodSuccess("getProvingPeriod", ctx, t, st, vms, address.TestAddress, minerAddr)
@@ -377,7 +377,7 @@ func TestMinerGetProvingPeriod(t *testing.T) {
 
 		st, vms := th.RequireCreateStorages(ctx, t)
 
-		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t))
+		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(240), t, st, vms, address.TestAddress, th.RequireRandomPeerID(t), 0)
 
 		// commit sector to set ProvingPeriodEnd
 		commR := th.MakeCommitment()
@@ -446,7 +446,7 @@ func TestMinerCommitSector(t *testing.T) {
 		amtCollateralForPledge := MinimumCollateralPerSector.CalculatePrice(types.NewBytesAmount(numSectorsToPledge))
 
 		origPid := th.RequireRandomPeerID(t)
-		minerAddr := th.CreateTestMinerWith(amtCollateralForPledge, t, st, vms, address.TestAddress, origPid)
+		minerAddr := th.CreateTestMinerWith(amtCollateralForPledge, t, st, vms, address.TestAddress, origPid, 0)
 
 		commR := th.MakeCommitment()
 		commRStar := th.MakeCommitment()
@@ -476,7 +476,7 @@ func TestMinerCommitSector(t *testing.T) {
 		st, vms := th.RequireCreateStorages(ctx, t)
 
 		origPid := th.RequireRandomPeerID(t)
-		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(100), t, st, vms, address.TestAddress, origPid)
+		minerAddr := th.CreateTestMinerWith(types.NewAttoFILFromFIL(100), t, st, vms, address.TestAddress, origPid, 0)
 
 		commR := th.MakeCommitment()
 		commRStar := th.MakeCommitment()
--- a/actor/builtin/storagemarket/storagemarket.go
+++ b/actor/builtin/storagemarket/storagemarket.go
@@ -92,21 +92,21 @@ var storageMarketExports = exec.Exports{
 		Params: []abi.Type{abi.BytesAmount, abi.PeerID},
 		Return: []abi.Type{abi.Address},
 	},
-	"getMiners": &exec.FunctionSignature{
-		Params: nil,
-		Return: []abi.Type{abi.Addresses},
-	},
-	"getProofsMode": &exec.FunctionSignature{
-		Params: []abi.Type{},
-		Return: []abi.Type{abi.ProofsMode},
+	"updateStorage": &exec.FunctionSignature{
+		Params: []abi.Type{abi.BytesAmount},
+		Return: nil,
 	},
 	"getTotalStorage": &exec.FunctionSignature{
 		Params: []abi.Type{},
 		Return: []abi.Type{abi.BytesAmount},
 	},
-	"updateStorage": &exec.FunctionSignature{
-		Params: []abi.Type{abi.BytesAmount},
-		Return: nil,
+	"getProofsMode": &exec.FunctionSignature{
+		Params: []abi.Type{},
+		Return: []abi.Type{abi.ProofsMode},
+	},
+	"getMiners": &exec.FunctionSignature{
+		Params: nil,
+		Return: []abi.Type{abi.Addresses},
 	},
 }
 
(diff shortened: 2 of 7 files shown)


actor/builtin/storagemarket/storagemarket.go
```diff
@@ -92,17 +92,21 @@ var storageMarketExports = exec.Exports{
 		Params: []abi.Type{abi.BytesAmount, abi.PeerID},
 		Return: []abi.Type{abi.Address},
 	},
-	"updateStorage": &exec.FunctionSignature{
-		Params: []abi.Type{abi.BytesAmount},
-		Return: nil,
+	"getMiners": &exec.FunctionSignature{
```
shannonwells: alphabetized

plumbing/api.go
```diff
@@ -118,6 +118,9 @@ func (api *API) ActorLs(ctx context.Context) (<-chan state.GetAllActorsResult, e
 	return api.chain.LsActors(ctx)
 }
 
+// StorageMinerLs returns a list of storage miner addresses
+func (api *API) StorageMinerLs(ctx context.Context) {}
```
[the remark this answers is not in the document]
shannonwells: Yes, I forgot about it. It's deleted locally and I hadn't pushed the changes.

actor/builtin/miner/miner.go
```diff
@@ -516,6 +519,25 @@ func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
 	return big.NewInt(result), 0, nil
 }
 
+// GetSlashable returns whether this miner is slashable. At this time this means only that
```
ZenGround0: This name is a bit misleading.  The comment does a good job of explaining it but if I only saw the name I would for sure think this returns true when the miner actor can actually be slashed.  Because this returns true when there are sectors in the miner's set what about `GetProving`, `Proving`, `ActivelyProving` etc, something along those lines?

actor/builtin/miner/miner_test.go
```diff
@@ -1133,6 +1146,30 @@ func TestMinerGetPoStState(t *testing.T) {
...
+		done := types.EmptyIntSet()
+		mal.requirePoSt(firstCommitBlockHeight+5, done)
+		// even though they are late, they have never submitted a post so
+		mal.assertSlashability(true, firstCommitBlockHeight+6)
+	})
+
+	t.Run("is reported as not slashable if no proving set", func(t *testing.T) {
+		mal := setupMinerActorLiason(t)
+		mal.assertSlashability(false, firstCommitBlockHeight)
+	})
```
ZenGround0: It might also be good to test the edge case where a miner has a single sector in its proving set after its first commitSector but before its first submitPoSt

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -193,6 +197,53 @@ func (sma *Actor) UpdateStorage(vmctx exec.VMContext, delta *types.BytesAmount)
 	return 0, nil
 }
 
+func (sma *Actor) GetSlashableMiners(vmctx exec.VMContext) (*[]address.Address, uint8, error) {
```
ZenGround0: Same comment about naming as above propagates to here.  Consider `GetProvingMiners`

actor/builtin/storagemarket/storagemarket_test.go
```diff
@@ -234,3 +299,52 @@ func deriveMinerAddress(creator address.Address, nonce uint64) (address.Address,
 
 	return address.NewActorAddress(buf.Bytes())
 }
+
+// mustCreateStorageMiner creates a storage miner for the given state tree & storage map
+// and returns the address of the created miner.
+func mustCreateStorageMiner(t *testing.T, st state.Tree, vms vm.StorageMap, height uint64) address.Address {
```
ZenGround0: Please consider building on top of existing work in testhelpers/core.go to consolidate our testing framework.  There are already `CreateTestMiner` and `CreateTestMinerWith`.  Neither is exactly what you want because you need to specify height but it should be straightforward to add this param to `CreateTestMinerWith` and then update callers.

testhelpers/core.go
```diff
@@ -151,7 +151,8 @@ func CreateTestMinerWith(
 
 	result, err := ApplyTestMessage(stateTree, vms, msg, types.NewBlockHeight(0))
 	require.NoError(t, err)
-
+	require.NotNil(t, result)
+	require.NoError(t, result.ExecutionError)
```
ZenGround0: 🙏

actor/builtin/storagemarket/storagemarket.go
shannonwells: So I had thought about this -- I was allowing for the possibility that other factors besides nonzero proving set may be used to determine slashability.  I'm not married to the name though.

actor/builtin/miner/miner_test.go
shannonwells: I thought that was the last example. What's missing?

ZenGround0: It's similar to the last example but you would run `mal.assertIsProving(true. firstCommitBlockHeight+1)` after the first `mal.requireCommit`

shannonwells: ah. I was thinking about storagemarket_test where I'm doing that already, but I'll put an explicit one here.

actor/builtin/miner/miner.go
```diff
@@ -516,6 +519,25 @@ func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
...
+		return state.ProvingSet.Size() != 0, nil
+	})
+
+	if err != nil {
+		return false, errors.CodeError(err), err
+	}
+	result, ok := out.(bool)
+	if !ok {
+		return false, 1, errors.NewFaultErrorf("expected a bool, but got %T instead", out)
+	}
```
anorth: FYI I added some other helpers to make read-only methods less verbose. See `GetProvingPeriod` for an example.

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -237,6 +288,19 @@ func (sma *Actor) GetProofsMode(vmctx exec.VMContext) (types.ProofsMode, uint8,
...
+	if err != nil {
+		return false, err
+	}
+
+	isProving, err := abi.Deserialize(msgResult[0], abi.Boolean)
+	if err != nil {
+		return false, err
+	}
+	return isProving.Val.(bool), nil
+}
```
anorth: @acruikshank FWIW I think it's a bit silly that read-only methods that are not part of consensus have to go through this message-passing and ABI-deserialization junk. We're doing something wrong.

acruikshank: (non-blocking)
I agree, but it's also kind of silly that state-manipulating methods that are part of consensus go through ABI-deserialization. Nothing about consensus requires this right now. It's all part of a long term plan to move the state machine into a separate VM that we haven't been able to prioritize.

There are advantages to keeping code that accesses actor-state within the actor. There are also advantages to limiting actors to message processing (actors not objects). There's no real advantage to serializing parameters and deserializing results (until we have a VM). We can simplify this interface in the near term.

One option would be a function that could generate a populated VMContext at the chain head. With this, you could call the actor method directly. See #3067 for an example of doing this in tests.

actor/builtin/miner/miner.go
shannonwells: ah, I didn't even see loadState.  Thanks for the pointer.

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -237,6 +288,19 @@ func (sma *Actor) GetProofsMode(vmctx exec.VMContext) (types.ProofsMode, uint8,
...
 
+func (sma *Actor) getMinerIsProving(vmctx exec.VMContext, minerAddr address.Address) (bool, error) {
+	msgResult, _, err := vmctx.Send(minerAddr, "isProving", types.ZeroAttoFIL, nil)
+	if err != nil {
+		return false, err
+	}
+
+	isProving, err := abi.Deserialize(msgResult[0], abi.Boolean)
+	if err != nil {
+		return false, err
```
acruikshank: Blocking: This needs to be wrapped in a fault errror

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -193,6 +197,53 @@ func (sma *Actor) UpdateStorage(vmctx exec.VMContext, delta *types.BytesAmount)
...
+	ctx := context.Background()
+	ret, err := actor.WithState(vmctx, &state, func() (interface{}, error) {
+		lu, err := actor.LoadLookup(ctx, vmctx.Storage(), state.Miners)
+		if err != nil {
+			return nil, err
+		}
+
+		vals, err := lu.Values(ctx)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
shannonwells: this functionality is now gone and GetPoStState already followed the error reporting protocol.

actor/builtin/storagemarket/storagemarket.go
shannonwells: GetProofsMode and GetTotalStorage don't wrap in fault errors -- should these be updated as well?

actor/builtin/storagemarket/storagemarket.go
```diff
@@ -193,6 +198,56 @@ func (sma *Actor) UpdateStorage(vmctx exec.VMContext, delta *types.BytesAmount)
...
 }
 
+func (sma *Actor) GetLateMiners(vmctx exec.VMContext) (*map[string]uint64, uint8, error) {
+	if err := vmctx.Charge(actor.DefaultGasCost); err != nil {
+		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
+	}
+	var state State
+	ctx := context.Background()
+
+	ret, err := actor.WithState(vmctx, &state, func() (interface{}, error) {
```
acruikshank: This read-only method can be simplified a bit by using the new `actor.ReadState` method instead of `actor.WithState`

testhelpers/core.go
```diff
@@ -113,11 +113,26 @@ func VMStorage() vm.StorageMap {
 	return vm.NewStorageMap(blockstore.NewBlockstore(datastore.NewMapDatastore()))
 }
 
-// CreateTestMiner creates a new test miner with the given peerID and miner
+// MustSign signs a given address with the provided mocksigner or panics if it
+// cannot.
+func MustSign(s types.MockSigner, msgs ...*types.Message) []*types.SignedMessage {
```
acruikshank: Why is this being added? It doesn't seem to be used in this PR.

shannonwells: I did not see that; it must have been pulled in when I did a rebase.

Issue #63, opened 2018-02-16
we should consider auto-generating the spec


phritz:
The spec is the ultimate commentary on the code, maybe it should live alongside/in the code? We should consider building a tool that generates the spec automatically from repo files, comments & code. Like godoc or javasdoc, but for the spec.

Spitballing here, hear me out. Looking over what's in the spec we have a few different things. 
* General commentary, motivation, and whatnot: this could live in a markdown file we use as the skeleton of the spec document. Additional content could be extracted from the source files into this skeleton. 
* Data structures: it's easy to annotate structures and fields that are top-level concepts either in comments ```// @TopLeveLConcept StateTree``` or using actual \``pl.spec`\` annotations on fields, and we could use annotations that map to spec types if we wanted. Labeled  comment sections and these data structures could be extracted directly.  
* Algorithms: should be in comments anyway. When we extract this description it's easy to automatically back-link to the code. 
* APIs are easy to annotate and extract. If we had an IDL that generated Go it'd be even easier, but we don't need it. 
* Other kinds of high-level descriptions like security properties: could be drawn from comments, readme/markdown files that live alongside the code

We could include composition hints in the comments or code for how to merge pieces from different source files (/at/Section Foo). 

Advantages: reduces risk of diversion between spec and implementation as we implement. You can see what is and is not implemented at a glace (annotations could differentiate them). We could flag changes that affect spec in code reviews. 

When we are "done" implementing we could detach the output spec from its inputs and just use it as a stand alone file if we wanted.

PR #3059, opened 2019-07-12
support version command over HTTP

acruikshank:
Addresses [issue #63 in the js-filecoin-api](https://github.com/filecoin-project/js-filecoin-api-client/issues/63).

### Problem

When we changed the `version` command to allow it to work without a running daemon, we ended up removing it from the HTTP API. This was unintentional.

### Solution

Restore the version command to the HTTP API. The `version` CLI command does not contact the daemon, and returns the version of the CLI client. Making a request to the /api/version endpoint now works and returns the version of the daemon.

This is maybe a little hacky, but it's a simple solution that works until we get a better HTTP API.

### DIFF
--- a/commands/main.go
+++ b/commands/main.go
@@ -183,6 +183,7 @@ var rootSubcmdsDaemon = map[string]*cmds.Command{
 	"stats":            statsCmd,
 	"swarm":            swarmCmd,
 	"wallet":           walletCmd,
+	"version":          versionCmd,
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


eingenito:
LGTM. Same challenge exists in IPFS.

mishmosh:
Thanks @eingenito. Could you go back and `Approve` in github? You should have access now, and that would be the requisite 2nd approval.

PR #3060, opened 2019-07-12
Removed duplicate block storage

### LINKED ISSUE #2636: Chain data stored twice on disk
### Description
Our design should not be storing chain data once in a bitswap datastore and then copying it all over to a chain datastore.  Originally I enforced this separation to reason about security.  Specifically I wanted to prevent code from ever assuming a block had been verified without verifying it first.  The clunky way the code does this now is by only writing verified blocks to the chain datastore and always assuming reads from that datastore are good.  

Right now this is utter overkill.  The datastore separation is not needed because (1) during node operation all tipsets added to the chainstore are kept in memory and (2) during load of a previously synced chain the node reads the head cids from the protected chain datastore.  Furthermore when we (3) stop keeping all chain store tipsets in memory we can simply store metadata in the protected chain store without putting all chain data in the protected store.

### Acceptance criteria
(1) Chain bits are only in one place on disk
(2) go-filecoin nodes still never ever read unverified blocks thinking that they are verified.

### Risks + pitfalls
Right now the default store *might* be going to disk in some cases to check that a block is verified.  All `Get` and `Has` chain store methods need to be carefully audited to make sure they don't violate AC (2) when doing away with chain copying.

### Where to begin
chain/default_store.go should be organized to never check membership in the conceptual chain based on data of a certain cid existing in the datastore.

ZenGround0:
This resolves #2636.  It removes writing blockchain data to the protected chain store.  `Load` now reads from the global IPLD store.

With this change the store's `*Block(s)` methods all no longer make sense and have been deleted.  This makes for a better interface.  Consumers should not be accessing individual blocks through the chain store, they should be accessing tipsets.  This descopes the chain store's interface to something more manageable and expected.  Consumers were already using this pattern for the most part.

### DIFF
--- a/chain/store.go
+++ b/chain/store.go
@@ -30,19 +30,30 @@ var logStore = logging.Logger("chain.store")
 
 var headKey = datastore.NewKey("/chain/heaviestTipSet")
 
+type ipldSource struct {
+	// bs is the blockstore from which IPLD objects are pulled
+	bs bstore.Blockstore
+}
+
+// GetBlock retrieves a filecoin block by cid from the IPLD source.
+func (source *ipldSource) GetBlock(ctx context.Context, c cid.Cid) (*types.Block, error) {
+	data, err := source.bs.Get(c)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get block %s", c.String())
+	}
+	return types.DecodeBlock(data.RawData())
+}
+
 // Store is a generic implementation of the Store interface.
 // It works(tm) for now.
 type Store struct {
-	// bsPriv is the on disk storage for blocks.  This is private to
-	// the Store to keep code that adds blocks to the Store's
-	// underlying storage isolated to this module.  It is important that only
-	// code with access to a Store can write to this storage to
-	// simplify checking the security guarantee that only tipsets of a
-	// validated chain are stored in the filecoin node's Store.
-	bsPriv bstore.Blockstore
-	// ds is the datastore backing bsPriv.  It is also accessed directly
-	// to set and get chain meta-data, specifically the tipset cidset to
-	// state root mapping, and the heaviest tipset cids.
+	// ipldSource is a wrapper around ipld storage.  It is used
+	// for reading filecoin block and state objects kept by the node.
+	ipldSource *ipldSource
+
+	// ds is the datastore for the chain's private metadata which consists
+	// of the tipset key to state root cid mapping, and the heaviest tipset
+	// key.
 	ds repo.Datastore
 
 	// genesis is the CID of the genesis block.
@@ -65,10 +76,9 @@ type Store struct {
 }
 
 // NewStore constructs a new default store.
-func NewStore(ds repo.Datastore, genesisCid cid.Cid) *Store {
-	priv := bstore.NewBlockstore(ds)
+func NewStore(ds repo.Datastore, bs bstore.Blockstore, genesisCid cid.Cid) *Store {
 	return &Store{
-		bsPriv:     priv,
+		ipldSource: &ipldSource{bs: bs},
 		ds:         ds,
 		headEvents: pubsub.New(128),
 		tipIndex:   NewTipIndex(),
@@ -103,7 +113,7 @@ func (store *Store) Load(ctx context.Context) (err error) {
 		return err
 	}
 
-	headTs, err := LoadTipSetBlocks(ctx, store, headTsKey)
+	headTs, err := LoadTipSetBlocks(ctx, store.ipldSource, headTsKey)
 	if err != nil {
 		return errors.Wrap(err, "error loading head tipset")
 	}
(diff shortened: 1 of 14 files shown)


chain/syncer_test.go
```diff
@@ -193,33 +195,33 @@ func requireSetTestChain(t *testing.T, con consensus.Protocol, mockStateRoots bo
 func loadSyncerFromRepo(t *testing.T, r repo.Repo, dstP *SyncerTestParams) (*chain.Syncer, *th.TestFetcher) {
 	powerTable := &th.TestView{}
 	bs := bstore.NewBlockstore(r.Datastore())
-	cst := hamt.NewCborStore()
+	cst := &hamt.CborIpldStore{Blocks: bserv.New(bs, offline.Exchange(bs))}
```
ZenGround0: These changes aren't strictly necessary but put our tests in a state where the global IPLD store approaches global.  I can change back if desired.

chain/syncer_test.go
```diff
@@ -770,6 +754,16 @@ func TestLoadFork(t *testing.T) {
 	require.NoError(t, err)
 	requireHead(t, chainStore, forklink3)
 
+	// Put blocks in global IPLD blockstore
+	// TODO #2128 make this cleaner along with broad test cleanup.
```
ZenGround0: We can make this cleaner (but more opaque) by forcing the blockSource (fetcher) to be backed by the same repo/globalIPLD store.

mining/worker.go
```diff
@@ -92,7 +91,6 @@ type DefaultWorker struct {
 	processor     MessageApplier
 	powerTable    consensus.PowerTableView
 	blockstore    blockstore.Blockstore
-	cstore        *hamt.CborIpldStore
```
ZenGround0: I removed this unused reference

chain/store.go
```diff
@@ -30,19 +30,30 @@ var logStore = logging.Logger("chain.store")
 
 var headKey = datastore.NewKey("/chain/heaviestTipSet")
 
+type globalIPLDSource struct {
+	// bs is the blockstore from which IPLD objects are pulled
+	bs bstore.Blockstore
+}
+
+// GetBlock retrieves a filecoin block by cid from the global IPLD source.
+func (source *globalIPLDSource) GetBlock(ctx context.Context, c cid.Cid) (*types.Block, error) {
```
frrist: Initially the addition of this method and structure seemed out of place to me. I _think_ we are adding this for `LoadTipSetBlocks` and thus the `TipsetIterator` to use, is that correct? I don't really like this, but I don't have a better suggestion. Gut feeling says `IterAncestors` could be a method on the ChainStore, but maybe there is a better way. 

Fine leaving this the way it is -- it's an improvement, but I think we can make this cleaner after a few more related changes land.

chain/store.go
```diff
@@ -251,50 +246,6 @@ func (store *Store) HasTipSetAndStatesWithParentsAndHeight(parentKey string, h u
 	return store.tipIndex.HasByParentsAndHeight(parentKey, h)
 }
 
-// GetBlocks retrieves the blocks referenced in the input cid set.
```
frrist: Loving the color red here! :smiley:

node/node.go
```diff
@@ -395,7 +394,7 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 	}
 
 	// set up chainstore
-	chainStore := chain.NewStore(nc.Repo.ChainDatastore(), genCid)
+	chainStore := chain.NewStore(nc.Repo.ChainDatastore(), bs, genCid)
 	chainState := cst.NewChainStateProvider(chainStore, &cstOffline)
```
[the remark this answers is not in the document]
ZenGround0: PR in draft at #3061 attempts to improve the state of this a bit.  Looking forward to having this discussion in more depth on that one.

chain/store.go
ZenGround0: >  I think we are adding this for LoadTipSetBlocks and thus the TipsetIterator to use, is that correct?

correct
>  I don't really like this

me neither

> I think we can make this cleaner after a few more related changes land.

It seems like we're having the same thoughts. Feels good!

chain/store.go
```diff
@@ -30,19 +30,30 @@ var logStore = logging.Logger("chain.store")
 
 var headKey = datastore.NewKey("/chain/heaviestTipSet")
 
+type globalIPLDSource struct {
```
anorth: I don't like the name "global" here. The chain store shouldn't be making broad assertions outside its function. Better could be just `ipldSource`, `chainAndStateSource`.

chain/store.go
anorth: This makes sense to me as as you have it, so far.

chain/syncer_test.go
anorth: I think the fetcher is supposed to represent network, so we won't be exercising the right paths if the stuff to be fetched is already local.

plumbing/cst/chain_state.go
```diff
@@ -28,8 +27,8 @@ type chainReader interface {
 // ChainStateProvider composes a chain and a state store to provide access to
 // the state (including actors) derived from a chain.
 type ChainStateProvider struct {
-	reader chainReader         // Provides chain blocks and tipsets.
-	cst    *hamt.CborIpldStore // Provides state trees.
+	reader chainReader         // Provides chain tipsets and state roots.
+	cst    *hamt.CborIpldStore // Provides chain blocks and state trees.
```
anorth: It's unfortunate that the ChainStateProvider is required to provide blocks at all. It's worth a quick investigation to see if you can fix the callers to not need that, then revert this documentation not to reference chain blocks at all.

ZenGround0: `go-filecoin show block` needs this right now so I'm leaving it for this PR.

chain/store.go
ZenGround0: I understand where you're coming from.  Naming it like this came from an impulse to identify an important and too easy to misunderstand property of the ipld store.  I'll change it back though.

PR #3061, opened 2019-07-12
Add state fetching to chain store

ZenGround0:
This PR adds the `GetTipSetState` endpoint, and updates consumers of state fetching and the chain store tests.

It also factors the blockstore out of the chain store so that all IPLD stuff is pulled out of the cborstore.

This change makes a tiny addition to the chain builder testing framework for setting stateCids in the block construction callback in order to use the new framework for the `GetTipSetState` test.


chain/syncer.go
```diff
@@ -75,7 +74,7 @@ type syncerChainReader interface {
 	BlockHeight() (uint64, error)
 	GetHead() types.TipSetKey
 	GetTipSet(tsKey types.TipSetKey) (types.TipSet, error)
-	GetTipSetStateRoot(tsKey types.TipSetKey) (cid.Cid, error)
+	GetTipSetState(ctx context.Context, tsKey types.TipSetKey) (state.Tree, error)
```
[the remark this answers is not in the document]
ZenGround0: See [here](https://github.com/filecoin-project/go-filecoin/pull/3064#pullrequestreview-261832764) #3064 invalidates a lot of this work in a big way and requires rethinking the approach here.

chain/store.go
```diff
@@ -30,26 +34,42 @@ var logStore = logging.Logger("chain.store")
...
 	bs bstore.Blockstore
+	// cst is an alternate view into the blockstore allowing for easy cbor
+	// (un)marshalling and interop with go-ipld-hamt.
+	cst *hamt.CborIpldStore
+}
+
+func newGlobalIPLDRef(bs bstore.Blockstore) *globalIPLDRef{
+	return &globalIPLDRef{
+		bs:  bs,
+		cst: &hamt.CborIpldStore{Blocks: bserv.New(bs, offline.Exchange(bs))},
```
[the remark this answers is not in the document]
ZenGround0: We could remove the blockstore if we accept that we will not run `types.DecodeBlock` during decoding.  The only downside to this is that we lose out on caching the block's raw bytes.

chain/store.go
```diff
@@ -31,25 +32,33 @@ var logStore = logging.Logger("chain.store")
 var headKey = datastore.NewKey("/chain/heaviestTipSet")
 
 type ipldSource struct {
-	// bs is the blockstore from which IPLD objects are pulled
-	bs bstore.Blockstore
+	// cst is a store allowing access
+	// (un)marshalling and interop with go-ipld-hamt.
+	cst *hamt.CborIpldStore
```
ZenGround0: The tradeoff to using this single store instead of this and the blockstore is that go-filecoin has less control over decoding process.  Practically this means that blocks decoded during load no longer cache their serialized bytes [here](https://github.com/filecoin-project/go-filecoin/blob/master/types/block.go#L127)

node/node.go
```diff
@@ -387,15 +386,15 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
...
 	genCid, err := readGenesisCid(nc.Repo.Datastore())
 	if err != nil {
 		return nil, err
 	}
 
 	// set up chainstore
-	chainStore := chain.NewStore(nc.Repo.ChainDatastore(), bs, genCid)
-	chainState := cst.NewChainStateProvider(chainStore, &cstOffline)
+	chainStore := chain.NewStore(nc.Repo.ChainDatastore(), &ipldCborStore, genCid)
+	chainState := cst.NewChainStateProvider(chainStore, &ipldCborStore)
```
anorth: In follow up, I think here and all the places below that get a chain store and cbor store separately should be able to use just the (wider) chain store interface now.

plumbing/cst/chain_state.go
```diff
@@ -117,7 +112,7 @@ func (chn *ChainStateProvider) GetActorAt(ctx context.Context, tipKey types.TipS
 
 // LsActors returns a channel with actors from the latest state on the chain
 func (chn *ChainStateProvider) LsActors(ctx context.Context) (<-chan state.GetAllActorsResult, error) {
-	st, err := chain.LatestState(ctx, chn.reader, chn.cst)
+	st, err := chn.reader.GetTipSetState(ctx, chn.reader.GetHead())
```
anorth: 👍

chain/store.go
ZenGround0: Check out #3078 for a potential plan on improving the state of this.

PR #3063, opened 2019-07-15
Implement a chain.Builder and use it to replace a bunch of uses of initSyncTestDefault

anorth:
This PR is in two commits, though reviewable all at once. The first commit implements the chain.Builder and replaces existing uses of FakeChainProvider. The second uses this to replace a bunch of chain package tests that needlessly depend on the complicated common test setup in syncer_test.go, getting us closer to cleaning that up. These tests no longer construct block/state/chain stores.

There's still a `th.TestFetcher` which I hope this will also replace, but it's more tightly bound in with the syncer tests.


anorth:
Future work may be able to replace the fake state tree calculation with a real one, but I don't think that anything we're currently doing should depend on details about the state, so my preferred approach is to try to factor that out of the syncer.

chain/testing.go
```diff
@@ -0,0 +1,243 @@
...
+		fakeState = append(fakeState, mCid)
+	}
+
+	return makeCid(fakeState)
+}
+
+///// Interface implementations /////
+
+// GetBlock returns the block identified by `c`.
+func (f *Builder) GetBlock(ctx context.Context, c cid.Cid) (*types.Block, error) {
```
ZenGround0: This doesn't seem to be needed by syncFetcher or TipSetProvider.  Is there another interface this is trying to satisfy?

plumbing/cst/chain_state.go
```diff
@@ -75,7 +77,13 @@ func (chn *ChainStateProvider) GetBlock(ctx context.Context, id cid.Cid) (*types
 
 // SampleRandomness samples randomness from the chain at the given height.
 func (chn *ChainStateProvider) SampleRandomness(ctx context.Context, sampleHeight *types.BlockHeight) ([]byte, error) {
-	tipSetBuffer, err := chain.GetRecentAncestorsOfHeaviestChain(ctx, chn.reader, sampleHeight)
+	head := chn.reader.GetHead()
```
frrist: These changes appear unrelated to the rest of the changes in this PR, is this an optimization?

chain/reorg_test.go
```diff
@@ -68,79 +68,29 @@ func TestReorgDiffSubset(t *testing.T) {
 // block is 'a' blocks after the genesis block.  The  main chain has an
 // additional 'b' blocks, the fork has an additional 'c' blocks.  This function
 // returns the forked head, the main head and the common ancestor
```
ZenGround0: Let's add a precondition that `a > 0` in the comment.  This method should fail more gracefully than before this change when `a==0`(I think it fails [here](https://github.com/filecoin-project/go-filecoin/pull/3063/files#diff-8a85ecf0d5b5586324ba95f4530a6a0dR82) now) but still worth a warning.

chain/testing.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// BuildToTip creates and returns anew block child of `parent`.
+// The `build` function is invoked to modify the block before it is stored.
+func (f *Builder) BuildToTip(parent types.TipSet, build func(b *BlockBuilder)) *types.Block {
+	ticket := make([]byte, binary.Size(f.seq))
+	binary.BigEndian.PutUint64(ticket, f.seq)
+	f.seq++
+
+	// Sum weight of parents' parent weight, plus one for each parent.
```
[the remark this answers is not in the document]
anorth: I was trying to maintain just the easy property that weight increased. I agree with your note elsewhere that it should be configurable, and I've left a comment to that effect.

chain/testing.go
```diff
@@ -0,0 +1,243 @@
...
+func (f *Builder) AppendTo(parents ...*types.Block) *types.Block {
+	tip := types.UndefTipSet
+	if len(parents) > 0 {
+		tip = types.RequireNewTipSet(f.t, parents...)
+	}
+	return f.BuildToTip(tip, nil)
+}
+
+// AppendManyTo appends `count` blocks to the chain.
+func (f *Builder) AppendManyTo(count int, parents ...*types.Block) *types.Block {
```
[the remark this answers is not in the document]
anorth: I agree, but can't think of anything better right now.

chain/testing.go
anorth: `BlockProvider` in traversal. I added another static check for it.

plumbing/cst/chain_state.go
anorth: This inlining GetRecentAncestorsOfHeaviestChain into its only call site narrows the interface of the chain reader passed on to GetRecentAncestors so that it can be satisfied by the new builder (no longer needs to know about head).

chain/testing.go
```diff
@@ -0,0 +1,243 @@
+package chain
```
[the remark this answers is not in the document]
anorth: It would not be overkill, but I'mma land this now without that.

Issue #2943, opened 2019-06-16
Placeholder: implement ArbitrateDeal actor method


anorth:
### Description
As part of storage protocol dispute resolution, the storage miner must have an `ArbitrateDeal` method, which transfers some part of the miner actor's storage collateral (distinct from pledge collateral) to the client of a failed deal.

See [faults and slashing design doc](https://docs.google.com/document/d/1U3b9GVNVOLoS_-q9kXU3a9dqHGvcrHprwDFzNYgLI0M/edit?pli=1#heading=h.anj0qf3angix) and [actor spec](https://github.com/filecoin-project/specs/blob/master/actors.md#arbitratedeal)

Work includes:
- actor method implementation
- CLI commands for invoking deal arbitration on behalf of a client

This should probably be broken out into more stories for implementation.

anorth:
Implementation is blocked on piece inclusion proofs from the prover. This is also waiting on https://github.com/filecoin-project/specs/issues/60, https://github.com/filecoin-project/specs/issues/159, https://github.com/filecoin-project/specs/issues/306

acruikshank:
When this feature is implemented, it will need to temproarily store a set of arbitrated deals in the miner actor `ArbitratedDealSet` to prevent a client from arbitrating the same deal twice. This set will need to be cleared in `SubmitPoSt` which is where this work intersects with the SubmitPoSt revamp #2813.

acruikshank:
This is no longer blocked on piece inclusion proofs, but it is blocked on #1211

anorth:
The October 2019 spec dispenses with this arbitration, in favour of automatic arbitration via a cron system actor.

PR #3065, opened 2019-07-15
implement miner.SlashStorageFault

### LINKED ISSUE #2945: Implement SlashStorageFault miner actor method
### Description
The `SlashStorageFault` method is invoked when a miner fails to submit a PoSt before the generation attack time threshold has passed after the end of the proving period. This burns all pledge collateral and sets power to zero.

See the [faults and slashing design doc](https://docs.google.com/document/d/1U3b9GVNVOLoS_-q9kXU3a9dqHGvcrHprwDFzNYgLI0M/edit?pli=1#) for context. The method is [specified here](https://github.com/filecoin-project/specs/blob/master/actors.md#slashstoragefault).

There are a number of spec changes to be aware of when implementing: https://github.com/filecoin-project/specs/issues/60, https://github.com/filecoin-project/specs/issues/329. See also https://github.com/filecoin-project/specs/pull/325 which defines pledge collateral dynamically.

https://github.com/filecoin-project/specs/issues/159 suggests a dutch auction mechanism for rewarding the slasher. If that's not resolved yet then implement slashing without the reward mechanism for now.

### Acceptance criteria
- StorageMiner.SlashStorageFault is implemented in accordance with the spec as much as possible

### Risks + pitfalls
- The definition of pledge collateral is in flux

acruikshank:
closes #2945

### Problem

SlashStorageFault in miner actor is currently unimplemented. This function has been in flux in [the spec](https://github.com/filecoin-project/specs/blob/master/actors.md#slashstoragefault). What is in the spec is fairly reasonable, but has some potential issues:

1. The spec defines a SlashedSet in actor state that is never used (implemented anyway).
2. The algorithm describes actions for after GenerationAttackTimeout, PoStTimout, and SectorFailureTimout in text, but not in the algorithm (this PR implements the algorithm).
3. The spec is in flux on what storage collateral means and how much should be set aside for deal arbitration (this implementation uses `getPledgeCollateralRequirement` with a TODO to revise).

Despite the issues, this method gives us something to call once we've identified late miners (#3057) and allows us to set aside funds to pay out in arbitrated deals (#2943).

### Solution

Implement `miner.SlashStorageFault` as faithfully to the algorithm described in spec as possible and test it. There are a few things to look for in this PR:

1. It adds 2 new error codes to miner concerning slashable status.
2. It adds `SlashedSet`, `SlashedAt`, and `OwedStorageCollateral` to miner state.
3. The `SlashStorageFault` method is the main contribution in this PR.
4. It adds a few test helpers and removes a few that are no longer used.
5. I was hitting a problem in GenGen where an IntSet was failing because it's internal data structure was nil. Returning a zero value when it is nil, seemed reasonable, so I did that rather than track down what was causing the nil value.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -210,6 +210,7 @@ func NewState(owner, worker address.Address, pid peer.ID, sectorSize *types.Byte
 		Power:             types.NewBytesAmount(0),
 		NextAskID:         big.NewInt(0),
 		SectorSize:        sectorSize,
+		SlashedSet:        types.EmptyIntSet(),
 		ActiveCollateral:  types.ZeroAttoFIL,
 	}
 }
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -76,6 +76,10 @@ const (
 	ErrGetProofsModeFailed = 42
 	// ErrInsufficientCollateral indicates that the miner does not have sufficient collateral to commit additional sectors.
 	ErrInsufficientCollateral = 43
+	// ErrMinerAlreadySlashed indicates that an attempt has been made to slash an already slashed miner
+	ErrMinerAlreadySlashed = 44
+	// ErrMinerNotSlashable indicates that an attempt has been made to slash a miner that does not meet the criteria for slashing.
+	ErrMinerNotSlashable = 45
 )
 
 // Errors map error codes to revert errors this actor may return.
@@ -177,6 +181,16 @@ type State struct {
 	// SectorSize is the amount of space in each sector committed to the network
 	// by this miner.
 	SectorSize *types.BytesAmount
+
+	// SlashedSet is a set of sector ids that have been slashed
+	SlashedSet types.IntSet
+
+	// SlashedAt is the time at which this miner was slashed
+	SlashedAt *types.BlockHeight
+
+	// OwedStorageCollateral is the collateral for sectors that have been slashed.
+	// This collateral can be collected from arbitrated deals, but not de-pledged.
+	OwedStorageCollateral types.AttoFIL
 }
 
 // NewActor returns a new miner actor with the provided balance.
@@ -254,6 +268,10 @@ var minerExports = exec.Exports{
 	},
 	"submitPoSt": &exec.FunctionSignature{
 		Params: []abi.Type{abi.PoStProofs, abi.IntSet},
+		Return: []abi.Type{},
+	},
+	"slashStorageFault": &exec.FunctionSignature{
+		Params: []abi.Type{},
 		Return: []abi.Type{},
 	},
 	"changeWorker": &exec.FunctionSignature{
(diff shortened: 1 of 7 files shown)


anorth:
FYI see PR https://github.com/filecoin-project/specs/pull/379, which adds more detail about those timeouts etc (tho not the pseudocode) but also raises more questions. I don't think it's ready for final implementation yet. That said, implementing something so we can exercise these paths is good, but we'll be back here to fix them up later.

actor/builtin/miner/miner.go
```diff
@@ -955,6 +973,70 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
...
+			return nil, Errors[ErrStoragemarketCallFailed]
+		}
+		state.Power = types.NewBytesAmount(0)
+
+		// record what has been slashed
+		state.SlashedSet = state.ProvingSet
+
+		// reserve collateral for arbitration
+		// TODO: This calculation is probably not correct: https://github.com/filecoin-project/go-filecoin/issues/3050
+		state.OwedStorageCollateral = ma.getPledgeCollateralRequirement(state, chainHeight)
```
anorth: This is definitely wrong. I think we should use zero here until a workable definition emerges: https://github.com/filecoin-project/specs/issues/385, https://github.com/filecoin-project/specs/issues/386

actor/builtin/miner/miner_test.go
```diff
@@ -1143,7 +1273,21 @@ func mustDeserializeAddress(t *testing.T, result [][]byte) address.Address {
...
+		panic(err)
+	}
+
+	minerState := &State{}
+	err = cbor.DecodeInto(data, minerState)
+	if err != nil {
+		panic(err)
+	}
+
+	return minerState
```
anorth: Super useful! Could we make something like this more available for worker code?

actor/builtin/miner/miner.go
```diff
@@ -177,6 +181,16 @@ type State struct {
 	// SectorSize is the amount of space in each sector committed to the network
 	// by this miner.
 	SectorSize *types.BytesAmount
+
+	// SlashedSet is a set of sector ids that have been slashed
+	SlashedSet types.IntSet
```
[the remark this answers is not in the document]
acruikshank: I think this is unambiguous, but let me lay out my understanding for clarity:

1. "Slash" is a verb. It's something one user does to another (technicially a miner can slash themself, but anyone can). Specifically slashing (storage slashing) is the act of sending a message to the miner actor's SlashStorageFault method to notify the network that the miner is out of compliance and take corrective action.
2. A user cannot slash a miner for a faulting on a sector. A slash covers all sectors in the proving set.
3. A miner may declare faults and pay a small penalty for them in SubmitPoSt. It's only after the miner fails to PoSt altogether (or PoSts unsuccessfully) that they are liable to be slashed.

This is all based on the [implementations in the spec](https://github.com/filecoin-project/specs/blob/master/actors.md) which used `SlashedAt` and `SlashedSet` only in `SlashStorageFault`. Other spec information might contradict it at the moment. The goal was to stick to the spec implementation.

PR #3066, opened 2019-07-15
Features/piece inclusion proofs

sidke:
## What’s in this PR?

- swap out stubbed comm_p and pip verification methods for rfp implementations
- adjust condition parameters to use `PieceSize` for pip verification

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -1,7 +1,6 @@
 package miner
 
 import (
-	"bytes"
 	"math/big"
 
 	"github.com/ipfs/go-cid"
@@ -80,20 +79,24 @@ const (
 	ErrMinerAlreadySlashed = 44
 	// ErrMinerNotSlashable indicates that an attempt has been made to slash a miner that does not meet the criteria for slashing.
 	ErrMinerNotSlashable = 45
+	// ErrInvalidPieceInclusionProof indicates that the piece inclusion proof was
+	// malformed or did not succesfully verify.
+	ErrInvalidPieceInclusionProof = 46
 )
 
 // Errors map error codes to revert errors this actor may return.
 var Errors = map[uint8]error{
-	ErrInvalidSector:           errors.NewCodedRevertErrorf(ErrInvalidSector, "sectorID out of range"),
-	ErrSectorIDInUse:           errors.NewCodedRevertErrorf(ErrSectorIDInUse, "sector already committed at this ID"),
-	ErrStoragemarketCallFailed: errors.NewCodedRevertErrorf(ErrStoragemarketCallFailed, "call to StorageMarket failed"),
-	ErrCallerUnauthorized:      errors.NewCodedRevertErrorf(ErrCallerUnauthorized, "not authorized to call the method"),
-	ErrInsufficientPledge:      errors.NewCodedRevertErrorf(ErrInsufficientPledge, "not enough pledged"),
-	ErrInvalidPoSt:             errors.NewCodedRevertErrorf(ErrInvalidPoSt, "PoSt proof did not validate"),
-	ErrAskNotFound:             errors.NewCodedRevertErrorf(ErrAskNotFound, "no ask was found"),
-	ErrInvalidSealProof:        errors.NewCodedRevertErrorf(ErrInvalidSealProof, "seal proof was invalid"),
-	ErrGetProofsModeFailed:     errors.NewCodedRevertErrorf(ErrGetProofsModeFailed, "failed to get proofs mode"),
-	ErrInsufficientCollateral:  errors.NewCodedRevertErrorf(ErrInsufficientCollateral, "insufficient collateral"),
+	ErrInvalidSector:              errors.NewCodedRevertErrorf(ErrInvalidSector, "sectorID out of range"),
+	ErrSectorIDInUse:              errors.NewCodedRevertErrorf(ErrSectorIDInUse, "sector already committed at this ID"),
+	ErrStoragemarketCallFailed:    errors.NewCodedRevertErrorf(ErrStoragemarketCallFailed, "call to StorageMarket failed"),
+	ErrCallerUnauthorized:         errors.NewCodedRevertErrorf(ErrCallerUnauthorized, "not authorized to call the method"),
+	ErrInsufficientPledge:         errors.NewCodedRevertErrorf(ErrInsufficientPledge, "not enough pledged"),
+	ErrInvalidPoSt:                errors.NewCodedRevertErrorf(ErrInvalidPoSt, "PoSt proof did not validate"),
+	ErrAskNotFound:                errors.NewCodedRevertErrorf(ErrAskNotFound, "no ask was found"),
+	ErrInvalidSealProof:           errors.NewCodedRevertErrorf(ErrInvalidSealProof, "seal proof was invalid"),
+	ErrGetProofsModeFailed:        errors.NewCodedRevertErrorf(ErrGetProofsModeFailed, "failed to get proofs mode"),
+	ErrInsufficientCollateral:     errors.NewCodedRevertErrorf(ErrInsufficientCollateral, "insufficient collateral"),
+	ErrInvalidPieceInclusionProof: errors.NewCodedRevertErrorf(ErrInvalidPieceInclusionProof, "piece inclusion proof did not validate"),
 }
 
 const (
@@ -281,7 +284,7 @@ var minerExports = exec.Exports{
 	},
 	// verifyPieceInclusion is not in spec, but should be.
 	"verifyPieceInclusion": &exec.FunctionSignature{
-		Params: []abi.Type{abi.Bytes, abi.SectorID, abi.Bytes},
+		Params: []abi.Type{abi.Bytes, abi.BytesAmount, abi.SectorID, abi.Bytes},
 		Return: []abi.Type{},
 	},
 	"getSectorSize": &exec.FunctionSignature{
@@ -672,7 +675,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 
 // VerifyPieceInclusion verifies that proof proves that the data represented by commP is included in the sector.
 // This method returns nothing if the verification succeeds and returns a revert error if verification fails.
-func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, sectorID uint64, proof []byte) (uint8, error) {
+func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, pieceSize *types.BytesAmount, sectorID uint64, proof []byte) (uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
@@ -699,13 +702,19 @@ func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, sectorID
 		// Verify proof proves CommP is in sector's CommD
 		var typedCommP types.CommP
 		copy(typedCommP[:], commP)
-		valid, err := verifyInclusionProof(typedCommP, commitment.CommD, proof)
-		if err != nil {
-			return nil, err
-		}
 
-		if !valid {
-			return nil, errors.NewRevertError("invalid inclusion proof")
+		res, err := ctx.Verifier().VerifyPieceInclusionProof(verification.VerifyPieceInclusionProofRequest{
+			CommD:               commitment.CommD,
+			CommP:               typedCommP,
+			PieceInclusionProof: proof,
+			PieceSize:           pieceSize,
+			SectorSize:          state.SectorSize,
+		})
+		if err != nil {
+			return nil, errors.RevertErrorWrap(err, "failed to verify piece inclusion proof")
+		}
+		if !res.IsValid {
+			return nil, Errors[ErrInvalidPieceInclusionProof]
 		}
 
 		return nil, nil
@@ -1181,17 +1190,4 @@ func lateState(provingPeriodEnd *types.BlockHeight, chainHeight *types.BlockHeig
 		return PoStStateAfterProvingPeriod, roundsLate
 	}
 	return PoStStateWithinProvingPeriod, roundsLate
-}
-
-// TODO: This is a fake implementation pending availability of the verification algorithm in rust proofs
-// see https://github.com/filecoin-project/go-filecoin/issues/2629
-func verifyInclusionProof(commP types.CommP, commD types.CommD, proof []byte) (bool, error) {
-	if len(proof) != 2*int(types.CommitmentBytesLen) {
-		return false, errors.NewRevertError("malformed inclusion proof")
-	}
-	combined := []byte{}
-	combined = append(combined, commP[:]...)
-	combined = append(combined, commD[:]...)
-
-	return bytes.Equal(combined, proof), nil
 }
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -718,9 +718,9 @@ func TestMinerSubmitPoStVerification(t *testing.T) {
 
 		// The 3 sector is not in the proving set, so its CommR should not appear in the VerifyPoSt request
 		minerState.ProvingSet = types.NewIntSet(1, 2)
-		verifier := verification.NewFakeVerifier(verification.FakeVerifierConfig{
+		verifier := &verification.FakeVerifier{
 			VerifyPoStValid: true,
-		})
+		}
 		vmctx := th.NewFakeVMContextWithVerifier(message, minerState, verifier)
 
 		miner := Actor{Bootstrap: false}
@@ -771,9 +771,9 @@ func TestMinerSubmitPoStVerification(t *testing.T) {
 		minerState.SectorCommitments.Add(1, types.Commitments{CommR: comm1.CommR})
 
 		minerState.ProvingSet = types.NewIntSet(1)
-		verifier := verification.NewFakeVerifier(verification.FakeVerifierConfig{
+		verifier := &verification.FakeVerifier{
 			VerifyPoStError: errors.New("verifier error"),
-		})
+		}
 
 		vmctx := th.NewFakeVMContextWithVerifier(message, minerState, verifier)
 
@@ -794,9 +794,9 @@ func TestMinerSubmitPoStVerification(t *testing.T) {
 		minerState.SectorCommitments.Add(1, types.Commitments{CommR: comm1.CommR})
 
 		minerState.ProvingSet = types.NewIntSet(1)
-		verifier := verification.NewFakeVerifier(verification.FakeVerifierConfig{
+		verifier := &verification.FakeVerifier{
 			VerifyPoStValid: false,
-		})
+		}
 
 		vmctx := th.NewFakeVMContextWithVerifier(message, minerState, verifier)
 
@@ -1243,10 +1243,10 @@ func TestVerifyPIP(t *testing.T) {
 			message:   "verifyPieceInclusionProof",
 			sectorSet: sectorSetWithOneCommitment,
 			lastPoSt:  types.NewBlockHeight(0),
-			verifier: verification.NewFakeVerifier(verification.FakeVerifierConfig{
+			verifier: &verification.FakeVerifier{
 				VerifyPieceInclusionProofValid: false,
 				VerifyPieceInclusionProofError: errors.New("wombat"),
-			}),
+			},
 		}).build()
 
 		code, err := minerActor.VerifyPieceInclusion(vmctx, commP, types.NewBytesAmount(66), 1, []byte{42})
@@ -1266,9 +1266,9 @@ func TestVerifyPIP(t *testing.T) {
 			message:   "verifyPieceInclusionProof",
 			sectorSet: sectorSetWithOneCommitment,
 			lastPoSt:  types.NewBlockHeight(0),
-			verifier: verification.NewFakeVerifier(verification.FakeVerifierConfig{
+			verifier: &verification.FakeVerifier{
 				VerifyPieceInclusionProofValid: false,
-			}),
+			},
 		}).build()
 
 		code, err := minerActor.VerifyPieceInclusion(vmctx, commP, types.NewBytesAmount(66), 1, []byte{42})
@@ -1289,9 +1289,9 @@ func TestVerifyPIP(t *testing.T) {
 			message:   "verifyPieceInclusionProof",
 			sectorSet: sectorSetWithOneCommitment,
 			lastPoSt:  types.NewBlockHeight(0),
-			verifier: verification.NewFakeVerifier(verification.FakeVerifierConfig{
+			verifier: &verification.FakeVerifier{
 				VerifyPieceInclusionProofValid: true,
-			}),
+			},
 		}).build()
 
 		code, err := minerActor.VerifyPieceInclusion(vmctx, commP, types.NewBytesAmount(66), 1, []byte{42})
@@ -1310,9 +1310,9 @@ func TestVerifyPIP(t *testing.T) {
 			message:   "verifyPieceInclusionProof",
 			sectorSet: sectorSetWithOneCommitment,
 			lastPoSt:  types.NewBlockHeight(PieceInclusionGracePeriodBlocks),
-			verifier: verification.NewFakeVerifier(verification.FakeVerifierConfig{
+			verifier: &verification.FakeVerifier{
 				VerifyPieceInclusionProofValid: true,
-			}),
+			},
 		}).build()
 
 		code, err := minerActor.VerifyPieceInclusion(vmctx, commP, types.NewBytesAmount(66), 1, []byte{42})
@@ -1515,7 +1515,7 @@ func (b *minerEnvBuilder) build() (exec.VMContext, *verification.FakeVerifier, *
 	}
 
 	if b.verifier == nil {
-		b.verifier = verification.NewFakeVerifier(verification.FakeVerifierConfig{})
+		b.verifier = &verification.FakeVerifier{}
 	}
 
 	vmctx := th.NewFakeVMContextWithVerifier(types.NewMessage(address.TestAddress, address.TestAddress2, 0, types.ZeroAttoFIL, b.message, nil), minerState, b.verifier)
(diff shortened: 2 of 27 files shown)


actor/builtin/miner/miner.go
```diff
@@ -76,20 +75,23 @@ const (
 	ErrGetProofsModeFailed = 42
 	// ErrInsufficientCollateral indicates that the miner does not have sufficient collateral to commit additional sectors.
 	ErrInsufficientCollateral = 43
+	// comment
```
sidke: todo: actual description

actor/builtin/miner/miner_test.go
```diff
@@ -935,32 +935,26 @@ func TestVerifyPIP(t *testing.T) {
 	minerAddr := th.CreateTestMiner(t, st, vms, address.TestAddress, origPid)
 
 	sectorId := uint64(1)
-	commD := th.MakeCommitment()
+	pip := []byte{24, 0, 0, 0, 0, 0, 0, 0, 92, 46, 114, 56, 244, 245, 178, 65, 110, 127, 140, 185, 130, 36, 177, 209, 113, 122, 121, 150, 227, 8, 36, 77, 9, 242, 66, 104, 144, 231, 230, 105, 229, 170, 199, 185, 252, 141, 159, 211, 47, 128, 21, 148, 42, 6, 242, 12, 91, 193, 150, 55, 181, 213, 107, 234, 29, 139, 241, 240, 236, 154, 176, 51}
```
sidke: @laser probably doesn't like this

actor/builtin/miner/miner_test.go
```diff
@@ -977,55 +971,55 @@ func TestVerifyPIP(t *testing.T) {
...
 		})
 
-		t.Run("PIP is invalid if proof is invalid", func(t *testing.T) {
-			wrongPIP := append([]byte{pip[0] + 1}, pip[1:]...)
-			err := runVerifyPIP(t, 3, commP, sectorId, wrongPIP)
-
-			require.Error(t, err)
-			assert.Equal(t, "invalid inclusion proof", err.Error())
-		})
+		//t.Run("PIP is invalid if proof is invalid", func(t *testing.T) {
```
sidke: uncomment this test back in and fix it

actor/builtin/miner_redeem_test.go
```diff
@@ -41,8 +41,10 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
 	// Create a miner actor with fake commitments
 	minerAddr := addrGetter()
 	sectorID := uint64(123)
-	commP := []byte{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
-	commD := []byte{0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7}
+	pip := []byte{24, 0, 0, 0, 0, 0, 0, 0, 92, 46, 114, 56, 244, 245, 178, 65, 110, 127, 140, 185, 130, 36, 177, 209, 113, 122, 121, 150, 227, 8, 36, 77, 9, 242, 66, 104, 144, 231, 230, 105, 229, 170, 199, 185, 252, 141, 159, 211, 47, 128, 21, 148, 42, 6, 242, 12, 91, 193, 150, 55, 181, 213, 107, 234, 29, 139, 241, 240, 236, 154, 176, 51}
```
sidke: @laser won't like again

link.sh
```diff
@@ -0,0 +1,7 @@
+rm ~/Projects/filecoin/go-filecoin/proofs/lib/libsector_builder_ffi.a
```
sidke: remove this file

proofs/libsectorbuilder/cgo_bindings.go
```diff
@@ -403,3 +405,50 @@ func GeneratePoSt(
...
+	defer C.sector_builder_ffi_destroy_verify_piece_inclusion_proof_response(resPtr)
+
+	if resPtr.status_code != 0 {
+		return false, errors.New(C.GoString(resPtr.error_msg))
+	}
+
+	return bool(resPtr.is_valid), nil
+}
+
+// GeneratePieceCommitment TODO
```
sidke: describe this method

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -1,5 +1,6 @@
 package sectorbuilder
 
+import "C"
```
sidke: delete

proofs/verification/interface.go
```diff
@@ -35,8 +36,33 @@ type VerifySealResponse struct {
 	IsValid bool
 }
 
+// @TODO
```
sidke: describe these

proofs/verification/interface.go
```diff
@@ -35,8 +36,33 @@ type VerifySealResponse struct {
...
+type VerifyPieceInclusionProofRequest struct {
+	CommD               types.CommD
+	CommP               types.CommP
+	PieceInclusionProof []byte
+	PieceSize           *types.BytesAmount
+	SectorSize          *types.BytesAmount
+}
+
+// @TODO
```
sidke: describe these

proofs/verification/interface.go
```diff
@@ -35,8 +36,33 @@ type VerifySealResponse struct {
...
+type VerifyPieceInclusionProofResponse struct {
+	IsValid bool
+}
+
+type GeneratePieceCommitmentRequest struct {
```
sidke: describe these

proofs/verification/interface.go
```diff
@@ -35,8 +36,33 @@ type VerifySealResponse struct {
...
+	PieceReader io.Reader
+}
+
+type GeneratePieceCommitmentResponse struct {
```
sidke: describe these

proofs/verification/rustverifier.go
```diff
@@ -1,6 +1,12 @@
 package verification
 
-import "github.com/filecoin-project/go-filecoin/proofs/libsectorbuilder"
+import "C"
```
sidke: delete

proofs/verification/rustverifier.go
```diff
@@ -43,3 +49,36 @@ func (rp *RustVerifier) VerifyPoSt(req VerifyPoStRequest) (VerifyPoStResponse, e
 		IsValid: isValid,
 	}, nil
 }
+
+// VerifyPieceInclusionProof TODO
```
sidke: describe these

proofs/verification/rustverifier.go
```diff
@@ -43,3 +49,36 @@ func (rp *RustVerifier) VerifyPoSt(req VerifyPoStRequest) (VerifyPoStResponse, e
...
+	if err != nil {
+		return VerifyPieceInclusionProofResponse{}, err
+	}
+
+	return VerifyPieceInclusionProofResponse{
+		IsValid: isValid,
+	}, nil
+}
+
+// GeneratePieceCommitment TODO
```
sidke: describe these

protocol/storage/client.go
```diff
@@ -179,10 +192,12 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		Duration:        duration,
 		MinerAddress:    miner,
 		CommP:           commP,
+		PieceSize:       &pieceSize,
 		PaymentInterval: VoucherInterval,
 		ChannelExpiry:   *chainHeight.Add(types.NewBlockHeight(duration + ChannelExpiryInterval)),
 		GasPrice:        types.NewAttoFIL(big.NewInt(CreateChannelGasPrice)),
 		GasLimit:        types.NewGasUnits(CreateChannelGasLimit),
+		// todo here
```
sidke: remove this

protocol/storage/client_test.go
```diff
@@ -102,7 +101,7 @@ func TestProposeDeal(t *testing.T) {
 			assert.Equal(t, testAPI.perPayment.MulBigInt(big.NewInt(int64(i+1))), voucher.Amount)
 			assert.Equal(t, testAPI.payer, voucher.Payer)
 			assert.Equal(t, minerAddr, voucher.Condition.To)
-			assert.Equal(t, commP[:], voucher.Condition.Params[0])
+			// assert.Equal(t, commP[:], voucher.Condition.Params[0])
```
sidke: remove this

actor/builtin/miner/miner.go
```diff
@@ -630,13 +632,21 @@ func (ma *Actor) VerifyPieceInclusion(ctx exec.VMContext, commP []byte, sectorID
...
+			CommD:               commitment.CommD,
+			CommP:               typedCommP,
+			PieceInclusionProof: proof,
+			PieceSize:           pieceSize,
+			SectorSize:          state.SectorSize,
 		}
 
-		if !valid {
-			return nil, errors.NewRevertError("invalid inclusion proof")
+		res, err := (&verification.RustVerifier{}).VerifyPieceInclusionProof(req)
```
anorth: FYI in parallel work @acruikshank has been working to abstract the rust verifier out of the actor code, instead exposing methods on the VMContext.

actor/builtin/miner/miner_test.go
anorth: @acruikshank is also working on a fake VMContext precisely so we can test things like this more easily. Consider a new miner_test2.go file that uses that fake setup for this purpose (and eventually we might port all the tests to it).

actor/builtin/miner_redeem_test.go
laser: Added [follow up/TODO](https://github.com/filecoin-project/go-filecoin/issues/3086).

proofs/libsectorbuilder/cgo_transforms.go
```diff
@@ -81,9 +81,14 @@ func goPieceMetadata(src *C.sector_builder_ffi_FFIPieceMetadata, size C.size_t)
 
 	ptrs := (*[1 << 30]C.sector_builder_ffi_FFIPieceMetadata)(unsafe.Pointer(src))[:size:size]
```
ZenGround0: nit: not this PR but this magic number could use some explaining for the uninitiated.

proofs/sectorbuilder/rustsectorbuilder.go
```diff
@@ -191,12 +191,6 @@ func (sb *RustSectorBuilder) findSealedSectorMetadata(sectorID uint64) (*SealedS
 			}
```
ZenGround0: Blocking: I don't have any background in this code so the existing way might be fine.  However the comment on 181 seems to indicate that this whole loop should go away with this change.  As real PIPs are now available.  It seems like this comment should be changed or removed if the loop should still exist even with real PIPs.

plumbing/api.go
```diff
@@ -339,7 +338,7 @@ func (api *API) DAGGetFileSize(ctx context.Context, c cid.Cid) (uint64, error) {
 
 // DAGCat returns an iostream with a piece of data stored on the merkeldag with
 // the given cid.
-func (api *API) DAGCat(ctx context.Context, c cid.Cid) (uio.DagReader, error) {
+func (api *API) DAGCat(ctx context.Context, c cid.Cid) (io.Reader, error) {
```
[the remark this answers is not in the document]
laser: @ZenGround0 - The loosening of this type from `uio.DagReader` to `io.Reader` means that we can dependency inject a fake (an `io.Reader` satisfied by some byte buffer) for testing. In non-test, we continue to use the `uio.DagReader`.

proofs/libsectorbuilder/cgo_bindings.go
```diff
@@ -49,6 +49,7 @@ type PieceMetadata struct {
 	Key            string
 	Size           uint64
 	InclusionProof []byte
+	CommP          [32]byte
```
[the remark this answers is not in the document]
laser: @ZenGround0 (cc @sidke) If you do introduce a constant, please leave the constant in `cgo_bindings.go`. Do not introduce a dependency on go-filecoin types and packages, as `cgo_bindings.go` and `cgo_transforms.go` are being extracted into a separate Git repository so that why's team can use them.

proofs/libsectorbuilder/cgo_bindings.go
```diff
@@ -403,3 +404,52 @@ func GeneratePoSt(
...
+	commDCBytes := C.CBytes(commD[:])
+	defer C.free(commDCBytes)
+
+	commPCBytes := C.CBytes(commP[:])
+	defer C.free(commPCBytes)
+
+	pieceInclusionProofCBytes := C.CBytes(proof)
+	defer C.free(pieceInclusionProofCBytes)
+
+	resPtr := (*C.sector_builder_ffi_VerifyPieceInclusionProofResponse)(unsafe.Pointer(C.sector_builder_ffi_verify_piece_inclusion_proof(
```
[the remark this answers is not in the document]
laser: @ZenGround0 - All of these C symbols are defined in a single C archive file, which is called `libsector_builder_ffi.a`, and are automatically generated as part of the `sector-builder-ffi` Rust crate, which re-exports stuff from the `rust-fil-proofs-ffi` crate. 

TLDR: The prefix is not a thing that we have the ability to change. cc @sidke

proofs/sectorbuilder/rustsectorbuilder.go
laser: @ZenGround0 - Good catch. We have:

1. removed the old comment
1. updated the software to correctly marshall PIP and CommP bytes from FFI to Go types
1. added a test (at the SectorBuilder level) which round-trips a CommP and PIP through the verification function

proofs/sectorbuilder/testing/interface_test.go
```diff
@@ -357,4 +357,38 @@ func TestSectorBuilder(t *testing.T) {
...
+		require.NoError(t, err)
+
+		timeout := time.After(MaxTimeToSealASector)
+
+		select {
+		case val := <-h.SectorBuilder.SectorSealResults():
+			require.NoError(t, val.SealingErr)
+			require.NotNil(t, val.SealingResult)
+			require.Equal(t, sectorID, val.SealingResult.SectorID)
+			require.Equal(t, 1, len(val.SealingResult.Pieces), "too many pieces")
```
ZenGround0: nit: maybe the case where the number of pieces is 0 is impossible, but if not the error message would cause some confusion in that case.

protocol/storage/client.go
```diff
@@ -104,23 +107,32 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		minerAlive <- smc.api.PingMinerWithTimeout(ctxSetup, pid, 15*time.Second)
 	}()
 
-	size, err := smc.api.DAGGetFileSize(ctxSetup, data)
+	pieceSize, err := smc.api.DAGGetFileSize(ctxSetup, data)
```
ZenGround0: lovely rename ❤️

proofs/libsectorbuilder/cgo_bindings.go
```diff
@@ -414,3 +415,52 @@ func GeneratePoSt(
 
 	return proofs, goUint64s(resPtr.faults_ptr, resPtr.faults_len), nil
 }
+
+// VerifyPieceInclusionProof returns true if the piece inclusion proof is valid
+// with the given arguments.
+func VerifyPieceInclusionProof(sectorSize uint64, pieceSize uint64, commP [32]byte, commD [32]byte, proof []byte) (bool, error) {
```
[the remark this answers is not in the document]
laser: We are a bit inconsistent here, aren't we? We also don't `VerifyPoRep`, but rather `VerifySeal`. Hrm.

proofs/libsectorbuilder/cgo_bindings.go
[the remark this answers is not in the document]
ZenGround0: Is renaming the sector-builder-ffi rust crate to something more reflective of its functionality something you could do?  I'm not saying it is important at all right now but want to make sure we're on the same page that it is tech debt that we'd prefer to correct.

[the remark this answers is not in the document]
laser: I think we may be speaking past each other here. I'll unpack things further...

`go-filecoin` links a single C archive file called `libsector_builder_ffi.a` with a header file `libsector_builder_ffi.h`. This archive/headerfile pair contains all of the sector builder and sector builder-nonspecific functions. The `sector_builder_ffi` prefix is a C convention; symbols are global in C, so libraries prefix their function names with the library name in order to avoid collision with other libraries. The source library in this case is `sector-builder-ffi`.

If we split the sector builder-nonspecific stuff into a separate archive file, go-filecoin (and any other non-Rust node implementation) would need to link in two libraries (each containing the Rust stdlib and other dependencies), each with their own headerfile. We'd need to modify go-filecoin to include a new submodule dependency on `rust-fil-proofs-ffi` (for the sector builder-nonspecific stuff), and we'd need to sure that it never used a version of that submodule which was incompatible with the `rust-fil-proofs-sector-builder` submodule it was pointed at. 

In all, combining the two sets of functions into a single C library keeps that archive file as small as possible. It reduces the complexity of linking the sector builder in to a node (there are no sector builder consumers who don't also want to verify PoSt and seal proofs) to a minimum.

Does that make sense?

actor/builtin/miner/miner_test.go
```diff
@@ -1178,107 +1187,145 @@ func assertSlashStatus(t *testing.T, st state.Tree, vms vm.StorageMap, minerAddr
...
-
-	origPid := th.RequireRandomPeerID(t)
-	minerAddr := th.CreateTestMiner(t, st, vms, address.TestAddress, origPid)
-
-	sectorId := uint64(1)
-	commD := th.MakeCommitment()
+	sectorSetWithOneCommitment, commDForSectorIdOne := func() (SectorSet, types.CommD) {
+		set := NewSectorSet()
+		comms := th.MakeCommitments()
+		set.Add(1, comms)
```
ZenGround0: nit: to make this a bit clearer I would give `1` a variable name so you can pass that into `VerifyPieceInclusion` and lead readers back to the setup.

actor/builtin/miner/miner_test.go
```diff
@@ -1178,107 +1187,145 @@ func assertSlashStatus(t *testing.T, st state.Tree, vms vm.StorageMap, minerAddr
...
+		require.Nil(t, verifier.LastReceivedVerifyPieceInclusionProofRequest)
+	})
 
-			assert.NoError(t, err)
-		})
+	t.Run("PIP is invalid if miner's PoSt is too far into the past", func(t *testing.T) {
+		vmctx, verifier, minerActor := newBuilder().
+			withMessage("verifyPieceInclusionProof").
+			withSectorSet(sectorSetWithOneCommitment).
+			withLastPoSt(types.NewBlockHeight(0).Sub(types.NewBlockHeight(PieceInclusionGracePeriodBlocks * 2))).
```
ZenGround0: This negative block height makes me nervous.  The miner env builder should have a `withHeight` option function that sets the height and then you can set the vmctx up with a big enough block height to do this subtraction.

commands/client_daemon_test.go
```diff
@@ -230,7 +231,12 @@ func TestPieceRejectionInProposeStorageDeal(t *testing.T) {
 func TestSelfDialStorageGoodError(t *testing.T) {
 	tf.IntegrationTest(t)
 
-	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fast.FilecoinOpts{})
+	// set block time sufficiently high that client can import its piece
```
ZenGround0: Do I understand correctly that this was introduced to reduce the chances of getting bitten by a race?  If so please document that a race (that we usually win) still exists.  Also consider referencing or creating an issue to remove this flaky behavior from daemon tests/this test.

proofs/libsectorbuilder/cgo_bindings.go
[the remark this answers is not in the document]
ZenGround0: I think understand what you are saying.  Let me summarize to make sure I understand.  What I want is for all of the places that say `sector_builder_ffi_(PoSt|PIP|verifySeal)` to instead say `proofs_ffi_(PoSt|PIP|verifySeal)`.  We can't do that without (1) linking two libraries and duplicating standard libs, or (2) renaming the rust sector builder source library `proofs-ffi`.  Neither of these options is better than just keeping the imperfectly scoped but functional `sector_builder_ffi` prefix, so there is no plan to change the prefix.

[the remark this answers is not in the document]
laser: Yessiree! You got it.

actor/builtin/miner/miner_test.go
```diff
@@ -1447,3 +1494,53 @@ func mustGetMinerState(st state.Tree, vms vm.StorageMap, a address.Address) *Sta
...
+}
+
+func (b *minerEnvBuilder) build() (exec.VMContext, *verification.FakeVerifier, *Actor) {
+	minerState := NewState(address.TestAddress, address.TestAddress, peer.ID(""), b.sectorSize)
+	minerState.SectorCommitments = b.sectorSet
+	minerState.LastPoSt = b.lastPoSt
+
+	vmctx := th.NewFakeVMContextWithVerifier(types.NewMessage(address.TestAddress, address.TestAddress2, 0, types.ZeroAttoFIL, b.message, nil), minerState, b.verifier)
+
+	return vmctx, b.verifier, &Actor{}
```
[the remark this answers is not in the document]
acruikshank: Actually, I can see how a variable, minerActor, is slightly cleaner than (&Actor). I rescend this comment.

actor/builtin/miner/miner_test.go
```diff
@@ -768,7 +771,10 @@ func TestMinerSubmitPoStVerification(t *testing.T) {
 		minerState.SectorCommitments.Add(1, types.Commitments{CommR: comm1.CommR})
 
 		minerState.ProvingSet = types.NewIntSet(1)
-		verifier := &th.FakeVerifier{Err: errors.New("verifier error")}
+		verifier := verification.NewFakeVerifier(verification.FakeVerifierConfig{
```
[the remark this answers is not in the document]
laser: I didn't want to export `FakeVerifier`'s; I wanted it to be immutable after construction. That said, I don't not-want to export those fields enough to push back. Your wish is my command!

node/testing.go
```diff
@@ -191,7 +191,11 @@ func MakeOfflineNode(t *testing.T) *Node {
 // DefaultTestingConfig returns default configuration for testing
 func DefaultTestingConfig() []ConfigOpt {
 	return []ConfigOpt{
-		VerifierConfigOption(verification.NewFakeVerifier(true, nil)),
+		VerifierConfigOption(verification.NewFakeVerifier(verification.FakeVerifierConfig{
```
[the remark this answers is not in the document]
laser: > Is this necessary?

It may not be necessary, but it's also not something we introduced. Note the line which we replaced:

```
VerifierConfigOption(verification.NewFakeVerifier(true, nil)), 
```

In order to limit the chance of regressing the application, we simply retrofitted the old config to conform to the new API.

proofs/piece_commitment.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// GeneratePieceCommitmentResponse represents the commitment bytes
+type GeneratePieceCommitmentResponse struct {
+	CommP types.CommP
+}
+
+// GeneratePieceCommitment produces a piece commitment for the provided data
+// stored at a given piece path.
+func GeneratePieceCommitment(req GeneratePieceCommitmentRequest) (res GeneratePieceCommitmentResponse, retErr error) {
```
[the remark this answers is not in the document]
laser: > Also, I know this gets covered by storage.Client tests, but it's a pretty significant chunk of code to have no direct testing.

Cool. We now exercise this code as part of `integration_test.go`. Just pushed.

protocol/storage/client.go
```diff
@@ -56,6 +58,7 @@ type clientPorcelainAPI interface {
 	CreatePayments(ctx context.Context, config porcelain.CreatePaymentsParams) (*porcelain.CreatePaymentsReturn, error)
 	DealGet(context.Context, cid.Cid) (*storagedeal.Deal, error)
 	DAGGetFileSize(context.Context, cid.Cid) (uint64, error)
+	DAGCat(context.Context, cid.Cid) (io.Reader, error)
```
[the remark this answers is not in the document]
laser: Why does `DAGGetFileSize` belong in `clientPorcelainAPI` but `DAGCat` does not? I still don't have my "Should this go into porcelain?" heuristic dialed to the same setting as yours.

actor/builtin/miner/miner_test.go
laser: Done! Your wish is my command (to Sidney).

protocol/storage/client.go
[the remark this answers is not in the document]
acruikshank: Actually DAGGetFileSize should probably move into the porcelain call as well since it's used by the piece commitment logic. The point is that the process of putting together all the appropriate low level calls to determine the piece commitment of an imported piece is not something storage.Client should need to deal with. The whole point of porcelain is piecing together low-level calls to create a more humane interface.

I don't think Client should deal with DAG calls at all. I don't think the Reader -> tmp.File -> proofs.GeneratePieceCommitment belongs in proofs. This whole chain should be porcelain consumed by Client.

That would be better, but not better enough to be worth blocking this PR.

commands/client_daemon_test.go
laser: A storage client will wait a duration equal to [5 * `$BLOCKTIME`](https://github.com/filecoin-project/go-filecoin/blob/master/protocol/storage/client.go#L93) to complete the storage deal setup-stuff. This includes creating a payment, pinging the storage miner, etc.

If the environment is created without setting a block time explicitly, `$BLOCKTIME` [is equal to one millisecond](https://github.com/filecoin-project/go-filecoin/blob/5fbda21520cd282dbba60f0a9ddcc440e737e4f6/tools/fast/fastesting/basic.go#L66).

I don't think I'm masking over a race condition here. Rather that the default is too low for a storage client and miner to consummate a storage deal. Agree? Disagree?

proofs/verification/testing.go
```diff
@@ -1,23 +1,61 @@
 package verification
 
-// FakeVerifier is a simple mock Verifier for testing
+// FakeVerifierConfig configures the behavior of the FakeVerifier.
+type FakeVerifierConfig struct {
```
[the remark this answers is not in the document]
laser: I have terminated it with extreme prejudice as per your command.

commands/client_daemon_test.go
ZenGround0: thx for explaining!

laser: no prob bob

PR #3067, opened 2019-07-15
Use proving set instead of sector commitments for PoSt validation

### LINKED ISSUE #3003: Only use commitments from current proving set as arguments to PoSt verify
### Description
#2999 introduced the proving set but still computes PoSt verification with all sectors in the sectorset.  Instead only the sectorIDs listed in the proving set should be passed to the verification function.

### Acceptance criteria
Only sectors referenced by proving set go to PoSt verify.

### Where to begin
It might make sense to update the `SectorSet` abstraction introduced in #2999 to handle filtered fetching of commitments.

acruikshank:
closes #3003

### Problem

PoSt verification should only cover sectors in the current proving set. We are currently proving over all sectors. This includes sectors that have been added during the current proving set, which should be excluded.

### Solution

Only include sectors in the proving set in verification. This involves some rather minor changes to `SubmitPoSt` and `GetSectorCommitments` (now `GetProvingSetCommitments`). The problem is the part of the code that changed in `SubmitPoSt` was not under test. This prompted a much larger refactor to make all of miner actor testable. A consequence is that all actor code will now be more testable.

### What Changes in this PR

Major thrusts of work are fairly well encapsulated within 6 commits:

1. Commit 1: __Send only proving set CommRs to VerifyPoSt__: This completes the verification side of the issue, but at this stage, it is completely untested. Since the miner is coupled to rust proofs, testing this code will require a valid proof based off a real sealed sector. This is not something we want to do in a unit test, so the next step is to decouple rust proofs code from the actor.
2. Commit 2: __Introduce VMContext.Verifier abstraction, remove `ReadStorage` and `WriteStorage`:__ This is all in preparation for faking out the VMContext and it's verifier.
3. Commit 3: __Introduce FakeVMContext and use it to test SubmitPoSt verification:__ The FakeVMContext is designed to provide reasonable trade off between default functionality and flexibility. It allows us to fully mock out the state machine infrastructure and call actor methods directly for the first time. The `TestMinerSubmitPoStVerification` takes advantage of this to provide full coverage of all the code previously hidden in the non-bootstrap miner block.
4. Commit 4: __Convert getSectorCommitments to getProvingSetCommitments:__ This satisfies the proving side of this issue. The miner's prover calls this method to get all commitments it needs to prove on. This commit changes the miner method to only return commitments in the proving set and renames it to signal this functionality.
5. Commit 5: __Test `getProvingSetCommitments`:__ The new test takes advantage of the new FakeVMContext test infrastructure to provide full coverage of the query method.
6. Commit 6: __Refactor testVerfier into testhelpers:__ This is general purpose code (although probably only need by miner for now), and miner_test is way too long.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -910,7 +910,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 			for _, id := range state.ProvingSet.Values() {
 				commitment, found := state.SectorCommitments.Get(id)
 				if !found {
-					return nil, errors.NewFaultErrorf("miner ProvingSet contains sector id %d missing in SectorCommitments", id)
+					return nil, errors.NewFaultErrorf("miner ProvingSet sector id %d missing in SectorCommitments", id)
 				}
 				commRs = append(commRs, commitment.CommR)
 			}
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -612,7 +612,7 @@ func (ma *Actor) CommitSector(ctx exec.VMContext, sectorID uint64, commD, commR,
 			req.SectorID = sectorbuilder.SectorIDToBytes(sectorID)
 			req.SectorSize = state.SectorSize
 
-			res, err := (&verification.RustVerifier{}).VerifySeal(req)
+			res, err := ctx.Verifier().VerifySeal(req)
 			if err != nil {
 				return nil, errors.RevertErrorWrap(err, "failed to verify seal proof")
 			}
@@ -767,12 +767,8 @@ func (ma *Actor) GetPeerID(ctx exec.VMContext) (peer.ID, uint8, error) {
 
 	var state State
 
-	chunk, err := ctx.ReadStorage()
+	err := actor.ReadState(ctx, &state)
 	if err != nil {
-		return peer.ID(""), errors.CodeError(err), err
-	}
-
-	if err := actor.UnmarshalStorage(chunk, &state); err != nil {
 		return peer.ID(""), errors.CodeError(err), err
 	}
 
@@ -929,7 +925,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 				SectorSize:    state.SectorSize,
 			}
 
-			res, err := (&verification.RustVerifier{}).VerifyPoSt(req)
+			res, err := ctx.Verifier().VerifyPoSt(req)
 			if err != nil {
 				return nil, errors.RevertErrorWrap(err, "failed to verify PoSt")
 			}
@@ -1045,7 +1041,7 @@ func (ma *Actor) SlashStorageFault(ctx exec.VMContext) (uint8, error) {
 // GetProvingPeriod returns the proving period start and proving period end
 func (ma *Actor) GetProvingPeriod(ctx exec.VMContext) (*types.BlockHeight, *types.BlockHeight, uint8, error) {
 	var state State
-	err := loadState(ctx, &state)
+	err := actor.ReadState(ctx, &state)
 	if err != nil {
 		return nil, nil, errors.CodeError(err), err
 	}
@@ -1057,7 +1053,7 @@ func (ma *Actor) GetProvingPeriod(ctx exec.VMContext) (*types.BlockHeight, *type
 // power and proving period.
 func (ma *Actor) CalculateLateFee(ctx exec.VMContext, height *types.BlockHeight) (types.AttoFIL, uint8, error) {
 	var state State
-	err := loadState(ctx, &state)
+	err := actor.ReadState(ctx, &state)
 	if err != nil {
 		return types.ZeroAttoFIL, errors.CodeError(err), err
 	}
@@ -1166,19 +1162,6 @@ func latePoStFee(pledgeCollateral types.AttoFIL, provingPeriodEnd *types.BlockHe
 //
 // Internal functions
 //
-
-// Loads the actor's state into `state`.
-func loadState(ctx exec.VMContext, state *State) error {
-	chunk, err := ctx.ReadStorage()
-	if err != nil {
-		return err
-	}
-
-	if err := actor.UnmarshalStorage(chunk, &state); err != nil {
-		return err
-	}
-	return nil
-}
 
 // calculates proving period start from the proving period end and the proving period duration
 func provingPeriodStart(state State) *types.BlockHeight {
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -878,7 +878,7 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 
 		// The message value has been added to the actor's balance.
 		// Ensure this value fully covers the fee which will be charged to this balance so that the resulting
-		// balance (whichs forms pledge & storage collateral) is not less than it was before.
+		// balance (which forms pledge & storage collateral) is not less than it was before.
 		messageValue := ctx.Message().Value
 		if messageValue.LessThan(feeRequired) {
 			return nil, errors.NewRevertErrorf("PoSt message requires value of at least %s attofil to cover fees, got %s", feeRequired, messageValue)
@@ -911,8 +911,12 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProofs []types.PoStProof, do
 			}
 
 			var commRs []types.CommR
-			for _, v := range state.SectorCommitments {
-				commRs = append(commRs, v.CommR)
+			for _, id := range state.ProvingSet.Values() {
+				commitment, found := state.SectorCommitments.Get(id)
+				if !found {
+					return nil, errors.NewFaultErrorf("miner ProvingSet contains sector id %d missing in SectorCommitments", id)
+				}
+				commRs = append(commRs, commitment.CommR)
 			}
 
 			sortedCommRs := proofs.NewSortedCommRs(commRs...)
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -306,7 +306,7 @@ var minerExports = exec.Exports{
 		Params: nil,
 		Return: []abi.Type{abi.SectorID},
 	},
-	"getSectorCommitments": &exec.FunctionSignature{
+	"getProvingSetCommitments": &exec.FunctionSignature{
 		Params: nil,
 		Return: []abi.Type{abi.CommitmentsMap},
 	},
@@ -535,15 +535,23 @@ func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
 	return big.NewInt(result), 0, nil
 }
 
-// GetSectorCommitments returns all sector commitments posted by this miner.
-func (ma *Actor) GetSectorCommitments(ctx exec.VMContext) (map[string]types.Commitments, uint8, error) {
+// GetProvingSetCommitments returns all sector commitments posted by this miner.
+func (ma *Actor) GetProvingSetCommitments(ctx exec.VMContext) (map[string]types.Commitments, uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
 
 	var state State
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		return (map[string]types.Commitments)(state.SectorCommitments), nil
+		commitments := NewSectorSet()
+		for _, sectorID := range state.ProvingSet.Values() {
+			c, found := state.SectorCommitments.Get(sectorID)
+			if !found {
+				return nil, errors.NewFaultErrorf("proving set id, %d, missing in sector commitments", sectorID)
+			}
+			commitments.Add(sectorID, c)
+		}
+		return (map[string]types.Commitments)(commitments), nil
 	})
 	if err != nil {
 		return map[string]types.Commitments{}, errors.CodeError(err), err
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -542,27 +542,20 @@ func (ma *Actor) GetProvingSetCommitments(ctx exec.VMContext) (map[string]types.
 	}
 
 	var state State
-	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		commitments := NewSectorSet()
-		for _, sectorID := range state.ProvingSet.Values() {
-			c, found := state.SectorCommitments.Get(sectorID)
-			if !found {
-				return nil, errors.NewFaultErrorf("proving set id, %d, missing in sector commitments", sectorID)
-			}
-			commitments.Add(sectorID, c)
-		}
-		return (map[string]types.Commitments)(commitments), nil
-	})
+	err := actor.ReadState(ctx, &state)
 	if err != nil {
 		return map[string]types.Commitments{}, errors.CodeError(err), err
 	}
 
-	a, ok := out.(map[string]types.Commitments)
-	if !ok {
-		return map[string]types.Commitments{}, 1, errors.NewFaultErrorf("expected a map[string]types.Commitments, but got %T instead", out)
+	commitments := NewSectorSet()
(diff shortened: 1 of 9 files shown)


acruikshank:
cc @laser @sidke @ingar for changes to actor testing of proof verification.

actor/storage.go
```diff
@@ -38,27 +38,49 @@ func UnmarshalStorage(raw []byte, to interface{}) error {
...
+func ReadState(ctx exec.VMContext, st interface{}) error {
+	storage := ctx.Storage()
+
+	memory, err := storage.Get(storage.Head())
+	if err != nil {
+		return vmerrors.FaultErrorWrap(err, "Could not read actor storage")
+	}
+
+	chunk := make([]byte, len(memory))
+	copy(chunk, memory)
```
ZenGround0: Could you explain why we need to copy here?  Looking at the `storage.Get` method's implementation it is not clear what the benefit is.

actor/builtin/miner/miner_test.go
```diff
@@ -696,6 +699,116 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
+	t.Run("Reverts if verification errors", func(t *testing.T) {
+		minerState := *NewState(address.TestAddress, address.TestAddress, peer.ID(""), types.OneKiBSectorSize)
+		minerState.ProvingPeriodEnd = types.NewBlockHeight(ProvingPeriodDuration(types.OneKiBSectorSize))
+		minerState.SectorCommitments = NewSectorSet()
+		minerState.SectorCommitments.Add(1, types.Commitments{CommR: commR1})
+
+		minerState.ProvingSet = types.NewIntSet(1)
+		vmctx := th.NewFakeVMContext(message, minerState)
+
+		verifier := &th.FakeVerifier{Err: xerrors.New("verifier error")}
```
shannonwells: This looks unintentional

actor/builtin/miner/miner_test.go
```diff
@@ -1133,6 +1246,56 @@ func TestMinerGetPoStState(t *testing.T) {
...
+		commitments, code, err := miner.GetProvingSetCommitments(vmctx)
+		require.NoError(t, err)
+		require.Equal(t, uint8(0), code)
+
+		assert.Equal(t, 2, len(commitments))
+		assert.Equal(t, commR1, commitments["1"].CommR)
+		assert.Equal(t, commR2, commitments["2"].CommR)
+	})
+
+	t.Run("faults if proving set commitment not founc in sector commitments", func(t *testing.T) {
```
shannonwells: fyi typo: "founc"

testhelpers/core.go
```diff
@@ -184,3 +187,150 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+		TestRandomness:  randomness,
+		Charger: func(cost types.GasUnits) error {
+			return nil
+		},
+		Sampler: func(sampleHeight *types.BlockHeight) ([]byte, error) {
+			return randomness, nil
+		},
+		Sender: func(to address.Address, method string, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error) {
+			return [][]byte{}, 0, nil
+		},
```
ZenGround0: What is the reason to exclude the addressor, actorTyper, actorCreater and verifier from the default setup?  Is it ok that methods on the output of `NewFakeVMContext` panic when called?

actor/builtin/miner/miner_test.go
```diff
@@ -1133,6 +1246,56 @@ func TestMinerGetPoStState(t *testing.T) {
...
 }
 
+func TestGetProvingSetCommitments(t *testing.T) {
+	tf.UnitTest(t)
+
+	message := types.NewMessage(address.TestAddress, address.TestAddress2, 0, types.ZeroAttoFIL, "getProvingSetCommitments", nil)
+	var commR1, commR2, commR3 types.CommR
+	copy(commR1[:], th.MakeRandomBytes(32))
+	copy(commR2[:], th.MakeRandomBytes(32))
+	copy(commR3[:], th.MakeRandomBytes(32))
```
ZenGround0: nit: you could simplify these lines and the `types.Commitments{}` wrap below by calling `th.MakeCommitments()`

actor/builtin/miner/miner_test.go
acruikshank: It wasn't. It's a pattern that's been in place when "errors" and "github.com/filecoin-project/go-filecoin/vm/errors" are both in use. I can swap the namespaces to "errors" and "vmerrors" though.

actor/builtin/miner/miner_test.go
```diff
@@ -696,6 +699,116 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
+		minerState.ProvingSet = types.NewIntSet(1, 2)
+		vmctx := th.NewFakeVMContext(message, minerState)
+
+		verifier := &th.FakeVerifier{Valid: true}
+		vmctx.TestVerifier = verifier
+
+		miner := Actor{Bootstrap: false}
+
+		testProof := []types.PoStProof{th.MakeRandomPoStProofForTest()}
+		_, err := miner.SubmitPoSt(vmctx, testProof, types.NewIntSet())
```
ZenGround0: nit: I've been using `types.EmptyIntSet()` to try to be more idiomatic.  In retrospect it is redundant to this call which is also pretty clear but it would be good to be consistent.  If you want to keep it this way and drop an issue for removing `EmptyIntSet` feel free.

actor/builtin/miner/miner_test.go
acruikshank: good call

actor/builtin/miner/miner_test.go
acruikshank: thank you

actor/storage.go
acruikshank: I agree we don't. This is copied from really old code. I removed it.

testhelpers/core.go
acruikshank: pure, unadulterated laziness

actor/builtin/miner/miner_test.go
acruikshank: I'm unclear on the benefits of EmptyIntSet, but I've changed it to keep consistent.

actor/builtin/miner/miner.go
```diff
@@ -516,15 +516,23 @@ func (ma *Actor) GetPoStState(ctx exec.VMContext) (*big.Int, uint8, error) {
...
+// GetProvingSetCommitments returns all sector commitments posted by this miner.
+func (ma *Actor) GetProvingSetCommitments(ctx exec.VMContext) (map[string]types.Commitments, uint8, error) {
 	if err := ctx.Charge(actor.DefaultGasCost); err != nil {
 		return nil, exec.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
 	}
 
 	var state State
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		return (map[string]types.Commitments)(state.SectorCommitments), nil
+		commitments := NewSectorSet()
```
anorth: FYI consider `ReadState()` for more concise read-only methods (saves you some type assertions later too)

actor/builtin/miner/miner_test.go
```diff
@@ -696,6 +699,115 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
+		minerState.SectorCommitments.Add(1, types.Commitments{CommR: comm1.CommR})
+		minerState.SectorCommitments.Add(2, types.Commitments{CommR: comm2.CommR})
+		minerState.SectorCommitments.Add(3, types.Commitments{CommR: comm3.CommR})
+
+		// The 3 sector is not in the proving set, so its CommR should not appear in the VerifyPoSt request
+		minerState.ProvingSet = types.NewIntSet(1, 2)
+		vmctx := th.NewFakeVMContext(message, minerState)
+
+		verifier := &th.FakeVerifier{Valid: true}
+		vmctx.TestVerifier = verifier
```
anorth: Nit: inline into `NewFakeVMContextWithVerifier`?

actor/builtin/miner/miner_test.go
```diff
@@ -696,6 +699,115 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
+
+		miner := Actor{Bootstrap: false}
+
+		testProof := []types.PoStProof{th.MakeRandomPoStProofForTest()}
+		code, err := miner.SubmitPoSt(vmctx, testProof, types.NewIntSet())
+		require.Error(t, err)
+		assert.Equal(t, "miner ProvingSet sector id 4 missing in SectorCommitments", err.Error())
+		assert.True(t, vmerrors.IsFault(err))
+		assert.Equal(t, uint8(1), code)
+	})
```
anorth: Just confirming my understanding: this is testing an impossible case if the actor's abstraction boundary is respected, right. This should be an invariant maintained by the actor methods and only testable because we are reaching in to modify its state, and not reproducible as a "black box" test.

I'm not sure in general that we should attempt to test things like this exhaustively, but we *should* attempt all possible ways to get the miner state to be inconsistent by manipulating its exposed methods.

(Somewhere in the VM we should be doing a `recover()` to prevent a broken actor from taking down the vm/node entirely).

exec/exec.go
```diff
@@ -79,20 +80,23 @@ type VMContext interface {
 
 	CreateNewActor(addr address.Address, code cid.Cid, initalizationParams interface{}) error
 
-	// TODO: Remove these when Storage above is completely implemented
-	ReadStorage() ([]byte, error)
-	WriteStorage(interface{}) error
+	Verifier() Verifier
```
anorth: I'm on the fence about whether we should flatten the VMContext API entirely or not. In some ways a flat API is easier to fake/mock/satisfy

exec/exec.go
```diff
@@ -79,20 +80,23 @@ type VMContext interface {
...
 	Put(interface{}) (cid.Cid, error)
 	Get(cid.Cid) ([]byte, error)
 	Commit(cid.Cid, cid.Cid) error
 	Head() cid.Cid
 }
 
+// Verifier is the proof verification interface
+type Verifier interface {
+	VerifySeal(req verification.VerifySealRequest) (verification.VerifySealResponse, error)
+	VerifyPoSt(req verification.VerifyPoStRequest) (verification.VerifyPoStResponse, error)
```
anorth: Since we're abstracting it, I'd strongly consider removing the Request/Response structs in favour of simple parameters. It's not an RPC at this level.

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
 
 	return st, vms
 }
+
+type testStorage struct {
```
anorth: I would prefer that code like this live in the package alongside the things its faking, `vm/testing.go`. `testhelpers` is a dumping ground that does not make things easy to find and is not a scalable pattern.

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+
+}
+func (ts testStorage) Get(cid.Cid) ([]byte, error) {
+	node, err := cbor.WrapObject(ts.state, types.DefaultHashFunction, -1)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	return node.RawData(), nil
+}
```
anorth: Please document the intent of these fake methods. There's nontrivial code here so it  looks like `Get` is doing something, but `Put` isn't? How is the fake behaviour supposed to work, and is this interim or more or less complete?

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+	return nil
+}
+
+func (ts testStorage) Head() cid.Cid {
+	return cid.Cid{}
+}
+
+// FakeVMContext creates the scaffold for faking out the vm context for direct calls to actors
+type FakeVMContext struct {
+	TestMessage        *types.Message
```
anorth: `Test` prefix on these names isn't adding any information.

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+	Charger            func(cost types.GasUnits) error
+	Sampler            func(sampleHeight *types.BlockHeight) ([]byte, error)
+	ActorCreator       func(addr address.Address, code cid.Cid, initalizationParams interface{}) error
+}
+
+var _ exec.VMContext = &FakeVMContext{}
+
+// NewFakeVMContext fakes the state machine infrastructure so actor methods can be called directly
+func NewFakeVMContext(message *types.Message, state interface{}) *FakeVMContext {
+	randomness := MakeRandomBytes(32)
```
anorth: Please use a deterministic value, seeded from a parameter if necessary. Nondeterministic behaviour in tests makes debugging hard. Or maybe just leave this nil until a specific test calls for it.

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+	addressGetter := address.NewForTestGetter()
+	return &FakeVMContext{
+		TestMessage:        message,
+		TestStorage:        &testStorage{state: state},
+		TestBlockHeight:    types.NewBlockHeight(0),
+		TestBalance:        types.ZeroAttoFIL,
+		TestRandomness:     randomness,
+		TestIsAccountActor: true,
```
anorth: This one should be a parameter: it should be false for the actual tests you're using this in.

testhelpers/core.go
```diff
@@ -184,3 +187,158 @@ func RequireCreateStorages(ctx context.Context, t *testing.T) (state.Tree, vm.St
...
+// Verifier provides an interface to the proofs verifier
+func (tc *FakeVMContext) Verifier() exec.Verifier {
+	return tc.TestVerifier
+}
+
+// FakeVerifier is a fake implementation of a proof verifier
+type FakeVerifier struct {
+	Valid       bool
+	Err         error
+	SealRequest verification.VerifySealRequest
```
anorth: Please document that the first two determine behaviour, the latter two are captured from callers. Maybe call this `LastSealRequest` or `ReceivedSealRequest` or similar.

exec/exec.go
anorth: Oh, I see that this interface is intended to be satisfied by the existing RustVerifier. My comment probably applies at that level, but no need to do anything about it right now. Just FYI.

actor/builtin/miner/miner_test.go
acruikshank: This is correct. We are deliberately forcing a violation of the invariant and checking that the code correctly reports it as a fault. I agree that it's probably not necessary to test all these cases, but the ability to do this type of testing is new, so I was trying to get a feel for what 100% coverage looks like.

testhelpers/core.go
acruikshank: I'm trying to optimize for low configuration in the common case and enough flexibility for the general case. IsAccountActor is used only one place. If we use FakeVMContext to test that method, it's easy enough to change the parameter after construction. I'd rather cater the constructor to the common case.

actor/builtin/miner/miner_test.go
ZenGround0: Yeah I think the benefits are minimal and it's reasonable to remove it in a small PR for what you originally had.  I added it for a clear name.

actor/builtin/miner/miner.go
acruikshank: done.

testhelpers/core.go
acruikshank: The Prefix avoids a name collision with the accessor. I want to keep the relationship clear. I changed the "Test" prefix to a "Value" suffix, which at least hints at the relationship and the point of the suffix.

actor/builtin/miner/miner_test.go
```diff
@@ -696,6 +699,109 @@ func TestMinerSubmitPoStPowerUpdates(t *testing.T) {
...
 
+func TestMinerSubmitPoStVerification(t *testing.T) {
+	tf.UnitTest(t)
+
+	message := types.NewMessage(address.TestAddress, address.TestAddress2, 0, types.ZeroAttoFIL, "submitPoSt", nil)
+	comm1 := th.MakeCommitments()
+	comm2 := th.MakeCommitments()
+	comm3 := th.MakeCommitments()
+
+	t.Run("Sends correct parameters to post verifier", func(t *testing.T) {
```
laser: Nice.

Issue #3022, opened 2019-07-05
Remove implicit head state access in favour of explicit key


anorth:
### Description
A bunch of places in our code access parts of actor state without identifying which state snapshot is being accessed; the head state is implicit. This pattern is enabled and made widespread by plumbing and porcelain methods which implicitly access "latest" state: e.g. all the message query and preview stuff.

This pattern is unsafe if more than one piece of the state is to be expected and the bits expected to be consistent. The head tipset and state may change at any time, concurrent with state access, leading to inconsistent results. This is a broad undefined concurrency issue across the high-level protocol code.

This should be resolved by requiring an explicit tipset key (or possibly state root CID) be passed around and required by any code attempting to read state. Since the state identified by a tipset is effectively immutable, any concurrent change to the state store should not be visible to readers. An alternate pattern might be a "state view" which has the tipset/root CID curried in.

### Acceptance criteria
- All accessors of chain state must explicitly identify the tipset at which state is to be accessed
- Code patterns should make it difficult to do the wrong thing by accident

### Risks + pitfalls
There's lots of code that needs to change.

### Where to begin
Previewer and queryer in message plumbing.

ZenGround0:
A very related add on to this is to remove `GetHeight` from the chain store interface and instead get the height from the head tipset.  Block height is not technically state but the same inconsistency issues exist for this access.

PR #3076, opened 2019-07-17
Remove chain.Store's "BlockHeight" method

ZenGround0:
This moves the needle on issue #3022.  PR #3061 makes state accesses within plumbing require referencing state wrt a specific tipset key.  This PR enforces the same for block heights by removing `BlockHeight` from `chain.Store`.

It also un-swallows a swallowed error in the syncer's finality calculation.

@anorth I think #3022 is broader in scope than this PR + #3061 and is only complete when *callers* of plumbing reference a specific tipsetkey.  However I wasn't sure so let me know if I am wrong and the issue is addressed here and I'll connect it to this PR / close out on merge.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -70,7 +70,6 @@ const (
 )
 
 type syncerChainReaderWriter interface {
-	BlockHeight() (uint64, error)
 	GetHead() types.TipSetKey
 	GetTipSet(tsKey types.TipSetKey) (types.TipSet, error)
 	GetTipSetStateRoot(tsKey types.TipSetKey) (cid.Cid, error)
@@ -188,7 +187,14 @@ func (syncer *Syncer) collectChain(ctx context.Context, tsKey types.TipSetKey) (
 	// Continue collecting the chain if we're either not yet caught up or the
 	// number of new input blocks is less than the FinalityLimit constant.
 	// Otherwise, halt assuming the new blocks come from an invalid chain.
-	for (syncer.syncMode == Syncing) || !syncer.exceedsFinalityLimit(chain) {
+	for {
+		exceedsFinality, err := syncer.exceedsFinalityLimit(chain)
+		if err != nil {
+			return nil, err
+		}
+		if syncer.syncMode != Syncing && exceedsFinality {
+			break
+		}
 
 		// Finish traversal if the tipset made is tracked in the store.
 		if syncer.chainStore.HasTipSetAndState(ctx, tsKey.String()) {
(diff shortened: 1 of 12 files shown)


anorth:
Looks good so far, and I agree with your judgement.

ZenGround0:
This PR is now ready for review.  In addition to the description above this PR has also improved the state of core inbox tests as a side effect of getting those tests working with the reduced chain store interface.  Message carrying chains are now constructed by using a chain.Builder.

The deprecated `NewChainWithMessages` still exists because the `plumbing/msg/waiter_test.go` tests depend on it and I didn't want to further expand the scope of this PR.  Follow up issue #3110 has been added to our chain sync epic and tracks the work needed to use the new builder-based message carrying chain constructor in waiter tests and delete the deprecated function.

core/inbox_test.go
```diff
@@ -5,12 +5,9 @@ import (
 	"fmt"
```
frrist: This looks like it goes beyond the scope of removing `BlockHeight()`, in the future please do this in a separate commit at a minimum -- separate PR is fine too.

plumbing/msg/previewer.go
```diff
@@ -48,9 +48,13 @@ func (p *Previewer) Preview(ctx context.Context, optFrom, to address.Address, me
 	if err != nil {
 		return types.NewGasUnits(0), errors.Wrap(err, "could load tree for latest state root")
 	}
-	h, err := p.chainReader.BlockHeight()
+	head, err := p.chainReader.GetTipSet(p.chainReader.GetHead())
 	if err != nil {
-		return types.NewGasUnits(0), errors.Wrap(err, "couldnt get base tipset height")
+		return types.NewGasUnits(0), errors.Wrap(err, "couldnt get head tipset ")
```
frrist: nit: here and in `plumbing/msg/queryer.go`, lets's make these error message more consistent -- "failed to get head tipset"

chain/syncer.go
```diff
@@ -188,7 +187,14 @@ func (syncer *Syncer) collectChain(ctx context.Context, tsKey types.TipSetKey) (
 	// Continue collecting the chain if we're either not yet caught up or the
 	// number of new input blocks is less than the FinalityLimit constant.
 	// Otherwise, halt assuming the new blocks come from an invalid chain.
-	for (syncer.syncMode == Syncing) || !syncer.exceedsFinalityLimit(chain) {
+	for {
+		exceedsFinality, err := syncer.exceedsFinalityLimit(chain)
+		if err != nil {
+			return nil, err
+		}
+		if !(syncer.syncMode == Syncing) && exceedsFinality {
```
anorth: Simplify the first test to `syncMode != Syncing`

core/inbox_test.go
```diff
@@ -31,20 +28,18 @@ func TestUpdateMessagePool(t *testing.T) {
...
 		ib := core.NewInbox(p, 10, chainProvider)
 
 		m := types.NewSignedMsgs(2, mockSigner)
 		mustAdd(ib, m[0], m[1])
 
-		blk := types.Block{Height: 0}
-		parent := th.MustNewTipSet(&blk)
-
-		oldChain := core.NewChainWithMessages(store, parent, msgsSet{})
+		parent := builder.Genesis()
```
anorth: This creates a *new* genesis block, distinct from the one created inside newBuilderAndProvider. You probably want `chainProvider.GetHead()` instead, to avoid confusion. Alternatively, inline the newBuilderAndProvider method.

Consider renaming `Genesis` to `NewGenesis`.

core/testing.go
```diff
@@ -34,6 +37,40 @@ func MustConvertParams(params ...interface{}) []byte {
...
+	for _, tsMsgSet := range msgSets {
+		if len(tsMsgSet) == 0 {
+			parent = builder.BuildOn(parent, nil)
+		} else {
+			parent = builder.Build(parent, len(tsMsgSet), msgBuild(t, tsMsgSet))
+		}
+		tipSets = append(tipSets, parent)
+	}
+	return tipSets
+}
```
anorth: This is nice, I think you could add it to the chain builder directly for easy reuse.

`builder.AppendWithMessages(parent, msgSets...)`.

ZenGround0: Added to requirements for followup #3110

chain/syncer.go
ZenGround0: whoops thanks

PR #3081, opened 2019-07-17
Feat/chain bootstrap mode

frrist:
This PR is a sketch of adding a Syncing and Caught-Up sync mode to the syncer. There are no plans to merge this, just here to express ideas and show areas that may need work/refactoring. Review on general design would be appreciated.


chain/store.go
```diff
@@ -84,8 +83,13 @@ type Store struct {
 	tipIndex *TipIndex
 }
 
+type ipldStateThing interface {
```
frrist: I like the idea of this being an interface as it makes the test setup a bit easier, thoughts?

chain/evaluator.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// IsBadTipSet returns true if the tipset `ts` is known to be bad, true otherwise.
+func (cse *StateEvaluator) IsBadTipSet(ts types.TipSetKey) bool {
+	return cse.badTipSets.Has(ts.String())
+}
+
+// Evaluate is a wrapper around calls to consensus, it will either sync the chain `chain` to the store
+// or return an error
+func (cse *StateEvaluator) Evaluate(ctx context.Context, parent types.TipSet, chain []types.TipSet) error {
```
ZenGround0: `StateEvaluator` is a worse name than `Syncer` from my point of view.  Yes states are evaluated, but this is in order to *validate* tipsets and *sync* the chain store's internal state with the network.

chain/peer_manager.go
```diff
@@ -0,0 +1,106 @@
+package chain
```
ZenGround0: Nice, I'll use this code in the upcoming peer manager PR.

chain/store.go
```diff
@@ -250,6 +250,19 @@ func (store *Store) GetTipSetStateRoot(key types.TipSetKey) (cid.Cid, error) {
 	return store.tipIndex.GetTipSetStateRoot(key.String())
 }
 
+// FindCommonTipSet assumes `tips` are ordered based on height and will return the
+// first tipset the store and `tip` have in commmon in addition to the index.
+func (store *Store) FindCommonTipSet(ctx context.Context, tips []types.TipSet) (int, types.TipSet, bool) {
```
ZenGround0: Nice

frrist:
Closing this in favor of smaller PR's coming down the pipeline.

PR #3084, opened 2019-07-17
better error messaging for voucher decoding

### LINKED ISSUE #2909: [bug] Problems with paych redeem command
**Describe the bug**
Among other problems, `go-filecoin paych redeem` gives an error saying `Error: malformed stream: invalid appearance of uint token; expected start of map`

**To Reproduce**
1. Build and run localnet with `./localnet -blocktime=1s -miner-count=1`
1. Use ps to find out what go-filecoin processes are running so that you can see repo paths.
1. Run `go-filecoin wallet ls` to get a miner actor address for one of the repos.
1. From another terminal create a payment channel like:
```Shell
./go-filecoin paych create <minerActorAddr> 1.22 30 --gas-price=1 --gas-limit=1 --repodir=/var/folders/whatever/2/repo   
    zDPWYqFCrLQaTow26LBrpuGTp7VNgRahx4xiUXhn2tGEhSmvVyco
```
1. Ensure the miner node knows about the channel:
```
./go-filecoin paych ls --repodir=/var/folders/j0/k5m65hbx2dvby3y9bm6t2s940000gn/T/localnet152379070/0/repo
4: target: t12rcvqbmrxpj5pb33cq6ima7gusflwhjgyrsjiti, amt: 0.00001016, amt redeemed: 0, eol: 2014
```
1. Then
```
./go-filecoin paych redeem --gas-price=1 --gas-limit=1 zDPWYqFCrLQaTow26LBrpuGTp7VNgRahx4xiUXhn2tGEhSmvVyco --repodir=/var/folders/j0/k5m65hbx2dvby3y9bm6t2s940000gn/T/localnet152379070/0/repo

Error: malformed stream: invalid appearance of uint token; expected start of map
```

**Expected behavior**
* `Error: malformed stream: invalid appearance of uint token; expected start of map` appears to be the main bug, but also:
* "price option is required" is misleading. The option is "gas-price" not "price".
* Lastly, the CLI help conflicts with the error output. The CLI help shows that gas-price and gas-limit are optional (via square brackets).  Then if these options are missing, the error message refers to them as `price` and `limit` options, respectively.
```
./go-filecoin paych redeem -h --repodir=/var/folders/whatever/0/repo
USAGE
  ./go-filecoin paych redeem <voucher> - Redeem a payment voucher against a payment channel

  ./go-filecoin paych redeem [--from=<from>] [--gas-price=<gas-price>] [--gas-limit=<gas-limit>] [--preview] [--] <voucher>

Use './go-filecoin paych redeem --help' for more information about this command.
```

**Version information**
Occurs on master branch

acruikshank:
closes #2909

### Problem

Error messaging for redeeming a message voucher is bad. In the case of the bug, there was an attempt to redeem something other than a payment voucher, and the error mentions getting an int instead of a map in a stream. Prior to the decoding error, there was an attempt to call redeem without providing a gas-limit or gas-price, but the error only mentioned 'limit' and 'price'.

### Solution

Provide error message that simply suggest indicated that the payment voucher decoding has failed. Precisely why it has failed is going to be hard to recover, and precisely where in the decoding it failed is irrelevant. Also fix the error messages for required gas price and limit.

### DIFF
--- a/types/payment_voucher.go
+++ b/types/payment_voucher.go
@@ -1,6 +1,8 @@
 package types
 
 import (
+	"errors"
+
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/multiformats/go-multibase"
 	"sort"
@@ -54,13 +56,13 @@ type PaymentVoucher struct {
 func DecodeVoucher(voucherRaw string) (*PaymentVoucher, error) {
 	_, cborVoucher, err := multibase.Decode(voucherRaw)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("voucher could not be decoded")
 	}
 
 	var voucher PaymentVoucher
 	err = cbor.DecodeInto(cborVoucher, &voucher)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("voucher string could not be decoded into voucher")
 	}
 
 	return &voucher, nil
--- a/commands/main.go
+++ b/commands/main.go
@@ -334,7 +334,7 @@ var previewOption = cmdkit.BoolOption("preview", "Preview the Gas cost of this c
 func parseGasOptions(req *cmds.Request) (types.AttoFIL, types.GasUnits, bool, error) {
 	priceOption := req.Options["gas-price"]
 	if priceOption == nil {
-		return types.ZeroAttoFIL, types.NewGasUnits(0), false, errors.New("price option is required")
+		return types.ZeroAttoFIL, types.NewGasUnits(0), false, errors.New("gas-price option is required")
 	}
 
 	price, ok := types.NewAttoFILFromFILString(priceOption.(string))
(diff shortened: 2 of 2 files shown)


types/payment_voucher.go
```diff
@@ -54,13 +56,13 @@ type PaymentVoucher struct {
 func DecodeVoucher(voucherRaw string) (*PaymentVoucher, error) {
 	_, cborVoucher, err := multibase.Decode(voucherRaw)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("voucher could not be decoded")
```
shannonwells: error wrap to help with debugging, or just let the dev do it by hand?

acruikshank: I considered it. The info that comes out of the decoder are is unlikely to be very helpful for debugging and is likely to sow confusion, so I opted not to wrap.

Issue #1148, opened 2018-10-12
chain store in unexpected state


ZenGround0:
### Description
See the following error on my local node
`12:43:42.236 ERROR       node: processBlock(): processing block from network: parent tipset must be in store: the chain store is in an unexpected state blocks.go:56`

### Acceptance criteria
Determine how the node get in an unexpected state.

PR #3094, opened 2019-07-22
Port most syncer tests to use chain.Builder and non-shared setup

anorth:
Expand chain.Builder to keep track of computed state roots.

Sorry the diff here is a bit large. There is no change in syncer functionality or test coverage intended, though the tests themselves are now much shorter. The chain builder has more code in order to consistently compute state CIDs.

### DIFF
--- a/chain/syncer_integration_test.go
+++ b/chain/syncer_integration_test.go
@@ -1,0 +1,618 @@
+package chain_test
+
+import (
+	"context"
+	"testing"
+
+	bserv "github.com/ipfs/go-blockservice"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
+	bstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-ipfs-exchange-offline"
+	"github.com/libp2p/go-libp2p-peer"
+
+	"github.com/filecoin-project/go-filecoin/actor/builtin"
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/consensus"
+	"github.com/filecoin-project/go-filecoin/gengen/util"
+	"github.com/filecoin-project/go-filecoin/proofs/verification"
+	"github.com/filecoin-project/go-filecoin/repo"
+	"github.com/filecoin-project/go-filecoin/state"
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/types"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This file contains test that use a full syncer dependency structure, including store and
+// consensus implementations.
+// See syncer_test.go for most syncer unit tests.
+// The tests here should probably be reworked to follow the unit test patterns, and this
+// integration test setup reserved for a few "sunny day" tests of integration specifically.
+
+type SyncerTestParams struct {
+	// Chain diagram below.  Note that blocks in the same tipset are in parentheses.
+	//
+	// genesis -> (link1blk1, link1blk2) -> (link2blk1, link2blk2, link2blk3) -> link3blk1 -> (link4blk1, link4blk2)
+
+	// Blocks
+	genesis                         *types.Block
+	link1blk1, link1blk2            *types.Block
+	link2blk1, link2blk2, link2blk3 *types.Block
+	link3blk1                       *types.Block
+	link4blk1, link4blk2            *types.Block
+
+	// Cids
+	genCid                                                       cid.Cid
+	genStateRoot, link1State, link2State, link3State, link4State cid.Cid
+
+	// TipSets
+	genTS, link1, link2, link3, link4 types.TipSet
+
+	// utils
+	cidGetter         func() cid.Cid
+	minerAddress      address.Address
+	minerOwnerAddress address.Address
+	minerPeerID       peer.ID
+}
+
+func initDSTParams() *SyncerTestParams {
+	var err error
+	minerAddress, err := address.NewActorAddress([]byte("miner"))
+	if err != nil {
+		panic(err)
+	}
+	minerOwnerAddress, err := address.NewActorAddress([]byte("minerOwner"))
+	if err != nil {
+		panic(err)
+	}
+	minerPeerID, err := th.RandPeerID()
+	if err != nil {
+		panic(err)
+	}
+
+	// Set up the test chain
+	bs := bstore.NewBlockstore(repo.NewInMemoryRepo().Datastore())
+	cst := hamt.NewCborStore()
+	genesis, err := initGenesis(minerAddress, minerOwnerAddress, minerPeerID, cst, bs)
+	if err != nil {
+		panic(err)
+	}
+	genCid := genesis.Cid()
+	genTS := th.MustNewTipSet(genesis)
+
+	// mock state root cids
+	cidGetter := types.NewCidForTestGetter()
+
+	genStateRoot := genesis.StateRoot
+
+	return &SyncerTestParams{
+		minerAddress:      minerAddress,
+		minerOwnerAddress: minerOwnerAddress,
+		minerPeerID:       minerPeerID,
+		genesis:           genesis,
+		genCid:            genCid,
+		genTS:             genTS,
+		cidGetter:         cidGetter,
+		genStateRoot:      genStateRoot,
+	}
+}
+
+// This function sets global variables according to the tests needs.  The
+// test chain's basic structure is always the same, but some tests want
+// mocked stateRoots or parent weight calculations from different consensus protocols.
+func requireSetTestChain(t *testing.T, con consensus.Protocol, mockStateRoots bool, dstP *SyncerTestParams) {
+
+	var err error
+	// see powerTableForWidenTest
+	minerPower := types.NewBytesAmount(25)
+	totalPower := types.NewBytesAmount(100)
+	mockSigner, _ := types.NewMockSignersAndKeyInfo(1)
+	minerWorker := mockSigner.Addresses[0]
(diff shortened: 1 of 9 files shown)


chain/syncer_integration_test.go
```diff
@@ -0,0 +1,617 @@
...
+	for _, block := range blocks {
+		c := block.Cid()
+		cids = append(cids, c)
+	}
+	f.AddSourceBlocks(blocks...)
+	return types.NewTipSetKey(cids...)
+}
+
+// Syncer is capable of recovering from a fork reorg after Load.
+func TestLoadFork(t *testing.T) {
```
anorth: @ZenGround0 I don't think this is testing anything of the syncer, or at least I can't work out what it is really testing. Maybe it's testing `Store.Load`. I would like to delete it. After looking at the new syncer_test.go, can you either 👍 or describe how to set up the case to be exercised without the shared setup of this file?

chain/syncer_integration_test.go
```diff
@@ -0,0 +1,617 @@
...
+	forklink4 := th.RequireNewTipSet(t, forklink4blk1)
+	cidsFork4 := requirePutBlocks(t, blockSource, forklink4.ToSlice()...)
+	err = loadSyncer.HandleNewTipset(ctx, cidsFork4)
+	assert.NoError(t, err)
+}
+
+// Syncer handles MarketView weight comparisons.
+// Current issue: when creating miner mining with addr0, addr0's storage head isn't found in the blockstore
+// and I can't figure out why because we pass in the correct blockstore to createStorageMinerWithpower.
+func TestTipSetWeightDeep(t *testing.T) {
```
anorth: I have made no attempt to port this sole remaining test. Likely it can be done so, possibly with a custom `StateBuilder` given to the chain.Builder which defines a different `Weight` function. Left for a follow-up. Or maybe that's unnecessary. Or maybe it should remain an integration test (but we can still rewrite it to use chain.Builder).

chain/testing.go
```diff
@@ -172,28 +247,74 @@ func (bb *BlockBuilder) SetStateRoot(root cid.Cid) {
...
+	return parentWeight + uint64(tip.Len()), nil
+}
+
+///// State evaluator /////
 
-	return makeCid(fakeState)
+// Fake state evaluator that delegates to the FakeStateBuilder.
+type FakeStateEvaluator struct {
+	FakeStateBuilder
+}
```
anorth: This implements `syncStateEvaluator`, unexported from syncer.go.

chain/testing.go
```diff
@@ -74,64 +106,108 @@ func (f *Builder) BuildOn(parent *types.Block, build func(b *BlockBuilder)) *typ
...
+		if build != nil {
+			build(&BlockBuilder{b}, i)
+		}
 
-	if build != nil {
-		build(&BlockBuilder{b})
+		// Compute state root for this block.
+		tipSet, err := types.NewTipSet(b)
+		require.NoError(f.t, err)
+		b.StateRoot = f.ComputeState(tipSet)
```
ZenGround0: I'm a bit worried that we are treating the `builder` like a chain store (it implements the store interface that the syncer needs) and not keeping track of intermediate state roots.  As it stands right now the `builder.GetTipSetStateRoot` method will error in places the `GetTipSet` method will not because we track intermediate blocks and not intermediate state roots.

chain/syncer_integration_test.go
ZenGround0: This is an integration test I added after getting burned by bad assumptions between the store and the syncer taking nodes down.  Read [here](https://github.com/filecoin-project/go-filecoin/issues/1148#issuecomment-432008060) for a full description of what was being tested.

This particular edge case hasn't caused problems since this test was written and imo it is low risk to remove especially now that the chain store's block reading methods are gone.  

IMO we need to have integration tests that can hit specific cases like this without mocking out the chain store.  Ideally we move them to FAST.  I think this one could be ported to FAST without hassle.  `TipSetWeightDeep` would require more work but I think that it also belongs as a FAST integration test.  Having further syncer integration tests across a variety of edge cases is probably worthwhile, but may require significant CLI additions (i.e. a `sync start`/`sync once` subcommand) to do right.

chain/syncer_integration_test.go
ZenGround0: As mentioned above we could also turn this into a FAST test, though that might be considerable work.

chain/testing.go
```diff
@@ -21,13 +22,17 @@ import (
 // State root CIDs are computed by an abstract StateBuilder. The default FakeStateBuilder produces
 // state CIDs that are distinct but not CIDs of any real state tree. A more sophisticated
 // builder could actually apply the messages to a state tree (not yet implemented).
+// The builder is deterministic: two builders receiving the same sequence of calls will produce
+// exactly the same chain.
```
frrist: I know it is implied, but do you think it's worth adding a note stating that the miner address passed to `NewBuilder` is basically a "seed" for the chain it generates? I tripped over this bit the first time I used it.

chain/testing.go
```diff
@@ -56,16 +68,36 @@ func (f *Builder) AppendOn(parents ...*types.Block) *types.Block {
 	if len(parents) > 0 {
 		tip = types.RequireNewTipSet(f.t, parents...)
 	}
-	return f.BuildOnTip(tip, nil)
+	return f.AppendOnTip(tip)
+}
+
+// AppendOnTip creates and returns a new single-block tipset child of `parent`, with no messages.
```
frrist: This method returns a block, not a tipset. Is this comment a typo or should the method declaration be changed? I don't like the idea of calling a `types.Block` a single-block tipset since it cannot be treated as a `types.Tipset`.

chain/testing.go
```diff
@@ -56,16 +68,36 @@ func (f *Builder) AppendOn(parents ...*types.Block) *types.Block {
...
+func (f *Builder) AppendOnTip(parent types.TipSet) *types.Block {
+	return f.BuildOnTip(parent, 1, nil).At(0)
+}
+
+// AppendTipOn creates and returns a new `count`-block tipset child of `parents`, with no messages.
+func (f *Builder) AppendTipOn(parent types.TipSet, width int) types.TipSet {
+	return f.BuildOnTip(parent, width, nil)
 }
 
 // AppendManyOn appends `count` blocks to the chain.
```
frrist: typo: appends `height` blocks to the chain

chain/testing.go
```diff
@@ -74,64 +106,108 @@ func (f *Builder) BuildOn(parent *types.Block, build func(b *BlockBuilder)) *typ
...
-	f.seq++
-
-	// Sum weight of parents' parent weight, plus one for each parent.
-	// Note: as with the state builder, we should probably factor this out.
-	parentWeight := types.Uint64(0)
-	for i := 0; i < parent.Len(); i++ {
-		parentWeight += parent.At(i).ParentWeight + 1
-	}
+func (f *Builder) BuildOnTip(parent types.TipSet, width int, build func(b *BlockBuilder, i int)) types.TipSet {
+	require.True(f.t, width > 0)
```
frrist: baby nit: `require.NotZero` will give this a slightly better error message on failure (which is probably rare anyways)

chain/syncer_test.go
```diff
@@ -809,119 +223,62 @@ func TestLoadFork(t *testing.T) {
...
+	link4 := builder.AppendTipOn(link3, 2)
 
-	// Check that the widened tipset (link2intersect U dstP.link2) is tracked
-	link2Union := th.RequireNewTipSet(t, dstP.link2blk1, dstP.link2blk2, dstP.link2blk3, link2blkother)
-	assertTsAdded(t, chainStore, link2Union)
-}
+	// Build another block with parents link1, but not included in link2.
+	link2Alt := builder.AppendTipOn(link1, 1)
+	// Build a tipset containing one block from link2, plus this new sibling.
+	link2Sbuset := types.RequireNewTipSet(t, link2.At(0), link2Alt.At(0))
```
ZenGround0: nit: link2Subset is a bad name because it includes blocks not in link2

chain/syncer_integration_test.go
anorth: Thanks for the link. I'm not convinced FAST is the answer - we should be able to more directly put the store into some state then test the syncer deals with it. I think I understand the case now and will attempt to construct it in a unit test in a follow up.

*Specific* cases should be unit testable. Integration tests are good for the general case.

chain/testing.go
```diff
@@ -74,64 +106,108 @@ func (f *Builder) BuildOn(parent *types.Block, build func(b *BlockBuilder)) *typ
...
-	binary.BigEndian.PutUint64(ticket, f.seq)
-	f.seq++
-
-	// Sum weight of parents' parent weight, plus one for each parent.
-	// Note: as with the state builder, we should probably factor this out.
-	parentWeight := types.Uint64(0)
-	for i := 0; i < parent.Len(); i++ {
-		parentWeight += parent.At(i).ParentWeight + 1
-	}
+func (f *Builder) BuildOnTip(parent types.TipSet, width int, build func(b *BlockBuilder, i int)) types.TipSet {
```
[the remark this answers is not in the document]
anorth: Ok done. I added a couple more helpers and updated some existing tests to be tipset-based.

chain/testing.go
```diff
@@ -172,28 +248,75 @@ func (bb *BlockBuilder) SetStateRoot(root cid.Cid) {
...
+			return 0, err
 		}
-		fakeState = append(fakeState, mCid)
 	}
+	return parentWeight + uint64(tip.Len()), nil
+}
+
+///// State evaluator /////
+
+// FakeStateEvaluator is a syncStateEvaluator delegates to the FakeStateBuilder.
```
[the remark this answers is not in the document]
anorth: It's primarily answering a chain sync need. If something else consumes the same interface and we want to fake it the same way, then yes let's move it.

chain/testing.go
anorth: Actually I didn't think of it that way. We may want to add an explicit seed later.

In fact I don't think this miner address will last because it causes the chains to be "wrong" in that multiple blocks at the same height have the same miner. I expect to move to random miner addresses.

PR #3097, opened 2019-07-22
fix: mining start add get miner actor (#1544)

### LINKED ISSUE #1544: Commands `mining start` should not succeed if miner is not registered
### Description

In relation to issue #1539 it appears that the `miner create` call, while returns successful, may not actually be waiting for the message to have returned (this will probably be mitigated by #1535 as well?).

When running `mining start` immediately after, the miner is not verified on chain yet and fails with the errors from #1539.

The command `mining start` should probably verify that the default miner exists on chain, or fail otherwise.

It seems like we might not be properly returning the errors back up the the command, so it succeeds when it shouldn't.

/cc @whyrusleeping

deaswang:
Fixes #1544

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -771,6 +771,10 @@ func (node *Node) StartMining(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get mining address")
 	}
+	_, err = node.PorcelainAPI.ActorGet(ctx, minerAddr)
+	if err != nil {
+		return errors.Wrap(err, "failed to get miner actor")
+	}
 
 	// ensure we have a sector builder
 	if node.SectorBuilder() == nil {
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -771,6 +771,10 @@ func (node *Node) StartMining(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get mining address")
 	}
+	_, err = node.PorcelainAPI.ActorGet(ctx, minerAddr)
+	if err != nil {
+		return errors.Wrap(err, "failed to get miner actor")
```
frrist: A test for this should be added to `node_test.go` -- here is where I would start:
https://github.com/filecoin-project/go-filecoin/blob/199c9241887366995ea8493fc5d9136be9f09f3c/node/node_test.go#L142-L143

After the call to `GiveMiner` we should assert that calling `StartMining` returns the expected error, so something like this:
`	assert.Contains(t, minerNode.StartMining(ctx).Error(), "failed to get miner actor")`

deaswang:
Added test case.

PR #3100, opened 2019-07-22
Send message w/o broadcasting to network

shannonwells:
This adds a "publish" parameter to `Sender.Send` which tells Outbox whether to publish the message to the network.  It is hard-coded to `true` in the API, since it's intended to be used internally only, not exposed as a CLI or porcelain-level feature.

This is work in support of #2939 ; nodes want to slash miners but not tell the network that is what they are doing, but instead include this message in their next mined block. Fault monitor will call `outbox.Send` with `bcast` set to false.

### Alternatives considered 
Instead of changing the `Send` signature by adding a parameter to its already long list, I looked at creating two functions, `Send` and `SendWithoutBroadcast`, which each used an internal function and `Send` would broadcast.  However, since `Send` needs the block height and the signed message, the refactor got messy and clarity was lost. There is only one caller for `Outbox.Send`, so very little had to change.

### Other changes
I removed `msg/Sender` wrapper, because it wasn't being used. It looks like it was intended to be like `MessageWaiter` and friends, but this was ignored in favor of using Outbox directly. It was nothing but a thin wrapper anyway. Only using Sender as a wrapper instead of Outbox would require adding to it the three functions that are called directly on Outbox Queue (Queues, List and Clear), and I see no value add there, since these functions are obviously not private.  If someone thinks it should be added for consistency or some reason I didn't think of, I could be persuaded otherwise.


core/outbox.go
```diff
@@ -70,8 +70,9 @@ func (ob *Outbox) Queue() *MessageQueue {
 }
 
 // Send marshals and sends a message, retaining it in the outbound message queue.
+// If publish is true, it publishes the message to the network at the current block height.
 func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value types.AttoFIL,
-	gasPrice types.AttoFIL, gasLimit types.GasUnits, method string, params ...interface{}) (out cid.Cid, err error) {
+	gasPrice types.AttoFIL, gasLimit types.GasUnits, method string, publish bool, params ...interface{}) (out cid.Cid, err error) {
```
anorth: I propose moving the `publish` parameter one slot earlier so that the params directly follow the method name and the bool cannot be mistaken for a parameter.

core/outbox.go
```diff
@@ -120,9 +121,11 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
 		return cid.Undef, errors.Wrap(err, "failed to add message to outbound queue")
 	}
 
-	err = ob.publisher.Publish(ctx, signed, height)
-	if err != nil {
-		return cid.Undef, err
+	if publish {
```
anorth: Blocking: this doesn't do what you want. This will prevent the message reaching the message pool too: see `defaultMessagePublisher`. This parameter needs to be passed to the publisher. Given that, `broadcast` might be a better name for the parameter, though it's all just a matter of scopes.

Unfortunately this is a little difficult to integration test without adding an interface to abstract over the `pubsub.Publisher` so you can mock it. But you can improve the `mockPublisher` to record what value was the flag that it received.

Alternatively, the outbox could be refactored to hold a reference to (an interface to) the message pool, essentially inlining `defaultMessagePublisher` in the outbox. I'm ok with that too so long as the concrete message pool type isn't referenced. Slight preference for the first way.

core/outbox.go
```diff
@@ -115,12 +116,12 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
 		return cid.Undef, errors.Wrap(err, "failed to get block height")
 	}
 
-	// Add to the local message queue/pool at the last possible moment before broadcasting to network.
+	// Add to the local message queue at the last possible moment before
```
shannonwells: Fixed this misleading comment

core/outbox.go
shannonwells: I did sort of both: I made  node’s defaultMessagePublisher’s pubsub.Publisher refer instead to an interface/ It calls just one function on the publisher, so that the testing is more complete. We are just testing basic logic, so if you think it's too much I'll revert it.

core/outbox_test.go
```diff
@@ -129,22 +135,51 @@ func TestOutbox(t *testing.T) {
 
 		ob := core.NewOutbox(w, nullValidator{}, queue, publisher, nullPolicy{}, provider, provider)
 
-		_, err := ob.Send(context.Background(), sender, toAddr, types.ZeroAttoFIL, types.NewGasPrice(0), types.NewGasUnits(0), "")
+		_, err := ob.Send(context.Background(), sender, toAddr, types.ZeroAttoFIL, types.NewGasPrice(0), types.NewGasUnits(0), true, "")
 		assert.Error(t, err)
 		assert.Contains(t, err.Error(), "account or empty")
 	})
+	t.Run("enqueues and does not bcast if bcast set to false", func(t *testing.T) {
```
anorth: You could cut down on the test code by combining this into the test at line 40, and looping over a `bcast` value in the array `{true, false}`.

node/msg_publisher_test.go
```diff
@@ -0,0 +1,46 @@
...
+	require.NoError(t, err)
+
+	t.Run("Publish is called when bcast is true", func(t *testing.T) {
+		mnp := mockNetworkPublisher{}
+		pub := newDefaultMessagePublisher(&mnp, "Stuff", pool)
+		assert.NoError(t, pub.Publish(context.Background(), signed, 0, true))
+		assert.Equal(t, "Stuff", mnp.Topic)
+	})
+
+	t.Run("Publish is not called when bcast is false", func(t *testing.T) {
```
ZenGround0: It would also be good to check membership in the pool after calling publish.

node/msg_publisher_test.go
```diff
@@ -0,0 +1,46 @@
+package node
```
ZenGround0: Awesome that we now have tests for this thing!

PR #3109, opened 2019-07-23
Add optional func argument to InitAndStart series

travisperson:
go-filecoin can only be configured while the daemon is running using the cli. However, there are times when we need to change configuration prior to starting the daemon. This additional func argument allows for making changes to the on disk configuration using process.Config / process.WriteConfig.


tools/fast/series/init_and_start.go
```diff
@@ -6,12 +6,20 @@ import (
 	"github.com/filecoin-project/go-filecoin/tools/fast"
 )
 
-// InitAndStart is a quick way to run Init and Start for a filecoin process.
-func InitAndStart(ctx context.Context, node *fast.Filecoin) error {
+// InitAndStart is a quick way to run Init and Start for a filecoin process, between calls
+// to init and start, an optional function can be passed to perform any actions (generally
+// configuration) prior to the daemon starting.
+func InitAndStart(ctx context.Context, node *fast.Filecoin, fn func(*fast.Filecoin) error) error {
```
frrist: Would it be worth making `fn` a variadic argument instead?

PR #3111, opened 2019-07-23
add fetcher callback to halt fetching

frrist:
### What
This PR changes the Fetcher interface to accept a callback method that can be used to halt fetching blocks.
This PR then replaces calls to `collectChain` with this method.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -344,39 +344,48 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
 		return nil
 	}
 
-	haltOnState := func(t types.TipSet) bool {
-		return syncer.chainStore.HasTipSetAndState(ctx, t.String())
+	syncDoneCb := func(t types.TipSet) (bool, error) {
+		return syncer.chainStore.HasTipSetAndState(ctx, t.String()), nil
+	}
+
+	catchDoneCb := func(t types.TipSet) (bool, error) {
+		if syncer.chainStore.HasTipSetAndState(ctx, t.String()) {
+			return true, nil
+		}
+		// Only check if the head we are fetching exceeds finality
+		if t.Key().Equals(tsKey) {
+			ok, err := syncer.exceedsFinalityLimit(t)
+			if err != nil {
+				return true, err
+			}
+			if ok {
+				return true, ErrNewChainTooLong
+			}
+		}
+		return false, nil
 	}
 
 	var chain []types.TipSet
 	switch syncer.syncMode {
 	case Syncing:
-		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, haltOnState)
+		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, syncDoneCb)
 		if err != nil {
 			return err
 		}
-		chain = reverse(chain)
 	case CaughtUp:
 		// we cancel the fetch during caught up mode after a single blocktime.
 		fetchCtx, cancel := context.WithTimeout(ctx, blkWaitTime)
 		defer cancel()
-		chain, err = syncer.fetcher.FetchTipSets(fetchCtx, tsKey, haltOnState)
+
+		chain, err = syncer.fetcher.FetchTipSets(fetchCtx, tsKey, catchDoneCb)
 		if err != nil {
 			return err
-		}
-		chain = reverse(chain)
-		// TODO figure out a better check for finality than this
-		// https://github.com/filecoin-project/go-filecoin/issues/3112
-		ok, err := syncer.exceedsFinalityLimit(chain)
-		if err != nil {
-			return err
-		}
-		if !ok {
-			return ErrNewChainTooLong
 		}
 	default:
 		panic("syncer in unknown state")
 	}
+	// Fetcher returns chain in Traversal order, reverse it to height order
+	Reverse(chain)
 
 	parentCids, err := chain[0].Parents()
 	if err != nil {
@@ -422,9 +431,9 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
 	return nil
 }
 
-func (syncer *Syncer) exceedsFinalityLimit(chain []types.TipSet) (bool, error) {
-	if len(chain) == 0 {
-		return false, nil
+func (syncer *Syncer) exceedsFinalityLimit(tip types.TipSet) (bool, error) {
+	if !tip.Defined() {
+		return false, errors.New("cannon check finality limit on undefined tipset")
 	}
 	head, err := syncer.chainStore.GetTipSet(syncer.chainStore.GetHead())
 	if err != nil {
@@ -435,17 +444,9 @@ func (syncer *Syncer) exceedsFinalityLimit(chain []types.TipSet) (bool, error) {
 		return false, err
 	}
 	finalityHeight := types.NewBlockHeight(headHeight).Add(types.NewBlockHeight(uint64(FinalityLimit)))
-	chainHeight, err := chain[0].Height()
+	chainHeight, err := tip.Height()
 	if err != nil {
 		return false, err
 	}
 	return types.NewBlockHeight(chainHeight).LessThan(finalityHeight), nil
-}
-
-func reverse(tips []types.TipSet) []types.TipSet {
-	out := make([]types.TipSet, len(tips))
-	for i := 0; i < len(tips); i++ {
-		out[i] = tips[len(tips)-(i+1)]
-	}
-	return out
 }
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -145,89 +145,6 @@ func NewSyncer(e syncStateEvaluator, s syncerChainReaderWriter, f net.Fetcher, s
 	}
 }
 
-// fetchTipSetWithTimeout resolves a TipSetKey to a TipSet. This method will
-// request the TipSet from the network if it is not found locally in the
-// blockstore. This method is all or nothing, it will error if any of the cids
-// that make up the requested tipset cannot be resolved.
-func (syncer *Syncer) fetchTipSetWithTimeout(ctx context.Context, tsKey types.TipSetKey) (types.TipSet, error) {
-	ctx, cancel := context.WithTimeout(ctx, blkWaitTime)
-	defer cancel()
-
-	tss, err := syncer.fetcher.FetchTipSets(ctx, tsKey, 1)
-	if err != nil {
-		return types.UndefTipSet, err
-	}
-
-	// TODO remove this when we land: https://github.com/filecoin-project/go-filecoin/issues/1105
-	if len(tss) > 1 {
-		panic("programmer error, FetchTipSets ignored recur parameter, returned more than 1 tipset")
-	}
-
-	return tss[0], nil
-}
-
-// collectChain resolves the cids of the head tipset and its ancestors to
-// blocks until it resolves a tipset with a parent contained in the Store. It
-// returns the chain of new incompletely validated tipsets and the id of the
-// parent tipset already synced into the store.  collectChain resolves cids
-// from the syncer's fetcher.  In production the fetcher wraps a bitswap
-// session.  collectChain errors if any set of cids in the chain resolves to
-// blocks that do not form a tipset, or if any tipset has already been recorded
-// as the head of an invalid chain.  collectChain is the entrypoint to the code
-// that interacts with the network. It does NOT add tipsets to the chainStore..
-func (syncer *Syncer) collectChain(ctx context.Context, tsKey types.TipSetKey) (ts []types.TipSet, err error) {
-	ctx, span := trace.StartSpan(ctx, "Syncer.collectChain")
-	span.AddAttributes(trace.StringAttribute("tipset", tsKey.String()))
-	defer tracing.AddErrorEndSpan(ctx, span, &err)
-
-	var chain []types.TipSet
-	var count uint64
-	defer logSyncer.Infof("chain fetch from network complete %s", tsKey.String())
-
-	// Continue collecting the chain if we're either not yet caught up or the
-	// number of new input blocks is less than the FinalityLimit constant.
-	// Otherwise, halt assuming the new blocks come from an invalid chain.
-	for {
-		exceedsFinality, err := syncer.exceedsFinalityLimit(chain)
-		if err != nil {
-			return nil, err
-		}
-		if syncer.syncMode != Syncing && exceedsFinality {
-			break
-		}
-
-		// Finish traversal if the tipset made is tracked in the store.
-		if syncer.chainStore.HasTipSetAndState(ctx, tsKey.String()) {
-			return chain, nil
-		}
-
-		logSyncer.Debugf("CollectChain next link: %s", tsKey.String())
-
-		if syncer.badTipSets.Has(tsKey.String()) {
-			return nil, ErrChainHasBadTipSet
-		}
-
-		ts, err := syncer.fetchTipSetWithTimeout(ctx, tsKey)
-		if err != nil {
-			return nil, err
-		}
-
-		count++
-		if count%500 == 0 {
-			logSyncer.Infof("fetching the chain, %d blocks fetched", count)
-		}
-
-		// Update values to traverse next tipset
-		chain = append([]types.TipSet{ts}, chain...)
-		tsKey, err = ts.Parents()
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	return nil, ErrNewChainTooLong
-}
-
 // syncOne syncs a single tipset with the chain store. syncOne calculates the
 // parent state of the tipset and calls into consensus to run a state transition
 // in order to validate the tipset.  In the case the input tipset is valid,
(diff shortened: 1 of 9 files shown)


net/fetcher.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Fetcher defines an interface that may be used to fetch data from the network.
 type Fetcher interface {
-	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, recur int) ([]types.TipSet, error)
+	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, done func(ts types.TipSetKey) bool) ([]types.TipSet, error)
```
frrist: Operating over a TipSet**Key** here may prove problematic as it doesn't allow for attributes like `Height` to be inspected. We may want to change this to a full TipSet later on but currently there is no reason to.

Another option might be to provide another function as a parameter that is called after the tipset at tsKey has been fetcher.

chain/syncer.go
```diff
@@ -421,13 +345,42 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
-	// Walk the chain given by the input blocks back to a known tipset in
-	// the store. This is the only code that may go to the network to
-	// resolve cids to blocks.
-	chain, err := syncer.collectChain(ctx, tsKey)
-	if err != nil {
-		return err
+	var chain []types.TipSet
+	switch syncer.syncMode {
+	case Syncing:
+		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, func(t types.TipSetKey) bool {
```
ZenGround0: It looks like we could deduplicate the termination function by declaring it at the top level.

chain/testing.go
```diff
@@ -367,10 +367,12 @@ func (f *Builder) GetTipSet(key types.TipSetKey) (types.TipSet, error) {
 	return types.NewTipSet(blocks...)
 }
 
-// FetchTipSets returns `recur` tipsets from `key` by following parent keys.
-func (f *Builder) FetchTipSets(ctx context.Context, key types.TipSetKey, recur int) ([]types.TipSet, error) {
+// FetchTipSets fetchs the tipset at `tsKey` from the network using the fetchers blockStore backed by the Builder.
```
ZenGround0: You should scare quote "network" or somehow otherwise indicate that this doesn't actually go to the network.

chain/syncer.go
```diff
@@ -421,13 +345,39 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
 	}
 
-	// Walk the chain given by the input blocks back to a known tipset in
-	// the store. This is the only code that may go to the network to
-	// resolve cids to blocks.
-	chain, err := syncer.collectChain(ctx, tsKey)
-	if err != nil {
-		return err
+	haltOnState := func(t types.TipSetKey) bool {
+		if syncer.chainStore.HasTipSetAndState(ctx, t.String()) {
```
anorth: @ZenGround0 nit: the store interface would be nicer if it took tipset keys directly instead of strings, abstracting the underlying serialization format and limitations.

chain/syncer.go
```diff
@@ -421,13 +345,39 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
+			return true
+		}
+		return false
```
anorth: This should be a one-liner: `return syncer.ChainStore.HasTipSetAndState(...)`

chain/syncer.go
```diff
@@ -421,13 +345,39 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
+	}
+
+	var chain []types.TipSet
+	switch syncer.syncMode {
+	case Syncing:
+		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, haltOnState)
+		if err != nil {
+			return err
+		}
+		chain = reverse(chain)
```
anorth: Move this duplicated line after the case block.

You can ease this by changing `exceedsFinalityLimit` to only take a single block rather than a slice of which it only inspects the first element.

chain/testing.go
```diff
@@ -367,10 +367,12 @@ func (f *Builder) GetTipSet(key types.TipSetKey) (types.TipSet, error) {
 	return types.NewTipSet(blocks...)
 }
 
-// FetchTipSets returns `recur` tipsets from `key` by following parent keys.
-func (f *Builder) FetchTipSets(ctx context.Context, key types.TipSetKey, recur int) ([]types.TipSet, error) {
+// FetchTipSets fetchs the tipset at `tsKey` from the fetchers blockStore backed by the Builder.
+// FetchTipSets will only fetch TipSets whos TipSetKeys evaluate to `false` when passed to `done`,
+// this includes the provided `tsKey`.
```
anorth: Move the second sentence to the interface. Also document that the returned tipsets are in traversal order (which is reverse height order).

chain/syncer.go
```diff
@@ -421,13 +345,39 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
+	case CaughtUp:
+		// we cancel the fetch during caught up mode after a single blocktime.
+		fetchCtx, cancel := context.WithTimeout(ctx, blkWaitTime)
+		defer cancel()
+		chain, err = syncer.fetcher.FetchTipSets(fetchCtx, tsKey, haltOnState)
```
anorth: Blocking: it's more complicated than that. In this state, we aim to avoid fetching a giant chain that turns out not to join with ours, so the finality check has to come *before* the big fetch.

I think this should:
- fetch one tipset
- check whether it exceeds the finality limit
- if not, fetch the rest

chain/syncer.go
```diff
@@ -481,3 +431,11 @@ func (syncer *Syncer) exceedsFinalityLimit(chain []types.TipSet) bool {
 	chainHeight, _ := chain[0].Height()
 	return types.NewBlockHeight(chainHeight).LessThan(finalityHeight)
 }
+
+func reverse(tips []types.TipSet) []types.TipSet {
```
anorth: This exists already in core/policy.go, and without the unnecessary copy. Please extract that one to an exported function in `chain` somewhere and use it instead.

chain/testing.go
```diff
@@ -367,10 +367,12 @@ func (f *Builder) GetTipSet(key types.TipSetKey) (types.TipSet, error) {
...
+func (f *Builder) FetchTipSets(ctx context.Context, key types.TipSetKey, done func(t types.TipSetKey) bool) ([]types.TipSet, error) {
 	var tips []types.TipSet
-	for i := 0; i < recur; i++ {
+	for !done(key) {
```
anorth: Testing the key before fetching is slightly awkward, and I think worth changing.  There's no point invoking the predicate on the first `key` passed as parameter to the function: the called wouldn't have called if they didn't want to fetch something. It's more useful to inspect is the tipset that is fetched, with the predicate value indicating whether it wants to fetch another one (that tipset's parent). That will resolve your comment below too.

(This applies to the interface, which should clearly document this semantic).

net/fetcher.go
anorth: As noted above, I think *only* the latter is needed.

net/fetcher.go
```diff
@@ -36,11 +36,12 @@ func NewBitswapFetcher(ctx context.Context, bsrv bserv.BlockService, bv consensu
 }
 
 // FetchTipSets fetchs the tipset at `tsKey` from the network using the fetchers bitswap session.
-// FetchTipSets will fetch `recur` partens of `tsKey`. FetchTipSets does not return partial results.
-func (bsf *BitswapFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, recur int) ([]types.TipSet, error) {
+// FetchTipSets will only fetch TipSets whos TipSetKeys evaluate to `false` when passed to `done`,
+// this includes the provided `tsKey`.
```
anorth: Remove redundant comment in favour of the interface documentation

testhelpers/net.go
```diff
@@ -142,12 +142,13 @@ func (f *TestFetcher) AddSourceBlocks(blocks ...*types.Block) {
 	}
 }
 
-// FetchTipSets fetchs the tipset at `tsKey` from the network using the fetchers bitswap session.
-// FetchTipSets will fetch `recur` partens of `tsKey`. FetchTipSets does not return partial results.
-func (f *TestFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, recur int) ([]types.TipSet, error) {
+// FetchTipSets fetchs the tipset at `tsKey` from the network using the fetchers `sourceBlocks`
+// FetchTipSets will only fetch TipSets whos TipSetKeys evaluate to `false` when passed to `done`,
+// this includes the provided `tsKey`.
```
anorth: Ditto

chain/syncer.go
frrist: Created an issue for that here: https://github.com/filecoin-project/go-filecoin/issues/3112
will address in a follow on

consensus/block_validation.go
```diff
@@ -79,7 +84,7 @@ func (dv *DefaultBlockValidator) ValidateSyntax(ctx context.Context, blk *types.
 		return fmt.Errorf("block %s has nil StateRoot", blk.Cid().String())
 	}
 	if blk.Miner.Empty() {
-		return fmt.Errorf("block %s has nil miner address", blk.Miner.String())
+		return fmt.Errorf("block %s has nil miner address", blk.Cid().String())
```
frrist: Fixing a typo

consensus/block_validation.go
```diff
@@ -71,6 +71,11 @@ func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *ty
 
 // ValidateSyntax validates a single block is correctly formed.
 func (dv *DefaultBlockValidator) ValidateSyntax(ctx context.Context, blk *types.Block) error {
+	// TODO special handling for genesis block
+	// figure out in: https://github.com/filecoin-project/go-filecoin/issues/3121
+	if blk.Height == 0 {
```
frrist: Currently the genesis blocks we create fail validation, this check was added as a temporary fix to avoid running validation on genesis blocks.

consensus/block_validation_test.go
```diff
@@ -102,6 +102,7 @@ func TestBlockValidSyntax(t *testing.T) {
 		StateRoot: validSt,
 		Miner:     validAd,
 		Ticket:    validTi,
+		Height:    1,
```
frrist: if unset the default value for height is 0, setting it to 1 as we don't want to treat this as a genesis block.

core/policy.go
```diff
@@ -57,7 +57,7 @@ func (p *DefaultQueuePolicy) HandleNewHead(ctx context.Context, target PolicyTar
 
 	// Remove from the queue all messages that have now been mined in new blocks.
 	// Rearrange the tipsets into increasing height order so messages are discovered in nonce order.
-	reverse(newTips)
+	chain.Reverse(newTips)
```
frrist: method was extracted to the chain package

net/fetcher.go
```diff
@@ -15,7 +15,9 @@ import (
 
 // Fetcher defines an interface that may be used to fetch data from the network.
 type Fetcher interface {
-	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, recur int) ([]types.TipSet, error)
+	// FetchTipSets will only fetch TipSets whos TipSetKeys evaluate to `false` when passed to `done`,
+	// this includes the provided `tsKey`. The returns slice of TipSets is in Traversal order.
+	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, done func(ts types.TipSet) (bool, error)) ([]types.TipSet, error)
```
frrist: The `done` callback now returns an error. This is because we are performing a finality check with it and that method can error.

chain/syncer_integration_test.go
```diff
@@ -251,6 +251,7 @@ func initSyncTest(t *testing.T, con consensus.Protocol, genFunc func(cst *hamt.C
 	chainStore := chain.NewStore(chainDS, cst, &state.TreeStateLoader{}, calcGenBlk.Cid())
 
 	fetcher := th.NewTestFetcher()
+	fetcher.AddSourceBlocks(calcGenBlk)
```
frrist: The genesis block needs to exist in the store the fetcher pulls from, this is because the fetcher will stop fetching when it finds a block it has seen before and this can sometimes be the genesis block.

chain/syncer_integration_test.go
```diff
@@ -397,6 +398,10 @@ func TestLoadFork(t *testing.T) {
 
 	// Shut down store, reload and wire to syncer.
 	loadSyncer, blockSource := loadSyncerFromRepo(t, r, dstP)
+	// The fetcher will need to fetch these so that is may pass them to the
+	// done callback inorder to determine when to stop fetching.
+	// DONOTMERGE get specific approval on this before merging
+	blockSource.AddSourceBlocks(forklink3.ToSlice()...)
```
frrist: The fetcher needs this value in its store as it uses it to determine when it can stop fetching.

chain/syncer_integration_test.go
```diff
@@ -486,6 +491,9 @@ func TestTipSetWeightDeep(t *testing.T) {
 
 	// Setup a fetcher for feeding blocks into the syncer.
 	blockSource := th.NewTestFetcher()
+	// the block source needs the genesis
+	// DONOTMERGE
```
frrist: ditto the above statement about requiring the genesis block.

chain/syncer.go
```diff
@@ -478,22 +431,22 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
 	return nil
 }
 
-func (syncer *Syncer) exceedsFinalityLimit(chain []types.TipSet) (bool, error) {
-	if len(chain) == 0 {
-		return false, nil
+func (syncer *Syncer) exceedsFinalityLimit(newHead types.TipSet) (bool, error) {
```
frrist: There are 2 errors this method can return:
1. Calls to Tipset methods that return the undefined tipset error
2. The single call to `GetTipSet` that will return an error if chainStore is unable to find a block. A method on the ChainStore like `GetHeadTipSet` that doesn't error would be a "nice to have" here since it seems unlikely that we will fail to find the tipset for our current head.

This method is only used in the fetchers call back, and it would be great if we could remove the error from its return.

chain/syncer.go
```diff
@@ -423,17 +340,53 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
 	}
 
-	// Walk the chain given by the input blocks back to a known tipset in
-	// the store. This is the only code that may go to the network to
-	// resolve cids to blocks.
-	chain, err := syncer.collectChain(ctx, tsKey)
-	if err != nil {
-		return err
+	syncDoneCb := func(t types.TipSet) (bool, error) {
+		return syncer.chainStore.HasTipSetAndState(ctx, t.String()), nil
```
anorth: I think this can be `HasTipSetAndState(t.Parent.String())`, assuming the fetcher still returns the tipset for which the predicate returns `true`. This has the effect of stopping the fetch one tip sooner.

chain/syncer.go
```diff
@@ -423,17 +340,53 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
+
+	var chain []types.TipSet
+	switch syncer.syncMode {
+	case Syncing:
+		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, syncDoneCb)
+		if err != nil {
+			return err
+		}
+	case CaughtUp:
+		// we cancel the fetch during caught up mode after a single blocktime.
```
anorth: This comment will rot if `blkWaitTime` changes. 

`// Time out if this bounded fetch doesn't complete quickly`

chain/syncer.go
```diff
@@ -423,17 +340,53 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
+		if syncer.chainStore.HasTipSetAndState(ctx, t.String()) {
+			return true, nil
+		}
+		return false, nil
+	}
+
+	var chain []types.TipSet
+	switch syncer.syncMode {
+	case Syncing:
+		chain, err = syncer.fetcher.FetchTipSets(ctx, tsKey, syncDoneCb)
```
anorth: `// No time-out here, this can fetch an arbitrarily long chain`.

chain/syncer.go
anorth: I think it makes sense that the callback should be capable of propagating an error.

chain/syncer_integration_test.go
anorth: It doesn't make sense that the fetch is ever asked for the genesis block, `HasTipSetAndState` should always return true and hence it doesn't need to be fetched. Maybe my suggested change to the predicate above will fix this.

chain/testing.go
```diff
@@ -367,14 +367,21 @@ func (f *Builder) GetTipSet(key types.TipSetKey) (types.TipSet, error) {
...
 		if err != nil {
 			return nil, err
 		}
+		ok, err := done(tip)
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			break
+		}
```
anorth: Blocking: the last fetched tip should still be returned. I think the loop simplifies to `for !done {...}`.

net/fetcher.go
anorth: Blocking: this should always return at least one tipset, even if the predicate is `true`. The predicate means done *after* this tipset.

net/fetcher.go
```diff
@@ -51,15 +52,25 @@ func (bsf *BitswapFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetK
...
 		}
 
-		cur, err = ts.Parents()
+		ok, err := done(ts)
 		if err != nil {
 			return nil, err
 		}
+		if ok {
+			break
+		}
```
anorth: Ditto

testhelpers/net.go
```diff
@@ -158,12 +157,22 @@ func (f *TestFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, r
...
 
-		cur, err = ts.Parents()
+		ok, err := done(ts)
 		if err != nil {
 			return nil, err
 		}
 
+		if ok {
+			break
+		}
```
anorth: Ditto

chain/syncer_integration_test.go
frrist: Your suggestion on fetching parents fixes this -- thanks!

chain/testing.go
frrist: I have addressed the part about including the last fetched tipset in the result, however `done` will still need to be called inside the loop (as opposed to the `for {` declaration) since it operators on a `TipSet` and we don't yet have one at that point.

net/fetcher.go
frrist: agreed, addressed.

consensus/block_validation.go
ZenGround0: An alternative I'd prefer slightly to this is to make the validator aware of the genesis cid (pass it in during construction the same way expected gets it).  This way we are skipping only genesis.  Fine to leave that for #3121.

PR #3115, opened 2019-07-24
Split deal proposal context

travisperson:
The timeout context in deal proposal was being used to cover two different cases that should be handled independently.

First we need to verify that we are connected to the miner, and secondly we need to have our payment channel mined into a block. Everything else can use the regular context.

### DIFF
--- a/protocol/storage/client.go
+++ b/protocol/storage/client.go
@@ -183,14 +183,14 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data 
 	// Always set payer because it is used for signing
 	proposal.Payment.Payer = fromAddress
 
-	// The payment setup requires that the payment is mined into a block, currently we
-	// will wait for at most 5 blocks to be mined before giving up
-	ctxPaymentSetup, cancel := context.WithTimeout(ctx, 5*smc.api.BlockTime())
-	defer cancel()
-
 	// create payment information
 	totalCost := price.MulBigInt(big.NewInt(int64(pieceSize * duration)))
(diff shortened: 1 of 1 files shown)


protocol/storage/client.go
```diff
@@ -122,7 +119,7 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		return nil, fmt.Errorf("piece is %d bytes but sector size is %d bytes", pieceSize, maxUserBytes)
 	}
 
-	pieceReader, err := smc.api.DAGCat(ctxSetup, data)
+	pieceReader, err := smc.api.DAGCat(ctx, data)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to make piece reader")
 	}
```
anorth: Could you drop a comment below that `GeneratePieceCommitment` is computationally expensive and could take minutes for large pieces?

protocol/storage/client.go
```diff
@@ -179,8 +176,8 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 		if err != nil {
 			return nil, err
 		}
-	case <-ctxSetup.Done():
-		return nil, ctxSetup.Err()
+	case <-time.After(30 * time.Second):
```
anorth: There is already a timeout at line 104, the channel will hold an error if it expires. This additional timeout is confusing and I don't think necessary. I think we only want a simple read from the channel.

travisperson: ah interesting, makes this even simpler.

Issue #3098, opened 2019-07-22
Support HTTP REST API


eingenito:
### Description
Related spec issue: https://github.com/filecoin-project/specs/issues/400

The current HTTP RPC API from go-ipfs-cmds was never designed as a general purpose API framework and is awkward for developer-users:

* doesn’t support standard web patterns, e.g. for pagination
* doesn’t support client authentication
* not a widely-used or understood standard
* lacks rich tooling and support c.f. REST, gRPC or other common frameworks
* difficult to document as a wire format
* somewhat difficult to replicate outside of Go

A standard HTTP api could address these issues and offer the possibility of cross-implementation compatibility for tools that communicate with a node. 

### Scope
The goals of this effort would be:

* Design a v1 (or v0?) REST API for interacting with a filecoin node suitable for use across implementations
* Implement the API in go-filecoin
* Migrate the existing CLI commands to use the new API possibly eliminating the use of go-ipfs-cmds

### Risks + pitfalls
A lot of work to replace a working but somewhat awkward existing API. Helpful for longer term adoption and support of 3rd parties building on Filecoin

### Where to begin
@eingenito working on a draft of such an API for review by the team and any other interested parties.

ridewindx:
Any progress or plan on this?

eingenito:
Hey @ridewindx there is some progress - a WIP design document: https://docs.google.com/document/d/1ANnTHOU-8612ayvvS7Ru4B1L4voojLE0R0TQ8zF1x5s/edit#

shannonwells:
Status: See [filecoin-project/go-http-api](https://github.com/filecoin-project/go-http-api) -- further development is on hold for now.  There's a [go-filecoin branch](https://github.com/filecoin-project/go-filecoin/tree/feat/rest-api-part1) that needs to be updated to use the latest master, but it isn't far behind.

gargdeepak:
@shannonwells @anorth I can pick this up if you can share what's remaining to be done

anorth:
The time isn't right for this yet.

Issue #2523, opened 2019-04-09
Require non-zero gas price


acruikshank:
### Description

Allowing messages with zero gas price into the incoming message pool is a DoS vector since there is almost no barrier to creating accounts from which messages may be sent. In this respect a limit very slightly above zero (say 1 AttoFil) is still much better than zero.

see #2500

### Acceptance criteria

Messages with zero gas price fail. All documentation is updated to send messages with non-zero price.

### Risks + pitfalls

Almost everything is going to break. Not only to all our tests that currently send messages send them with zero price, very few of them send messages from an account that has a balance.

### Where to begin

Add a positive price check to `DefaultMessageValidator` and see what breaks.

PR #3118, opened 2019-07-24
add default for gas-price and gas-limit (#1699)

### LINKED ISSUE #1699: [bug] CLI Help when some "options" are required is self-contradicting
### Description
In several places in the CLI help, we have text that puts `--someOption` in square brackets, but such an option is required.  This is confusing, since the extended help has no indication that these things are required.  Example is the miner create help text quoted below.

If I run the miner create command as I have before (`go-filecoin miner create 10 100`), it fails, saying only that --price is required.  But both gas price and gas limit are required arguments.  In Unix help, square brackets around an argument mean it is optional, however these arguments are in square brackets in the help text.  In the "USAGE" section and the `miner --help` output, `--price` and `--limit` aren't shown at all (so the miner command help can't be used as a quick reference because it's incomplete).  In the "OPTIONS" shown below, `--price` and `--limit` also aren't marked required.

### Expected Behavior
All help text should show all required arguments. The only arguments in square brackets should be optional ones.  In the extended help (e.g. OPTIONS section), these arguments should say "required" in some way, and be the same way everywhere.

### Current Behavior
```
 $ gf miner create --help
USAGE
  go-filecoin miner create <pledge> <collateral> - Create a new file miner with <pledge> 1GB sectors and <collateral> FIL

SYNOPSIS
  go-filecoin miner create [--from=<from>] [--peerid=<peerid>] [--price=<price>] [--limit=<limit>] [--] <pledge> <collateral>

ARGUMENTS

  <pledge>     - The size of the pledge (in 1GB sectors) for the miner
  <collateral> - The amount of collateral in FIL to be sent (minimum 0.001 FIL per sector)

OPTIONS

  --from   string - Address to send from.
  --peerid string - Base58-encoded libp2p peer ID that the miner will operate.
  --price  string - Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message.
  --limit  uint64 - Maximum number of GasUnits this message is allowed to consume.
```

### Starting points
If there is a single default gas limit and gas price, then the default option text will need to be updated in [main.go](https://github.com/filecoin-project/go-filecoin/blob/4b6d10df1aa9ecebc6525e5a8e0d8bfbe5097660/commands/main.go#L316).

Also look in `commands/` for functions that call `parseGasOptions`:
* [message.go](https://github.com/filecoin-project/go-filecoin/blob/7cc6a6b790c3fd74e2f68c061a5e239348037b32/commands/message.go#L39)
* [miner.go](https://github.com/filecoin-project/go-filecoin/blob/4b6d10df1aa9ecebc6525e5a8e0d8bfbe5097660/commands/miner.go#L61) and other functions in this file
* [payment_channel.go](https://github.com/filecoin-project/go-filecoin/blob/4b6d10df1aa9ecebc6525e5a8e0d8bfbe5097660/commands/payment_channel.go#L68) and other functions in this file

deaswang:
Fixes #1699

### DIFF
--- a/commands/main.go
+++ b/commands/main.go
@@ -328,7 +328,7 @@ func isConnectionRefused(err error) bool {
 	return syscallErr.Err == syscall.ECONNREFUSED
 }
 
-var priceOption = cmdkit.StringOption("gas-price", "Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message").WithDefault("0.00013")
+var priceOption = cmdkit.StringOption("gas-price", "Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message").WithDefault("0")
 var limitOption = cmdkit.Uint64Option("gas-limit", "Maximum number of GasUnits this message is allowed to consume").WithDefault(300)
 var previewOption = cmdkit.BoolOption("preview", "Preview the Gas cost of this command without actually executing it")
 
--- a/commands/main.go
+++ b/commands/main.go
@@ -328,8 +328,8 @@ func isConnectionRefused(err error) bool {
 	return syscallErr.Err == syscall.ECONNREFUSED
 }
 
-var priceOption = cmdkit.StringOption("gas-price", "Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message")
-var limitOption = cmdkit.Uint64Option("gas-limit", "Maximum number of GasUnits this message is allowed to consume")
+var priceOption = cmdkit.StringOption("gas-price", "Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message").WithDefault("0.001")
+var limitOption = cmdkit.Uint64Option("gas-limit", "Maximum number of GasUnits this message is allowed to consume").WithDefault(1000)
 var previewOption = cmdkit.BoolOption("preview", "Preview the Gas cost of this command without actually executing it")
 
 func parseGasOptions(req *cmds.Request) (types.AttoFIL, types.GasUnits, bool, error) {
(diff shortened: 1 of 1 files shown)


deaswang:
already rebase master. @travisperson @frrist

deaswang:
strange, sometime pass, sometime fail at different test case here. test case should queued.

anorth:
The values have changed since this was reviewed. Please see my comment here: https://github.com/filecoin-project/go-filecoin/pull/1896#discussion_r256598999

travisperson:
We removed the ability for zero priced gas. (Issue https://github.com/filecoin-project/go-filecoin/issues/2523; PR https://github.com/filecoin-project/go-filecoin/pull/2605)

https://github.com/filecoin-project/go-filecoin/blob/3dbbf3a515ab9c9c48bb37da5dafc7bce77e17c0/consensus/validation.go#L67-L69

I'm not sure how these tests are passing unless the code above is dead.

anorth:
In that case I don't think we should have a default gas price baked into the code. Until we have something more sophisticated, I think we should require users to specify a gas price, and thus a bound on their cost. cc @mishmosh for comment.

Sorry for the misdirection from #1699, @deaswang. Can we fix up the help documentation to match behaviour instead?

deaswang:
OK. Wait for final decide. if gas price is required, it should move to Arguments. Options donot support require option.
https://godoc.org/github.com/ipfs/go-ipfs-cmdkit

anorth:
In *that* case I think maybe we should just not fix this, and wait for #3098 and a CLI implemented on top of it to solve all our problems.

frrist:
Sorry for the misdirection on this @deaswang, we are going to close this and reevaluate after #3098 lands.

PR #3131, opened 2019-07-25
Allow larger runs in RLE+

ingar:
## Problem
When RLE+ was implemented, the maximum run length was 16384, because of the very specific use case (fault sets) and how they were specified.  Now there is some flux in the specification of exactly what is encoded with RLE+ -- whether it is indices into a list of sector commitments or sector ids, and how many numbers can be encoded with RLE+.

## Solution
Allow runs of up to `MaxUint64`.  This will cover any use cases we will likely run across with go-filecoin, and still prevents attacks by limiting the allowed size of uvarints to 10 bytes.

### DIFF
--- a/rleplus/rleplus_test.go
+++ b/rleplus/rleplus_test.go
@@ -2,6 +2,7 @@ package rleplus_test
 
 import (
 	"fmt"
+	"math"
 	"sort"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


ingar:
This is definitely going to change, but things are in flux right now; most likely we will allow for runs up to `MaxUint64` long.

rleplus/rleplus_test.go
```diff
@@ -62,15 +63,11 @@ func TestRleplus(t *testing.T) {
...
 
-	t.Run("Encode limits run lengths to 2^14", func(t *testing.T) {
-		// Create a run larger than 2^14
-		var ints []uint64
-		for i := 0; i <= 1<<14; i++ {
-			ints = append(ints, uint64(i))
-		}
-
+	t.Run("Encode allows all runs sizes possible uint64", func(t *testing.T) {
+		// create a run of math.MaxUint64 - 1
```
ZenGround0: nit: comment seems off by one

steven004:
Glad to see this change after I just put some [comments](https://github.com/filecoin-project/go-filecoin/pull/2974#discussion_r319843184) for #2974 . 

In addition, I am thinking there might be a potential security risk which requires more change. That is, one miner could send out a message with very huge provingSet or faultSet presented by RLE+ to overflow all verifiers when running [decode function](https://github.com/filecoin-project/go-filecoin/blob/master/rleplus/rleplus.go#L88). To avoid this, I would suggest we still have some limitation, but on the size of ints array instead.

ingar:
@steven004 Could you open an issue for the large int slice DOS vulnerability?

steven004:
> @steven004 Could you open an issue for the large int slice DOS vulnerability?

Sure, will do.

PR #3132, opened 2019-07-25
Plumb peer ID to syncer's fetcher

ZenGround0:
@hannahhoward I plumbed the peer ID down to `FetchTipSets` let me know if I should push it further into `GetBlocks`.

### DIFF
--- a/chain/syncer_integration_test.go
+++ b/chain/syncer_integration_test.go
@@ -330,7 +330,7 @@ func TestLoadFork(t *testing.T) {
 	// Set up chain store to have standard chain up to dstP.link2
 	_ = requirePutBlocks(t, blockSource, dstP.link1.ToSlice()...)
 	cids2 := requirePutBlocks(t, blockSource, dstP.link2.ToSlice()...)
-	err := syncer.HandleNewTipset(ctx, cids2)
+	err := syncer.HandleNewTipset(ctx, cids2, peer.ID(""))
 	require.NoError(t, err)
 
 	// Now sync the store with a heavier fork, forking off dstP.link1.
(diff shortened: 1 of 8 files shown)


chain/syncer_integration_test.go
```diff
@@ -330,7 +330,7 @@ func TestLoadFork(t *testing.T) {
 	// Set up chain store to have standard chain up to dstP.link2
 	_ = requirePutBlocks(t, blockSource, dstP.link1.ToSlice()...)
 	cids2 := requirePutBlocks(t, blockSource, dstP.link2.ToSlice()...)
-	err := syncer.HandleNewTipset(ctx, cids2)
+	err := syncer.HandleNewTipset(ctx, cids2, th.RequireRandomPeerID(t))
```
anorth: Blocking: don't add nondeterminism to tests - use a fixed ID.

I suggest just `""` or `"testpeer"` if it's not to be used?

node/block.go
```diff
@@ -27,7 +27,7 @@ func (node *Node) AddNewBlock(ctx context.Context, b *types.Block) (err error) {
 	}
 
 	log.Debugf("syncing new block: %s", b.Cid().String())
-	if err := node.Syncer.HandleNewTipset(ctx, types.NewTipSetKey(blkCid)); err != nil {
+	if err := node.Syncer.HandleNewTipset(ctx, types.NewTipSetKey(blkCid), node.Host().ID()); err != nil {
```
frrist: Thinking out loud: It would be nice if the fetcher logic could be skipped for this case, else we'll have to check if this peerID is our own before fetching from it -- that can all happen later though.

net/fetcher.go
```diff
@@ -18,7 +19,7 @@ type Fetcher interface {
 	// FetchTipSets will only fetch TipSets that evaluate to `false` when passed to `done`,
 	// this includes the provided `ts`. The TipSet that evaluates to true when
 	// passed to `done` will be in the returned slice. The returns slice of TipSets is in Traversal order.
-	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, done func(ts types.TipSet) (bool, error)) ([]types.TipSet, error)
+	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, done func(ts types.TipSet) (bool, error), from peer.ID) ([]types.TipSet, error)
```
hannahhoward: Non-blocking: I generally feel a callback parameter (i.e. done func here) should go last, so that if it's inlined, the peer.ID is dangling several lines below.

PR #3134, opened 2019-07-26
Disallow mining start and once together (#2579)

### LINKED ISSUE #2579: [bug] Mining Once and Mining Start don't work as expected when used together
**Describe the bug**
Mining Once and Mining Start aren't safe to use together.  This is a placeholder issue that needs more work to describe the underlying race and exactly the conditions that cause this to work poorly.

**To Reproduce**
Try to include a message using mining once while running mining start and watch as it is sometimes not included.  FAST is an example of a place where this shows up because that framework uses once as part of some builtins.  I recently ran into behavior where running start on the FAST genesis miner and allocating funds using the builtin functions that wrap once ends up in lost data where the transfer of funds does not show up in the blockchain.  This is an unintuitive and tough to debug issue we should address sooner rather than later.

**Expected behavior**

once and the main loop triggered by start should never cause races on which blocks are included.  There should be a clear mental framework for understanding how to use these two commands together.

**Screenshots**

**Version information**
- Version: 8ea4ff036100e6960a91a19e01c9bfb72114d100

**Additional context**

@laser and I ran into this bug a while ago during post integration.  I promised I'd write it up but did not 👎 .  IIRC there is a race between including blocks with once and the main mining loop triggered by start that is not too hard to see if you go into the node code looking for it.

deaswang:
Fix #2579

### DIFF
--- a/protocol/block/mining_api_test.go
+++ b/protocol/block/mining_api_test.go
@@ -8,8 +8,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/protocol/storage"
 	"github.com/filecoin-project/go-filecoin/types"
 
-	ast "github.com/stretchr/testify/assert"
-	req "github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/node"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
@@ -18,58 +18,49 @@ import (
 func TestTrivialNew(t *testing.T) {
 	tf.UnitTest(t)
 
-	assert := ast.New(t)
-	require := req.New(t)
-
-	api, _ := newAPI(t, assert)
+	api, _ := newAPI(t)
 	require.NotNil(t, api)
 }
 
 func TestAPI_MineOnce(t *testing.T) {
 	tf.UnitTest(t)
 
-	assert := ast.New(t)
-	require := req.New(t)
 	ctx := context.Background()
-	api, nd := newAPI(t, assert)
-	require.NoError(nd.Start(ctx))
+	api, nd := newAPI(t)
+	require.NoError(t, nd.Start(ctx))
 	defer nd.Stop(ctx)
 
 	blk, err := api.MiningOnce(ctx)
-	require.Nil(err)
-	require.NotNil(blk)
-	assert.NotNil(blk.Ticket)
+	require.Nil(t, err)
+	require.NotNil(t, blk)
+	assert.NotNil(t, blk.Ticket)
 }
 
 func TestMiningAPI_MiningStart(t *testing.T) {
 	tf.UnitTest(t)
 
-	assert := ast.New(t)
-	require := req.New(t)
 	ctx := context.Background()
-	api, nd := newAPI(t, assert)
-	require.NoError(nd.Start(ctx))
+	api, nd := newAPI(t)
+	require.NoError(t, nd.Start(ctx))
 	defer nd.Stop(ctx)
 
-	require.NoError(api.MiningStart(ctx))
-	assert.True(nd.IsMining())
+	require.NoError(t, api.MiningStart(ctx))
+	assert.True(t, nd.IsMining())
 	nd.StopMining(ctx)
 }
 
 func TestMiningAPI_MiningIsActive(t *testing.T) {
 	tf.UnitTest(t)
 
-	assert := ast.New(t)
-	require := req.New(t)
 	ctx := context.Background()
-	api, nd := newAPI(t, assert)
-	require.NoError(nd.Start(ctx))
+	api, nd := newAPI(t)
+	require.NoError(t, nd.Start(ctx))
 	defer nd.Stop(ctx)
 
-	require.NoError(nd.StartMining(ctx))
-	assert.True(api.MiningIsActive())
+	require.NoError(t, nd.StartMining(ctx))
+	assert.True(t, api.MiningIsActive())
 	nd.StopMining(ctx)
-	assert.False(api.MiningIsActive())
+	assert.False(t, api.MiningIsActive())
 
 	nd.StopMining(ctx)
 }
@@ -77,16 +68,14 @@ func TestMiningAPI_MiningIsActive(t *testing.T) {
 func TestMiningAPI_MiningStop(t *testing.T) {
 	tf.UnitTest(t)
 
-	assert := ast.New(t)
-	require := req.New(t)
 	ctx := context.Background()
-	api, nd := newAPI(t, assert)
-	require.NoError(nd.Start(ctx))
+	api, nd := newAPI(t)
+	require.NoError(t, nd.Start(ctx))
 	defer nd.Stop(ctx)
 
-	require.NoError(nd.StartMining(ctx))
+	require.NoError(t, nd.StartMining(ctx))
 	api.MiningStop(ctx)
-	assert.False(nd.IsMining())
+	assert.False(t, nd.IsMining())
 }
 
 func TestMiningAPI_MiningAddress(t *testing.T) {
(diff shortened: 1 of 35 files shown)


anorth:
@ZenGround0 could you please shepherd this?

protocol/block/mining_api_test.go
```diff
@@ -90,6 +89,28 @@ func TestMiningAPI_MiningStop(t *testing.T) {
 	assert.False(nd.IsMining())
 }
 
+func TestMiningAPI_MiningTogether(t *testing.T) {
+	tf.UnitTest(t)
+
+	assert := ast.New(t)
+	require := req.New(t)
```
ZenGround0: Blocking: this used to be the way we set up testing but we recently moved to calling the top level package functions directly without setting up these types.  `assert.True(nd.IsMining()` becomes `assert.True(t, nd.IsMining())` calling the package instead of these types.

tools/fast/fastesting/basic.go
```diff
@@ -77,8 +77,18 @@ func NewTestEnvironment(ctx context.Context, t *testing.T, fastenvOpts fast.File
 	// Define a MiningOnce function which will bet set on the context to provide
 	// a way to mine blocks in the series used during testing
 	var MiningOnce series.MiningOnceFunc = func() {
-		_, err := genesis.MiningOnce(ctx)
+		isActive, err := genesis.MiningStatus(ctx)
 		require.NoError(t, err)
+		if isActive {
+			err = genesis.MiningStop(ctx)
```
ZenGround0: Blocking: this might be the right way but I am suspicious this will cause confusion.  It looks like this is not violating any existing test assumptions which is a good sign.  The alternative would be to make this fail when called during active mining and handle the complexity a level up.

cc @travisperson could you either sign off or suggest changes here?  @frrist you can make this call too if you feel high context.

frrist: I am a fan of handling this a level up as it feels like this could lead to some odd behaviour, but I will defer to @travisperson.

travisperson: I think the best thing to do here would be to have the user set their own mining once function when they are going to explicitly use the async mining by calling `ctx = series.SetCtxMiningOnce(ctx, func() {})` after they setup the environment, as the preferred method of using this package is through mining once, and the use of `mining start` anywhere at the moment is an issue that will be address in #2968.

So for now, we should update the other tests, and not change this.

tools/fast/action_mining.go
```diff
@@ -4,9 +4,21 @@ import (
...
+// MiningStatus runs the `mining status` command to get if mining
+func (f *Filecoin) MiningStatus(ctx context.Context) (bool, error) {
+	var out commands.MiningStatusResult
+
+	if err := f.RunCmdJSONWithStdin(ctx, nil, &out, "go-filecoin", "mining", "status"); err != nil {
+		return false, err
+	}
+
+	return out.Active, nil
+}
```
travisperson: This function should return the entire `commands.MiningStatusResult` structure.

deaswang: https://github.com/filecoin-project/go-filecoin/blob/1cab3122247e6a5c0f6a31227a649feb69f7aec2/tools/fast/action_miner.go#L18
https://github.com/filecoin-project/go-filecoin/blob/1cab3122247e6a5c0f6a31227a649feb69f7aec2/tools/fast/action_miner.go#L39
reference action_miner, it also return single value. and `commands.MiningStatusResult` have only one value.

protocol/block/mining_api_test.go
deaswang: I copy from other test case in the mining_api_test.go. Need all test case update?

frrist:
The contents of this PR have changed since approvals were given, we will need review the new code that has been added.

tools/fast/action_mining.go
travisperson: It could probably go either way at the moment, but the intent is that `miner status` will return additional information later. It's easier to return the status structure here now and now have to update everything down the road.

I would love to see `miner status` tell us the next proving period, proving set, etc.

deaswang:
@travisperson 
Thank for response. I pick up level up handle to fix it before. Need more to study.

travisperson:
@deaswang I apologize in the delay of getting you a response to this issue. I've been working through a strange bug with our nightly deployments that last several days.

I think this issue is getting a little more complex than originally though.

As the issue you are addressing points out, currently `mining start` and `mining once` aren't really suppose to be used together. Unfortunately some of these tests currently are trying to use both at the same time.

> If unset CtxMiningOnce. then `env.RequireNewNodeWithFunds` and `series.CreateStorageMinerWithAsk` won't get response for new block mined, then the message can't send in the `series.SendFilecoinFromDefault`, miner can't get the fund, also set price can't finish.

I don't know exactly what code you were using when you ran into this issue, but the issue I see is that if we use something like `ctx = series.SetCtxMiningOnce(ctx, func() error { return nil })` on the `ctx` returned by `fastesting.NewEnvironment` we aren't actually changing the context that is used when the environment. This results in the mining once func still be executed when we create new processes.

So we need to make sure we are updating the context for the environment, and using that new context everywhere else. I think this basically means we need to be sure to unset it quickly.

Unfortunately some of the tests do not opt into `mining start` immediately, so the changes aren't super easy. We need to restructure some of them.

Because this is a bit of a temporary work around till we resolve #2968 (as we should not be using `mining start` at all in these tests). I think we should make this an opt out kind of deal.

For that I think we should add a function to the `fastesting` structure called `RunAsyncMiner` which will perform the `series.SetCtxMiningOnce` with the noop function. It should also update the `ctx` on the `fastesting` structure so that other calls to `env.RequireNewNodeWithFunds` will be using the correct context. This function will also start mining for the `e.GenesisMiner` miner.

```
func (env *TestEnvironment) RunAsyncMiner() {
	var MiningOnce series.MiningOnceFunc = func() {}
	env.ctx = series.SetCtxMiningOnce(ctx, MiningOnce)
	err := env.GenesisMiner.MiningStart()
	require.NoError(env.t, err)
}
```

On any test that needs to make deals (will be using `mining start`) we just setup them up like so

```
ctx, env := fastesting.NewTestEnvironment(context.Background(), t, opts)
ctx = env.RunAsyncMiner()
defer func() {
  require.NoError(t, env.Teardown(ctx))
}()
```

So basically we are immediately clearing the `CtxMiningOnce` func upon construction of the environment and telling the `GenesisMiner` to just start mining.

deaswang:
@travisperson  Fixed. Thank.

deaswang:
@travisperson Rebase.

tools/fast/series/set_price_ask.go
```diff
@@ -21,7 +21,8 @@ func SetPriceGetAsk(ctx context.Context, miner *fast.Filecoin, price *big.Float,
 		return porcelain.Ask{}, err
 	}
 
-	CtxMiningOnce(ctx)
+	// Remove for issue #2579 mining start and once disallow together
+	// CtxMiningOnce(ctx)
```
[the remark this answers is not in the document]
deaswang: delete the code?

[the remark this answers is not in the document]
deaswang: The `CtxMiningOnce(ctx)` here is call after `miner.MiningStart(ctx)` at `CreateStorageMinerWithAsk` in `github.com/filecoin-project/go-filecoin/tools/fast/series/create_miner_with_ask.go:21`. so it always conflict with miningOnce. commented it only change one line, but unset it need to unset CtxMiningOnce at many test cases. also some test case use new daemon and ctx call `CreateStorageMinerWithAsk`, there are some trouble to unset CtxMiningOnce.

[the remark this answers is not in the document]
deaswang: @travisperson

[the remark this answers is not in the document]
travisperson: Okay, I'll deal with that bit later.

PR #3136, opened 2019-07-26
MinerPower bug fix & test

shannonwells:
## Problem
FAST's `Filecoin.MinerPower`, it is borked.

Based on @travisperson 's comment I've filed an [issue to change the output of the `miner power` command](https://github.com/filecoin-project/go-filecoin/issues/3144
) -- feel free to update to your liking.

## Solution
MinerPower was not updated to parse the latest output consisting of `<minerpower> / <totalpower>`.  It was missed because it hadn't been used yet. Fix this, include a regression test.

### DIFF
--- a/tools/fast/action_miner.go
+++ b/tools/fast/action_miner.go
@@ -3,6 +3,8 @@ package fast
 import (
 	"context"
 	"math/big"
+	"strconv"
+	"strings"
 
 	"github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p-core/peer"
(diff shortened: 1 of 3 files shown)


tools/fast/action_miner.go
```diff
@@ -65,17 +67,27 @@ func (f *Filecoin) MinerOwner(ctx context.Context, minerAddr address.Address) (a
 	return out, nil
 }
 
-// MinerPower runs the `miner power` command against the filecoin process
-func (f *Filecoin) MinerPower(ctx context.Context, minerAddr address.Address) (*big.Int, error) {
-	var out big.Int
+// MinerPower runs the `miner power` command against the filecoin process and returns the result
+// as two *big.Ints
+func (f *Filecoin) MinerPower(ctx context.Context, minerAddr address.Address) (int64, int64, error) {
```
shannonwells: I know this is technically out of line with power's type, but when writing the test for this, it seemed silly to output big.Ints;  in the test we'd be creating big.Ints from uint64s, we'd be parsing to (u)ints from strings, and converting that to big.Int, so .... skipping the middleman for testing purposes seemed reasonable.

tools/fast/action_miner_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	minerAddress := requireGetMinerAddress(ctx, t, minerDaemon)
+
+	dealDuration := uint64(5)
+	dealResponse, err := clientDaemon.ClientProposeStorageDeal(ctx, dataCid, minerAddress, 0, dealDuration, true)
+	require.NoError(t, err)
+	return dealResponse, dealDuration
+}
+
+func requireGetMinerAddress(ctx context.Context, t *testing.T, daemon *fast.Filecoin) address.Address {
```
shannonwells: Should `requireGetMinerAddress` be an export somewhere instead? It seems to be a thing we need.

travisperson:
I think the output of `miner power` should not be a string by default. We should update it to some structure with the two values and use the text encoder to get the `{power} / {total}`. This will make it a lot easier to parse on the FAST side, as well as parse for others consuming the API. I'm sure it will change as well in the new API work ever that lands.

shannonwells:
> I think the output of `miner power` should not be a string by default. We should update it to some structure with the two values and use the text encoder to get the `{power} / {total}`. This will make it a lot easier to parse on the FAST side, as well as parse for others consuming the API. I'm sure it will change as well in the new API work ever that lands.

So do you think the PR should also change the output of the command?  If so, I'd prefer to call this a temporary fix since I need it to write this slashing test.

tools/fast/action_miner.go
anorth: Ok for now, but note that the comment above is wrong.

FAST is not just for testing, so this is a shortcut that will trip someone up later.

shannonwells:
Thinking on this a bit, this is a bug fix for existing behavior. I think that since it had no test to begin with, this PR should not be for changing MinerPower output. If MinerPower behavior should change, that should be in a separate PR and then the test will have to change to match the new behavior.  Also, it's effectively blocking me from working on storage fault slashing.

shannonwells:
>FAST is not just for testing, so this is a shortcut that will trip someone up later.

If that's true it's easy to just output big.Ints and then it will never trip anyone up.  I'll just change it back.

Issue #1105, opened 2018-10-11
Differentiate initial/catch-up chain sync from maintaining sync


ZenGround0:
### Description
Change the syncer interface to expose an explicit initial (or catch-up) sync method, which is given a head CID to sync to. Distinguish this from the maintenance of sync, which happens via the HandleNewTipset method.

At boot, some node activities (connecting to pubsubs, mining) should be gated behind the initial sync completing.

The head CID is to be obtained by a low-trust bootstrapping mechanism (outside scope here).

### Acceptance criteria
When a node restarts or goes to sleep for a long time, it should traverse an arbitrarily long head. This "catch up" should happen automatically on restart. (In future, probably also in case a node detects it has been partitioned from the majority network for some time).

During normal operation a node should respect the maxNewChainLen limit and not swap its head for a fork that exceeds this.

### Risks + pitfalls
Chain sync is security-sensitive work. It's important for us to have a thorough understanding of behaviour and good tets.

### Where to begin
Modify the Syncer interface to differentiate between the two cases with a new method or an argument flag.  Now call the syncer during the node startup process, probably after `Load` completes on the head of the nodes being bootstrapped off of.  This requires messing with the startup flow so that bootstrapping occurs after chain loading.

anorth:
This issue was filed a long time ago - it's still basically something we want, I'll update the description to match our intentions.

frrist:
With the merge of #3111 we are getting closer to pushing this issue to completion. However, one largish piece of incomplete works that still sticks out to me is the ability for a node to determine if it has "CaughtUp" yet. 

In #3111 we have cases for a "Syncing" mode and a "CaughtUp" mode but are still lacking the ability to determine when to switch from one to the other. Here is how I imagine that working:

Whenever a Filecoin node starts, it creates a Syncer and places it in “Syncing” mode with block/message pubsub turned off. The filecoin node will then “Say Hello” to the bootstrap peers and receive their current heads. The node then begins fetching and processing the chain head it gets from the bootstrappers. During this fetching/processing time the network will continue to grow the chain, and If the node takes longer than finality blocks to fetch and process it’s chain it will need to remain in “Syncing” mode. I believe this means at the end of every “Syncing” run we will need to “Say Hello..again” to bootstrap nodes in order to get the CID of the latest head. If the height of the new chain head is [some-number-probably-600-ish] ahead of our own we will want to reamine in “Syncing” mode, else we can switch to “CaughtUp”.

[...]

Issue #2309, opened 2019-03-18
Replace chain store's headEvents "pubsub" with a synchronous event bus


anorth:
### Description
The chain store carries a `headEvents` reference which is a channel-based "pubsub" object from https://github.com/cskr/pubsub. The object supports notification of when a new head tipset is discovered. 

The asynchrony forced by using channels complicates, things, and in particular has led to a mutable `HeaviestTipSetHandled` callback func on the Node purely so that `node.TestUpdateMessagePool` can tell when the subscriber to that pubsub has finished.

Forcing the use of channels here unnecessary. This use would be much better served by a simple, synchronous event bus like [this one](https://github.com/asaskevich/EventBus/blob/master/event_bus.go). The message pool test could then simply update the head and know that the node has responded when the event post returns.

Note that this pattern doesn't prevent asynchrony for any particular event/subscriber, but makes it opt-in. Reducing unnecessary concurrency, and scoping it tightly, could also help us avoid work and/or headaches of handing concurrency in objects that are transitively reachable from any goroutine (we almost certainly don't handle this properly at present).

Since an event bus is a relatively simple pattern, I suggest we write our own, a simpler version of that linked above. We can then limit & customise functionality to exactly our needs. I expect an event bus will be useful for more than this single use case.

### Acceptance criteria
- New heaviest tipsets are notified via a synchronous event bus
- `Node.HeaviestTipSetHandled` is removed
- `github.com/cskr/pubsub` is removed from dependencies

### Risks + pitfalls
- This will involve a little refactor of the current tipset handling code.
- The referenced event bus does not plumb a `Context`, but we should

### Where to begin
In `node.go`, understand usage of `HeaviestTipSetCh` and `HeaviestTipSetHandled`.

anorth:
Libp2p have produced a local event bus, see https://github.com/libp2p/go-libp2p/issues/653. On my first reading, it seems to involve mandatory channels (and hence asynchrony), but worth investigating.

anorth:
The ugly `HeaviestTipSetHandled` callback has since been purged, but the forced concurrency here is still undesirable.

PR #3139, opened 2019-07-29
Defer subscription to block pubsub until after initial chain sync.

### LINKED ISSUE #3138: Defer block pubsub subscription until chain initial chain sync complete
### Description
At present, the node subscribes to a pubsub topic for blocks (headers) immediately at startup. It also begins chain sync (via callback from the hello protocol), which will block the syncer until it's caught up, potentially hours. Pubsub message callbacks will just queue up against that lock but achieve nothing useful. After the initial sync is complete, the node only needs to receive the most recent block in order to initiate catch up of whatever it missed while syncing.

This may become a bit more sophisticated after we implement principled peer & chain bootstrapping #2674.

See also #3137.

### Acceptance criteria
- A node does not participate in block header pubsub until it's ready to process those blocks

anorth:
Also cleans up some other sequencing and context in node boot.

A consequence of this is that after the chain syncer properly distinguishes between initial/trusted/bootstrap syncing and regular "caught up" syncing of close-to-head blocks (#1105), we won't see a slew of error messages from blocks that were received over pubsub but were waiting for that initial sync to complete and are now "too old". Note that since sync takes a long time, that bootstrap sync may need to be repeated a few times to get the trusted chain in range of the blocks appearing on pubsub (which would be needed anyway, not a consequence of this change).

Closes #3138.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
+	"runtime"
 	"sync"
 	"time"
 
@@ -65,6 +67,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/sampling"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/filecoin-project/go-filecoin/util/moresync"
 	vmerr "github.com/filecoin-project/go-filecoin/vm/errors"
 	"github.com/filecoin-project/go-filecoin/wallet"
 )
@@ -76,7 +79,7 @@ var (
 	ErrNoMinerAddress = errors.New("no miner addresses configured")
 )
 
-type pubSubProcessorFunc func(ctx context.Context, msg pubsub.Message) error
+type pubSubHandler func(ctx context.Context, msg pubsub.Message) error
 
 type nodeChainReader interface {
 	GenesisCid() cid.Cid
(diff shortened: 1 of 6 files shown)


node/node.go
```diff
@@ -516,61 +518,95 @@ func (node *Node) Start(ctx context.Context) error {
...
+		node.HelloSvc = hello.New(node.Host(), node.ChainReader.GenesisCid(), syncCallBack, node.PorcelainAPI.ChainHead, node.Repo.Config().Net, flags.Commit)
+
+		// Subscribe to block pubsub after the initial sync completes.
+		go func() {
+			chainSynced.Wait()
+			if syncCtx.Err() == nil {
+				// Subscribe to block pubsub topic to learn about new chain heads.
+				node.BlockSub, err = node.pubsubscribe(syncCtx, net.BlockTopic, node.processBlock)
+				if err != nil {
+					log.Errorf("failed to subscribe to %s: %s", net.MessageTopic, err)
```
frrist:
```suggestion
					log.Errorf("failed to subscribe to %s: %s", net.BlockTopic, err)
```

frrist: Is there any point continuing if this operation fails? Part of me thinks the node should exit.

node/node.go
```diff
@@ -649,7 +685,10 @@ func (node *Node) handleNewMiningOutput(miningOutCh <-chan mining.Output) {
 
 }
 
-func (node *Node) handleNewHeaviestTipSet(ctx context.Context, head types.TipSet) {
+func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet) {
+	node.HeaviestTipSetHandled = func() {}
```
ZenGround0: This appears to be unused here and in tests. If so it would be nice to just remove it.

node/node.go
anorth: Let's defer that to an operator for now. Longer term I think chain syncing should be runtime switchable on or off, independently of other node functions, so no failures should shut down a node.

node/node.go
anorth: Oh wonderful, thanks. Re https://github.com/filecoin-project/go-filecoin/issues/2309

node/node.go
frrist: Alright, that sounds reasonable to me -- the above error message is still wrong (should state that we failed to subscribe to the BlockTopic) but we change that in a follow on.

PR #3141, opened 2019-07-29
Rewrite syncer TestLoadFork to use chain builder and fakes

anorth:
The huge common test setup is now only required for the store tests, from which it should be pretty easy to remove.

### DIFF
--- a/chain/store_test.go
+++ b/chain/store_test.go
@@ -11,6 +11,8 @@ import (
 	"github.com/ipfs/go-hamt-ipld"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 
+	"github.com/libp2p/go-libp2p-core/peer"
+
 	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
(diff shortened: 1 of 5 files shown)


chain/store_test.go
```diff
@@ -23,7 +25,201 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// Note: many of these tests use the test chain defined in the init function of default_syncer_test.
+// This setup stuff is a horrendous legacy holdover from when these store tests shared a
+// common monolithic setup with the syncer tests. There's no good reason for it to still exist.
```
anorth: This is just a direct move, indicating that the syncer tests no longer use this.

chain/testing.go
```diff
@@ -175,9 +175,8 @@ func (f *Builder) Build(parent types.TipSet, width int, build func(b *BlockBuild
 		}
 
 		// Compute state root for this block.
-		tipSet, err := types.NewTipSet(b)
-		require.NoError(f.t, err)
-		b.StateRoot = f.ComputeState(tipSet)
+		prevState := f.StateForKey(parent.Key())
+		b.StateRoot, err = f.stateBuilder.ComputeState(prevState, [][]*types.SignedMessage{b.Messages})
```
anorth: Previously the block was placed in a tipset before the state root was calculated, which computed and cached the wrong CID in the block 😬. That took me quite a while to track down - the cached CID is a bit of a trap!

chain/syncer_integration_test.go
```diff
@@ -27,400 +27,78 @@ import (
...
-	fakeChildParams.Nonce = uint64(2)
-	forklink2blk3 := th.RequireMkFakeChild(t, fakeChildParams)
-	forklink2 := th.RequireNewTipSet(t, forklink2blk1, forklink2blk2, forklink2blk3)
+	assert.True(t, newStore.HasTipSetAndState(ctx, left.Key().String()))
+	assert.False(t, newStore.HasTipSetAndState(ctx, right.Key().String()))
 
-	fakeChildParams.Nonce = uint64(0)
-	fakeChildParams.Parent = forklink2
-	forklink3blk1 := th.RequireMkFakeChild(t, fakeChildParams)
+	newRight := builder.AppendManyOn(2, right)
```
ZenGround0: Blocking: these new blocks need to be available within the offlineSyncer's fetcher.  Right now they are placed in the original fetcher so the offlineSyncer will just fail because it couldn't fetch the two new blocks.  This test should instead check that the `HandleNewTipset` call fails because the *`right`* chain is not fetchable or considered already "stored" in the chain store even though all the blocks are kept in the `Store`'s internal repo.

In the previous version this intent is captured by putting `link1` and `link2`'s blocks to the cborStore while putting `link3`'s blocks to the `loadSyncer`'s `blockSource`.

chain/syncer_integration_test.go
```diff
@@ -27,400 +27,78 @@ import (
...
-	err = loadSyncer.HandleNewTipset(ctx, cidsFork4, peer.ID(""))
-	assert.NoError(t, err)
 }
 
 // Syncer handles MarketView weight comparisons.
 // Current issue: when creating miner mining with addr0, addr0's storage head isn't found in the blockstore
 // and I can't figure out why because we pass in the correct blockstore to createStorageMinerWithpower.
 func TestTipSetWeightDeep(t *testing.T) {
-	tf.UnitTest(t)
+	// This test takes many seconds, not sure exactly why.
```
ZenGround0: Doing a little bit of Printf investigation it looks like the call to gengen takes the majority of time.  Wrapping the gengen call with "start" and "done" print messages the "start" message shows up as soon as the go test output logs `CONT  TestTipSetWeightDeep`.  The output then hangs for 5 seconds and as soon as "done" gets printed `--- PASS: TestTipSetWeightDeep ([5-6]s)` gets printed.

I'm confident enough in this little inspection to recommend replacing "not sure exactly why" with "gengen call dominates/is bottleneck"

frrist: iirc this is related to gengen allocating power to the miners it defines in the genesis block.

chain/testing.go
frrist: I believe this was the purpose of the `paranoid` field in the block -- the comment above it hints at a similar strategy to the one @ZenGround0 has proposed: https://github.com/filecoin-project/go-filecoin/blob/44b95abe40a20441561516b797dd3e46ab0f3939/types/block.go#L63-L66

chain/syncer_integration_test.go
anorth: Ok, I think I fixed it, and have written an essay attempting to communicate this. 

I think this is a bad test. In particular, I think it would fail if the code and syncer/store interactions were correct, but just different to how they are now - i.e. if the syncer could read individual blocks, but didn't make invalid inferences about the state it has thus computed.

The interactions of the syncer, fetcher, and store are weird - I wouldn't yet call it "detangled", even after we clean up these tests. I will write more in an issue.

ZenGround0: @anorth as I mentioned [here](https://github.com/filecoin-project/go-filecoin/pull/3094#discussion_r305959284) we should consider ourselves free to remove this test.

PR #3142, opened 2019-07-29
fix Chain store takes tipset keys (#3119)

### LINKED ISSUE #3119: Chain store takes tipset keys
### Description
As pointed out [here](https://github.com/filecoin-project/go-filecoin/pull/3111#discussion_r306561887) the tipindex accessor methods on the chain store API take in strings in places where they should take in tipset keys.

### Acceptance criteria
Move to tipset keys in chain store.

### Risks + pitfalls
Should be very straightforward.

### Where to begin
`chain/store.go`: `HasTipSetAndState`, `GetTipSetAndStatesByParentsAndHeight`, `HasTipSetAndStatesWithParentsAndHeight`

deaswang:
Fix #3119

Add Serialize func for TipSetKey. It can replace String() if need.

### DIFF
--- a/types/tipset_key.go
+++ b/types/tipset_key.go
@@ -1,6 +1,9 @@
 package types
 
 import (
+	"bytes"
+	"encoding/base64"
+	"encoding/gob"
 	"encoding/json"
 	"fmt"
 	"sort"
@@ -126,6 +129,17 @@ func (s TipSetKey) String() string {
 		out = fmt.Sprintf("%s %s", out, it.Value().String())
 	}
 	return out + " }"
+}
+
+// Serialize returns a serialize string of TipSetKey use gob and base64.
+func (s TipSetKey) Serialize() string {
+	b := bytes.Buffer{}
+	e := gob.NewEncoder(&b)
+	err := e.Encode(s)
+	if err != nil {
+		return ""
+	}
+	return base64.StdEncoding.EncodeToString(b.Bytes())
 }
 
 // MarshalJSON serializes the key to JSON.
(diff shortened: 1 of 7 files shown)


types/tipset_key.go
```diff
@@ -128,6 +131,17 @@ func (s TipSetKey) String() string {
 	return out + " }"
 }
 
+// Serialize returns a serialize string of TipSetKey use gob and base64.
```
ZenGround0: Blocking: We already have a canonical encoding for `TipSetKey` into cbor defined at the top of this file.  I don't see the value in defining a gob serialization like this right now.  Perhaps I'm missing something so feel free to explain why you think this is important.

types/tipset_key.go
```diff
@@ -128,6 +131,17 @@ func (s TipSetKey) String() string {
...
+func (s TipSetKey) Serialize() string {
+	b := bytes.Buffer{}
+	e := gob.NewEncoder(&b)
+	err := e.Encode(s)
+	if err != nil {
+		return ""
```
frrist: in the off chance that we leave this method on `TipSetKey`, we should **not** be swallowing this error.

types/tipset_key.go
deaswang: OK, Remove it. and the method just for review to if it need.
https://filecoinproject.slack.com/archives/CEHHJNJS3/p1564377470109500

anorth:
Thanks, let's land this and then consider serialization later.

@ZenGround0 I requested a simpler serialisation for use as a database/map key - there's little reason to pay the cost of pretty-printed braces and base32 encoding for that use. This is not for exchange with other nodes.

However, such a serialisation could be encoded external to the tipset key, e.g. in the store module that intends to use it.

PR #3151, opened 2019-07-29
Bugs/action client ignores allow duplicates

### LINKED ISSUE #3150: [bug] FAST action_client/ClientProposeStorageDeal ignores allowDuplicates flag
**Describe the bug**
As the subject line says

**To Reproduce**
I have rewritten the test for duplicate deals to use FAST so it will show red/green test behavior: putting back the old behavior the test fails, and with the new code that uses the allowDuplicates flag it passes.  Logging the error output shows that the test fails because duplicate deals aren't allowed because the CLI was not passed `--allow-duplicates`.

shannonwells:
# Problem
Closes #3150 , FAST `ClientProposeDeal` was ignoring the allowDuplicates flag.  Discovered while writing an integration test for another story.

# Solution
- Pass --allow-duplicates as a parameter according to the flag.  
- Update `TestDuplicateDeals` from old-style daemon test to FAST, so it now uses this function `ClientProposeDeal`.

with @ingar

### DIFF
--- a/commands/client_daemon_test.go
+++ b/commands/client_daemon_test.go
@@ -3,25 +3,24 @@ package commands_test
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"math/big"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/ipfs/go-ipfs-files"
+	logging "github.com/ipfs/go-log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/fixtures"
-	"github.com/filecoin-project/go-filecoin/protocol/storage"
 	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/tools/fast"
 	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
 	"github.com/filecoin-project/go-filecoin/tools/fast/series"
-	"github.com/ipfs/go-ipfs-files"
-	logging "github.com/ipfs/go-log"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestListAsks(t *testing.T) {
--- a/commands/client_daemon_test.go
+++ b/commands/client_daemon_test.go
@@ -135,41 +135,6 @@ func minerClientMakeDealWithAllowDupes(ctx context.Context, t *testing.T, allowD
 	return dealResponse, err
 }
 
-func TestDuplicateDealsOld(t *testing.T) {
-	tf.IntegrationTest(t)
-
-	miner := th.NewDaemon(t,
-		th.WithMiner(fixtures.TestMiners[0]),
-		th.KeyFile(fixtures.KeyFilePaths()[0]),
-		th.DefaultAddress(fixtures.TestAddresses[0]),
-	).Start()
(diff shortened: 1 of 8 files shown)


commands/client_daemon_test.go
```diff
@@ -78,39 +80,61 @@ func TestStorageDealsAfterRestart(t *testing.T) {
...
-	defer miner.ShutdownSuccess()
-
-	client := th.NewDaemon(t, th.KeyFile(fixtures.KeyFilePaths()[2]), th.DefaultAddress(fixtures.TestAddresses[2])).Start()
-	defer client.ShutdownSuccess()
+	// Give the deal time to complete
+	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fast.FilecoinOpts{
+		InitOpts:   []fast.ProcessInitOption{fast.POAutoSealIntervalSeconds(1)},
+		DaemonOpts: []fast.ProcessDaemonOption{fast.POBlockTime(50 * time.Millisecond)},
+	})
+	defer func() {
```
frrist: I think we can just do:
`defer 		require.NoError(t, env.Teardown(ctx))`
Same with the below.

tools/fast/action_client.go
```diff
@@ -44,7 +44,16 @@ func (f *Filecoin) ClientProposeStorageDeal(ctx context.Context, data cid.Cid,
 	sAsk := fmt.Sprintf("%d", ask)
 	sDuration := fmt.Sprintf("%d", duration)
 
-	if err := f.RunCmdJSONWithStdin(ctx, nil, &out, "go-filecoin", "client", "propose-storage-deal", sMiner, sData, sAsk, sDuration); err != nil {
+	// RunCmd does not allow empty arguments (e.g. in the cast that allowDuplicates is false, use an
```
frrist: I think you can just do `--allow-duplicates=true` and  `--allow-duplicates=false`

frrist: also cast -> case if we keep this comment

tools/fast/process.go
```diff
@@ -199,6 +200,16 @@ func (f *Filecoin) LastCmdStdErr() io.ReadCloser {
 	return f.lastCmdOutput.Stderr()
 }
 
+// LastCmdStdErrStr is a shortcut to just get the output as string
+func (f *Filecoin) LastCmdStdErrStr() string {
```
frrist: Blocking: This is nice to have, but let's return an error here for the case that reading fails.

commands/client_daemon_test.go
```diff
@@ -78,39 +79,61 @@ func TestStorageDealsAfterRestart(t *testing.T) {
 
 func TestDuplicateDeals(t *testing.T) {
 	tf.IntegrationTest(t)
+	logging.SetDebugLogging()
 
-	miner := th.NewDaemon(t,
```
frrist: :champagne: :fireworks: :tada: :medal_sports: :1st_place_medal: :100:

tools/fast/action_client.go
travisperson: We should probably move this to the other pattern by using a variadic argument for flags.

@shannonwells good example: https://github.com/filecoin-project/go-filecoin/blob/master/tools/fast/action_miner.go#L18-L36

You can then add a function in https://github.com/filecoin-project/go-filecoin/blob/master/tools/fast/process_action_options.go

commands/client_daemon_test.go
```diff
@@ -78,39 +79,61 @@ func TestStorageDealsAfterRestart(t *testing.T) {
 
 func TestDuplicateDeals(t *testing.T) {
 	tf.IntegrationTest(t)
+	logging.SetDebugLogging()
```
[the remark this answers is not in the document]
shannonwells: yep thanks

tools/fast/process.go
shannonwells: meh, I was hoping to leave it since returning the error string will cause a test failure, and the dev will see immediately why, however it does break patterns...

PR #3153, opened 2019-07-30
Graphsync Fetcher

hannahhoward:
A WIP implementation of the Fetcher interface in graphsync, based on where it's headed in terms of interface

The main thing to look at is GraphSyncFetcher.FetchTipSet, and it's sub-calls

Since we no longer know the total recursion depth, it uses a compromise approach of increasing the amount of levels fetched per call up to a max level (64) -- this provides a compromise between fetching quickly in the case of a long sync vs not fetching too much extra in the case of a short sync.

Also, Graphsync generally starts queries from a single cid, vs here each query starts from a set of CIDs. We use a potentially inefficient approach fetching the first layer with non-recursive graphsync selectors, effectively mimic-ing Bitswap. After that, we use recursive calls to graphsync with increasing recursion depth


taoshengshi:
@hannahhoward , is here `recursive calls to graphsync` means flooding message to network along a graph path?

hannahhoward:
@taoshengshi no this does not mean flooding messages on a graph path. The intent of this new fetcher is to make syncing of the blockchain faster.

chain/testing.go
```diff
@@ -314,7 +312,7 @@ type FakeStateEvaluator struct {
 
 // RunStateTransition delegates to StateBuilder.ComputeState.
 func (e *FakeStateEvaluator) RunStateTransition(ctx context.Context, ts types.TipSet, ancestors []types.TipSet, stateID cid.Cid) (cid.Cid, error) {
-	return e.ComputeState(stateID, ts)
```
anorth: Please rebase, I think I've fixed what you were going for here.

core/policy_test.go
```diff
@@ -192,7 +192,8 @@ func TestMessageQueuePolicy(t *testing.T) {
 		b2 := blocks.BuildOnBlock(root, func(b *chain.BlockBuilder) {
 			b.AddMessage(msgs[1], &types.MessageReceipt{})
 			b.SetTicket([]byte{1})
-			b.SetTimestamp(2) // Tweak if necessary to force CID ordering opposite ticket ordering.
+			// Tweak Timestamp on block if necessary to force CID ordering
+			// opposite ticket ordering.
```
anorth: Do not delete this line. Change the value to zero if necessary.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
 	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, from peer.ID, done func(ts types.TipSet) (bool, error)) ([]types.TipSet, error)
 }
 
+// GraphSyncFetcher is used to fetch data over the network.  It is implemented with
+// a persistent bitswap session on a networked blockservice.
+type GraphSyncFetcher struct {
```
anorth: Please move this to a new file `graphsync.go`. We shouldn't have put the bitswap fetcher in here. Also the comment is wrong (I hope).

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+	gs        *graphsync.GraphSync
+	validator consensus.BlockSyntaxValidator
+	bridge    ipldbridge.IPLDBridge
+	bs        bstore.Blockstore
```
anorth: `bs` -> `store`?

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+}
+
+// NewGraphSyncFetcher returns a BitswapFetcher wired up to the input BlockService and a newly
+// initialized persistent session of the block service.
```
anorth: There are no block services visible here, so I don't understand. Can you clarify whether this constructor has side effects like starting goroutines or making network requests?

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+	gsf := &GraphSyncFetcher{
+		bs:        blockstore,
+		validator: bv,
+		bridge:    bridge,
+	}
+	gsf.gs = graphsync.New(ctx, network, bridge, gsf.loader, gsf.storer)
+	return gsf
+}
+
+func (gsf *GraphSyncFetcher) loader(lnk ipldp.Link, lnkCtx ipldp.LinkContext) (io.Reader, error) {
```
anorth: Nit: please move non-exported methods below exported ones so that the flow is more top-down to someone reading for the first time.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+		if err != nil {
+			return err
+		}
+		return gsf.bs.Put(block)
+	}
+	return &buffer, committer, nil
+}
+
+const maxRecursionDepth = 64
+const recursionMultiplier = 4
```
anorth: Please document what these mean and the implications of changing them.

taoshengshi:
> @taoshengshi no this does not mean flooding messages on a graph path. The intent of this new fetcher is to make syncing of the blockchain faster.

@hannahhoward  I mean how to understand  `recursive ` in this context?

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+// NewGraphSyncFetcher returns a BitswapFetcher wired up to the input BlockService and a newly
+// initialized persistent session of the block service.
+func NewGraphSyncFetcher(ctx context.Context, network gsnet.GraphSyncNetwork, bridge ipldbridge.IPLDBridge, blockstore bstore.Blockstore,
+	bv consensus.BlockSyntaxValidator) *GraphSyncFetcher {
+	gsf := &GraphSyncFetcher{
+		bs:        blockstore,
+		validator: bv,
+		bridge:    bridge,
+	}
+	gsf.gs = graphsync.New(ctx, network, bridge, gsf.loader, gsf.storer)
```
anorth: Blocking: this method of construction makes it very difficult to unit test the fetcher, since a fake graphsync cannot be injected. Please re-arrange construction so that we can exhaustively test the fetcher's behaviour and response to various things graphsync could return, like failures and partial responses, via a fake implementation of a consumer-defined interface for the fetcher's needs from graphsync.

I can see that the store interface methods are going to make that difficult, and you'll need to couple the store to the graphsync before wrapping it in a fetcher.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+				}
+
+				out = append(out, ts)
+				isDone, err := done(ts)
+				if err != nil {
+					return nil, err
+				}
+
+				if isDone {
+					break makeRecursion
```
anorth: This unstructured break seems unnecessary: the outer loop can be `for !isDone`, and the containing `if` become unnecessary then too

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+func (gsf *GraphSyncFetcher) LoadTipsetFromCids(ctx context.Context, cidResponseLayer []cid.Cid) (types.TipSet, error) {
+	blockResponseLayer := make([]blocks.Block, 0, len(cidResponseLayer))
+	for _, cid := range cidResponseLayer {
+		block, err := gsf.bs.Get(cid)
+		if err != nil {
+			return types.UndefTipSet, err
+		}
+		blockResponseLayer = append(blockResponseLayer, block)
+	}
+	validatedBlocks, err := sanitizeBlocks(ctx, blockResponseLayer, gsf.validator)
```
anorth: This implies that blocks are written to the store without being validated, and that seems bad to me. Can we validate in the storer instead? Or is the intention to avoid deserialising the stored bytes?

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+				recursionDepth *= recursionMultiplier
+			}
+		}
+	}
+	return out, nil
+}
+
+// LoadTipsetFromCids loads and sanitizes an already downloaded TipSet
+// from an array of cids
+func (gsf *GraphSyncFetcher) LoadTipsetFromCids(ctx context.Context, cidResponseLayer []cid.Cid) (types.TipSet, error) {
```
anorth: `ResponseLayer` etc in the names here makes it all sound more complicated than it is. I suggest just `cids`, `blocks` etc.

Also I don't think this needs to be exported.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+			return types.UndefTipSet, err
+		}
+		blockResponseLayer = append(blockResponseLayer, block)
+	}
+	validatedBlocks, err := sanitizeBlocks(ctx, blockResponseLayer, gsf.validator)
+	if err != nil {
+		return types.UndefTipSet, err
+	}
+
+	ts, err := types.NewTipSet(validatedBlocks...)
```
anorth: Just `return types.NewTipSet(...)`.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+	if err != nil {
+		return types.UndefTipSet, err
+	}
+
+	return ts, nil
+}
+
+type resultSet struct {
```
anorth: Please document what this is intended to do. The name seems a bit odd to me - would just "response" work?

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+	for _, c := range cids {
+		responseChan, errChan := gsf.gs.Request(ctx, from, cidlink.Link{Cid: c}, selector)
+		resultSets = append(resultSets, resultSet{responseChan, errChan})
+	}
+	for _, rs := range resultSets {
+		for err := range rs.errChan {
+			return err
+		}
+		for range rs.responseChan {
+		}
```
anorth: I think we don't need to consume these channels in full since their references don't escape this function, and they'll be collected, in which case this second loop is misleading. If we *do*, then this code is wrong because returning an error will cause leaks. Please add a comment either way.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+}
+
+type resultSet struct {
+	responseChan <-chan graphsync.ResponseProgress
+	errChan      <-chan error
+}
+
+// GetSingleLayer requests a single layer of cids as individual bocks, fetching
+// non-recursively
+func (gsf *GraphSyncFetcher) GetSingleLayer(ctx context.Context, cids []cid.Cid, from peer.ID) error {
```
anorth: Naming nit: I prefer methods which might do network requests to be named `Fetch...`, clearly indicating the potential latency involved. Reserve `Get` for looking things up in a map. I like that you used `Load` for reading from the store.

Also unexport.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+		}
+		for range rs.responseChan {
+		}
+	}
+	return nil
+}
+
+// GetBlocks gets multiple layers of parent blocks starting from a baseCid, up to
+// the given recursion depth parameter
+func (gsf *GraphSyncFetcher) GetBlocks(ctx context.Context, baseCid cid.Cid, from peer.ID, recursionDepth int) ([][]cid.Cid, error) {
```
anorth: `FetchLayers`? Also unexport

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+
+	selector, err := gsf.bridge.BuildSelector(func(ssb selector.SelectorSpecBuilder) selector.SelectorSpec {
+		return ssb.ExploreRecursive(recursionDepth, ssb.ExploreFields(func(efsb selector.ExploreFieldsSpecBuilder) {
+			efsb.Insert("parents", ssb.ExploreUnion(
+				ssb.ExploreAll(ssb.Matcher()),
+				ssb.ExploreIndex(0, ssb.ExploreRecursiveEdge()),
```
anorth: Please document the intent of this query in English, maybe copying from the doc where we worked them out.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+		return nil, err
+	}
+
+	cidResponseLayers := make([][]cid.Cid, recursionDepth)
+	hasCids := make([]map[cid.Cid]struct{}, recursionDepth)
+
+	responseChan, errChan := gsf.gs.Request(ctx, from, cidlink.Link{Cid: baseCid}, selector)
+	for err := range errChan {
+		return nil, err
+	}
```
anorth: Does this block until the entire request is complete (and the errChan closed), buffering all results in memory? It seems like there should be a better pattern.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+	for response := range responseChan {
+		// we are only interested in specific blocks
+		depth := (len(response.Path.Segments()) / 2) - 1
+		if depth < 0 {
+			continue
+		}
```
anorth: The above needs more explanation, I don't understand at all.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+		hasCidsLayer := hasCids[depth]
+		if hasCidsLayer == nil {
+			hasCidsLayer = make(map[cid.Cid]struct{})
+			hasCids[depth] = hasCidsLayer
+		}
+		asCidLink, ok := response.LastBlock.Link.(cidlink.Link)
+		if !ok {
+			continue
```
anorth: Needs more explanation. When can this happen?

net/fetcher_test.go
```diff
@@ -84,3 +94,71 @@ func TestFetchNotBlockFormat(t *testing.T) {
 	require.Error(t, err)
 	require.Nil(t, blocks)
 }
+
+func TestGraphsyncFetcherHappyPath(t *testing.T) {
```
anorth: I would call this an "integration test", of multiple graphsyncs/fetchers. We need it, but we also need unit tests for the fetcher code in isolation - see my comment in the constructor.

Also please extract to a graphsync_test.go file, to match moving the implementation.

node/node.go
```diff
@@ -384,7 +386,11 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 	//nwork := bsnet.NewFromIpfsHost(innerHost, router)
 	bswap := bitswap.New(ctx, nwork, bs)
 	bservice := bserv.New(bs, bswap)
-	fetcher := net.NewBitswapFetcher(ctx, bservice, blkValid)
+	//fetcher := net.NewBitswapFetcher(ctx, bservice, blkValid)
+
+	bridge := ipldbridge.NewIPLDBridge()
+	gsn := gsnet.NewFromLibp2pHost(peerHost)
+	fetcher := net.NewGraphSyncFetcher(ctx, gsn, bridge, bs, blkValid)
```
anorth: So that we can keep PRs small and make rapid progress on the path here, I suggest not checking these changes in yet, which means we can check in code above before it's necessarily full and complete.

hannahhoward:
@ZenGround0 Are you familiar with the Graphsync protocol & IPLD Selectors? If not, I would familiarize yourself with these specs and that may help provide some background:
https://github.com/ipfs/go-graphsync
https://github.com/ipfs/go-graphsync/blob/master/docs/architecture.md
https://github.com/ipld/specs/blob/master/block-layer/graphsync/graphsync.md
https://github.com/ipld/specs/blob/master/selectors/selectors.md

net/fetcher.go
hannahhoward: outdated comment will fix.

net/fetcher.go
hannahhoward: for sure

net/fetcher.go
hannahhoward: what is the general guideline for you all on variable names? I tend to lean towards the verbose, but I know go lang folks tend to like things short. I noticed some short names in the closest code and started heading that way, but I think maybe from your comment longer is ok?

net/fetcher.go
hannahhoward: The good news is the next step -- I don't think the loader/storer should be in fetcher nor should the bridge or the network. The method on the bridge is just a convenience method and probably not needed. I posted the PR mostly cause this is the first time I got a test to pass.

net/fetcher.go
hannahhoward: yea I was trying to minimize the number serializations/deserializations to store. 

One thing I would point out is that while it may not be obvious -- Bitswap is doing the same thing:
https://github.com/ipfs/go-bitswap/blob/master/bitswap.go#L282
That's in the code path every time you get a wanted block from the network.

Arguably with Bitswap it's less bad because only one layer at a time. 

One other complication of putting things in the storer is that currently the loader/storer can still be extracted pretty easily from the fetcher so that Graphsync is initialized outside the fetcher. Which, again, was the next step. However, this is more complicated if we're trying to do validation inside the storer. Though I guess if we're don't mind the extra serialize/deserialize we can just return an error in the storer when we're unable to verify format which should abort the Graphsync call with an error.

net/fetcher.go
ZenGround0: Along those lines replacing "layer" with "tipset" will make this code more idiomatic.  i.e. `cidsResponseTipset`

net/fetcher.go
ZenGround0: Guarding block store stores with a custom validation function sounds great for the future, but as @hannahhoward says it would be a new feature.  Unless it is very simple we should wait for another PR to make this improvement.

hannahhoward: Also, longer term I believe we may be able to use some neat facilities of `go-ipld-prime` to do this without extra serialization/deserialization, when we are ready to start integrating parts of it into the rest of the codebase.

net/fetcher.go
```diff
@@ -22,6 +30,212 @@ type Fetcher interface {
...
+}
+
+const maxRecursionDepth = 64
+const recursionMultiplier = 4
+
+// FetchTipSets gets Tipsets starting from the given tipset key and continuing till
+// the done function returns true or errors
+func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, from peer.ID, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
+	cids := tsKey.ToSlice()
+	err := gsf.GetSingleLayer(ctx, cids, from)
```
frrist: I don't understand why we are doing a single request here. I initially thought we needed the CID's of `tsKey` parents, but that doesn't seem to be the case since we are passing `cids` to `GetBlocks` below. Can you add a comment explaining what is going on here?

taoshengshi:
> @taoshengshi Are you familiar with the Graphsync protocol & IPLD Selectors? If not, I would familiarize yourself with these specs and that may help provide some background:
> https://github.com/ipfs/go-graphsync
> https://github.com/ipfs/go-graphsync/blob/master/docs/architecture.md
> https://github.com/ipld/specs/blob/master/block-layer/graphsync/graphsync.md
> https://github.com/ipld/specs/blob/master/selectors/selectors.md

copy that, will dive deep, thanks!

net/fetcher.go
anorth: A wide range is acceptable and this is by no means a blocking comment. I think my preference is slightly more descriptive than some.

In general I consider it a function of scope. For a local in a loop or short function, small is great. Struct members need to make sense perhaps 100s of lines away from their definition, so more information helps. Exported members even more so, such that I tend to use one or more actual words for them.

net/fetcher.go
anorth: Happy for your discretion, but please leave a comment describing this to the future.

Aside: my preference for in-code documentation is definitely for more than the average.

net/fetcher.go
frrist: Per our last synchronous chat, this is required since graphsync operates over a single CID. No clarification/reply here needed, but a comment stating this would be appreciated.

net/graphsync_fetcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	t.Run("happy path returns correct tipsets", func(t *testing.T) {
+		parentBlock := types.NewBlockForTest(nil, uint64(0))
+		block1 := types.NewBlockForTest(parentBlock, uint64(3))
+		block2 := types.NewBlockForTest(parentBlock, uint64(5))
+		block3 := types.NewBlockForTest(parentBlock, uint64(6))
+		requireBlockStorePut(t, bs, parentBlock.ToNode())
+		requireBlockStorePut(t, bs, block1.ToNode())
+		requireBlockStorePut(t, bs, block2.ToNode())
+		requireBlockStorePut(t, bs, block3.ToNode())
```
anorth: I find this test quite confusing because the blocks are already in the store. I would expect them to be put there only as a side-effect of the graphsync request. I think I understand what's going on and why this test works, but it's very white-box, and very legitimate changes to the internals of the fetcher could break this test without changing essential functionality.

Since you have started with a mock graphsync (thanks!) let's make it better and have the mock responsible for placing blocks in the store. This will then remain a good test if we dispense with the fetcher return value and lean in to relying on the storage side-effect.

net/graphsync_fetcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+	return graphsync.ResponseProgress{
+		Path: ipld.ParsePath(path),
+		LastBlock: struct {
+			Path ipld.Path
+			Link ipld.Link
+		}{
+			Link: cidlink.Link{Cid: blockCid},
+		},
+	}
+}
```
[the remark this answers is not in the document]
hannahhoward: for sure

hannahhoward:
Will make changes. FYI I left the error tests and handling very simple and somewhat untested just cause I am really waiting to see how the PeerManager shakes out first, cause it will like play big roll in error recovery.

net/graphsync_fetcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+			ssb.ExploreIndex(0, ssb.ExploreRecursiveEdge()),
+		))
+	})).Selector()
+	require.NoError(t, err)
+
+	t.Run("happy path returns correct tipsets", func(t *testing.T) {
+		parentBlock := types.NewBlockForTest(nil, uint64(0))
+		block1 := types.NewBlockForTest(parentBlock, uint64(3))
+		block2 := types.NewBlockForTest(parentBlock, uint64(5))
+		block3 := types.NewBlockForTest(parentBlock, uint64(6))
```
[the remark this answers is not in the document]
hannahhoward: will switch. Just for reference, I was emulating fetcher_test.go, which I assume was written before the chain builder. will switch over.

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,219 @@
...
+			if depth < 0 {
+				continue
+			}
+			cidMap := cidMaps[depth]
+			if cidMap == nil {
+				cidMap = make(map[cid.Cid]struct{})
+				cidMaps[depth] = cidMap
+			}
+			asCidLink := response.LastBlock.Link.(cidlink.Link)
+			cidMap[asCidLink.Cid] = struct{}{}
```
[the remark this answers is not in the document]
hannahhoward: you are 100% right and this is much simpler!

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
+package net
+
+import (
```
frrist: nit: We use the following import ordering (per [conventions and style](https://github.com/filecoin-project/go-filecoin/blob/master/CONTRIBUTING.md#conventions-and-style)):
```
import (
        [stdlib packages, alpha-sorted]
        <single, empty line>
        [external packages]
        <single, empty line>
        [go-filecoin packages]
)
```

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+//
+// To do this, the Graphsync fetcher starts fetching a single tipset at a time,
+// then gradually ramps up to fetch lots of tipsets at once, up to a fixed limit
+//
+// The constants below determine the maximum number of tipsets fetched at once
+// (maxRecursionDepth) and how fast the ramp up is (recursionMultipler)
+const maxRecursionDepth = 64
+const recursionMultiplier = 4
+
+// FetchTipSets gets Tipsets starting from the given tipset key and continuing till
```
ZenGround0: nit: till -> until

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+// the done function returns true or errors
+//
+// For now FetchTipSets operates in two parts:
+// 1. It fetches relevant blocks through Graphsync, which writes them to the block store
+// 2. It reads them from the block store and validates there syntax as blocks
```
ZenGround0: nit: their

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+// and constructs a tipset
+// This does have a potentially unwanted side effect of writing blocks to the block store
+// that later don't validate (bitswap actually does this as well)
+// In the future, the blocks will be validated directly through graphsync as
+// go-filecoin migrates to the same IPLD library used by go-graphsync (go-ipld-prime)
```
ZenGround0: This reads like a TODO.  Would you mind filing a github issue for the work migrating go-filecoin to go-ipld-prime and then referencing the issue number in this comment?  This is a convention we have recently adopted to prevent hidden TODOs from piling up.

net/graphsync_fetcher_test.go
```diff
@@ -0,0 +1,296 @@
...
+	})
+}
+
+func TestRealWorldGraphsyncFetchAcrossNetwork(t *testing.T) {
+	tf.IntegrationTest(t)
+	ctx := context.Background()
+	// setup a chain
+	builder := chain.NewBuilder(t, address.Undef)
+	gen := builder.NewGenesis()
+	final := builder.AppendManyOn(30, gen)
```
frrist: nit: Can the `30` be put into a variable and then make `31` ([here](https://github.com/filecoin-project/go-filecoin/pull/3153/files#diff-883ff976e50c0f0232bc503f90d13111R220)) `variable+1`?

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	out := []types.TipSet{startingTipset}
+	isDone, err := done(startingTipset)
+	if err != nil {
+		return nil, err
+	}
+	recursionDepth := 1
+	ts := startingTipset
+	for !isDone {
+		err := gsf.fetchBlocksRecursively(ctx, ts.At(0).Cid(), from, recursionDepth)
```
ZenGround0: I think what is happening here is that we include the first block of our previously fetched tipset in our query in order to avoid the waste that comes from querying multiple starting cids.  If I've got that correct then this is great, but it is the kind of thing that looks like an error and will cause confusion and mistaken "I found a bug" moments for future readers.  It would be awesome to have a comment explaining that this is intentional and why we need it.

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	for _, errChan := range errChans {
+		for err := range errChan {
+			return err
+		}
+	}
+	return nil
+}
+
+// fetchBlocksRecursively gets multiple sets of parent blocks starting from a baseCid, up to
```
ZenGround0: More idiomatically "...gets the blocks from recursionDepth ancestor tipsets starting from baseCid..."

net/graphsync_fetcher.go
```diff
@@ -0,0 +1,180 @@
...
+// the given recursion depth parameter
+func (gsf *GraphSyncFetcher) fetchBlocksRecursively(ctx context.Context, baseCid cid.Cid, from peer.ID, recursionDepth int) error {
+
+	// recursive selector to fetch n sets of parent blocks
+	// starting from block matching base cid:
+	//   - fetch all parent blocks
+	//   - with exactly the first parent block, repeat again for it's parents
```
ZenGround0: nit it's --> its

PR #3154, opened 2019-07-30
Rename devnet test to devnet staging

travisperson:
Once this merges we will need to update the wiki for the references there. Should be fairly limited.


travisperson:
> Much better, thanks. I trust the DNS records for staging.kittyhawk.wtf are in place?

They will be updated when we deploy the staging devnet.

Issue #3054, opened 2019-07-11
Maintain peer list for fetching blockchain data


anorth:
### Description
The current bitswap-based chain fetch mechanism does its own peer management. We intend to switch to using graphsync (or maybe #2065) which will need peers from which to fetch to be explicitly specified.

Implement a small object which maintains a list of peers discovered (through bootstrap config and DHT traversal). This should store the results of "hello" protocol exchanges so it knows the head and height (later: protocol version information) about each peer. Decouple a bit from the so that the syncer can distinguish initial chain sync from ongoing (#1105).

This object should expose a method for selecting one or more peers from which to fetch chain data, to be consumed by a graphsync-based fetcher. There should be some functionality for the fetcher to report a peer as not responding and hence suspending or evicting it.

Later, this object will run the secure bootstrapping mechanism of discovering many random peers and seeking a majority agreement on head CID before syncing a chain (#2674). Right now, we just rely on configured trusted bootstrap nodes.

cc @frrist @hannahhoward 

### Acceptance criteria
- An object maintains a list of peers with some state, and can provide peers to a fetcher

### Risks + pitfalls
Libp2p does some implicit things, we don't want the higher level peer list to fall out of sync with peers actually connected.

### Where to begin
Begin with the current hello protocol and understanding existing peer bootstrap and discovery. Decouple from the chain syncer.

anorth:
See also some discussion in #1105. I'm not sure if that means that bootstrapping and peer management need necessarily be linked now, though.

taoshengshi:
Who will provide this kind of peer list ? How to mitigate eclipse attacks to new comers and DDoS attacks to these peers ?

anorth:
That's a problem for #2674, which we are not addressing yet (but will soon).

taoshengshi:
@anorth clear, thanks!

anorth:
Done in #3163, #3184

PR #3163, opened 2019-07-31
Peer Tracker #3054

ZenGround0:
I changed the name from "manager" to "tracker" because our first draft of this is a passive listener to the actual management done by libp2p and the hello protocol.

- [x] peer tracking object & tests (commit 1)
- [x] integration into hello protocol and node's networking stack & tests (commit 2)

### DIFF
--- a/net/peer_tracker.go
+++ b/net/peer_tracker.go
@@ -1,0 +1,81 @@
+package net
+
+import (
+	"sync"
+
+	logging "github.com/ipfs/go-log"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+var logPeerTracker = logging.Logger("peer-tracker")
+
+// PeerTracker is used to record a subset of peers. Its methods are thread safe.
+// It is designed to plug directly into libp2p disconnect notifications to
+// automatically register dropped connections.
+type PeerTracker struct {
+	// mu protects peers
+	mu sync.RWMutex
+	// peers maps stringified peer.IDs to info about their chains
+	peers map[string]*types.ChainInfo
+}
+
+// NewPeerTracker creates a peer tracker.
+func NewPeerTracker() *PeerTracker {
+	return &PeerTracker{
+		peers: make(map[string]*types.ChainInfo),
+	}
+}
+
+// Track adds information about a given peer.ID
+func (tracker *PeerTracker) Track(ci *types.ChainInfo) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	pidKey := peer.IDB58Encode(ci.Peer)
+	if _, tracking := tracker.peers[pidKey]; tracking {
+		logPeerTracker.Warningf("unexpected duplicate track on peer: %s", pidKey)
+	}
+	tracker.peers[pidKey] = ci
+}
+
+// List returns the chain info of the currently tracked peers.  The info
+// tracked by the tracker can change arbitrarily after this is called -- there
+// is no guarantee that the peers returned will be tracked when they are used
+// by the caller and no guarantee that the chain info is up to date.
+func (tracker *PeerTracker) List() []*types.ChainInfo {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	var tracked []*types.ChainInfo
+	for _, ci := range tracker.peers {
+		tracked = append(tracked, ci)
+	}
+	out := make([]*types.ChainInfo, len(tracked))
+	copy(out, tracked)
+	return out
+}
+
+// Remove removes a peer ID from the tracker.
+func (tracker *PeerTracker) Remove(pid peer.ID) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	pidKey := peer.IDB58Encode(pid)
+	if _, tracking := tracker.peers[pidKey]; tracking {
+		delete(tracker.peers, pidKey)
+	}
+}
+
+// TrackerRegisterDisconnect registers a tracker remove operation as a libp2p
+// "Disconnected" network event callback.
+func TrackerRegisterDisconnect(ntwk network.Network, tracker *PeerTracker) {
+	notifee := &network.NotifyBundle{}
+	notifee.DisconnectedF = func(network network.Network, conn network.Conn) {
+		pid := conn.RemotePeer()
+		tracker.Remove(pid)
+	}
+	ntwk.Notify(notifee)
+}
--- a/net/peer_tracker_test.go
+++ b/net/peer_tracker_test.go
@@ -1,0 +1,110 @@
+package net_test
+
+import (
+	"context"
+	"sort"
+	"testing"
+
+	"github.com/libp2p/go-libp2p/p2p/net/mock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-filecoin/net"
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+func TestPeerTrackerTracks(t *testing.T) {
+	tf.UnitTest(t)
+
+	tracker := net.NewPeerTracker()
+	pid0 := th.RequireIntPeerID(t, 0)
+	pid1 := th.RequireIntPeerID(t, 1)
+	pid3 := th.RequireIntPeerID(t, 3)
+	pid7 := th.RequireIntPeerID(t, 7)
+
+	ci0 := types.NewChainInfo(pid0, types.NewTipSetKey(types.SomeCid()), 6)
+	ci1 := types.NewChainInfo(pid1, types.NewTipSetKey(), 0)
+	ci3 := types.NewChainInfo(pid3, types.NewTipSetKey(), 0)
+	ci7 := types.NewChainInfo(pid7, types.NewTipSetKey(), 0)
+
+	tracker.Track(ci0)
+	tracker.Track(ci1)
+	tracker.Track(ci3)
+	tracker.Track(ci7)
+
+	tracked := tracker.List()
+	sort.Sort(types.CISlice(tracked))
+	expected := []*types.ChainInfo{ci0, ci1, ci3, ci7}
+	sort.Sort(types.CISlice(expected))
+	assert.Equal(t, expected, tracked)
+}
+
+func TestPeerTrackerRemove(t *testing.T) {
+	tf.UnitTest(t)
+
+	tracker := net.NewPeerTracker()
+	pid0 := th.RequireIntPeerID(t, 0)
+	pid1 := th.RequireIntPeerID(t, 1)
+	pid3 := th.RequireIntPeerID(t, 3)
+	pid7 := th.RequireIntPeerID(t, 7)
+
+	ci0 := types.NewChainInfo(pid0, types.NewTipSetKey(types.SomeCid()), 6)
+	ci1 := types.NewChainInfo(pid1, types.NewTipSetKey(), 0)
+	ci3 := types.NewChainInfo(pid3, types.NewTipSetKey(), 0)
+	ci7 := types.NewChainInfo(pid7, types.NewTipSetKey(), 0)
+
+	tracker.Track(ci0)
+	tracker.Track(ci1)
+	tracker.Track(ci3)
+	tracker.Track(ci7)
+
+	tracker.Remove(pid1)
+	tracker.Remove(pid3)
+	tracker.Remove(pid7)
+
+	tracked := tracker.List()
+	expected := []*types.ChainInfo{ci0}
+	assert.Equal(t, expected, tracked)
+}
+
+func TestPeerTrackerNetworkDisconnect(t *testing.T) {
+	tf.UnitTest(t)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	mn, err := mocknet.FullMeshConnected(ctx, 4)
+	require.NoError(t, err)
+
+	self := mn.Hosts()[0]
+	a := mn.Hosts()[1]
+	b := mn.Hosts()[2]
+	c := mn.Hosts()[3]
+
+	selfID := self.ID()
+	aID := a.ID()
+	bID := b.ID()
+	cID := c.ID()
+
+	aCI := types.NewChainInfo(aID, types.NewTipSetKey(), 0)
+	bCI := types.NewChainInfo(bID, types.NewTipSetKey(), 0)
+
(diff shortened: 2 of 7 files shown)


types/chain_info.go
```diff
@@ -0,0 +1,33 @@
+package types
+
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// ChainInfo is used to track metadata about a peer and its chain.
+type ChainInfo struct {
```
frrist: Now we are on the same page!! In #3167 I have defined a similar type [here](https://github.com/filecoin-project/go-filecoin/pull/3167/files#diff-1f94e63da3661a7cf4c493bbbb999c86R307), once you land this I'll rebase my changes on it.

net/peer_tracker.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Track adds information about a given peer.ID
+func (tracker *PeerTracker) Track(pid peer.ID) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	pidKey := peer.IDB58Encode(pid)
+	if _, tracking := tracker.peers[pidKey]; tracking {
+		log.Warningf("unexpected duplicate track on peer: %s", pid)
```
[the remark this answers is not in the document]
ZenGround0: A duplicate hello without a disconnection shouldn't happen so I don't want to try to handle it like it is an expected thing.

In the unlikely event this does end up happening we should assume the remote node is reporting its current head in the latest invocation and not try to keep around stale state.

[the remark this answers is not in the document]
ZenGround0: ^ Contradicting myself a little.  Perhaps we should error explicitly if its a dup?  Waddayou think?

net/peer_tracker.go
```diff
@@ -0,0 +1,67 @@
...
+		peers: make(map[string]*types.ChainInfo),
+	}
+}
+
+// Track adds information about a given peer.ID
+func (tracker *PeerTracker) Track(ci *types.ChainInfo) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	pidKey := peer.IDB58Encode(ci.Peer)
```
frrist: any reason we can't use `ci.Peer.ShortString()`?

net/peer_tracker.go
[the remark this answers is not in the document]
frrist: > In the unlikely event this does end up happening we should assume the remote node is reporting its current head in the latest invocation and not try to keep around stale state.

Agreed, I have redacted my previous comment.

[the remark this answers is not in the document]
frrist: Although if the PeerTracker is used to track peers it hears about outside the hello protocol this may need rethinking -- hard to tell if that will be the case now though. Looks good.

testhelpers/net.go
```diff
@@ -122,6 +125,17 @@ func RandPeerID() (peer.ID, error) {
 	return peer.ID(h), nil
 }
 
+// RequireIntPeerID takes in an integer and creates a unique peer id for it.
+func RequireIntPeerID(t *testing.T, i int64) peer.ID {
```
frrist: I would prefer using this pattern for peerId generation, your call.
https://github.com/filecoin-project/go-filecoin/blob/72652f1c1336d091d08252899b0ff5872dd5a459/address/testing.go#L7-L20

net/peer_tracker.go
ZenGround0: Just trying to avoid collisions, are the guarantees for `ShortString()` pretty good?

protocol/hello/hello.go
```diff
@@ -112,7 +119,7 @@ func (h *Handler) handleNewStream(s net.Stream) {
 	}
 }
 
-// ErrBadGenesis is the error returned when a mismatch in genesis blocks happens.
+// ErrBadGenesis is the error return ed when a mismatch in genesis blocks happens.
```
frrist: typo?

net/peer_tracker.go
```diff
@@ -0,0 +1,67 @@
...
+		logPeerTracker.Warningf("unexpected duplicate track on peer: %s", pidKey)
+	}
+	tracker.peers[pidKey] = ci
+}
+
+// List returns the chain info of the currently tracked peers.  The info
+// tracked by the tracker can change arbitrarily after this is called -- there
+// is no guarantee that the peers returned will be tracked when they are used
+// by the caller and no guarantee that the chain info is up to date.
+func (tracker *PeerTracker) List() []*types.ChainInfo {
```
frrist: We need a lock here

net/peer_tracker_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	tracker.Track(ci0)
+	tracker.Track(ci1)
+	tracker.Track(ci3)
+	tracker.Track(ci7)
+
+	tracked := tracker.List()
+	sort.Sort(types.CISlice(tracked))
+	expected := []*types.ChainInfo{ci0, ci1, ci3, ci7}
+	sort.Sort(types.CISlice(expected))
```
frrist: If we only implemented the sorting interface on CISlice for this test I think we can get away with `assert.Equals`, what do you think?

net/peer_tracker_test.go
```diff
@@ -0,0 +1,85 @@
...
+	assert.Equal(t, expected, tracked)
+}
+
+func TestPeerTrackerDisconnect(t *testing.T) {
+	tf.UnitTest(t)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
frrist: not sure this is needed, but I haven't used mocknet before.

node/node.go
```diff
@@ -156,6 +157,9 @@ type Node struct {
 	// SectorBuilder is used by the miner to fill and seal sectors.
 	sectorBuilder sectorbuilder.SectorBuilder
 
+	// PeerTracker maintains a list of peers good for fetching.
+	PeerTracker *net.PeerTracker
```
frrist: Can we make this an interface instead? I think it should make changes easier down the road.

protocol/hello/hello.go
```diff
@@ -33,13 +33,17 @@ var log = logging.Logger("/fil/hello")
 
 // Message is the data structure of a single message in the hello protocol.
 type Message struct {
-	HeaviestTipSetCids   []cid.Cid
+	HeaviestTipSetCids   types.TipSetKey
```
frrist: :+1: this brings us inline with the [hello protocol definition](https://github.com/filecoin-project/specs/blob/94a3cbbc6c1b3f848957730f5fff860ec8db1907/network-protocols.md#hello-handshake), nice!

types/chain_info.go
```diff
@@ -0,0 +1,35 @@
...
+	Peer   peer.ID
+	Head   TipSetKey
+	Height uint64
```
frrist: Mentioned in slack, a `Trusted bool` filed on this would ease conflicts later (#3167), can just hardcode it to `true` as nothing uses yet.

types/chain_info.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Len returns the number of chain infos in the slice.
+func (cis CISlice) Len() int { return len(cis) }
+
+// Swap swaps chain infos.
+func (cis CISlice) Swap(i, j int) { cis[i], cis[j] = cis[j], cis[i] }
+
+// Less compares chain infos on peer ID.  There should only ever be one chain
+// info per peer in a CISlice.
+func (cis CISlice) Less(i, j int) bool { return string(cis[i].Peer) < string(cis[j].Peer) }
```
frrist: If we only added these for testing I would prefer to find a different way to make the assertion.

net/peer_tracker.go
frrist: Ehh good point (I was thinking of `KeyString` from go-cid), you could just call `ci.Peer.String()`, it's the same `IDB58Encode` call under the hood, just more familiar imo -- your call.

net/peer_tracker.go
ZenGround0: lol wot?  How did I miss this? thx

net/peer_tracker_test.go
ZenGround0: Does `assert.Equals` sort on its own?

net/peer_tracker_test.go
ZenGround0: It's not strictly needed to test the tracker API but it would be annoying to get connections without it.

It's nice to have here because it ensures that the callback signature fits into the libp2p callback system (and that the callingback works)

node/node.go
ZenGround0: I want to follow the dominant pattern which is still concrete structs.  Are you talking about refactoring node construction for better mocking?  I can see that point of view but don't know about concrete plans to make that refactor.  If we do it should be easy enough to convert this one to an iface along with the 20 other concrete structs if we do that.

types/chain_info.go
ZenGround0: Is your thinking that this is unnecessary for such small test code?  I can see that but argue that it is not much code, properly isolated, a standard go / PL pattern (I lifted it from peer ids), and likely to be used in tests (maybe logging) down the line.

net/peer_tracker_test.go
frrist: opps, I meant to type `assert.Contains`

net/peer_tracker.go
ZenGround0: Fun fact I started off using `String()` and `IDFromString()` and got burned for over an hour because it turns out those two are not inverses.

node/node.go
frrist: I don't view that as the dominant pattern here, a majority of these fields are interfaces (over half). But as you have states we don't have a concrete plan for a node refactor. I'll leave this up to you.

net/peer_tracker.go
```diff
@@ -0,0 +1,70 @@
...
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+
+	var out []*types.ChainInfo
+	for _, ci := range tracker.peers {
+		out = append(out, ci)
```
anorth: I think this should clone the ChainInfo too since it's mutable. Later improvements can reduce the (small) cost by not getting the whole list every time if it's not needed.

net/peer_tracker_test.go
```diff
@@ -0,0 +1,85 @@
...
+	tracker := net.NewPeerTracker()
+	pid0 := th.RequireIntPeerID(t, 0)
+	pid1 := th.RequireIntPeerID(t, 1)
+	pid3 := th.RequireIntPeerID(t, 3)
+	pid7 := th.RequireIntPeerID(t, 7)
+
+	ci0 := types.NewChainInfo(pid0, types.NewTipSetKey(), 0)
+	ci1 := types.NewChainInfo(pid1, types.NewTipSetKey(), 0)
+	ci3 := types.NewChainInfo(pid3, types.NewTipSetKey(), 0)
+	ci7 := types.NewChainInfo(pid7, types.NewTipSetKey(), 0)
```
anorth: Use at least one non-empty tipset key. At present, the tracker could ignore them and still pass tests.

net/peer_tracker.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return out
+}
+
+// OnDisconnect handles network disconnection events for the tracker. It
+// removes a peer ID from the tracker when the peer is disconnected.
+func (tracker *PeerTracker) OnDisconnect(network network.Network, conn network.Conn) {
```
anorth: Remove unused `network` parameter.

I would slightly prefer the tracker's interface to deal only in ChainInfo, no direct reference to the `network` package, and for this to be simply `Remove(peerId)`. Then a hookup function (not a member method) would connect the network notifications to this `Remove` method. This would then be unit-testable without a mock network (the hook-up of which isn't actually covered by the current tests anyway).

If you do this, rename `Track` to `Add`, since the responsibility of hooking up notifications doesn't really lie within this object.

node/node.go
anorth: I think concrete types are ok here, for now, but the various components that get references to these objects should primarily get interfaces. This could change after we clean things up a lot here.

node/node.go
```diff
@@ -536,13 +544,17 @@ func (node *Node) Start(ctx context.Context) error {
 		// Start bootstrapper.
 		node.Bootstrapper.Start(context.Background())
 
+		// Register peer tracker disconnect function with network.
+		notifee := &p2pnet.NotifyBundle{}
+		notifee.DisconnectedF = node.PeerTracker.OnDisconnect
+		node.host.Network().Notify(notifee)
```
anorth: If this were in a free function somewhere, you could test it with mocknet.

node/node.go
```diff
@@ -560,7 +572,7 @@ func (node *Node) Start(ctx context.Context) error {
 			// See https://github.com/filecoin-project/go-filecoin/issues/1105
 			chainSynced.Done()
 		}
-		node.HelloSvc = hello.New(node.Host(), node.ChainReader.GenesisCid(), syncCallBack, node.PorcelainAPI.ChainHead, node.Repo.Config().Net, flags.Commit)
+		node.HelloSvc = hello.New(node.Host(), node.ChainReader.GenesisCid(), node.PeerTracker, syncCallBack, node.PorcelainAPI.ChainHead, node.Repo.Config().Net, flags.Commit)
```
anorth: Nearly-blocking: instead of coupling hello with the tracker, add a peer to the tracker in the existing syncCallBack. 
- Perhaps change the callback to use the new ChainInfo type
- Perhaps rename that callback to `helloCallback` so that hello is less prescriptive.

protocol/hello/hello_test.go
```diff
@@ -230,4 +268,7 @@ func TestHelloMultiBlock(t *testing.T) {
 
 	msc1.AssertExpectations(t)
 	msc2.AssertExpectations(t)
+
+	assert.True(t, trackerA.contains(b.ID()))
+	assert.True(t, trackerB.contains(a.ID()))
```
anorth: If refactored as I suggest, I expect this diff to disappear. It's an issue for integration testing (probably sufficiently covered by daemon tests for now), not for hello unit testing.

types/chain_info.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// ChainInfo is used to track metadata about a peer and its chain.
+type ChainInfo struct {
+	Peer    peer.ID
+	Head    TipSetKey
+	Height  uint64
+	Trusted bool
```
anorth: I don't think the `hello` to a single node should indicate trustworthiness. This is a step we will have to revert when we do a more principled bootstrapping. The bootstrapper/peertracker will compute a trusted head from *many* peers (though it's ok in the interim to "trust" all of them).

In @frrist's downstream changes, I think it's fine that this results in two arguments to the syncer's HandleNewTipSet.

ZenGround0: Got it.  Since I did this to make @frrist's rebase easier I'll leave it up to him in #3167

frrist: I'm good with taking it out of `ChainInfo`, conflicts should be manageable.

PR #3167, opened 2019-07-31
Replace syncing modes with trusted boolean value

frrist:
### What
This PR can be reviewed in 2 steps:
- cd16520 adds a type called `PeerHeadInfo` that contains information about a peer, its current head state and its trustworthiness. `HandleNewTipset` is then modified to accept this type. This is a mechanical change that doesn't effect the code logic.
- bf0e2c3 does a couple things:
    - replaces the syncers `syncMode` with the `Trusted` field in `PeerHeadInfo`. The syncers logic is changed to check if a TipSet exceeds finality if the peer is untrusted.
    - modifies `exceedsFinalityLimit` to accept 2 Uint64's, the current head and the new Head we get from `PeerHeadInfo`. This change simplifies the Fethers callback to a single case.
    - removes the `fetchCtx` as that was mode dependent.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -78,12 +78,12 @@ type syncStateEvaluator interface {
 // tipset in the incoming chain, and assumptions regarding the existence of
 // grandparent state in the store.
 type Syncer struct {
-	// This mutex ensures at most one call to HandleNewTipset executes at
+	// This mutex ensures at most one call to HandleNewTipSet executes at
 	// any time.  This is important because at least two sections of the
 	// code otherwise have races:
 	// 1. syncOne assumes that chainStore.Head() does not change when
 	// comparing tipset weights and updating the store
-	// 2. HandleNewTipset assumes that calls to widen and then syncOne
+	// 2. HandleNewTipSet assumes that calls to widen and then syncOne
 	// are not run concurrently with other calls to widen to ensure
 	// that the syncer always finds the heaviest existing tipset.
 	mu sync.Mutex
@@ -289,13 +289,13 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
 	return wts, nil
 }
 
-// HandleNewTipset extends the Syncer's chain store with the given tipset if they
+// HandleNewTipSet extends the Syncer's chain store with the given tipset if they
 // represent a valid extension. It limits the length of new chains it will
 // attempt to validate and caches invalid blocks it has encountered to
 // help prevent DOS.
-func (syncer *Syncer) HandleNewTipset(ctx context.Context, ci *types.ChainInfo, trusted bool) (err error) {
+func (syncer *Syncer) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo, trusted bool) (err error) {
 	logSyncer.Debugf("Begin fetch and sync of chain with head %v", ci.Head)
-	ctx, span := trace.StartSpan(ctx, "Syncer.HandleNewTipset")
+	ctx, span := trace.StartSpan(ctx, "Syncer.HandleNewTipSet")
 	span.AddAttributes(trace.StringAttribute("tipset", ci.Head.String()))
 	defer tracing.AddErrorEndSpan(ctx, span, &err)
 
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
-	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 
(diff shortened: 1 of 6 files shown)


chain/syncer.go
```diff
@@ -324,14 +291,34 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
 	return wts, nil
 }
 
+// NewPeerHeadInfo accepts a peer, its current head state and its trustworthiness
+// and returns a PeerHeadInfo.
+func NewPeerHeadInfo(head types.TipSetKey, height types.Uint64, from peer.ID, trusted bool) *PeerHeadInfo {
```
frrist: This Method and its corresponding type need a better home, looking for suggestions, else will leave it here.

frrist: On second thought, going to wait for #3163 to land and rebase. `types.ChainInfo` will then replace this.

anorth: As I noted there, I expect the `Trusted` field to be removed from that structure, but that it's perfectly fine for that to come as an explicit argument to `HandleNewTipSet`.

chain/syncer.go
```diff
@@ -324,14 +290,34 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
...
+	Trusted bool
+}
+
 // HandleNewTipset extends the Syncer's chain store with the given tipset if they
 // represent a valid extension. It limits the length of new chains it will
 // attempt to validate and caches invalid blocks it has encountered to
 // help prevent DOS.
-func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey, from peer.ID) (err error) {
-	logSyncer.Debugf("Begin fetch and sync of chain with head %v", tsKey)
+func (syncer *Syncer) HandleNewTipset(ctx context.Context, ph *PeerHeadInfo) (err error) {
```
anorth: While you're here, consider renaming to `HandleNewTipSet` to match project capitalization convention.

chain/syncer.go
```diff
@@ -434,29 +395,14 @@ func (syncer *Syncer) HandleNewTipset(ctx context.Context, tsKey types.TipSetKey
...
-	}
-	finalityHeight := types.NewBlockHeight(curHeight).Add(types.NewBlockHeight(uint64(FinalityLimit)))
-	newHeight, err := newHead.Height()
-	if err != nil {
-		return false, err
-	}
-	return types.NewBlockHeight(newHeight).GreaterThan(finalityHeight), nil
+func (syncer *Syncer) exceedsFinalityLimit(curHeight, newHeight types.Uint64) bool {
+	finalityHeight := types.NewBlockHeight(uint64(curHeight)).Add(types.NewBlockHeight(uint64(FinalityLimit)))
+	return types.NewBlockHeight(uint64(newHeight)).GreaterThan(finalityHeight)
```
anorth: Just use `uint64` arithmetic. This doesn't need to be encoded on chain and 64 bits ought to be enough for anyone's chain length.

chain/syncer_integration_test.go
```diff
@@ -104,11 +108,16 @@ func TestLoadFork(t *testing.T) {
 	// Test that the syncer can't sync a block chained from on the right (originally shorter) chain
 	// without getting old blocks from network. i.e. the store index has been trimmed
 	// of non-heaviest chain blocks.
-	err = offlineSyncer.HandleNewTipset(ctx, newRight.Key(), "")
+	nrH, err := newRight.Height()
+	require.NoError(t, err)
+
+	err = offlineSyncer.HandleNewTipset(ctx, chain.NewPeerHeadInfo(newRight.Key(), types.Uint64(nrH), "", true))
```
anorth: Consider reducing the repeated verbosity by adding a `headFromTip(types.TipSet)` method to do the type wrangling just once

chain/syncer_test.go
```diff
@@ -25,7 +25,9 @@ func TestOneBlock(t *testing.T) {
 	genesis := builder.RequireTipSet(store.GetHead())
 
 	t1 := builder.AppendOn(genesis, 1)
-	assert.NoError(t, syncer.HandleNewTipset(ctx, t1.Key(), peer.ID("")))
+	t1H, err := t1.Height()
+	require.NoError(t, err)
+	assert.NoError(t, syncer.HandleNewTipset(ctx, chain.NewPeerHeadInfo(t1.Key(), types.Uint64(t1H), peer.ID(""), true)))
```
anorth: 😢 definitely refactor so these are back to one-liners.

node/node.go
```diff
@@ -543,8 +543,7 @@ func (node *Node) Start(ctx context.Context) error {
 		// Start up 'hello' handshake service
 		syncCallBack := func(pid peer.ID, cids []cid.Cid, height uint64) {
 			cidSet := types.NewTipSetKey(cids...)
-			err := node.Syncer.HandleNewTipset(context.Background(), cidSet, pid)
-			if err != nil {
+			if err := node.Syncer.HandleNewTipset(context.Background(), chain.NewPeerHeadInfo(cidSet, types.Uint64(height), pid, true)); err != nil {
```
anorth: Add an explicit named variable (if the linter will let you) and prominent comment about this innocuous boolean stating that this indicates we are trusting any node, and this should be replaced with [link to issue about bootstrapping].

chain/syncer_integration_test.go
```diff
@@ -104,11 +108,16 @@ func TestLoadFork(t *testing.T) {
 	// Test that the syncer can't sync a block chained from on the right (originally shorter) chain
 	// without getting old blocks from network. i.e. the store index has been trimmed
 	// of non-heaviest chain blocks.
-	err = offlineSyncer.HandleNewTipset(ctx, newRight.Key(), "")
+	nrH, err := newRight.Height()
+	require.NoError(t, err)
```
ZenGround0: A `RequireHeight(tipset)` function would help compact this.

chain/syncer_test.go
frrist: :pray: will do in a follow on, this hurts.

PR #3168, opened 2019-07-31
Prep for breaking messages out of blocks

### LINKED ISSUE #3165: Prepare for message separation without breaking block format
### Description
[This document](https://docs.google.com/document/d/1xOHUeM-svZoE3qQtOPCx1U96EQ6CuKsfB2GJYOHOk18/edit?usp=sharing) sketches out a plan for how to separate messages out of blocks.  This is going to be a large amount of code: doing everything in one PR would make for a painful review.  However as soon as we flip the switch and break the block format we need to land everything at once to keep master working.

This story captures all the work that can be done for the above plan without breaking the block format.  This way reviewers can digest code in more reasonable chunks

### Acceptance criteria
(1) Add a message/receipt collection type for grouping and serializing messages
(2) Implement the message collection store 
(3) Hook up the wiring to get a message collection store to every place that we foresee needing it in order to read / write messages
(4) The chain builder will implement the message collection store interface

not in scope: reading or writing to/from the store, modifying the internal semantics of the chain builder's message handling, modifying the block structure

### Risks + pitfalls
- We will likely miss unforeseen issues that will only arise when we change the block structure
- Review will require investment in understanding the full plan in order to make sense of the preliminary changes
- The code will have a lot of unused message collection readers floating around (doesn't seem like a problem, just FYI)

### Where to begin
The linked document

ZenGround0:
Closes #3165 

- [x] message collection store and basic tests
- [x] pipe message collection stores to all the needed places 
- [x] update the chain builder so that it implements the MessageCollectionStore interface

### DIFF
--- a/chain/message_store.go
+++ b/chain/message_store.go
@@ -1,0 +1,72 @@
+package chain
+
+import (
+	"context"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// MessageReader is an interface exposing the read methods of the
+// MessageStore.
+type MessageReader interface {
+	LoadMessages(context.Context, cid.Cid) ([]*types.SignedMessage, error)
+	LoadReceipts(context.Context, cid.Cid) ([]*types.MessageReceipt, error)
+}
+
+// MessageWriter is an interface exposing the write methods of the
+// MessageStore.
+type MessageWriter interface {
+	StoreMessages(context.Context, []*types.SignedMessage) (cid.Cid, error)
+	StoreReceipts(context.Context, []*types.MessageReceipt) (cid.Cid, error)
+}
+
+// MessageStore stores and loads collections of signed messages and receipts.
+type MessageStore struct {
+	ipldStore *hamt.CborIpldStore
+}
+
+// NewMessageStore creates and returns a new store
+func NewMessageStore(cst *hamt.CborIpldStore) *MessageStore {
+	return &MessageStore{
+		ipldStore: cst,
+	}
+}
+
+// LoadMessages loads the signed messages in the collection with cid c from ipld
+// storage.
+func (ms *MessageStore) LoadMessages(ctx context.Context, c cid.Cid) ([]*types.SignedMessage, error) {
+	// TODO #1324 message collection shouldn't be a slice
+	var out []*types.SignedMessage
+	err := ms.ipldStore.Get(ctx, c, &out)
+	return out, err
+}
+
+// StoreMessages puts the input signed messages to a collection and then writes
+// this collection to ipld storage.  The cid of the collection is returned.
+func (ms *MessageStore) StoreMessages(ctx context.Context, msgs []*types.SignedMessage) (cid.Cid, error) {
+	// For now the collection is just a slice (cbor array)
+	// TODO #1324 put these messages in a merkelized collection
+	return ms.ipldStore.Put(ctx, msgs)
+}
+
+// LoadReceipts loads the signed messages in the collection with cid c from ipld
+// storage and returns the slice implied by the collection
+func (ms *MessageStore) LoadReceipts(ctx context.Context, c cid.Cid) ([]*types.MessageReceipt, error) {
+	var out []*types.MessageReceipt
+	err := ms.ipldStore.Get(ctx, c, &out)
+	if err != nil {
+		return nil, err
+	}
+	return out, nil
+}
+
+// StoreReceipts puts the input signed messages to a collection and then writes
+// this collection to ipld storage.  The cid of the collection is returned.
+func (ms *MessageStore) StoreReceipts(ctx context.Context, receipts []*types.MessageReceipt) (cid.Cid, error) {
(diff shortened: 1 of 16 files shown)


chain/message_collection_store.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+
+	"github.com/ipfs/go-hamt-ipld"
+	"github.com/ipfs/go-cid"	
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+type MessageCollectionStore struct {
```
anorth: I suggest just `MessageStore` - we'll very likely want to add access for individual messages later.

types/testing_messages.go
```diff
@@ -58,3 +58,19 @@ func (mm *MessageMaker) NewSignedMessage(from address.Address, nonce uint64) *Si
...
+func NewReceiptMaker() *ReceiptMaker {
+	return &ReceiptMaker{0}
+}
+
+// NewReceipt creates a new distinct receipt.
+func (rm *ReceiptMaker) NewReceipt() *MessageReceipt {
+	seq := rm.seq
+	rm.seq++
+	return &MessageReceipt{Return: [][]byte{[]byte(fmt.Sprintf("%d",seq))}}
+}
```
anorth: Put this in the single test file that uses it until its needed elsewhere. `types` is a dumping ground already.

ZenGround0:
Note that this does not add the MessageStore to the Syncer or the consensus module interface.  I'm thinking of doing that plumbing in another quick prep PR.

chain/message_store.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+// LoadMessages loads the signed messages in the collection with cid c from ipld
+// storage.
+func (ms *MessageStore) LoadMessages(ctx context.Context, c cid.Cid) ([]*types.SignedMessage, error) {
+	// TODO #1324 message collection shouldn't be a slice
+	var out []*types.SignedMessage
+	err := ms.ipldStore.Get(ctx, c, &out)
+	if err != nil {
```
anorth: Would it be too radical to suggest `return out, err`?

chain/testing.go
```diff
@@ -31,7 +31,9 @@ type Builder struct {
 	stateBuilder StateBuilder
 	seq          uint64 // For unique tickets
 
-	blocks map[cid.Cid]*types.Block
+	blocks   map[cid.Cid]*types.Block
+	messages map[cid.Cid][]*types.SignedMessage
+	receipts map[cid.Cid][]*types.MessageReceipt
```
anorth: Any messages/receipts created in the `build` function below should be stored here at the same time as the block is stored. That will need a little further refactoring when we split the messages out.

I would prefer that the `StoreXxx` methods below did not exist, in favour of this mechanism. For things that want a (fake) message store but will not be using messages attached to blocks built by this builder, just use the MessageStore, or a distinct FakeMessageStore with these maps.

mining/worker_test.go
```diff
@@ -70,7 +73,9 @@ func Test_Mine(t *testing.T) {
 			MessageSource: pool,
 			Processor:     th.NewTestProcessor(),
 			PowerTable:    mining.NewTestPowerTableView(1),
-			Blockstore:    bs}, CreatePoSTFunc)
+			Blockstore:    bs,
+			MessageStore:  messages,
```
anorth: Unlike `policy_test`, I think this one is abusing the builder a bit.

chain/testing.go
ZenGround0: I also would prefer it this way.  If anyone has ideas I could use some help.

ZenGround0: Current idea: private fields on block builder share messages and receipts cache.  The chain builder passes them in during builder construction.  `AddMessage` / `AddReceipt` callbacks defined in chain add messages / receipts directly.  Out of package callers can't mess up setup because builder has control of which caches to share.  I think this should work pretty well, speculatively including it in this PR.

chain/message_store.go
ZenGround0: :trollface:

chain/testing.go
```diff
@@ -31,7 +31,9 @@ type Builder struct {
...
 	// Cache of the state root CID computed for each tipset key.
 	tipStateCids map[string]cid.Cid
 }
```
frrist: Blocking (can't comment on the exact line): 
Since the `Builder` implements `MessageReader` we should:
```go
var _ MessageReader = (*Builder)(nil)
```

core/policy_test.go
```diff
@@ -73,7 +75,10 @@ func TestMessageQueuePolicy(t *testing.T) {
 			b.IncHeight(103)
 		})
 		b1 := blocks.BuildOn(root, func(b *chain.BlockBuilder) {
-			b.AddMessage(msgs[0], &types.MessageReceipt{})
+			b.AddMessages(
+				[]*types.SignedMessage{msgs[0]},
+				types.EmptyReceipts(1),
+			)
```
anorth: Even though it's clumsy code in the builder, it would be better if the old `b.AddMessage` interface were retained for users to avoid the noise of the slice literal construction and the possibility of non-parallel slices.

Issue #3144, opened 2019-07-29
Make `miner power` output JSON not a string


shannonwells:
## Description
As noted in [Pull 3136](https://github.com/filecoin-project/go-filecoin/pull/3136), the `miner power` command outputs "N / M", miner power / total power, which was implemented [as originally designed](https://github.com/filecoin-project/go-filecoin/issues/1128), however we would like to have the CLI output JSON results which can be more easily parsed.

## Acceptance criteria
`miner power <mineraddr>` outputs two big.Ints, one for miner power and one for total power (of the network).  This is serialized as:
{
  minerPower: N
  totalPower: N
}

travisperson:
To add some details to this, there can be a text encoder to provide the `n / m` output, but the emitted value by the `Run` func should be a structure and not the formatted string.

frrist:
closed via #3169

Issue #3159, opened 2019-07-30
Add get proving period to CLI


shannonwells:
### Description
Provide a way to get proving period start and end for a given miner

### Acceptance criteria
`miner proving-period <mineraddr>` returns a JSON struct that shows the start and end of the current proving period in blocks

There is already a GetProvingPeriod export on miner actor

frrist:
closed via  #3169

PR #3169, opened 2019-07-31
mining status command shows proving period

frrist:
### What
This PR addresses #3159, #3144 and also extends the `mining status` command to contain the proving period output in addition to the output that may be had from the `miner power/collateral/owner` commands.

#### Example Output:
Miner that is mining
```
$ go-filecoin mining status
Mining Status
Active:     true
Address:    t2udhmwzokf4hq7oe4zj4g3faah7oz4dnbxtxaqoy
Owner:      t1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy
Collateral: 0.001
Power:      1024 / 1024

Proving Period
Start: 1000
End:   2000
Set:   [1]
```
As JSON:
```
{"active":true,"minerAddress":"t2udhmwzokf4hq7oe4zj4g3faah7oz4dnbxtxaqoy","owner":"t1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy","collateral":"0.001","provingPeriod":{"start":1000,"end":2000,"set":{"1":{"CommD":[177,13,57,70,81,133,15,212,161,120,137,46,226,133,236,225,81,20,85,120,8,117,214,78,226,211,208,208,222,107,248,249],"CommR":[180,76,232,95,240,68,198,177,248,59,142,136,59,191,133,122,171,153,197,178,82,199,66,156,50,243,168,174,183,158,248,86],"CommRStar":[246,89,193,143,13,206,204,119,199,94,122,129,191,222,39,95,103,207,226,66,207,60,195,84,243,237,226,214,190,204,78,163]}}},"minerPower":{"Power":"1024","Total":"1024"}}
```

Miner that is not mining
```
$ go-filecoin mining status
Mining Status
Active:     true
Address:    t2q7fy4aqm4yone7lmttcrz6eget6uhsgdtwp45qa
Owner:      t17k6pqfq5dcu53sc5e7iwd2ogqpz7d4yiqkpl3zi
Collateral: 0
Power:      0 / 1024

Proving Period
Start: 0
End:   0
Set:   []
```
As JSON:
```
{"active":false,"minerAddress":"t2q7fy4aqm4yone7lmttcrz6eget6uhsgdtwp45qa","owner":"t17k6pqfq5dcu53sc5e7iwd2ogqpz7d4yiqkpl3zi","collateral":"0","provingPeriod":{"start":0,"end":0},"minerPower":{"Power":null,"Total":null}}
```


commands/mine_util.go
```diff
@@ -0,0 +1,135 @@
+package commands
```
frrist: If a strong argument can be made for moving these to the Porcelain API I would be open to it. However they only appear to be used by the CLI so here they remain.

commands/miner.go
```diff
@@ -376,17 +377,7 @@ var minerOwnerCmd = &cmds.Command{
 		if err != nil {
 			return err
 		}
-
-		bytes, err := GetPorcelainAPI(env).MessageQuery(
```
frrist: these calls and the below were moved to `mine_util.go` since the `mining` and `miner` use them

commands/miner.go
```diff
@@ -416,37 +407,19 @@ Values will be output as a ratio where the first number is the miner power and s
...
-		if err != nil {
-			return err
-		}
-		total := types.NewBytesAmountFromBytes(bytes[0])
-
-		str := fmt.Sprintf("%s / %s", power, total) // nolint: govet
-		return re.Emit(str)
+		return re.Emit(minerPower)
 	},
+	Type: MinerPowerResult{},
```
frrist: closes #3144

commands/mine_util.go
shannonwells: I thought this was the whole purpose of porcelain: "// API is the porcelain implementation, a set of convenience calls written on the plumbing api, to be used to build user facing features and protocols."

All of these calls are on miner actors, so why not have it be part of miner porcelain?  We already have MinerGetOwnerAddress there.  Unless I'm missing something.

node/node.go
```diff
@@ -773,9 +773,9 @@ func (node *Node) addNewlyMinedBlock(ctx context.Context, b *types.Block) {
 	}
 }
 
-// miningAddress returns the address of the mining actor mining on behalf of
+// MiningAddress returns the address of the mining actor mining on behalf of
 // the node.
-func (node *Node) miningAddress() (address.Address, error) {
+func (node *Node) MiningAddress() (address.Address, error) {
```
shannonwells: so this is just exported for testing, I take it?

tools/fast/action_miner.go
```diff
@@ -67,27 +65,17 @@ func (f *Filecoin) MinerOwner(ctx context.Context, minerAddr address.Address) (a
...
-	var out string
+// MinerPower runs the `miner power` command against the filecoin process
+func (f *Filecoin) MinerPower(ctx context.Context, minerAddr address.Address) (commands.MinerPowerResult, error) {
+	var out commands.MinerPowerResult
 
 	sMinerAddr := minerAddr.String()
 
 	if err := f.RunCmdJSONWithStdin(ctx, nil, &out, "go-filecoin", "miner", "power", sMinerAddr); err != nil {
-		return big.NewInt(0), big.NewInt(0), err
+		return commands.MinerPowerResult{}, err
```
shannonwells: ✅

node/node.go
frrist: Used here as a command:
https://github.com/filecoin-project/go-filecoin/pull/3169/files#diff-642261538295948b5c2a6d350214f8e6R31

Used here: https://github.com/filecoin-project/go-filecoin/pull/3169/files#diff-642261538295948b5c2a6d350214f8e6R89

frrist:
I think you are right @shannonwells, these would probably go better on porcelain, I just didn't put them there because the logic was already in the commands/ package. I'll take a look at moving them to porcelain now.

commands/mine_util.go
anorth: Absolutely move them, there should be no real code in the commands. Recall that we intend to discard this CLI eventually, but still need this functionality. I think it is an error that there are any `messageQuery`'s directly in commands. cc @acruikshank

porcelain/miner.go
```diff
@@ -397,3 +397,105 @@ func MinerGetPeerID(ctx context.Context, plumbing mgpidAPI, minerAddr address.Ad
 	}
 	return pid, nil
 }
+
+// MinerProvingPeriod contains a miners proving period start and end as well
+// as a set of their commitments.
+type MinerProvingPeriod struct {
+	Start types.BlockHeight
+	End   types.BlockHeight
+	Set   map[string]types.Commitments
```
anorth: Since there are two possible things this could be, I think we should be explicit and call this `ProvingSet`. Say "proving set" instead of "commitments" in the comments too.

node/node.go
shannonwells: ah thanks. I looked for it and didn't see it.

PR #3170, opened 2019-08-01
WIP: Add loggers for repo to keep track log (#2443)

### LINKED ISSUE #2443: Log commands issued to a node to keep track of piece→miner mappings
## Motivation
In the near term, we don't intend to implement any discovery mechanism for retrieval. A client must keep track of which miner has stored their files.

The relevant information is embedded in the commands issued to a node by a client accepting a deal. As a cheap way of doing something here, we wish to log all the commands issued to a node to disk, where they can be inspected later by a client who has otherwise failed to keep track.

We probably don't want this log to be subject to log level suppression, so it's probably a distinct path from the usual application logs. We should log to a file somewhere, perhaps in `.filecoin`.

Plain text logs should suffice, but a little structure wouldn't hurt. CSV? JSON?

The first part of this story is proposing a specific mechanism in this issue.

deaswang:
fixes #2443

This is just one proposal, Not finished. look forward Any suggest.

### DIFF
--- a/commands/client.go
+++ b/commands/client.go
@@ -149,6 +149,10 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 		if err != nil {
 			return err
 		}
+		logger, err := GetInspectorAPI(env).Logger("deal.log")
+		if err == nil {
+			logger.Printf("deal: data %s store at miner %s askid %d for %d blocks time", data, miner, askid, duration)
+		}
 
 		return re.Emit(resp)
 	},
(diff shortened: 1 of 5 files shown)


mishmosh:
@deaswang Thanks for all the recent contributions, your pace is awesome :) 

In order to get your PRs reviewed & merged more quickly, could you start describing your proposed solutions in the issue first?

Generally, we follow the principle of "[design before code](https://github.com/filecoin-project/go-filecoin/blob/master/CONTRIBUTING.md#design-before-code)". For big changes, that usually means a separate design doc. For smaller changes, that can be done in the issue itself (example: #3177). This helps everyone agree on direction first, so code review goes smoothly.

Issue #3052, opened 2019-07-11
[bug] Same actor and nonce but different cid 


steven004:
**Describe the bug**
This error has been seen in 0.2.4 and 0.3.0 version several times, when there are staged sectors waiting in line for sealing. 

`[0;37m16:29:28.595 [31mERROR [0;34m      node: [0mfailed to send commitSector message from t1kbn...yny to t2kab...xvi for sector with id 33....26: failed to add message to message pool: validation error adding message to pool: message pool contains message with same actor and nonce but different cid [0;37mnode.go:1110[0m`

And, I talked with some other miners, they got this too.

**To Reproduce**
Do not need to do any special steps, just send asks as fast as you can, and check the output log, then you will see this some time. 

**Expected behavior**
The nonce is increasing always, no same nonces for two messages in one miner's system.

**Version information**
Found in both test-devnet-0.2.2 and test-devnet-0.3.2

**Additional context**
Could there be race condition happened when multiple sectors for seal, especially when multiple workers are running?

anorth:
Yes, I think I can see the potential race in `outbox.go`. 

I would expect you to receive an error from the message queue (not pool), "Invalid nonce %d, expected %d", and this would prevent you from adding the message to the pool. Nevertheless, I'll attempt to repro and fix.

anorth:
Here's what I think is happening: it's not as simple as a race in the outbox.

The inbound message pool and outbound message queue are distinct structures. Outbound messages are held in the queue (per-actor) until they are observed in a heaviest tipset, and then forgotten. Nonces for new messages are calculated as the max of the actor's nonce, or one greater than the largest nonce for the actor in the queue.

The message pool stores *all* messages that are candidates for mining into a block, not just those sent by the node itself. They aren't organised into per-actor queues, but do have an index for efficient lookup of address/nonce pairs. Messages are also removed from the pool when observed in a heaviest tipset.

When a re-org occurs (the chain head swaps from one tip to another), messages in the losing chain effectively become "un-mined", and the actors state reverts to a prior state before messages in the new chain are replayed. The new chain/state may be missing some messages that the old chain/state had. Those missing messages are added back to the pool as candidates to be mined, but are *not returned to the outbound queue*. 

When a new message is sent in this state, the nonce will be calculated from actor state, the queue for that actor will be empty, and the calculated nonce will duplicate that of an "un-mined" message that returned to the pool.

What to do?
[...]

PR #3178, opened 2019-08-02
Slashing for unreported storage faults

shannonwells:
Continuing from previous PRs, here the Storage Fault Slasher slashes miners that have not submitted PoSts by the end of the GenerationAttackThreshold grace period.

There are a decent amount of unit tests, but the one integration (FAST) test to check that a late miner really gets slashed takes, IMO too long to leave it for CI.


tools/fast/tests/slashing_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// Not passing, and also too slow to let run in CI
+func TestSlashing(t *testing.T) {
+	tf.IntegrationTest(t)
+	t.SkipNow()
```
shannonwells: This proves that slashing works, but it runs way too long for CI.

consensus/storage_fault_monitor.go
```diff
@@ -0,0 +1,89 @@
...
+	lateMiners, err := abi.Deserialize(res[0], abi.MinerPoStStates)
+	if err != nil {
+		return errors.FaultErrorWrap(err, "deserializing MinerPoStStates failed")
+	}
+
+	lms, ok := lateMiners.Val.(*map[string]uint64)
+	if !ok {
+		return errors.FaultErrorWrapf(err, "expected *map[string]uint64 but got %T", lms)
+	}
+	sfm.log.Debugf("there are %d late miners", len(*lms))
```
acruikshank: It's ok to keep this debug line (or remove it), but it should have a new line under it.

consensus/storage_fault_monitor.go
```diff
@@ -0,0 +1,89 @@
...
+	// Slash late miners.
+	for minerStr, state := range *lms {
+		minerAddr, err := address.NewFromString(minerStr)
+		if err != nil {
+			return errors.FaultErrorWrap(err, "could not create minerAddr string")
+		}
+
+		// send slash message, don't broadcast it, and don't wait for message to appear on chain.
```
acruikshank: Without broadcasting, this isn't really sending. Perhaps: // add slash message to outbox without broadcasting it...

consensus/storage_fault_monitor.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/filecoin-project/go-filecoin/vm/errors"
+)
+
+// monitorPlumbing is an interface for the functionality StorageFaultSlasher needs
+type monitorPlumbing interface {
+	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
+}
+
+// slashingMsgOutbox is the interface for the functionality of Outbox StorageFaultSlasher needs
+type slashingMsgOutbox interface {
```
acruikshank: Why bother to add message to outbox if we're not going to broadcast it? Wouldn't it be just as easy to add it to the inbox? Isn't that all adding all the outbox does aside from broadcasting?

consensus/storage_fault_monitor_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+func TestStorageFaultSlasher_Slash(t *testing.T) {
+	tf.UnitTest(t)
+
+	ctx := context.Background()
+	signer, _ := types.NewMockSignersAndKeyInfo(1)
+
+	t.Run("When no miners, empty LateMiners", func(t *testing.T) {
```
acruikshank: I'm having difficulty seeing what this test is testing. It looks like its just testing that Slash doesn't error when slashable miners is empty. If that's the case the test description should say that.

consensus/storage_fault_monitor_test.go
```diff
@@ -0,0 +1,144 @@
...
+	t.Run("When 3 late miners, sends 3 messages", func(t *testing.T) {
+		getf := address.NewForTestGetter()
+		height := types.NewBlockHeight(100)
+
+		addr1 := getf().String()
+		addr2 := getf().String()
+		addr3 := getf().String()
+
+		data, err := cbor.DumpObject(&map[string]uint64{
+			addr1: miner.PoStStateAfterProvingPeriod,
```
acruikshank: `Slash` shouldn't be sending slash messages unless we're past the generation attack time. The `Slash` code doesn't care, but this is confusing.

consensus/storage_fault_monitor_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	t.Run("When Send fails, get collection of all errors + no messages", func(t *testing.T) {
+		getf := address.NewForTestGetter()
+		height := types.NewBlockHeight(100)
+		addr1 := getf().String()
+		addr2 := getf().String()
+
+		data, err := cbor.DumpObject(&map[string]uint64{
+			addr1: miner.PoStStateAfterProvingPeriod,
+			addr2: miner.PoStStateAfterProvingPeriod,
```
acruikshank: Should be `PoStStateAfterGenerationAttackThreshold`

acruikshank: Perhaps the storage market actor should only return a slice of addresses.

consensus/storage_fault_monitor_test.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+		require.NoError(t, err)
+
+		queryer := makeQueryer([][]byte{data})
+		ob := outbox{failSend: true, failErr: "Boom"}
+		minerOwnerAddr := signer.Addresses[0]
+		fm := NewStorageFaultMonitor(&storageFaultMonitorPorcelain{false, false, queryer}, &ob, minerOwnerAddr)
+		err = fm.Slash(ctx, height)
+		assert.Error(t, err, "Boom\nBoom")
```
acruikshank: This assertion is misleading. It's just testing that the error is non nil (and it'll message "Boom\nBoom" if it is nil. Better to `asserrt.Error(t, err)` and then `assert.Contains(t, "message", err.Error()`.

consensus/storage_fault_monitor_test.go
```diff
@@ -0,0 +1,144 @@
...
+		queryer := makeQueryer([][]byte{data})
+		ob := outbox{}
+		minerOwnerAddr := signer.Addresses[0]
+		fm := NewStorageFaultMonitor(&storageFaultMonitorPorcelain{false, false, queryer}, &ob, minerOwnerAddr)
+		err = fm.Slash(ctx, height)
+		assert.NoError(t, err)
+		assert.Equal(t, 3, ob.msgCount)
+	})
+
+	t.Run("When Send fails, get collection of all errors + no messages", func(t *testing.T) {
```
acruikshank: This test description is not true.

node/node.go
```diff
@@ -724,10 +729,23 @@ func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet
 			prevHead = newHead
 
 			if node.StorageMiner != nil {
-				err := node.StorageMiner.OnNewHeaviestTipSet(newHead)
-				if err != nil {
+				if err := node.StorageMiner.OnNewHeaviestTipSet(newHead); err != nil {
 					log.Error(err)
 				}
+				if node.StorageFaultMonitor != nil {
+					height, err := newHead.Height()
```
acruikshank: BLOCKING: This is too much logic to put in node. It's also untested. At the very least, StorageFaultMiner should have its own (unit tested) `OnNewHeaviestTipSet` metthod. Since fault handling it intimately tied to storage mining, it might be better for the SlashMonitor to be constructed withing the StorageMiner and have it trigger slashing from its `OnNewHeaviestTipSet` method.

tools/fast/tests/slashing_test.go
acruikshank: If it takes too long, it should be a functional test.

node/node.go
shannonwells: good call, this is from the weeks-old original invocation and the design changed a lot while you were gone.  it went unexamined, and I agree that's a better idea.

consensus/storage_fault_monitor.go
shannonwells: Outbox Send does a bunch of other checks as well as determining current height and nonce, validates the message,  then it Enqueues the message into the MessageQueue, and _then_ calls Publish which is strictly not true; it not only adds to mpool but also itself will broadcast or not.  Inbox's `Add` adds directly to the message pool and doesn't add to the MessageQueue.  It seems like the slashing message should be following the same code path as every other message sent to actors.

consensus/storage_fault_monitor_test.go
shannonwells: Since we/I realized yesterday this should slash only in the case of `PoStStateAfterGenerationAttackThreshold`, I agree, but I'd rather not change this now since this needs to go in this week. It's not doing any harm right now, and I can follow up with a PR after the release.

acruikshank: It would be nice to at least have the right PoSt state in this PR, though.

consensus/storage_fault_monitor.go
acruikshank: Non-Blocking:
Ok, looking a little closer at it, I sort of get what your saying. I know this was mostly covered in the setup PR that introduced the bcast flag, but I didn't get a chance to comment on that. Needing to add flags like this always make me question our abstractions

Our goal is to create a message and add it directly to our message pool. The weird thing here for me is that we do this through our `Outbox` interface. It's even weirder that Outbox calls `Publish` on the message publisher with a flag that tells it not to publish, just so we can add it to the message pool. 

`Outbox.Send` is responsible for:
* Message construction from components
* Message signing
* Nonce generation
* Retrieving the block height
* "Publishing" through the `Publisher` interface

`defaultMessagePublisher.Publish` is responsible for:
* Encoding the message for publishing
* Adding the message to the message pool ensuring that miner's get their own messages. I think this is due to a bug in libp2p.
* Possibly publishing the encoded message to the network

In my ideal world, there would be a new `MessageSigner` struct that would export a `NewSignedMessage` method to construct, sign and nonce a message. Also, message pool would get its access to the block height back, since the need for it is an implementation detail that doesn't really need to be shared with Outbox or defaultMessagePublisher. It could be this goes into something like `Policy` and that is what satisfies something like a `AddToMessagePool` interface.

Then `Outbox.Send` would:
* Construct the signed, nonce'd message with a call to `MessageSigner`.
* Add the message to the message pool (assuming this is still necessary)
* Publish the message through the message publisher.

`defaultMessagePublisher.Publish` would just encode and publish the message to the network as its name implies.

Then code like `StorageFaultMonitor` would could add messages to the local message pool with dependencies on `NewSignedMessage` and `AddToMessagePool`.

CC @anorth

consensus/storage_fault_monitor_test.go
shannonwells: Ah, yes I  was talking only about returning a slice of addresses. Using the right PoSt state is already in my local change set.

tools/fast/tests/slashing_test.go
shannonwells: True, and I fear that it will still make CI unacceptably long, but we'll try it.

consensus/storage_fault_monitor.go
shannonwells: ok I think this means I'm to consider these changes in another PR @acruikshank ?

consensus/storage_fault_monitor.go
```diff
@@ -0,0 +1,89 @@
+package consensus
```
ZenGround0: I think this deserves its own package, say `fault` or `fault_monitor`

consensus/storage_fault_monitor.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+// Slash checks for miners with unreported faults, then slashes them
+// Slashing messages are not broadcast to the network, but included in the next block mined by the slashing
+// node.
+func (sfm *StorageFaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHeight) error {
+	res, err := sfm.plumbing.MessageQuery(ctx, sfm.msgSender, address.StorageMarketAddress, "getLateMiners")
+	if err != nil {
+		return errors.FaultErrorWrap(err, "getLateMiners message failed")
```
ZenGround0: nit: Are fault errors the right thing to return right here?  My impression is that they are used internally within state machine processing but not useful for querying.

node/node.go
```diff
@@ -98,16 +98,17 @@ type nodeChainSyncer interface {
...
 type Node struct {
 	host     host.Host
 	PeerHost host.Host
 
-	Consensus    consensus.Protocol
-	ChainReader  nodeChainReader
-	MessageStore *chain.MessageStore
-	Syncer       nodeChainSyncer
-	PowerTable   consensus.PowerTableView
+	Consensus           consensus.Protocol
```
ZenGround0: nit: this seems incorrect wrt standard linting.

node/node.go
```diff
@@ -98,16 +98,17 @@ type nodeChainSyncer interface {
 	HandleNewTipSet(ctx context.Context, ci *types.ChainInfo, trusted bool) error
 }
 
+
```
ZenGround0: nit: extra newline

state/tree.go
```diff
@@ -283,17 +283,6 @@ func GetAllActors(ctx context.Context, t Tree) <-chan GetAllActorsResult {
 	return out
 }
 
-// GetAllActorsFromStore loads a StateTree and returns a channel with addresses
```
ZenGround0: Nice!

protocol/storage/miner.go
```diff
@@ -114,6 +121,7 @@ func NewMiner(minerAddr, ownerAddr address.Address, workerAddr address.Address,
 		node:                nd,
 		proposalAcceptor:    acceptProposal,
 		proposalRejector:    rejectProposal,
+		storageFaultSlasher: slasher,
```
ZenGround0: Non-blocking: I have a feeling this should be operating in a separate event loop than the storage miner's.  Slashing other miners is as separate from sealing/posting as block mining is.  It would make sense for the slasher to be run in its own small "protocol" along with storage, retrieval, hello and block mining.  Fine for now.

tools/fast/tests/slashing_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
+	"github.com/filecoin-project/go-filecoin/tools/fast/series"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// Too slow to let run in CI. Use for debugging locally.
```
ZenGround0: It would be worth looking for an issue in the backlog or filing a new one that tracks work for parameter setting.  Running the actual code with a much lower proving period would be very useful for CI.  We already do something like this for blocktime, generalizing seems worth thinking about anyway.

consensus/storage_fault_monitor.go
shannonwells: @acruikshank  ? this was a request from you I think

protocol/storage/miner.go
shannonwells: @acruikshank

consensus/storage_fault_monitor.go
shannonwells: I was waffling on whether it should be in its own package from a dev standpoint or whether it rightfully belongs in consensus. I could go either way though.

tools/fast/tests/slashing_test.go
shannonwells: I should have taken this comment out since it's running in CI now, but I think the best issue for that would be #3180

consensus/storage_fault_monitor.go
acruikshank: I can't remember the context under which I would have said that. At any rate, Fault and Revert errors are meaningless outside of message execution, so wrapping the error in a Fault doesn't help and doesn't hurt. It might be better to wrap it some other way to avoid the type of confusion we're experiencing now.

node/node.go
shannonwells: Yes, to put it succinctly I had some environment issues and had to redo my git clone and pre-commit hooks, so it should be fixed now.

consensus/storage_fault_monitor.go
shannonwells: I have changes for this file anyway and I'll just convert it to a plain wrap

tools/fast/tests/slashing_test.go
shannonwells: and it was, my "that" referred to tracking work for parameter setting specifically for proving period.

consensus/storage_fault_monitor.go
shannonwells: pong (coming)

protocol/storage/testing.go
```diff
@@ -15,3 +15,11 @@ func (p *FakeProver) CalculatePoSt(ctx context.Context, start, end *types.BlockH
 		Proofs: []types.PoStProof{[]byte("test proof")},
 	}, nil
 }
+
+// TrivialTestSlasher is a storage fault slasher that does nothing
+type TrivialTestSlasher struct{}
+
+// Slash is a required function for storageFaultSlasher interfaces and is intended to do nothing.
+func (ts *TrivialTestSlasher) Slash(context.Context, *types.BlockHeight) error {
+	return nil
```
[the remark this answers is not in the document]
shannonwells: Fixed, and used now. Good idea, thanks

consensus/storage_fault_monitor.go
anorth: Ack. See also #3052 - the pool and queue may become the same thing.

consensus/storage_fault_slasher.go
```diff
@@ -0,0 +1,90 @@
+package consensus
```
anorth: I would say this *definitely* shouldn't be part of consensus. If you weren't about to land this, I would say blocking. Please follow up with a move to somewhere like `protocol/storage`.

consensus/storage_fault_slasher.go
```diff
@@ -0,0 +1,90 @@
...
+	for minerStr, state := range *lms {
+		minerAddr, err := address.NewFromString(minerStr)
+		if err != nil {
+			return errors.FaultErrorWrap(err, "could not create minerAddr string")
+		}
+
+		// add slash message to message pull w/o broadcasting
+		sfm.log.Debugf("Slashing %s with state %d\n", minerStr, state)
+
+		_, err = sfm.outbox.Send(ctx, sfm.msgSender, minerAddr, types.ZeroAttoFIL, types.NewAttoFILFromFIL(1),
```
anorth: Please factor this buried gas price constant out at least to a top-level constant where it can be well named, but preferably to a construction parameter to push the pricing decision out further toward the node.

consensus/storage_fault_slasher.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+// Slash checks for miners with unreported faults, then slashes them
+// Slashing messages are not broadcast to the network, but included in the next block mined by the slashing
+// node.
+func (sfm *StorageFaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHeight) error {
+	res, err := sfm.plumbing.MessageQuery(ctx, sfm.msgSender, address.StorageMarketAddress, "getLateMiners")
+	if err != nil {
+		return errors.FaultErrorWrap(err, "getLateMiners message failed")
```
anorth: "Faults" are for the VM and not relevant outside it. Please wrap with a simple `errors.Wrapf`. While doing so, please add more detail to each message with the actual values that failed. E.g. for this one, the sender and storage market addresses.

consensus/storage_fault_slasher_test.go
```diff
@@ -0,0 +1,146 @@
...
+		return cid.Undef, errors.New("gas limit must be >= 300")
+	}
+
+	if ob.failSend {
+		return cid.Undef, errors.New(ob.failErr)
+	}
+	ob.msgCount++
+	// we ignore the CID returned from Send anyway
+	return cid.Undef, nil
+}
```
anorth: Nice work with these fakes for unit testing 💯

protocol/storage/miner.go
anorth: @shannonwells please document this reasoning here. I would in fact advocate for moving it out anyway, as this is unwelcome coupling between two processes that have only a slight interaction - it should at least be possible (easy) to mine *without* running a slashing monitor, depending on node configuration. One smell here is all the unnecessary diffs related to passing a `TrivialTestSlasher`everywhere that a miner is constructed.

Edit: and the miner test below! Please move it out in a follow-up.

protocol/storage/miner_test.go
```diff
@@ -507,6 +509,143 @@ func TestOnNewHeaviestTipSet(t *testing.T) {
...
+}
+
+func callOnNewHeaviestTipSet(api *minerTestPorcelain, t *testing.T, miner *Miner, height uint64) {
+	api.blockHeight = types.NewBlockHeight(height)
+	block := &types.Block{Height: types.Uint64(height)}
+	ts, err := types.NewTipSet(block)
+	require.NoError(t, err)
+	_ = miner.OnNewHeaviestTipSet(ts)
+	// Sleep to make sure submitPoSt is not called
+	time.Sleep(1 * time.Second)
```
anorth: This will be flaky!

consensus/storage_fault_slasher.go
shannonwells: Whoops, this was meant to be changed and I forgot about it... will be in the follow up.

anorth:
I see in https://circleci.com/gh/filecoin-project/go-filecoin/26170?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link:

> Slowest test: storage TestSlasherCalls (took 9.02 seconds).

This is unacceptably long for a unit test. I am guessing it's the `time.Sleep`. Please also address this in the follow-up - but if that follow-up moves the slashing worker out of the miner, the offending test will probably become unnecessary.

shannonwells:
Applies to #2939

PR #3181, opened 2019-08-02
Update circleci docker container to circleci/golang:1.12.1-stretch

travisperson:
This is a bit of a hail-mary to move us past this current issue. We unfortunately still have no idea what is causing the panics to occur as only containers built in circleci has this issue. 

I'm hoping locking into `stretch` will fix this in the short term, as the rust proofs are built from a [`debian:stretch` base image](https://github.com/filecoin-project/rust-fil-proofs/blob/master/Dockerfile-ci#L5), and I was successful in building go-filecoin from `golang:1.12.1-stretch` and coping it into a `busybox:1-glibc` container.

The docker file I used was a slightly modified version  I created for the new k8s deployments.

https://github.com/filecoin-project/go-filecoin/blob/feat/devnet-fast-k8s/Dockerfile.devnet.base

I'm hoping because circleci uses this [same image to build their `circleci/golang:1.12.1-stretch`](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/ce5a26df3b256dad024122a0016b4dd19cc1ccc8/golang/images/1.12.1-stretch/Dockerfile#L5) container everything should be built in the same environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ jobs:
 
   deps_linux:
     docker:
-      - image: circleci/golang:1.12.1
+      - image: circleci/golang:1.12.1-stretch
     working_directory: /go/src/github.com/filecoin-project/go-filecoin
     resource_class: xlarge
     steps:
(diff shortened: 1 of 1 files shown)


anorth:
Invoking maintainer privilege to land this by request.

PR #3182, opened 2019-08-02
Prepare consensus for messages out of blocks

ZenGround0:
This follows onto PR #3168, making changes that will be needed to the consensus and processor interfaces for message separation.

Please do not review until #3168 is in as it is based on that branch.

### DIFF
--- a/consensus/expected.go
+++ b/consensus/expected.go
@@ -84,10 +84,10 @@ const AncestorRoundsNeeded = miner.LargestSectorSizeProvingPeriodBlocks + miner.
 // A Processor processes all the messages in a block or tip set.
 type Processor interface {
 	// ProcessBlock processes all messages in a block.
-	ProcessBlock(ctx context.Context, st state.Tree, vms vm.StorageMap, blk *types.Block, ancestors []types.TipSet) ([]*ApplicationResult, error)
+	ProcessBlock(context.Context, state.Tree, vm.StorageMap, *types.Block, []*types.SignedMessage, []types.TipSet) ([]*ApplicationResult, error)
 
 	// ProcessTipSet processes all messages in a tip set.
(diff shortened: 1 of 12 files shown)


consensus/expected.go
```diff
@@ -84,10 +84,10 @@ const AncestorRoundsNeeded = miner.LargestSectorSizeProvingPeriodBlocks + miner.
 // A Processor processes all the messages in a block or tip set.
 type Processor interface {
 	// ProcessBlock processes all messages in a block.
-	ProcessBlock(ctx context.Context, st state.Tree, vms vm.StorageMap, blk *types.Block, ancestors []types.TipSet) ([]*ApplicationResult, error)
+	ProcessBlock(context.Context, state.Tree, vm.StorageMap, *types.Block, []*types.SignedMessage, []types.TipSet) ([]*ApplicationResult, error)
```
shannonwells: We seem to vary as to whether to have param names in an interface declaration.  I waffle on it but generally I think I've decided in the code base I'd prefer it, because so often it clarifies purpose.  Do we have a stated preference?

shannonwells: I've noticed some inconsistencies on whether to use param names in interface declarations. Is there a desired convention in this code base?

shannonwells: I only think it's better if there are say two big.Ints, or a base type where it's not obvious (e.g. "WaitForSecs(int)" is obvious what the int is for).

ZenGround0:
> There is a part of me that wants to reference messages and their receipts by the CID of the block they correspond to instead of the index of the block in the tipset. There's also a part of me that thinks messages and their receipts could be placed into a single structure (MsgWithRcpts)

These all sound like better ways to do things.  Don't forget these ideas as we'll likely come back and implement them.  It definitely feels a bit dirty to implicitly assume the tipset blocks and messages are in the correct order in the consensus / processor interfaces.  Down the road we might want a software "full block" grouping things together (maybe tipsets will be made of these and not blocks).

PR #3184, opened 2019-08-03
Integrate PeerTracker and GraphsyncFetcher

### LINKED ISSUE #3177: Design: Integrate GraphsyncFetcher w/ PeerTracker
### Description

When I use a GraphsyncFetcher to fetch tipsets, if the peer I am initially given fails to return the tipsets, I want to use the PeerTracker to failover to a different peer, so that I can still sync the chain.

### Acceptance criteria

- GraphsyncFetcher makes requests to the peer specified by FetchTipSets initially
- When a Graphsync request to a peer fails inside the GraphsyncFetcher
  - GraphsyncFetcher should find an appropriate replacement peer using the PeerTracker
  - GraphsyncFetcher should attempt the request again with the new peer
  - GraphsyncFetcher should repeat a fixed number of times (or up to a fixed timeout?) before failing entirely

### Risks + pitfalls

- Ideally, not making frivolous requests to peers that do not have the tipset you're interested in

### Where to begin

There are a couple key components to this task:

1.  Finding an identifying appropriate peers. Currently, the PeerTracker just returns a list of chains it knows about (in ChainInfo structs). The simplest approach would be to just take the first of these, then move on to the second, etc. Alternatively, we could attempt to use chain height in ChainInfos to try to identify a peer who at least has the current tipset we're interested in. Another question to consider is whether this process would live in the PeerManager (producing a method with more intelligence than `List`) or in the GraphsyncFetcher. The downside of putting things in the PeerManager is you can't easily track which peers you've already tried. The downside of putting things in the GraphsyncFetcher is as soon as you query the PeerTracker, the list of peers becomes out of date, and you might need to re-query.

Proposal: For the first version, keep it extra simple.  Every time a request fails in the GraphsyncFetcher, query the PeerManager's existing `List` method, try the peers one by one in order. Iterate and optimize from there.

2. How many retries to make -- could be a fixed amount, or a fixed time.

Proposal: For now, every time a request fails, just try all the peers, fail if none succeed.

hannahhoward:
# Goals

Integrate the peer tracker with the graphsync fetcher to provide fallback resiliency if peer goes
offline during a fetch

# Implementation

- During a request:
  - track which peers we've already tried to request from
  - every time a request fails:
    - query the peer tracker for peers we haven't tried yet
    - if we find one we haven't tried, retry with that peer.  if any full tipsets were fetched, verify them and start at the first incomplete tipset
    - if we exhaust all peers we know about, fail with a message about the tipset we failed on
- multiple unit tests of different potential error cases during fetching process

# For Discussion

This is getting to be a very complicated and stateful iteration. We have to deal with:
  - tracking peers we've tried
  - tracking which peer we trying currently
  - gradually increasing our recursion depth
  - building a tipset list
  - lots and lots of error checking, all of which aborts the function
  - checking every tipset to see if we're done yet

I'm not entirely sure how to simplify all of it. I starting tracking a small bit of that state independently with the requestPeerFinder. I wonder if other parts could be broken out to abstract parts of the state tracking.

fix #3177

### DIFF
--- a/net/graphsync_fetcher.go
+++ b/net/graphsync_fetcher.go
@@ -96,6 +96,35 @@ func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSe
 	if err != nil {
 		return nil, err
 	}
+
+	// fetch remaining tipsets recursively
+	return gsf.fetchRemainingTipsets(ctx, startingTipset, rpf, done)
+}
+
+func (gsf *GraphSyncFetcher) fetchFirstTipset(ctx context.Context, tsKey types.TipSetKey, rpf *requestPeerFinder) (types.TipSet, error) {
+	cids := tsKey.ToSlice()
+	remainingCids := cids
+	for {
+		err := gsf.fetchBlocks(ctx, remainingCids, rpf.CurrentPeer())
+		if err == nil {
+			break
+		}
+		// something went wrong in a graphsync request, but we want to keep trying other peers, so
+		// just log error
+		logGraphsyncFetcher.Infof("Error occurred during Graphsync request: %s", err)
+		remainingCids, err = gsf.missingCids(remainingCids)
+		if err != nil {
+			return types.UndefTipSet, err
+		}
+		err = rpf.FindNextPeer()
+		if err != nil {
+			return types.UndefTipSet, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
+		}
+	}
+	return gsf.loadTipsetFromCids(ctx, cids)
+}
+
+func (gsf *GraphSyncFetcher) fetchRemainingTipsets(ctx context.Context, startingTipset types.TipSet, rpf *requestPeerFinder, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
 	out := []types.TipSet{startingTipset}
 	isDone, err := done(startingTipset)
 	if err != nil {
@@ -146,29 +175,6 @@ func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSe
 		}
 	}
 	return out, nil
-}
-
-func (gsf *GraphSyncFetcher) fetchFirstTipset(ctx context.Context, tsKey types.TipSetKey, rpf *requestPeerFinder) (types.TipSet, error) {
-	cids := tsKey.ToSlice()
-	remainingCids := cids
-	for {
-		err := gsf.fetchBlocks(ctx, remainingCids, rpf.CurrentPeer())
-		if err == nil {
-			break
-		}
-		// something went wrong in a graphsync request, but we want to keep trying other peers, so
-		// just log error
-		logGraphsyncFetcher.Infof("Error occurred during Graphsync request: %s", err)
-		remainingCids, err = gsf.missingCids(remainingCids)
-		if err != nil {
-			return types.UndefTipSet, err
-		}
-		err = rpf.FindNextPeer()
-		if err != nil {
-			return types.UndefTipSet, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
-		}
-	}
-	return gsf.loadTipsetFromCids(ctx, cids)
 }
 
 // fetchBlocks requests a single set of cids as individual bocks, fetching
--- a/net/graphsync_fetcher.go
+++ b/net/graphsync_fetcher.go
@@ -88,8 +88,8 @@ const recursionMultiplier = 4
 // go-filecoin migrates to the same IPLD library used by go-graphsync (go-ipld-prime)
 //
 // See: https://github.com/filecoin-project/go-filecoin/issues/3175
-func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, from peer.ID, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
-	rpf := newRequestPeerFinder(gsf.peerTracker, from)
+func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, initialPeer peer.ID, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
+	rpf := newRequestPeerFinder(gsf.peerTracker, initialPeer)
 
 	// fetch initial tipset
 	startingTipset, err := gsf.fetchFirstTipset(ctx, tsKey, rpf)
@@ -103,16 +103,15 @@ func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSe
 
 func (gsf *GraphSyncFetcher) fetchFirstTipset(ctx context.Context, tsKey types.TipSetKey, rpf *requestPeerFinder) (types.TipSet, error) {
 	cids := tsKey.ToSlice()
-	remainingCids := cids
 	for {
-		err := gsf.fetchBlocks(ctx, remainingCids, rpf.CurrentPeer())
+		err := gsf.fetchBlocks(ctx, cids, rpf.CurrentPeer())
 		if err == nil {
 			break
 		}
 		// something went wrong in a graphsync request, but we want to keep trying other peers, so
 		// just log error
 		logGraphsyncFetcher.Infof("Error occurred during Graphsync request: %s", err)
-		remainingCids, err = gsf.missingCids(remainingCids)
+		cids, err = gsf.missingCids(cids)
 		if err != nil {
 			return types.UndefTipSet, err
 		}
@@ -121,7 +120,7 @@ func (gsf *GraphSyncFetcher) fetchFirstTipset(ctx context.Context, tsKey types.T
 			return types.UndefTipSet, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
 		}
 	}
-	return gsf.loadTipsetFromCids(ctx, cids)
+	return gsf.loadTipsetFromCids(ctx, tsKey.ToSlice())
 }
 
 func (gsf *GraphSyncFetcher) fetchRemainingTipsets(ctx context.Context, startingTipset types.TipSet, rpf *requestPeerFinder, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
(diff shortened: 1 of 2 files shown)


net/graphsync_fetcher.go
```diff
@@ -122,20 +138,42 @@ func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSe
...
+		}
+		err = rpf.FindNextPeer()
+		if err != nil {
+			return types.UndefTipSet, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
+		}
+	}
+	return gsf.loadTipsetFromCids(ctx, cids)
+}
+
 // fetchBlocks requests a single set of cids as individual bocks, fetching
```
shannonwells: typo: bocks

net/graphsync_fetcher.go
```diff
@@ -185,3 +224,45 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
...
+	pri := &requestPeerFinder{peerTracker, from, make(map[peer.ID]struct{})}
+	pri.triedPeers[from] = struct{}{}
+	return pri
+}
+
+func (pri *requestPeerFinder) CurrentPeer() peer.ID {
+	return pri.currentPeer
+}
+
+func (pri *requestPeerFinder) FindNextPeer() error {
```
shannonwells: how does the peer tracker get its list?  any reason not to cache/memoize this  list in requestPeerFinder just for the current invocation?

net/graphsync_fetcher_test.go
```diff
@@ -88,14 +98,232 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+		ts, err := fetcher.FetchTipSets(ctx, final.Key(), pid0, done)
+		require.NoError(t, err, "the request completes successfully")
+		require.Equal(t, 7, len(mgs.receivedRequests), "all expected graphsync requests are made")
+		require.Equal(t, pid0, mgs.receivedRequests[0].p, "asks first peer for everything in first tipset")
+		require.Equal(t, pid0, mgs.receivedRequests[1].p, "asks first peer for everything in first tipset")
+		require.Equal(t, pid0, mgs.receivedRequests[2].p, "asks first peer for everything in first tipset")
+		require.Equal(t, pid1, mgs.receivedRequests[3].p, "asks second peer for failed responses in first tipset")
+		require.Equal(t, pid1, mgs.receivedRequests[4].p, "asks second peer for failed responses in first tipset")
+		require.Equal(t, pid2, mgs.receivedRequests[5].p, "asks third peer for failed responses in first tipset")
+		require.Equal(t, pid2, mgs.receivedRequests[6].p, "asks third peer for when continuing the query")
```
shannonwells: "asks third peer for___________ when continuing the query"?

net/graphsync_fetcher_test.go
```diff
@@ -88,14 +98,232 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+	t.Run("initial request fails on a block but fallback peer succeeds", func(t *testing.T) {
+		gen := builder.NewGenesis()
+		final := builder.AppendOn(gen, 3)
+		height, err := final.Height()
+		require.NoError(t, err)
+		chain1 := types.NewChainInfo(pid1, final.Key(), height)
+		chain2 := types.NewChainInfo(pid2, final.Key(), height)
+		pt := &fakePeerTracker{[]*types.ChainInfo{chain1, chain2}}
+
+		stubs := []requestResponse{
```
shannonwells: what is this a fake request for / response from?

net/graphsync_fetcher.go
hannahhoward: according to @ZenGround0 it could change on each call.

net/graphsync_fetcher_test.go
hannahhoward: this would be a request to another peer with graphsync and a response from that peer.

net/graphsync_fetcher.go
```diff
@@ -79,37 +89,75 @@ const recursionMultiplier = 4
 //
 // See: https://github.com/filecoin-project/go-filecoin/issues/3175
 func (gsf *GraphSyncFetcher) FetchTipSets(ctx context.Context, tsKey types.TipSetKey, from peer.ID, done func(types.TipSet) (bool, error)) ([]types.TipSet, error) {
```
ZenGround0: We should change the name "from" to something more meaningful at this point in the callstack.  `fetchPeer`, `peerToRequest`, `firstRequest`, I can't think of anything really good but those are some ideas.

net/graphsync_fetcher.go
```diff
@@ -79,37 +89,75 @@ const recursionMultiplier = 4
...
+	remainingCids := cids
+	for {
+		err := gsf.fetchBlocks(ctx, remainingCids, rpf.CurrentPeer())
+		if err == nil {
+			break
+		}
+		// something went wrong in a graphsync request, but we want to keep trying other peers, so
+		// just log error
+		logGraphsyncFetcher.Infof("Error occurred during Graphsync request: %s", err)
+		remainingCids, err = gsf.missingCids(remainingCids)
```
ZenGround0: Using these two names together is confusing.  `remaining` means that they remain to be fetched, and `missing` means that they are missing from the set already fetched, so they are actually synonyms here which is pretty unexpected.  You should chose one.  Also you could potentially avoid the rename on line 106, leaving the value as `cids` throughout and filtering it down with your choice of `missingCids` or `remainingCids`.

net/graphsync_fetcher.go
```diff
@@ -79,37 +89,75 @@ const recursionMultiplier = 4
...
+	startingTipset, err := gsf.fetchFirstTipset(ctx, tsKey, rpf)
 	if err != nil {
 		return nil, err
 	}
-	startingTipset, err := gsf.loadTipsetFromCids(ctx, cids)
-	if err != nil {
-		return nil, err
+
+	// fetch remaining tipsets recursively
+	return gsf.fetchRemainingTipsets(ctx, startingTipset, rpf, done)
```
ZenGround0: Decomposing into these two functions makes reading this code pleasant ✨

net/graphsync_fetcher.go
```diff
@@ -185,3 +237,60 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
...
+	return false, nil
+}
+
+type requestPeerFinder struct {
+	peerTracker graphsyncFallbackPeerTracker
+	currentPeer peer.ID
+	triedPeers  map[peer.ID]struct{}
+}
+
+func newRequestPeerFinder(peerTracker graphsyncFallbackPeerTracker, from peer.ID) *requestPeerFinder {
```
ZenGround0: I'd change from here to to `firstPeer` or something

net/graphsync_fetcher.go
```diff
@@ -185,3 +237,60 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
...
+	return missing, nil
+}
+
+func (gsf *GraphSyncFetcher) areCidsMissing(tsKey types.TipSetKey, rpf *requestPeerFinder) (bool, error) {
+	missing, err := gsf.missingCids(tsKey.ToSlice())
+	if err != nil {
+		return false, err
+	}
+	if len(missing) != 0 {
+		err := rpf.FindNextPeer()
```
ZenGround0: Blocking: I have a strong preference to put this on line 158.  This is a hidden side effect I didn't expect from the name.

net/graphsync_fetcher.go
```diff
@@ -185,3 +237,60 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
...
+		if err != nil {
+			return false, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+type requestPeerFinder struct {
```
ZenGround0: Wonderful abstraction

net/graphsync_fetcher.go
```diff
@@ -78,38 +88,79 @@ const recursionMultiplier = 4
...
+		// something went wrong in a graphsync request, but we want to keep trying other peers, so
+		// just log error
+		logGraphsyncFetcher.Infof("Error occurred during Graphsync request: %s", err)
+		cids, err = gsf.missingCids(cids)
+		if err != nil {
+			return types.UndefTipSet, err
+		}
+		err = rpf.FindNextPeer()
+		if err != nil {
+			return types.UndefTipSet, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
```
anorth: The "no more peers" error is swallowed, but should be wrapped

net/graphsync_fetcher.go
```diff
@@ -78,38 +88,79 @@ const recursionMultiplier = 4
...
-			cids := tsKey.ToSlice()
-			ts, err = gsf.loadTipsetFromCids(ctx, cids)
+			cidsAreMissing, err = gsf.areCidsMissing(tsKey)
+			if err != nil {
+				return nil, err
+			}
+			if cidsAreMissing {
+				err := rpf.FindNextPeer()
+				if err != nil {
+					return nil, fmt.Errorf("Failed fetching tipset: %s", tsKey.String())
```
anorth: Ditto swallowed error

net/graphsync_fetcher.go
```diff
@@ -185,3 +240,56 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
 
 	return types.NewTipSet(validatedBlocks...)
 }
+
+func (gsf *GraphSyncFetcher) missingCids(cids []cid.Cid) ([]cid.Cid, error) {
```
anorth: Please document the intent of this method

net/graphsync_fetcher.go
```diff
@@ -185,3 +240,56 @@ func (gsf *GraphSyncFetcher) loadTipsetFromCids(ctx context.Context, cids []cid.
...
+	if err != nil {
+		return false, err
+	}
+	if len(missing) != 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+type requestPeerFinder struct {
```
anorth: Please document this abstraction, especially its stateful nature and intended use pattern.

PR #3185, opened 2019-08-04
Break messages and receipts out of block header

ZenGround0:
Compiles.  Tests are still broken.

### DIFF
--- a/chain/message_store.go
+++ b/chain/message_store.go
@@ -5,13 +5,14 @@ import (
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-hamt-ipld"
+	"github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// MessageReader is an interface exposing the read methods of the
+// MessageProvider is an interface exposing the load methods of the
 // MessageStore.
-type MessageReader interface {
+type MessageProvider interface {
 	LoadMessages(context.Context, cid.Cid) ([]*types.SignedMessage, error)
 	LoadReceipts(context.Context, cid.Cid) ([]*types.MessageReceipt, error)
 }
@@ -39,9 +40,12 @@ func NewMessageStore(cst *hamt.CborIpldStore) *MessageStore {
 // storage.
 func (ms *MessageStore) LoadMessages(ctx context.Context, c cid.Cid) ([]*types.SignedMessage, error) {
 	// TODO #1324 message collection shouldn't be a slice
-	var out []*types.SignedMessage
+	var out types.MessageCollection
 	err := ms.ipldStore.Get(ctx, c, &out)
-	return out, err
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not load message collection with cid: %s", c.String())
+	}
+	return []*types.SignedMessage(out), nil
 }
 
 // StoreMessages puts the input signed messages to a collection and then writes
@@ -49,18 +53,19 @@ func (ms *MessageStore) LoadMessages(ctx context.Context, c cid.Cid) ([]*types.S
 func (ms *MessageStore) StoreMessages(ctx context.Context, msgs []*types.SignedMessage) (cid.Cid, error) {
 	// For now the collection is just a slice (cbor array)
 	// TODO #1324 put these messages in a merkelized collection
-	return ms.ipldStore.Put(ctx, msgs)
+	msgCollection := types.MessageCollection(msgs)
+	return ms.ipldStore.Put(ctx, msgCollection)
 }
 
 // LoadReceipts loads the signed messages in the collection with cid c from ipld
 // storage and returns the slice implied by the collection
 func (ms *MessageStore) LoadReceipts(ctx context.Context, c cid.Cid) ([]*types.MessageReceipt, error) {
-	var out []*types.MessageReceipt
+	var out types.ReceiptCollection
 	err := ms.ipldStore.Get(ctx, c, &out)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "could not load receipt collection with cid: %s", c.String())
 	}
-	return out, nil
+	return []*types.MessageReceipt(out), nil
 }
 
 // StoreReceipts puts the input signed messages to a collection and then writes
@@ -68,5 +73,6 @@ func (ms *MessageStore) LoadReceipts(ctx context.Context, c cid.Cid) ([]*types.M
 func (ms *MessageStore) StoreReceipts(ctx context.Context, receipts []*types.MessageReceipt) (cid.Cid, error) {
 	// For now the collection is just a slice (cbor array)
 	// TODO #1324 put these messages in a merkelized collection
-	return ms.ipldStore.Put(ctx, receipts)
+	rcptCollection := types.ReceiptCollection(receipts)
+	return ms.ipldStore.Put(ctx, rcptCollection)
 }
--- a/chain/testing.go
+++ b/chain/testing.go
@@ -41,7 +41,7 @@ type Builder struct {
 var _ BlockProvider = (*Builder)(nil)
 var _ TipSetProvider = (*Builder)(nil)
 var _ net.Fetcher = (*Builder)(nil)
-var _ MessageReader = (*Builder)(nil)
+var _ MessageProvider = (*Builder)(nil)
 
 // NewBuilder builds a new chain faker.
 // Blocks will have `miner` set as the miner address, or a default if empty.
@@ -61,6 +61,9 @@ func NewBuilder(t *testing.T, miner address.Address) *Builder {
 		messages:     make(map[cid.Cid][]*types.SignedMessage),
 		receipts:     make(map[cid.Cid][]*types.MessageReceipt),
 	}
+
+	b.messages[types.EmptyMessagesCID] = []*types.SignedMessage{}
+	b.receipts[types.EmptyReceiptsCID] = []*types.MessageReceipt{}
 
 	nullState, err := makeCid("null")
 	require.NoError(t, err)
@@ -166,8 +169,8 @@ func (f *Builder) Build(parent types.TipSet, width int, build func(b *BlockBuild
 			ParentWeight:    types.Uint64(parentWeight),
 			Parents:         parent.Key(),
 			Height:          height,
-			Messages:        []*types.SignedMessage{},
-			MessageReceipts: []*types.MessageReceipt{},
+			Messages:        types.EmptyMessagesCID,
+			MessageReceipts: types.EmptyReceiptsCID,
 			// Omitted fields below
 			//StateRoot:       stateRoot,
 			//Proof            PoStProof
@@ -181,7 +184,9 @@ func (f *Builder) Build(parent types.TipSet, width int, build func(b *BlockBuild
 
 		// Compute state root for this block.
 		prevState := f.StateForKey(parent.Key())
-		b.StateRoot, err = f.stateBuilder.ComputeState(prevState, [][]*types.SignedMessage{b.Messages})
+		msgs, ok := f.messages[b.Messages]
+		require.True(f.t, ok)
+		b.StateRoot, err = f.stateBuilder.ComputeState(prevState, [][]*types.SignedMessage{msgs})
 		require.NoError(f.t, err)
 
 		f.blocks[b.Cid()] = b
@@ -212,9 +217,21 @@ func (f *Builder) ComputeState(tip types.TipSet) cid.Cid {
 	require.NoError(f.t, err)
 	// Load the state of the parent tipset and compute the required state (recursively).
 	prev := f.StateForKey(parentKey)
-	state, err := f.stateBuilder.ComputeState(prev, tipMessages(tip))
+	state, err := f.stateBuilder.ComputeState(prev, f.tipMessages(tip))
 	require.NoError(f.t, err)
 	return state
+}
+
+// tipMessages returns the messages of a tipset.  Each block's messages are
+// grouped into a slice and a slice of these slices is returned.
+func (f *Builder) tipMessages(tip types.TipSet) [][]*types.SignedMessage {
+	var msgs [][]*types.SignedMessage
+	for i := 0; i < tip.Len(); i++ {
+		blkMsgs, ok := f.messages[tip.At(i).Messages]
+		require.True(f.t, ok)
+		msgs = append(msgs, blkMsgs)
+	}
+	return msgs
 }
 
 // Wraps a simple build function in one that also accepts an index, propagating a nil function.
@@ -261,8 +278,8 @@ func (bb *BlockBuilder) AddMessages(msgs []*types.SignedMessage, rcpts []*types.
 	require.NoError(bb.t, err)
 	bb.receipts[cR] = rcpts
 
-	bb.block.Messages = msgs
-	bb.block.MessageReceipts = rcpts
+	bb.block.Messages = cM
+	bb.block.MessageReceipts = cR
 }
 
 // SetStateRoot sets the block's state root.
@@ -329,7 +346,7 @@ type FakeStateEvaluator struct {
 
 // RunStateTransition delegates to StateBuilder.ComputeState.
 func (e *FakeStateEvaluator) RunStateTransition(ctx context.Context, tip types.TipSet, messages [][]*types.SignedMessage, receipts [][]*types.MessageReceipt, ancestors []types.TipSet, stateID cid.Cid) (cid.Cid, error) {
-	return e.ComputeState(stateID, tipMessages(tip))
+	return e.ComputeState(stateID, messages)
 }
 
 // IsHeavier compares chains weighed with StateBuilder.Weigh.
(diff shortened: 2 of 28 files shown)


hannahhoward:
@ZenGround0 Early review comment:

I wonder if you would benefit from having a `CompleteBlock` type struct that looks something like this:

```golang
type CompleteBlock struct {
  header *types.Block
  messages []*types.SignedMessage
  receipts []*types.MessageReceipt
}
```

It seems like you're in a lot of situations where you're having to expand parameter lists a lot and I wonder if this would help (esp with a constructor or a converter)

chain/message_store.go
```diff
@@ -41,7 +42,7 @@ func (ms *MessageStore) LoadMessages(ctx context.Context, c cid.Cid) ([]*types.S
 	// TODO #1324 message collection shouldn't be a slice
 	var out []*types.SignedMessage
 	err := ms.ipldStore.Get(ctx, c, &out)
-	return out, err
+	return out, errors.Wrapf(err, "could not load message collection with cid: %s", c.String())
```
hannahhoward: This seems to give the impression all of the messages go into a piece of hashable content. I was under the impression messages would be broken up over multiple hashes. Did I misunderstand? No big worry either way. It makes the changes to Graphsync easier.

ZenGround0:
@hannahhoward it probably makes sense to do this along the syncer / consensus / processor boundary in this PR.

I thought about doing this everywhere by changing tipsets to be a list of `CompleteBlock`s.  It seems like a good idea.  (One cool thing we'd get from doing that is a natural way to track aggregate messages and receipts per tipset in the same way we track aggregate state.  This would clean up message wait code in a big way).  I stayed away from it because it seemed like it would be more work to change both the block and tipset definition and I anticipated this change being pretty big already.  This is why I ended up piping the message store to so many places.  I am very open to making that change in a future PR.

hannahhoward:
> I thought about doing this everywhere by changing tipsets to be a list of `CompleteBlock`s. It seems like a good idea. (One cool thing we'd get from doing that is a natural way to track aggregate messages and receipts per tipset in the same way we track aggregate state. This would clean up message wait code in a big way). I stayed away from it because it seemed like it would be more work to change both the block and tipset definition and I anticipated this change being pretty big already. This is why I ended up piping the message store to so many places. I am very open to making that change in a future PR.

All makes sense. Figure this is the only part likely to make it into release, so we should make sure to document that work still needs to be done somewhere.

chain/message_store.go
ZenGround0: You've got it.  If I understand correctly we want a single hash to reference the message collection from the block header to keep header size low.  We eventually are planning to do some [merkelized data structure](https://github.com/filecoin-project/specs/blob/dd3bc99c76b2b6bd0baeb23687b0fe5e01c2db57/sharray.md) which will require graphsync traversals across several links, but we'll always have a single hash in the header and that merkelized data structure is further out.

ZenGround0:
@shannonwells @acruikshank this is almost the changeset I want to merge.  The one thing that is still not ready is the modification to the fetching code.  I've hacked in an extra round trip into the bitswap fetcher so that the node can sync properly and pass all integration tests.  I will replace this with an updated graph sync query.  I'm planning to do that after I rebase this on @hannahhoward's work integrating the graphsync fetcher into the node. I'll get review from her on that specialized work.  

In the meantime everything else could use careful review from veterans of the wider go-filecoin code.  In the charge to get this in by end of week I probably got too sloppy in places so I would appreciate getting called out for the worst instances of that sloppiness.

core/inbox.go
```diff
@@ -39,7 +39,12 @@ type MessageProvider interface {
 
 // NewInbox constructs a new inbox.
 func NewInbox(pool *MessagePool, maxAgeRounds uint, chain InboxChainProvider, messages MessageProvider) *Inbox {
-	return &Inbox{pool: pool, maxAgeTipsets: maxAgeRounds, chain: chain}
+	return &Inbox{
+		pool:          pool,
+		maxAgeTipsets: maxAgeRounds,
+		chain:         chain,
+		messages:      messages,
```
acruikshank: This should be `messageProvider` (or `msgGetter` or something like that). `messages` looks too much like a collection of messages.

mining/worker_test.go
```diff
@@ -516,7 +523,9 @@ func TestGenerateWithoutMessages(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Len(t, pool.Pending(), 0) // This is the temporary failure.
-	assert.Len(t, blk.Messages, 0)
+
+	assert.Equal(t, types.MessageCollection{}.Cid(), blk.Messages)
```
acruikshank: You generate a CID for an empty collection differently in `GenGen`. Is this a different CID? Could you do this there?  It would be nice to have a `types.EmptyMessagesCID` and `types.EmptyReceiptsCID` defined.

net/fetcher.go
```diff
@@ -91,11 +91,29 @@ func (bsf *BitswapFetcher) GetBlocks(ctx context.Context, cids []cid.Cid) ([]*ty
...
 
-	blocks, err := sanitizeBlocks(ctx, unsanitized, bsf.validator)
+	filBlocks, err := sanitizeBlocks(ctx, unsanitized, bsf.validator)
 	if err != nil {
 		return nil, err
 	}
-	return blocks, nil
+
+	// Hackity hack hack for getting messages and receipts
+	// TODO delete this and rebase off of integrated graphsync fetcher
```
acruikshank: Before merging, this must either go away or contain a link to a github issue.

net/fetcher_test.go
```diff
@@ -36,6 +37,15 @@ func TestFetchHappyPath(t *testing.T) {
 	block2 := builder.NewGenesis().At(0)
 	block3 := builder.NewGenesis().At(0)
 
+	// TODO this goes away in this PR after moving to graphsync fetcher
```
acruikshank: This TODO needs an issue URL.

plumbing/msg/waiter.go
```diff
@@ -47,9 +47,10 @@ type ChainMessage struct {
 // NewWaiter returns a new Waiter.
 func NewWaiter(chainStore waiterChainReader, messageReader chain.MessageReader, bs bstore.Blockstore, cst *hamt.CborIpldStore) *Waiter {
 	return &Waiter{
-		chainReader: chainStore,
-		cst:         cst,
-		bs:          bs,
+		chainReader:   chainStore,
+		cst:           cst,
+		bs:            bs,
+		messageReader: messageReader,
```
acruikshank: We seem to have a lot of names for things that retrieve messages. Do messageReaders and messageProviders have the same interface? Inteface and implementation?

plumbing/msg/waiter.go
```diff
@@ -109,7 +110,11 @@ func (w *Waiter) findMessage(ctx context.Context, ts types.TipSet, msgCid cid.Ci
 		}
 		for i := 0; i < iterator.Value().Len(); i++ {
 			blk := iterator.Value().At(i)
-			for _, msg := range blk.Messages {
+			msgs, err := w.messageReader.LoadMessages(ctx, blk.Messages)
```
acruikshank: Does this get replaced with a graphsync query? If so, it would be good to get a TODO and an issue about that.

net/fetcher.go
ZenGround0: Thx for the reminder.  This is going away after I rebase on the integrated graphsync fetcher

net/fetcher_test.go
ZenGround0: As with the other one this TODO and code are going away before this PR merges.

plumbing/msg/waiter.go
ZenGround0: Thanks for calling out.  Yeah it's mostly the same thing +/- a method.  I'll choose a standard name and use it throughout.

plumbing/msg/waiter.go
ZenGround0: No this is unrelated to the graphsync query.

shannonwells:
> I stayed away from it because it seemed like it would be more work to change both the block and tipset definition and I anticipated this change being pretty big already. This is why I ended up piping the message store to so many places. I am very open to making that change in a future PR.

I really wonder if this isn't putting the cart before the horse. It's a little late now given you want to get this in by Friday.  But maybe for next time.

types/testing.go
```diff
@@ -171,19 +172,25 @@ func NewSignedMessageForTestGetter(ms MockSigner) func() *SignedMessage {
 // SomeCid generates a Cid for use in tests where you want a Cid but don't care
 // what it is.
 func SomeCid() cid.Cid {
-	b := &Block{}
-	return b.Cid()
+	obj, err := cbor.WrapObject([]int{42}, DefaultHashFunction, -1)
```
shannonwells: this is always going to return the same value, no?  Why not just find out what it is and return that as a constant?  Then you don't have to have this panic code which will almost certainly never get called. Or see my other comment.

chain/testing.go
```diff
@@ -212,11 +217,21 @@ func (f *Builder) ComputeState(tip types.TipSet) cid.Cid {
 	require.NoError(f.t, err)
 	// Load the state of the parent tipset and compute the required state (recursively).
 	prev := f.StateForKey(parentKey)
-	state, err := f.stateBuilder.ComputeState(prev, tipMessages(tip))
+	state, err := f.stateBuilder.ComputeState(prev, f.tipMessages(tip))
 	require.NoError(f.t, err)
 	return state
 }
 
+func (f *Builder) tipMessages(tip types.TipSet) [][]*types.SignedMessage {
```
shannonwells: This function could use a short explanatory comment re: purpose/intent

commands/chain.go
```diff
@@ -93,7 +93,8 @@ var chainLsCmd = &cmds.Command{
 					output.WriteString("\t")
 					output.WriteString(strconv.FormatUint(uint64(block.Height), 10))
 					output.WriteString("\t")
-					output.WriteString(strconv.Itoa(len(block.Messages)))
+					output.WriteString(block.Messages.String())
+					//					output.WriteString(block.MessageReceipts.String())
```
shannonwells: Why is this commented out rather than deleted?

metrics/heartbeat_test.go
```diff
@@ -186,14 +186,12 @@ func TestHeartbeatRunSuccess(t *testing.T) {
...
-		Height:          height,
-		Nonce:           0,
-		Messages:        nil,
-		MessageReceipts: nil,
+		Miner:        address.NewForTestGetter()(),
+		Ticket:       nil,
+		Parents:      types.TipSetKey{},
+		ParentWeight: 0,
+		Height:       height,
+		Nonce:        0,
```
shannonwells: Why aren't you setting Messages and MessageReceipts here to the empty(Messages|Receipts)Cid as above?

types/message.go
```diff
@@ -107,3 +107,34 @@ func (msg *Message) Equals(other *Message) bool {
...
+type MessageCollection []*SignedMessage
+
+// TODO #3078 the panics here and in types.Block should go away.  We need to
+// keep them in order to use the ipld cborstore with the default hash function
+// because we need to implement hamt.cidProvider which doesn't handle errors.
+// We can clean all this up when we can use our own CborIpldStore with the hamt.
+
+// Cid returns the cid of the message collection.
+func (mC MessageCollection) Cid() cid.Cid {
+	nd, err := cbor.WrapObject(mC, DefaultHashFunction, -1)
```
shannonwells: You could shorten this code by having them each call one private free function with themselves as a parameter.

types/block_test.go
```diff
@@ -130,26 +117,26 @@ func cidFromString(input string) (cid.Cid, error) {
...
-	newSignedMessage := NewSignedMessageForTestGetter(mockSigner)
 	t.Run("successfully decodes raw bytes to a Filecoin block", func(t *testing.T) {
 		addrGetter := address.NewForTestGetter()
 
 		c1, err := cidFromString("a")
 		assert.NoError(t, err)
 		c2, err := cidFromString("b")
 		assert.NoError(t, err)
+		cM, err := cidFromString("messages")
+		assert.NoError(t, err)
```
shannonwells: Maybe it would just be better to force the caller to always provide a string to cbor encode, make `cidFromString` an export/test helper and get rid of `SomeCid` entirely.

ZenGround0:
> I really wonder if this isn't putting the cart before the horse. It's a little late now given you want to get this in by Friday.

In retrospect I too wouldn't be too surprised if it would have been a better change without that much more work.  For the future the time to catch this would have been during review of [the plan doc](https://docs.google.com/document/u/1/d/1xOHUeM-svZoE3qQtOPCx1U96EQ6CuKsfB2GJYOHOk18/edit).

plumbing/msg/waiter.go
shannonwells: my vote is for MessageProvider if that counts for anything.

plumbing/msg/waiter.go
```diff
@@ -181,12 +190,17 @@ func (w *Waiter) receiptFromTipSet(ctx context.Context, msgCid cid.Cid, ts types
 		// TODO: this should return an error if a receipt doesn't exist.
```
shannonwells: it would be really cool if somebody filed an issue for this and noted it here . . .

plumbing/msg/waiter.go
acruikshank: I thought the point of graphsync was to allow us to retrieve only the data we need. In this case it seems wasteful to load all the block messages when we only care about one message and we know its CID. It is not possible to use graphsync here?

ZenGround0: There must have been a communication gap on the purpose of graphsync.  The point of graphsync is to fetch lots of ipld nodes over the network in one request.  This code is a local search of the blockchain data that has already been fetched.  There are other efforts to merkelize the message collection which could lead to better targeted searches in this code.  However these efforts are orthogonal.  Message collections are currently just cbor arrays of messages and not setup for any sort of smart searching.  The only way to find a message is to follow the hash link and search all of the messages underneath it.

types/testing.go
ZenGround0: Agreed, will add to a types-testing tech debt follow on issue.

ZenGround0: #3193

types/block_test.go
ZenGround0: #3193

plumbing/msg/waiter.go
ZenGround0: Thank you for flagging #3194

gengen/util/gengen.go
```diff
@@ -146,8 +146,24 @@ func GenGen(ctx context.Context, cfg *GenesisCfg, cst *hamt.CborIpldStore, bs bl
 		return nil, err
 	}
 
+	emptyMessagesCid, err := cst.Put(ctx, types.MessageCollection{})
```
acruikshank: Couldn't this be replaced with types.EmptyMessagesCid?

net/graphsync_fetcher_test.go
```diff
@@ -64,6 +73,8 @@ func TestGraphsyncFetcher(t *testing.T) {
 	t.Run("happy path returns correct tipsets", func(t *testing.T) {
 		gen := builder.NewGenesis()
 		final := builder.AppendOn(gen, 3)
+		//		emptyMsgsNode, err := cbor.WrapObject(types.MessageCollection{}, types.DefaultHashFunction, -1)
+		//		require.NoError(t, err)
```
acruikshank: delete me?

node/block_propagate_test.go
```diff
@@ -110,10 +112,18 @@ func TestChainSync(t *testing.T) {
 	minerWorker, err := ki[0].Address()
 	require.NoError(t, err)
 	stateRoot := baseTS.ToSlice()[0].StateRoot
+	msgsCid := types.MessageCollection{}.Cid()
+	rcptsCid := types.ReceiptCollection{}.Cid()
```
acruikshank: `types.EmptyMessageCid` and `types.EmptyRecieptCid`

plumbing/msg/waiter.go
```diff
@@ -178,15 +187,21 @@ func (w *Waiter) receiptFromTipSet(ctx context.Context, msgCid cid.Cid, ts types
 	var rcpt *types.MessageReceipt
 	if ts.Len() == 1 {
 		b := ts.At(0)
-		// TODO: this should return an error if a receipt doesn't exist.
+		// TODO #3194: this should return an error if a receipt doesn't exist.
```
acruikshank: 🙏

gengen/util/gengen.go
ZenGround0: This call is not really about the cid.  We need to put the empty collection to cbor storage so that the node can look it up.  To make sure that we catch issues where gengen and types/messages drift apart in implementation I brought in the explicit check below.

acruikshank: 👍

shannonwells:
> > I really wonder if this isn't putting the cart before the horse. It's a little late now given you want to get this in by Friday.
> 
> In retrospect I too wouldn't be too surprised if it would have been a better change without that much more work. For the future the time to catch this would have been during review of [the plan doc](https://docs.google.com/document/u/1/d/1xOHUeM-svZoE3qQtOPCx1U96EQ6CuKsfB2GJYOHOk18/edit).

Definitely. Did not occur to me then though.

ZenGround0:
> Definitely. Did not occur to me then though.

Yeah me neither 😄 .  Was more a note to self / the whole team than directed at you specifically.  In any case thanks for bringing critical questions to the design.

PR #3190, opened 2019-08-06
Switch to Graphsync Fetcher

hannahhoward:
# Goals

Use Graphsync Fetcher in the rest of the project!

# Implementation

I decided to add some utilities to Graphsync to make it easier to setup a Graphsync loader and storer function from a blockstore -- to simplify the code in Filecoin, and cause this will probably be valuable elsewhere.

Otherwise, this just changes a few lines in the actual node constructor.

### DIFF
--- a/net/graphsync_fetcher_test.go
+++ b/net/graphsync_fetcher_test.go
@@ -14,13 +14,13 @@ import (
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
-	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
 	dss "github.com/ipfs/go-datastore/sync"
 	"github.com/ipfs/go-graphsync"
 	"github.com/ipfs/go-graphsync/ipldbridge"
 	gsnet "github.com/ipfs/go-graphsync/network"
+	gsstoreutil "github.com/ipfs/go-graphsync/storeutil"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	format "github.com/ipfs/go-ipld-format"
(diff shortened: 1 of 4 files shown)


node/node.go
```diff
@@ -29,8 +33,8 @@ import (
 	"github.com/libp2p/go-libp2p-core/host"
 	p2pmetrics "github.com/libp2p/go-libp2p-core/metrics"
 	"github.com/libp2p/go-libp2p-core/routing"
-	"github.com/libp2p/go-libp2p-kad-dht"
-	"github.com/libp2p/go-libp2p-kad-dht/opts"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
```
hannahhoward: I have no idea why my editor did this? Anyone?

net/graphsync_fetcher_test.go
```diff
@@ -392,33 +392,8 @@ func TestRealWorldGraphsyncFetchAcrossNetwork(t *testing.T) {
...
-			block, err := blocks.NewBlockWithCid(buffer.Bytes(), asCidLink.Cid)
-			if err != nil {
-				return err
-			}
-			return bs.Put(block)
-		}
-		return &buffer, committer, nil
-	}
+	localLoader := gsstoreutil.LoaderForBlockstore(bs)
+	localStorer := gsstoreutil.StorerForBlockstore(bs)
```
ZenGround0: 🥇

node/node.go
ZenGround0: They like to do this sometimes.  Please change it back

node/node.go
```diff
@@ -388,7 +392,13 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
 	//nwork := bsnet.NewFromIpfsHost(innerHost, router)
 	bswap := bitswap.New(ctx, nwork, bs)
 	bservice := bserv.New(bs, bswap)
-	fetcher := net.NewBitswapFetcher(ctx, bservice, blkValid)
```
ZenGround0: Once we are done with the next release and the graphsync fetcher has been tried by fire there will be very little reason to keep a bitswap fetcher around in the code.  Would you mind filing an issue for removing that code once the next release comes and goes?

hannahhoward: #3191

node/node.go
```diff
@@ -388,7 +392,13 @@ func (nc *Config) Build(ctx context.Context) (*Node, error) {
...
+
+	graphsyncNetwork := gsnet.NewFromLibp2pHost(peerHost)
+	bridge := ipldbridge.NewIPLDBridge()
+	loader := gsstoreutil.LoaderForBlockstore(bs)
+	storer := gsstoreutil.StorerForBlockstore(bs)
+	gsync := graphsync.New(ctx, graphsyncNetwork, bridge, loader, storer)
+	fetcher := net.NewGraphSyncFetcher(ctx, gsync, bs, blkValid, peerTracker)
```
acruikshank: All this DI is appropriate for this function, but it's a lot of detail added to an already too long method. It would be nice if we could expose a fetcher constructor that takes a peerHost, peerTracker and a block store from the net package. I'm not familiar enough with the abstractions to say if this is a bad idea.

hannahhoward: gonna make a seperate ticket for this. Thanks!

hannahhoward: Follow-up: I think it would be better not to do this, because the high odds are we are going to find other uses for Graphsync soon, and may want it available on the node, similar to how the block service (backed by Bitswap) is availabe on the node.

acruikshank: Just to be clear. I wasn't suggesting we remove this from Node altogether, just that we push the construction of the GraphSyncFetcher dependencies into the net package.

PR #3195, opened 2019-08-07
Client validates PIP

### LINKED ISSUE #2676: Client Validates piece inclusion proof in deal
Blocked on #2571

### Description

Once a deal is in the POSTED state, it should include a sector id and a piece inclusion proof (#2571). The proof proves that the client's piece is included within the sector represented by the sector id.

Upon making a deal state request and seeing the deal is in the posted state, the client should check for the existence of the pip and check it's validity. This can be done using `verifyPieceInclusion` on the storage miner's miner actor (#2625).

If the piece is not valid, the client should immediately cancel payment by calling cancel on the payment channel (#2621). If the miner cannot respond to the cancel by posting a voucher with a valid proof on chain, the client will be able to retrieve their funds.

### Acceptance criteria

1. Upon receiving a POSTED deal state, the client waits for the commitSector message and then runs a `verifyPieceInclusion` query. If the query succeeds, halt.
2. If the query fails, the client immediately calls `cancel` on their payment channel.
3. This flow with both a success and failure case is integration tested.

ingar:
## Problem
PIP generation is happening during deal creation, but it isn't being verified.

## Solution
Add a new client command, `verify-deal-proposal`, that checks that the returned deal from the storage miner is in the [`Complete`](https://github.com/filecoin-project/go-filecoin/blob/fe3d41143e81d461d8f049f0ac739b391b5aa342/protocol/storage/storagedeal/state.go#L33) state, and that the PIP returned in the [`ProofInfo`](https://github.com/filecoin-project/go-filecoin/blob/fe3d41143e81d461d8f049f0ac739b391b5aa342/protocol/storage/storagedeal/types.go#L113) is valid.

## Note
This command doesn't cancel the payment channel if the PIP is invalid; the client must call `paych cancel` with the channel id, which is stored in the proposal.

Resolves #2676

### DIFF
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -1260,15 +1260,22 @@ func TestVerifyPIP(t *testing.T) {
 		msgParams := actor.MustConvertParams(commP, pieceSize, firstSectorID, pip)
 		message := types.NewMessage(address.TestAddress, address.TestAddress2, 0, types.ZeroAttoFIL, "verifyPieceInclusion", msgParams)
 
+		comm1 := th.MakeCommitments()
+		comm2 := th.MakeCommitments()
+
+		commitments := NewSectorSet()
+		commitments.Add(1, types.Commitments{CommR: comm1.CommR})
+		commitments.Add(2, types.Commitments{CommR: comm2.CommR})
+
 		minerState := *NewState(address.TestAddress, address.TestAddress, peer.ID(""), types.OneKiBSectorSize)
-		minerState.SectorCommitments = nil
+		minerState.SectorCommitments = commitments
 
 		vmctx := th.NewFakeVMContext(message, minerState)
 		miner := Actor{}
 
 		code, err := miner.VerifyPieceInclusion(vmctx, commP, pieceSize, firstSectorID, pip)
 
-		require.Error(t, err, "not active")
+		require.EqualError(t, err, "miner not active")
 		require.NotEqual(t, uint8(0), code)
 	})
 
@@ -1298,7 +1305,7 @@ func TestVerifyPIP(t *testing.T) {
 
 		code, err := miner.VerifyPieceInclusion(vmctx, commP, pieceSize, firstSectorID, pip)
 
-		require.Error(t, err, "miner is tardy")
+		require.EqualError(t, err, "miner is tardy")
 		require.NotEqual(t, uint8(0), code)
 	})
 
--- a/actor/builtin/miner/miner_test.go
+++ b/actor/builtin/miner/miner_test.go
@@ -1373,6 +1373,40 @@ func TestVerifyPIP(t *testing.T) {
 	})
 }
 
+func TestDoVerifyPIP(t *testing.T) {
+	tf.UnitTest(t)
+
+	firstSectorCommitments := th.MakeCommitments()
+	firstSectorCommD := firstSectorCommitments.CommD
+	firstSectorID := uint64(1)
+
+	sectorSetWithOneCommitment := NewSectorSet()
+	sectorSetWithOneCommitment.Add(firstSectorID, firstSectorCommitments)
+
+	commP := th.MakeCommitment()
+
+	t.Run("Sends correct parameters to PIP verifier", func(t *testing.T) {
+		vmctx, verifier, minerActor := (&minerEnvBuilder{
+			message:   "verifyPieceInclusionProof",
+			sectorSet: sectorSetWithOneCommitment,
+			lastPoSt:  types.NewBlockHeight(PieceInclusionGracePeriodBlocks),
+			verifier: &verification.FakeVerifier{
+				VerifyPieceInclusionProofValid: true,
+			},
+		}).build()
+
+		code, err := minerActor.DoVerifyPieceInclusion(vmctx, commP, types.NewBytesAmount(66), firstSectorID, []byte{42})
+		require.NoError(t, err)
+		require.Equal(t, 0, int(code), "should be successful")
+
+		require.NotNil(t, verifier.LastReceivedVerifyPieceInclusionProofRequest)
+		require.Equal(t, verifier.LastReceivedVerifyPieceInclusionProofRequest.CommP[:], commP)
+		require.Equal(t, verifier.LastReceivedVerifyPieceInclusionProofRequest.CommD, firstSectorCommD)
+		require.Equal(t, verifier.LastReceivedVerifyPieceInclusionProofRequest.PieceSize, types.NewBytesAmount(66))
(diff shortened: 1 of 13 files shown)


commands/client.go
```diff
@@ -186,6 +188,8 @@ format is specified with the --enc flag.
 			return err
 		}
 
+		fmt.Printf("storagedeal response: %+v", resp)
+
```
travisperson: Extra print statement

commands/client.go
```diff
@@ -198,6 +202,48 @@ format is specified with the --enc flag.
...
+			return errors.New("storage deal not in Complete state")
+		}
+
+		validateError := GetPorcelainAPI(env).ClientValidateDeal(req.Context, proposalCid, resp.ProofInfo)
+
+		if err := re.Emit(VerifyStorageDealResult{validateError == nil}); err != nil {
+			return err
+		}
+
+		return err
```
travisperson: The context of this err value is not apparent, just returning `re.Emit(...)` will help make this easier to read.

porcelain/client.go
```diff
@@ -91,6 +94,39 @@ func listAsksFromActorResult(ctx context.Context, plumbing claPlubming, actorRes
...
+	params := []interface{}{
+		deal.CommP[:],
+		deal.Proposal.Size,
+		proofInfo.SectorID,
+		proofInfo.PieceInclusionProof,
+	}
+
+	_, err = plumbing.MessageQuery(ctx, address.Undef, deal.Miner, "doVerifyPieceInclusion", params...)
+	if err != nil {
+		fmt.Printf("verify error: %v", err)
```
travisperson: Extra print statement

tools/fast/tests/retrieval_test.go
```diff
@@ -119,6 +119,9 @@ func TestRetrievalLocalNetwork(t *testing.T) {
 	RunRetrievalTest(ctx, t, miner, client, sectorSize)
 }
 
+// Current issue:
+// miner state LastPoSt is nil during deal finalization, so verifyPieceInclusion will never succeed
+
```
travisperson: What is the fix for this? Do we need to be able to wait for a PoSt for the sector containing the deal piece to show up on chain?

ingar: We are going to not rely on `LastPoSt` when checking PIPs.  We are simply going to ensure that the miner is in a valid proving period and that the commitment is on-chain before we verify the PIP.

commands/client.go
ingar: Thanks, that was unintentional; the result of code thrashing.

actor/builtin/miner/miner_test.go
```diff
@@ -1254,41 +1256,41 @@ func TestVerifyPIP(t *testing.T) {
 		require.Nil(t, verifier.LastReceivedVerifyPieceInclusionProofRequest, "should have never called the verifier")
 	})
 
-	t.Run("PIP is invalid if miner has not submitted any PoSt proofs", func(t *testing.T) {
-		vmctx, verifier, minerActor := (&minerEnvBuilder{
-			message:   "verifyPieceInclusionProof",
-			sectorSet: sectorSetWithOneCommitment,
-			lastPoSt:  nil,
-		}).build()
+	t.Run("PIP is invalid if miner is tardy/slashable", func(t *testing.T) {
```
acruikshank: Non-Blocking: It would be nice to have an additional test for the condition where the miner is not active.

porcelain/client.go
```diff
@@ -2,12 +2,14 @@ package porcelain
...
-	"math/big"
-
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+	"github.com/pkg/errors"
+	"math/big"
```
acruikshank: This belongs in the section with "context"

porcelain/client.go
```diff
@@ -2,12 +2,14 @@ package porcelain
...
 	"context"
-	"math/big"
-
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/address"
+	"github.com/filecoin-project/go-filecoin/protocol/storage/storagedeal"
 	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
+	"github.com/ipfs/go-cid"
+	"github.com/pkg/errors"
```
acruikshank: These belong with go-ipld-cbor down below.

tools/fast/action_client.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/filecoin-project/go-filecoin/commands"
```
acruikshank: This belongs with address down below.

tools/fast/tests/retrieval_test.go
acruikshank: So does this comment go away?

tools/fast/series/wait_for_deal_state.go
```diff
@@ -9,12 +9,12 @@ import (
 
 // WaitForDealState will query the storage deal until its state matches the
 // passed in `state`, or the context is canceled.
-func WaitForDealState(ctx context.Context, client *fast.Filecoin, deal *storagedeal.Response, state storagedeal.State) error {
+func WaitForDealState(ctx context.Context, client *fast.Filecoin, deal *storagedeal.Response, state storagedeal.State) (dr *storagedeal.Response, err error) {
```
acruikshank: Non-Blocking: We're moving away from named return values in general. I think it's adding an extra layer of indirection here. My preference would be to not name the parameter and to return `dr` directly on line 21.

actor/builtin/miner/miner_test.go
```diff
@@ -1254,41 +1256,57 @@ func TestVerifyPIP(t *testing.T) {
...
+			VerifyPieceInclusionProofValid: true,
+		}
+
+		vmctx := th.NewFakeVMContextWithVerifier(message, minerState, verifier)
+		vmctx.BlockHeightValue = minerState.ProvingPeriodEnd.Add(GenerationAttackTime(minerState.SectorSize)).Add(types.NewBlockHeight(1))
+		miner := Actor{}
+
+		code, err := miner.VerifyPieceInclusion(vmctx, commP, pieceSize, firstSectorID, pip)
+
+		require.Error(t, err, "miner is tardy")
```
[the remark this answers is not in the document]
ingar: Thanks

PR #3200, opened 2019-08-08
misc: load git submodules before deps

### LINKED ISSUE #3211: Deps step requires loading submodules before running when go dependencies present
### Description
There is a dependency issue when adding _go_ submodules that causes our **go run ./build deps** to fail because it can't find the unloaded packages when executing the `./build/main.go` script. 

We run into this issue when attempting to move the CGO bindings for _sectorbuilder_ and _bls-signatures_ into their own repos and integrating them as submodules.  

### Acceptance criteria

- update readme with new steps 
- a deps / build sequence that allows for having go submodules. 

### Risks + pitfalls

- an extra step when downloading the code for the first time 

### Where to begin

- readme
- circleci, docker, appveyor scripts.

icorderi:
Closes #3211

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,9 @@ jobs:
             mkdir $HOME/.bin
             curl --location https://github.com/stedolan/jq/releases/download/jq-1.6/jq-osx-amd64 --output $HOME/.bin/jq
             chmod +x $HOME/.bin/jq
+      - run:
+          name: Load submodules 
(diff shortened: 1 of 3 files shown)


acruikshank:
@icorderi Could you create and/or reference an issue for this change? Up until now, running `deps` has been sufficient to load the submodules. I'd like to understand why that is no longer the case.

icorderi:
@acruikshank #3211 describes the issue

PR #3203, opened 2019-08-08
Update CLI show command

### LINKED ISSUE #3196: Update show command to handle messages out of blocks
### Description
After some initial exploration into updating the block explorer and the network stats dashboard and some helpful hints from @travisperson it looks like decoding the values output from `dag get` on the front end will be more onerous than it is worth.  More importantly that decoding is really go-filecoin's job so the node's API should do that for us.

### Acceptance criteria
Three new pieces of CLI functionality
(1) a --full flag on `show block` which returns the block header with messages and receipts embedded directly in it.  We'll use this flag to keep the old behavior
(2) a `show messages` command that returns a properly decoded SignedMessage array given the `Messages` cid from a block header
(3) a `show receipts` command.  Ditto (2) for receipts

These three userfacing endpoints will rely on the following internals:
(a) a new `FullBlock` type which may be useful elsewhere 
(b) `GetMessages` function on the chain state plumbing/porcelain
(c) `GetReceipts` function on the chain state plumbing/porcelain

b & c are very close to being done because the prep work already piped a message store to the chain state.

### Where to begin
start with the internals

ZenGround0:
Closes #3196.  

Some changes since writing that issue: it makes more sense to use a separate command for the header and the full block rather than controlling with a flag because the cmds lib can only return a single type from each CLI call.  I kept `block` as the name of the full block command and now `header` is the block header show command.

Note that the old`show block` api is still broken because now the structure of the output is like 

```
{
    Header {
        Parents: ...
        Ticket: ...
        Timestamp: ...
        ...
    }
    Messages: [msg1, msg2...]
    Receipts: [rcpt1, rcpt2,...]
}
```

instead of the previous:

```
{
    Parents: ...
    Ticket: ...
    Timestamp: ...
    ... 
    Messages: [msg1, msg2...]
    Receipts: [rcpt1, rcpt2...]
}
```

### DIFF
--- a/commands/show.go
+++ b/commands/show.go
@@ -149,7 +149,8 @@ var showMessagesCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Show a filecoin message collection by its CID",
 		ShortDescription: `Prints info for all messages in a collection,
-at the given CID.`,
+at the given CID.  Message collection CIDs are found in the "Messages" field of 
+the filecoin block header.`,
 	},
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("cid", true, false, "CID of message collection to show"),
@@ -184,7 +185,8 @@ var showReceiptsCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Show a filecoin receipt collection by its CID",
 		ShortDescription: `Prints info for all receipts in a collection,
-at the given CID.`,
+at the given CID.  Receipt collection CIDs are found in the "MessageReceipts"
+field of the filecoin block header.`,
 	},
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("cid", true, false, "CID of receipt collection to show"),
--- a/commands/show.go
+++ b/commands/show.go
@@ -23,7 +23,7 @@ var showCmd = &cmds.Command{
 
 var showBlockCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
-		Tagline: "Show a filecoin block by its CID",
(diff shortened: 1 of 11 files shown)


travisperson:
Awesome! I think we can work through this structure change pretty easily though.

commands/show.go
```diff
@@ -68,14 +140,76 @@ Timestamp:  %s
...
-			if err != nil {
-				return err
+			return err
+		}),
+	},
+}
+
+var showMessagesCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Show a filecoin message collection by its CID",
```
shannonwells: I think it would be helpful to add a quick line about how to get the message collection CID, similarly do the same with showReceiptsCmd.  Just something like "(obtained from the Block Header)" or something like that.

commands/show_test.go
```diff
@@ -3,10 +3,12 @@ package commands_test
 import (
 	"encoding/json"
 	"testing"
+	//	"fmt"
```
shannonwells: remove comment

commands/show_test.go
```diff
@@ -62,15 +64,121 @@ func TestBlockDaemon(t *testing.T) {
...
+		// ensure that we were returned the correct block
+
+		require.Equal(t, minedBlockCidStr, headerGetBlock.Cid().String())
+
+		// ensure that the JSON we received from block get conforms to schema
+
+		requireSchemaConformance(t, []byte(headerGetLine), "filecoin_block")
+	})
+
+	t.Run("show messages <empty-collection-cid> returns empty message collection", func(t *testing.T) {
```
shannonwells: need a similar test for show receipts <empty-collection-cid>

PR #3214, opened 2019-08-09
Have miner confirm that payment conditions are valid before continuing with payment deal

### LINKED ISSUE #2572: Miner confirms client's Data Commitment
### Description

The client will be including a data commitment within each of the payments they make to a miner. If this data commitment isn't correct, the miner may not get paid for storage. The miner must validate the data commitment before staging the piece.

After the miner receives all the data from the client, they will generate the data commitment using the same rust proofs function used by the client to generate the data commitment in the first place. If the commitment doesn't match the commitment in all the vouchers, the deal is off.

### Acceptance criteria

Miner rejects deal if data commitment in any of the payment vouchers does not match the data commitment the miner generates from the received data, the miner will terminate the deal.

acruikshank:
closes #2572

### Problem

A miner does quite a bit of validation of payment data prior to accepting the deal. Unfortunately, the miner cannot confirm the client has specified the correct commitment until they are in possession of the client's data.

### Solution

Once the client has exported their data, have the miner generate a piece commitment from it. Then have the miner validate all the payments go the the right method on the right miner with the right CommP and piece size.

### DIFF
--- a/porcelain/api.go
+++ b/porcelain/api.go
@@ -60,9 +60,9 @@ func (a *API) CreatePayments(ctx context.Context, config CreatePaymentsParams) (
 	return CreatePayments(ctx, a, config)
 }
 
-// ValidateStoragePaymentCondition validates that the given condition is a payment condition and has the right values
-func (a *API) ValidateStoragePaymentCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
-	return ValidateStoragePaymentCondition(ctx, condition, minerAddr, commP, pieceSize)
+// ValidatePaymentVoucherCondition validates that the given condition is a payment condition and has the right values
+func (a *API) ValidatePaymentVoucherCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
+	return ValidatePaymentVoucherCondition(ctx, condition, minerAddr, commP, pieceSize)
 }
 
 // DealGet returns a single deal matching a given cid or an error
--- a/porcelain/api.go
+++ b/porcelain/api.go
@@ -60,6 +60,11 @@ func (a *API) CreatePayments(ctx context.Context, config CreatePaymentsParams) (
 	return CreatePayments(ctx, a, config)
 }
 
+// ValidateStoragePaymentCondition validates that the given condition is a payment condition and has the right values
+func (a *API) ValidateStoragePaymentCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
+	return ValidateStoragePaymentCondition(ctx, condition, minerAddr, commP, pieceSize)
+}
+
 // DealGet returns a single deal matching a given cid or an error
 func (a *API) DealGet(ctx context.Context, proposalCid cid.Cid) (*storagedeal.Deal, error) {
 	return DealGet(ctx, a, proposalCid)
--- a/porcelain/payments.go
+++ b/porcelain/payments.go
@@ -185,6 +185,60 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
 	return response, nil
 }
 
+// ValidateStoragePaymentCondition validates that condition of a voucher created for a storage payment meets expectations
+func ValidateStoragePaymentCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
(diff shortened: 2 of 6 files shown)


porcelain/api.go
```diff
@@ -55,6 +55,11 @@ func (a *API) CreatePayments(ctx context.Context, config CreatePaymentsParams) (
 	return CreatePayments(ctx, a, config)
 }
 
+// ValidateStoragePaymentCondition validates that the given condition is a payment condition and has the right values
+func (a *API) ValidateStoragePaymentCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
+	return ValidateStoragePaymentCondition(ctx, condition, minerAddr, commP, pieceSize)
+}
```
anorth: I don't see why this is in porcelain since it's just a free function, with no dependency on plumbing or the `a` receiver. I think it should be moved to `protocol/storage`. It's exposed unnecessarily widely here.

porcelain/payments.go
```diff
@@ -185,6 +185,60 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
...
 
+// ValidateStoragePaymentCondition validates that condition of a voucher created for a storage payment meets expectations
+func ValidateStoragePaymentCondition(ctx context.Context, condition *types.Predicate, minerAddr address.Address, commP types.CommP, pieceSize *types.BytesAmount) error {
+	// a nil condition is always valid
+	if condition == nil {
+		return nil
+	}
+
+	if condition.To != minerAddr {
+		return errors.New("voucher condition addressed to wrong address")
```
anorth: "... addressed to %s, expected %s"

porcelain/payments.go
```diff
@@ -185,6 +185,60 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
...
+	// lack of type data in params means commP isn't necessarily cbor decoded
+	clientCommPBytes, ok := condition.Params[0].([]byte)
+	if ok {
+		copy(clientCommP[:], clientCommPBytes)
+	} else {
+		clientCommP, ok = condition.Params[0].(types.CommP)
+		if !ok {
+			return errors.New("piece commitment is not a CommP")
+		}
+	}
```
anorth: Surely this should be exactly one or the other. We're writing the code at both ends. Even if it's necessarily opaque, can we pick one?

porcelain/payments.go
```diff
@@ -185,6 +185,60 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
...
+
+	if clientCommP != commP {
+		return errors.New("piece commitment does not match commitment supplied in payment condition")
```
anorth: Include a string serialisation the actual and expected in the message

porcelain/payments.go
```diff
@@ -185,6 +185,60 @@ func CreatePayments(ctx context.Context, plumbing cpPlumbing, config CreatePayme
...
+	// lack of type data in params means piece size isn't necessarily cbor decoded
+	clientPieceSizeBytes, ok := condition.Params[1].([]byte)
+	if ok {
+		clientPieceSize = types.NewBytesAmountFromBytes(clientPieceSizeBytes)
+	} else {
+		clientPieceSize, ok = condition.Params[1].(*types.BytesAmount)
+		if !ok {
+			return errors.New("piece size is not a bytes amount")
+		}
+	}
```
anorth: Ditto my comment above about supporting just one serialisation.

protocol/storage/miner.go
```diff
@@ -476,6 +487,32 @@ func (sm *Miner) processStorageDeal(proposalCid cid.Cid) {
...
+	if err != nil {
+		return err
+	}
+
+	// Generating the piece commitment is a computationally expensive operation and can take
+	// many minutes depending on the size of the piece.
+	pieceCommitmentResponse, err := proofs.GeneratePieceCommitment(proofs.GeneratePieceCommitmentRequest{
+		PieceReader: pieceReader,
+		PieceSize:   types.NewBytesAmount(deal.Proposal.Size.Uint64()),
+	})
```
anorth: This will have a very noticeable operational effect for miners - those doing automated large deals will see a big CPU spike. We must make sure they are not surprised by this.

porcelain/api.go
acruikshank: I was hoping to keep the payment creation logic and the payment validation logic co-located. It also reduces clutter in the overcrowded client file. It's not inconceivable, that you'd want this type of validation in the form of command.

All that said, I agree it's not very porcelain-y and it doesn't really need to be here. I'm ambivalent. I'll be happy to move it in a quick follow-on PR if you still think it doesn't belong here.

PR #3222, opened 2019-08-12
sign and verify deal response

### LINKED ISSUE #1211: Secure Deal Protocol
### Description

Currently, the the miner responding to a storage deal does not actually sign the response. The miner must sign the response to allow deal arbitration.

### Acceptance criteria

1. Miner signs deal response with a signature that validates against the miner's miner actor's owner.
2. Client validates the signature and rejects the deal if it's invalid (reject only does what storage deal errors currently do).
3. Signing and validation are tested.

acruikshank:
closes #1211

### Problem

The storage miner's deal response has contained a `Signature` field for some time, but we've never actually populated it. It's critical that the miner signs the deal, because the response forms a contract that can be posted on chain for deal arbitration.

### Solution

Sign the deal proposal when the deal is accepted, when it's rejected, and when it's updated. Also have the client check the signature whenever a deal response is received.

### Refactors

This PR refactors `protocols/storage/miner` to abstract just the processing of the deal. This allows us to test the deal acceptance and deal rejection.

This PR also adds porcelain to retrieve the miner worker address because that is the address used to sign the response.

### DIFF
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -75,10 +75,9 @@ type Miner struct {
 	porcelainAPI minerPorcelain
 	node         node
 
-	proposalAcceptor func(m *Miner, p *storagedeal.Proposal) (*storagedeal.Response, error)
-	proposalRejector func(m *Miner, p *storagedeal.Proposal, reason string) (*storagedeal.Response, error)
-
 	storageFaultSlasher storageFaultSlasher
+
+	proposalProcessor func(context.Context, *Miner, cid.Cid)
 }
 
 // minerPorcelain is the subset of the porcelain API that storage.Miner needs.
@@ -96,6 +95,7 @@ type minerPorcelain interface {
 	MessageSend(ctx context.Context, from, to address.Address, value types.AttoFIL, gasPrice types.AttoFIL, gasLimit types.GasUnits, method string, params ...interface{}) (cid.Cid, error)
 	MessageQuery(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
 	MessageWait(ctx context.Context, msgCid cid.Cid, cb func(*types.Block, *types.SignedMessage, *types.MessageReceipt) error) error
+	SectorBuilder() sectorbuilder.SectorBuilder
 	types.Signer
 }
 
@@ -110,7 +110,6 @@ type prover interface {
 type node interface {
 	BlockService() bserv.BlockService
 	Host() host.Host
-	SectorBuilder() sectorbuilder.SectorBuilder
 }
 
 // NewMiner is for construction of a new storage miner.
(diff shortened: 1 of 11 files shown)


protocol/storage/miner.go
```diff
@@ -375,23 +379,29 @@ func rejectProposal(sm *Miner, p *storagedeal.Proposal, reason string) (*storage
 }
 
 func (sm *Miner) updateDealResponse(ctx context.Context, proposalCid cid.Cid, f func(*storagedeal.Response)) error {
```
shannonwells: There needs to be a comment explaining what the purpose of `f` is.  The way it's called in the code does not help.

protocol/storage/miner_test.go
```diff
@@ -66,31 +54,26 @@ func TestReceiveStorageProposal(t *testing.T) {
 		_, err := miner.receiveStorageProposal(context.Background(), proposal)
 		require.NoError(t, err)
 
-		assert.True(t, accepted, "Proposal has been accepted")
-		assert.False(t, rejected, "Proposal has not been rejected")
-		assert.Equal(t, "", message)
+		// one deal should be stored and it should have been accepted
```
shannonwells: add 'and signed'  since this is being checked here

protocol/storage/miner_test.go
```diff
@@ -99,12 +82,15 @@ func TestReceiveStorageProposal(t *testing.T) {
...
+		require.Len(t, porcelainAPI.deals, 1)
+		for _, deal := range porcelainAPI.deals {
+			assert.Equal(t, storagedeal.Accepted, deal.Response.State)
```
shannonwells: should there be a signature check here too or is this covered by the other tests?

protocol/storage/miner.go
```diff
@@ -344,12 +346,12 @@ func acceptProposal(sm *Miner, p *storagedeal.Proposal) (*storagedeal.Response,
 	}
 
 	// TODO: use some sort of nicer scheduler
```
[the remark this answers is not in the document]
acruikshank: We don't. This is old. I'm not really sure what a better scheduler would look like. Some of these comments are just to remind us that this miner won't work at scale. It's ok for an initial release, though.

protocol/storage/miner_test.go
acruikshank: I broke out the assertion that the response is signed when the deal is accepted into a separate test.

protocol/storage/miner_test.go
acruikshank: This comment convinced me that the signing of accepted deals should have it's own test, so I took it out of the first test and added a new test to check signing specifically. I also refactored a bit to reduce the test setup in these tests.

ingar:
When we had looked at this issue previously (before the worker address was a thing), we had proposed a change to the spec [and got it accepted](https://github.com/filecoin-project/specs/blob/master/network-protocols.md#storage-deal-make) to have a signature envelope over the proposal response, so that we wouldn't have to do the `response.signature = nil`, sign response, etc dance.

```
type SignedStorageDealResponse struct {
	response StorageDealResponse

	## Signature is a signature from the miner over the cbor encoded response
	signature Signature
}
```
Seems like this PR should include that?

acruikshank:
@ingar I went ahead and created the SignedResponse wrapper. It created a lot of churn, but seems to work ok.

Issue #3220, opened 2019-08-12
[bug] Failure to sync tipset received from hello, could not load message collection


anorth:
**Describe the bug**
On the 0.4.0 staging network, node are failing to sync the chain past height 55, because the message collection for a block at height 56 is not found in the store.

First reported in #3207, this issue is focussed on that specific problem.

It appears that there's a chain of ~3k blocks, but it doesn't validate.

```
13:44:29.027  INFO chain.sync: processing block 0 of 3040 for chain with head at { bafy2bzaceacrcn3ye7izsr7v26xwenzchjz5ocobtt3svpvmlm662b5p3ys7q bafy2bzaceazddktvonc3qijgiceri6tqnrzezi2n23yptucndojvugu3z52dc bafy2bzacecwnkysjfixkpuf5lvqyq3sho67vaksvu2l7yuzkow27633fbg4ay } syncer.go:396
```

Later...
```
13:44:29.580  INFO       node: error handling blocks: { bafy2bzaceacrcn3ye7izsr7v26xwenzchjz5ocobtt3svpvmlm662b5p3ys7q bafy2bzaceazddktvonc3qijgiceri6tqnrzezi2n23yptucndojvugu3z52dc bafy2bzacecwnkysjfixkpuf5lvqyq3sho67vaksvu2l7yuzkow27633fbg4ay } node.go:573
13:44:29.955  INFO       node: error handling blocks: { bafy2bzaceacrcn3ye7izsr7v26xwenzchjz5ocobtt3svpvmlm662b5p3ys7q bafy2bzaceazddktvonc3qijgiceri6tqnrzezi2n23yptucndojvugu3z52dc bafy2bzacecwnkysjfixkpuf5lvqyq3sho67vaksvu2l7yuzkow27633fbg4ay } node.go:573
13:44:30.428  INFO       node: error handling blocks: { bafy2bzaceacrcn3ye7izsr7v26xwenzchjz5ocobtt3svpvmlm662b5p3ys7q bafy2bzaceazddktvonc3qijgiceri6tqnrzezi2n23yptucndojvugu3z52dc bafy2bzacecwnkysjfixkpuf5lvqyq3sho67vaksvu2l7yuzkow27633fbg4ay } node.go:573
[...]

PR #3226, opened 2019-08-13
Graphsync fetcher fixed recursive selector to include messages/receipts, verifies their fetch

anorth:
With these changes, I have synced a node to the 0.4.0 staging devnet.

Fixes #3220

PR #3227, opened 2019-08-13
fix(net): fix bug w/ message fetching via graphsync

hannahhoward:
Preliminary fix for #3220

The test (integration test) is crappy but proves the fix. BTW, it requires a very specific combination of things to fail -- the height of 22 is intentional as it's
first layer + 1 recursive layer + 4 recursive layers +16 recursive layers


anorth:
Thank you! The fixed query works for me and I can sync against our running staging 0.4.0 network. I will integrate it into my PR #3226 which also extends the checking within the fetcher that all expected CIDs are stored.

PR #3228, opened 2019-08-13
FAST Updates

travisperson:
Set of FAST updates 

**New Actions**
- ShowMessages
- ShowReceipts
- ClientPayments

**Features**
- InitAndStart series takes a variadic set of functions to apply between `InitDaemon` and `StartDaemon`. This helps with applying configuration changes before the daemon starts.
- CtxSleepDelay now returns a channel which receives a value after the delay time. This allows the function to be more versatile allowing for use inside of `select` statements.
- CreateStorageMinerWithAsk now requires a sector size for the miner.

**Bugs**
- When tearing down processes with `TeardownProcess`, new processes will not use the same directory as a previous process.
- DHTFindProvs now returns a `*json.Decoder` so more than one value can be parsed from the LDJSON stream.

### DIFF
--- a/tools/fast/series/init_and_start.go
+++ b/tools/fast/series/init_and_start.go
@@ -6,10 +6,17 @@ import (
 	"github.com/filecoin-project/go-filecoin/tools/fast"
 )
 
-// InitAndStart is a quick way to run Init and Start for a filecoin process.
-func InitAndStart(ctx context.Context, node *fast.Filecoin) error {
+// InitAndStart is a quick way to run Init and Start for a filecoin process. A variadic set of functions
+// can be passed to run between init and the start of the daemon to make configuration changes.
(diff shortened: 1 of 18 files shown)


tools/fast/series/init_and_start.go
```diff
@@ -7,11 +7,17 @@ import (
 )
 
 // InitAndStart is a quick way to run Init and Start for a filecoin process.
-func InitAndStart(ctx context.Context, node *fast.Filecoin) error {
+func InitAndStart(ctx context.Context, node *fast.Filecoin, fns ...func(context.Context, *fast.Filecoin) error) error {
```
ingar: Maybe just a comment about at what point `fns` are intended to be run (even though it's right there in the code)

Issue #3027, opened 2019-07-08
CollectTipSetOfHeightAtLeast new OBO error


ZenGround0:
### Description
Because of the recent fix in #3026 we are now missing one tipset in the ancestor subchain in the case when the sample height falls on a null block.  This bug should only be an issue in the case a PoSt is submitted at the very last possible block.  In that case validators will incorrectly fail with ["sample height out of range"](https://github.com/filecoin-project/go-filecoin/pull/3026/files#diff-0169262ea0c745894301dc43c1446163R56)

### Acceptance criteria
- Modify `CollectTipSetsOfHeightAtLeast` to handle this OBO error
- Simplify the `GetRecentAncestors` code which has some complexity that will go away when making this change.  Specifically we no longer need two iterators because `CollectTipSetsOfHeightAtLeast` will no longer drop a value needed by `CollectAtMostNTipSets`
- Test this edge case

anorth:
See also #3025

ZenGround0:
@deaswang to clarify a bit to help you with #3250 the major fix that this issue is tracking is to make `CollectTipSetsOfHeightAtLeast` return all tipsets up to and including the first tipset with a height less than or equal to the input height.  The reason is that [this change](https://github.com/filecoin-project/go-filecoin/pull/3026/files#diff-0169262ea0c745894301dc43c1446163R27) in PR #3026 modified the sampling behavior during validation to expect all of these tipsets but did not modify the sampling behavior during tipset collection.

tl;dr `CollectTipSetsOfHeightAtLeast` should traverse up to `minHeight` (as it does now) *and additionally* include the first tipset less than `minHeight` if there is no tipset at `minHeight`.

PR #3234, opened 2019-08-14
add empty slice check for TipSets

### LINKED ISSUE #3188: [bug] panic: index out of range in chain/get_ancestors
The log errors during the running of the program are as follows: My server is configured with 24 cores, 48 threads and 24G memory. I don't know what caused it. It's doing storage transactions.

deaswang:
fixes #3188

this issue should be miss empty slice check for iterator is empty.

### DIFF
--- a/chain/get_ancestors.go
+++ b/chain/get_ancestors.go
@@ -99,6 +99,9 @@ func CollectTipSetsOfHeightAtLeast(ctx context.Context, iterator *TipsetIterator
 		}
 		ret = append(ret, iterator.Value())
 	}
+	if len(ret) == 0 {
+		return nil, errors.New("empty TipSet")
+	}
 	return ret, nil
 }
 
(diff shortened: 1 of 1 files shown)


anorth:
I don't think this is the right fix. An empty slice is a reasonable return value for this function, a check for it might be appropriate at some call sites if they're expecting non-empty.

If it is the problem described in #3027 then I expect the fix as @ZenGround0 describes it. I'll hand this over to @ZenGround0 to shepherd.

deaswang:
@anorth because only when collect empty tipsets slice, then `	firstExtraRandomnessAncestorsCids, err := provingPeriodAncestors[len(provingPeriodAncestors)-1].Parents()` will produce index out of range panic. and I have fix #3027 by pull #3250 .

ZenGround0:
Thanks @deaswang.  As @anorth suggests, the "right place" for this check is in the caller.  In this case, `GetRecentAncestors` should perform the check.  However in the change I have in mind for the finished #3250 you won't even need to calculate `firstExtraRandomnessAncestorsCids` as you can just reuse the first iterator in the call to `CollectAtMostNTipSets`.

Because this change is in conflict with #3250 I'm going to close it down in favor of that one.  

If we somehow converge on a solution in #3250 that *does* need to calculate `firstExtraRandomnessAncestorsCids` you can either add the slice length check to that PR, or reopen this PR after we merge #3250.

PR #3238, opened 2019-08-14
Feat/move slasher to storage

shannonwells:
Follow-up to #3178 Slashing for unreported storage faults.  Includes:
* moving `StorageFaultSlasher` into protocol/storage
* pulling slasher call back up to node and removing the tests that were in miner_test as they are no longer necessary
* wrapping original Slash call in an OnNewHeaviestTipSet

There is a difficulty with unit testing whether the fault slasher is instantiated in node due to the difficulty with creating a miner that is not a bootstrap miner in unit tests.  I would like to add an integration test to prove that slashing really occurs, but the difficulty there is getting it to pass in CI. In the previous PR this is exactly what happened; the integration test that proved slashing happens fails CI 100% of the time.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -98,6 +98,11 @@ type nodeChainSyncer interface {
 	HandleNewTipSet(ctx context.Context, ci *types.ChainInfo, trusted bool) error
 }
 
+// storageFaultSlasher is the interface for needed StorageFaultSlasher functionality
+type storageFaultSlasher interface {
+	OnNewHeaviestTipSet(context.Context, types.TipSet) error
+}
+
 // Node represents a full Filecoin node.
 type Node struct {
 	host     host.Host
@@ -109,10 +114,11 @@ type Node struct {
 	Syncer       nodeChainSyncer
 	PowerTable   consensus.PowerTableView
 
-	BlockMiningAPI *block.MiningAPI
-	PorcelainAPI   *porcelain.API
-	RetrievalAPI   *retrieval.API
-	StorageAPI     *storage.API
+	BlockMiningAPI      *block.MiningAPI
+	PorcelainAPI        *porcelain.API
+	RetrievalAPI        *retrieval.API
+	StorageAPI          *storage.API
+	StorageFaultSlasher storageFaultSlasher
 
 	// HeavyTipSetCh is a subscription to the heaviest tipset topic on the chain.
 	// https://github.com/filecoin-project/go-filecoin/issues/2309
@@ -745,6 +751,9 @@ func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet
 				if err := node.StorageMiner.OnNewHeaviestTipSet(newHead); err != nil {
 					log.Error(err)
 				}
+				if err := node.StorageFaultSlasher.OnNewHeaviestTipSet(ctx, newHead); err != nil {
+					log.Error(err)
+				}
 			}
 		case <-ctx.Done():
 			return
@@ -877,11 +886,16 @@ func (node *Node) StartMining(ctx context.Context) error {
 	go node.handleNewMiningOutput(miningCtx, outCh)
 
 	// initialize a storage miner
-	storageMiner, err := initStorageMinerForNode(ctx, node)
+	storageMiner, ownerAddress, err := initStorageMinerForNode(ctx, node)
 	if err != nil {
 		return errors.Wrap(err, "failed to initialize storage miner")
 	}
+
+	// initialize the storage fault slasher if appropriate
 	node.StorageMiner = storageMiner
+	if err = node.initStorageFaultSlasherForNode(ctx, ownerAddress); err != nil {
+		return errors.Wrap(err, "failure in initStorageFaultSlasherForNode")
+	}
 
 	// loop, turning sealing-results into commitSector messages to be included
 	// in the chain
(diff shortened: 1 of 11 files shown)


anorth:
Thanks - I will review shortly. @acruikshank I requested this move (but ack your comments in the original)

node/node.go
```diff
@@ -737,6 +743,11 @@ func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet
 				if err := node.StorageMiner.OnNewHeaviestTipSet(newHead); err != nil {
 					log.Error(err)
 				}
+				if node.StorageFaultSlasher != nil {
```
acruikshank: This shouldn't be nested in the StorageMiner condition. StorageFaultSlasher doesn't depend on StorageMiner not being nil. StorageMiner != nil should imply that StorageFaultSlash != nil, so this condition could go away, but my preference would be to have botth conditions side by side.

node/node.go
```diff
@@ -869,11 +880,16 @@ func (node *Node) StartMining(ctx context.Context) error {
 	go node.handleNewMiningOutput(miningCtx, outCh)
 
 	// initialize a storage miner
-	storageMiner, err := initStorageMinerForNode(ctx, node)
+	storageMiner, ownerAddress, err := initStorageMinerForNode(ctx, node)
```
acruikshank: BLOCKING: This needs to be the worker address. This probably just means renaming this variable here and elsewhere for now.

node/node.go
```diff
@@ -1026,13 +1044,27 @@ func initStorageMinerForNode(ctx context.Context, node *Node) (*storage.Miner, e
...
+}
+
+// initStorageFaultSlasherForNode sets node.FaultSlasher only if the node's miner actor
+// is not a bootstrap miner
+func (node *Node) initStorageFaultSlasherForNode(ctx context.Context, ownerAddress address.Address) error {
+	isBootstrapMinerActor, err := node.StorageMiner.IsBootstrapMinerActor(ctx)
+	if err != nil {
+		return errors.Wrap(err, "could not get bootstrap status of miner actor")
+	}
+	if !isBootstrapMinerActor {
```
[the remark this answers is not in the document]
acruikshank: To be clear, I don't think it's bad if the bootstrap miner doesn't slash, but we can get rid of this extra bit of complexity (and avoid future confusion) by removing this logic.

shannonwells:
One extra comment, which I'm sure @acruikshank  and @anorth you are had some idea about, currently the miner worker address is the same as the owner address, however, the story I'm working on is the one that will let you change the worker, and I added a to-do item to make sure the slasher's address gets updated.

node/node.go
```diff
@@ -109,10 +114,11 @@ type Node struct {
...
 
-	BlockMiningAPI *block.MiningAPI
-	PorcelainAPI   *porcelain.API
-	RetrievalAPI   *retrieval.API
-	StorageAPI     *storage.API
+	BlockMiningAPI      *block.MiningAPI
+	PorcelainAPI        *porcelain.API
+	RetrievalAPI        *retrieval.API
+	StorageAPI          *storage.API
+	StorageFaultSlasher storageFaultSlasher
```
anorth: Nit: I would put this after some other "worker" implementations, like the mining worker or storage/retrieval miners.

protocol/storage/fault_slasher_test.go
```diff
@@ -13,12 +13,41 @@ import (
...
+		ts := types.UndefTipSet
+		err := fm.OnNewHeaviestTipSet(ctx, ts)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "failed to get tipset height")
+	})
+
+	t.Run("calls slasher if tipset can get height", func(t *testing.T) {
+		store := chain.NewBuilder(t, minerOwnerAddr)
+		root := store.AppendBlockOnBlocks()
+		t0 := types.RequireNewTipSet(t, root)
```
anorth: `store.NewGenesis()`

protocol/storage/miner.go
```diff
@@ -815,7 +807,7 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) error {
 		}
 	}
 	// slash any late miners w/ unreported storage faults
```
anorth: Delete this comment.

protocol/storage/testing.go
```diff
@@ -21,7 +21,8 @@ type TrivialTestSlasher struct {
 	SendCalls uint64
 }
 
-// Slash is a required function for storageFaultSlasher interfaces and is intended to do nothing.
+// Slash is a required function for storageFaultSlasher interfaces and does nothing but track
+// how many times it's called.
 func (ts *TrivialTestSlasher) Slash(context.Context, *types.BlockHeight) error {
 	ts.SendCalls++
 	return nil
```
anorth: Can you delete this whole file now?

node/node_test.go
```diff
@@ -164,6 +164,11 @@ func TestNodeStartMining(t *testing.T) {
 		assert.Error(t, err, "node is already mining")
 	})
 
+	t.Run("bootstrap miner does not set a storage fault slasher", func(t *testing.T) {
```
acruikshank: This description is wrong. The test appears to be testing that "mining start initializes a storage fault slasher".

protocol/storage/testing.go
shannonwells: No, CalculatePoSt is still being used.  But I can get rid of the test slasher.

shannonwells:
Closes #2939

PR #3245, opened 2019-08-15
update miner redeem test to use real sealing

acruikshank:
### Problem 
The miner redeem test tests that a miner can redeem a voucher with a verify piece inclusion condition so long as they have the sector and a piece inclusion proof. This test had been written by specifying the bytes of an actual PIP. Something about the way PIPs are generated changed in a recent rust proofs build, and the test will no longer pass once the changes are checked in.

### Solution
We looked into mocking this test out, but a lot of testing infrastructure would need to be created. It was much simpler to actually seal a sector and use the proof provided by sealing, so changes to proof construction will always be carried over to the test.

Since takes much longer now to accommodate sealing, it is now a functional test.

### DIFF
--- a/actor/builtin/miner_redeem_test.go
+++ b/actor/builtin/miner_redeem_test.go
@@ -5,9 +5,10 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/filecoin-project/go-sectorbuilder"
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-hamt-ipld"
-	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-ipfs-blockstore"
 
 	"github.com/filecoin-project/go-filecoin/actor/builtin"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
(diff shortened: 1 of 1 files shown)


acruikshank:
@anorth: I think the quickest way to get this back to a unit test is to use a FakeVMContextWithVerifier test approach. The VMContext should be populated with a memory datastore backed storage object so the payment broker can read and write from it. paymentbroker.CrreateChannel should be called to establish the channel. The VMContext will aslo need a custom Send function that checks the message is going to the miner address with the right method, and then it should use a new FakeVMContext that can supply a faked out miner state from storage (as we do in the other FakeVMContext tests). This new context should have the verier code to demonstrate the expected parameters are being fed to the proof engine.

Not too bad, but a lot of steps, some of which are novel.

actor/builtin/miner_redeem_test.go
```diff
@@ -39,19 +41,32 @@ func TestVerifyPieceInclusionInRedeem(t *testing.T) {
...
+	sbh := sbtesting.NewBuilder(t).Build()
+	defer sbh.Close()
+
+	_, _, err := sbh.AddPiece(ctx, pieceData)
+	require.NoError(t, err)
+
+	sealResults := sbh.SectorBuilder.SectorSealResults()
+
+	// wait for sector to be sealed
+	sealedSectorResult := <-sealResults
```
laser: I observe that there's no timeout here. Is that OK?

PR #3250, opened 2019-08-16
fix collect tipsets error, add test case

### LINKED ISSUE #3027: CollectTipSetOfHeightAtLeast new OBO error

deaswang:
fixes #3027


anorth:
@ZenGround0 could you please shepherd this one?

chain/get_ancestors.go
```diff
@@ -51,38 +51,14 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
-	}
-	firstExtraRandomnessAncestorsCids, err := provingPeriodAncestors[len(provingPeriodAncestors)-1].Parents()
-	if err != nil {
-		return nil, err
-	}
-	// no parents means hit genesis so return the whole chain
-	if firstExtraRandomnessAncestorsCids.Len() == 0 {
-		return provingPeriodAncestors, nil
-	}
+	return CollectRecentTipSets(ctx, iterator, earliestAncestorHeight, lookback)
```
ZenGround0: Blocking: please don't merge together these two functions.  When I wrote "we no longer need two iterators" I didn't mean we don't want two functions operating over the iterator, I meant we don't need two calls to `IterAncestors`.  

We could have done this merging together of these functions before the change in #3026 (in fact I originally did this) but separated out the two functions to help readability and reusability.

deaswang: @ZenGround0 separated them. and rebase.

chain/get_ancestors.go
```diff
@@ -51,28 +51,12 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
 		earliestAncestorHeight = types.NewBlockHeight(uint64(0))
 	}
 
-	// Step 1 -- gather all tipsets with a height greater than the earliest
```
ZenGround0: Please keep this comment for readability and update it to be accurate: "gather all tipsets up to the first tipset with a height less than or equal to the earliest possible proving period start"

chain/get_ancestors.go
```diff
@@ -51,28 +51,12 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
-	firstExtraRandomnessAncestorsCids, err := provingPeriodAncestors[len(provingPeriodAncestors)-1].Parents()
-	if err != nil {
-		return nil, err
-	}
-	// no parents means hit genesis so return the whole chain
-	if firstExtraRandomnessAncestorsCids.Len() == 0 {
-		return provingPeriodAncestors, nil
-	}
 
-	// Step 2 -- gather the lookback tipsets directly preceding provingPeriodAncestors.
```
ZenGround0: Please keep this comment

chain/get_ancestors.go
```diff
@@ -51,28 +51,12 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
-	lookBackTS, err := provider.GetTipSet(firstExtraRandomnessAncestorsCids)
-	if err != nil {
-		return nil, err
-	}
-	iterator = IterAncestors(ctx, provider, lookBackTS)
 	extraRandomnessAncestors, err := CollectAtMostNTipSets(ctx, iterator, lookback)
 	if err != nil {
 		return nil, err
```
ZenGround0: Please update the function comment for `CollectTipSetsOfHeightAtLeast` to accurately reflect the new behavior.

chain/get_ancestors.go
```diff
@@ -99,6 +83,7 @@ func CollectTipSetsOfHeightAtLeast(ctx context.Context, iterator *TipsetIterator
 		}
 		ret = append(ret, iterator.Value())
```
ZenGround0: Blocking: Hey @deaswang in order to close out #3027 this function needs a behavior change.  It looks like the original issue wasn't clear about this.  I've updated the issue to [explain this in better detail](https://github.com/filecoin-project/go-filecoin/issues/3027#issuecomment-523106427).  Please follow up with questions here or there if it is still unclear.

chain/get_ancestors.go
```diff
@@ -51,28 +51,12 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
-		return provingPeriodAncestors, nil
-	}
 
-	// Step 2 -- gather the lookback tipsets directly preceding provingPeriodAncestors.
-	lookBackTS, err := provider.GetTipSet(firstExtraRandomnessAncestorsCids)
-	if err != nil {
-		return nil, err
-	}
-	iterator = IterAncestors(ctx, provider, lookBackTS)
 	extraRandomnessAncestors, err := CollectAtMostNTipSets(ctx, iterator, lookback)
```
ZenGround0: Blocking: @deaswang without the behavior change discussed [below](https://github.com/filecoin-project/go-filecoin/pull/3250/files#r315803061) this change will skip a tipset in the return array which will cause rare errors which will halt the chain's progress.

Once you implement the described behavior change to `CollectTipSetsOfHeightAtLeast` then this change to `GetRecentAncestors` will be correct.

deaswang:
the requested change have update. @ZenGround0

chain/get_ancestors.go
```diff
@@ -94,11 +89,12 @@ func CollectTipSetsOfHeightAtLeast(ctx context.Context, iterator *TipsetIterator
 		if err != nil {
 			return nil, err
 		}
-		if types.NewBlockHeight(h).LessThan(minHeight) {
-			return ret, nil
-		}
 		ret = append(ret, iterator.Value())
+		if types.NewBlockHeight(h).LessEqual(minHeight) {
+			break
```
ZenGround0: 👍

chain/get_ancestors.go
```diff
@@ -51,37 +51,32 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
-	// no parents means hit genesis so return the whole chain
-	if firstExtraRandomnessAncestorsCids.Len() == 0 {
-		return provingPeriodAncestors, nil
-	}
 
-	// Step 2 -- gather the lookback tipsets directly preceding provingPeriodAncestors.
-	lookBackTS, err := provider.GetTipSet(firstExtraRandomnessAncestorsCids)
-	if err != nil {
-		return nil, err
+	// Step 2 -- skip the next tipset which added if iterator not complete
```
ZenGround0: nit: it would be helpful to explain that the iterator is currently pointing at the lowest height element of `provingPeriodAncestors` and needs to be updated to avoid duplicates.

chain/get_ancestors.go
```diff
@@ -51,37 +51,32 @@ func GetRecentAncestors(ctx context.Context, base types.TipSet, provider TipSetP
...
 	if err != nil {
 		return nil, err
 	}
 	return append(provingPeriodAncestors, extraRandomnessAncestors...), nil
 }
 
-// CollectTipSetsOfHeightAtLeast collects all tipsets with a height greater
-// than or equal to minHeight from the input tipset.
+// CollectTipSetsOfHeightAtLeast collects all tipsets up to the first tipset with a height less than
+// or equal to the earliest possible proving period start
```
anorth: "up to and including"

Two blocking issues:
- The name of this function is now misleading. Maybe `CollectTipSetsPastHeight`? Not sure - it's hard to name this behaviour
- This function has another caller in `chain/traversal.go` and this behaviour changes appears to be incompatible with its needs (excluding the equal-height block). Unfortunately, `CollectTipsToCommonAncestor` does not have direct unit tests.

Some suggestions for resolution:
- Leave this function alone and write more code in `GetRecentAncestors` to traverse the additional block
- Leave this function alone and write a new one for the needs of `GetRecentAncestors`
- Revert this change, then replace `minheight` with a predicate and rename to `CollectTipSetsWhile`.

deaswang:
rebased and test case passed at local. @anorth @ZenGround0

PR #3251, opened 2019-08-16
remove nonce field from block header

### LINKED ISSUE #3248: Remove nonce field from block header
### Description
Our block header structure has a `Nonce` field that exists only as a shortcut for testing, to produce distinct headers. We have better fake block generation with the `chain.Builder` and other support code, so let's wean off the nonce.

### Acceptance criteria
- Block header nonce is removed while retaining existing test coverage

### Risks + pitfalls
- This will require a painful protocol upgrade unless we do it before a new network launch
- There may be quite a few old tests relying on it

### Where to begin
- Search for usages of the nonce

deaswang:
fixes #3248

### DIFF
--- a/core/policy_test.go
+++ b/core/policy_test.go
@@ -234,7 +234,7 @@ func TestMessageQueuePolicy(t *testing.T) {
 			b.SetTicket([]byte{2})
 			b.SetTimestamp(0) // Tweak if necessary to force CID ordering opposite ticket ordering.
 		})
-		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) > 0)
+		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) != 0)
 
 		// With blocks ordered [b1, b2], everything is ok.
 		err := policy.HandleNewHead(ctx, q, nil, []types.TipSet{requireTipset(t, b1, b2)})
--- a/core/policy_test.go
+++ b/core/policy_test.go
@@ -234,7 +234,8 @@ func TestMessageQueuePolicy(t *testing.T) {
 			b.SetTicket([]byte{2})
 			b.SetTimestamp(0) // Tweak if necessary to force CID ordering opposite ticket ordering.
 		})
-		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) != 0)
+
+		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) > 0)
 
 		// With blocks ordered [b1, b2], everything is ok.
 		err := policy.HandleNewHead(ctx, q, nil, []types.TipSet{requireTipset(t, b1, b2)})
--- a/mining/worker_test.go
+++ b/mining/worker_test.go
@@ -196,7 +196,7 @@ func TestApplyMessagesForSuccessTempAndPermFailures(t *testing.T) {
 	act1 := th.RequireNewFakeActor(t, vms, addr1, fakeActorCodeCid)
 	_, st := th.RequireMakeStateTree(t, cst, map[address.Address]*actor.Actor{
 		address.NetworkAddress: th.RequireNewAccountActor(t, types.NewAttoFILFromFIL(1000000)),
-		addr1:                  act1,
+		addr1: act1,
 	})
 
 	ctx := context.Background()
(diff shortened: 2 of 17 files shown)


types/block_test.go
```diff
@@ -151,10 +150,10 @@ func TestEquals(t *testing.T) {
 	var n1 Uint64 = 1234
 	var n2 Uint64 = 9876
 
-	b1 := &Block{Parents: NewTipSetKey(c1), Nonce: n1}
-	b2 := &Block{Parents: NewTipSetKey(c1), Nonce: n1}
-	b3 := &Block{Parents: NewTipSetKey(c1), Nonce: n2}
-	b4 := &Block{Parents: NewTipSetKey(c2), Nonce: n1}
```
anorth: This is a weak test, but you'll need to do something to restore it. I suggest setting (at least) the state root in place of the nonce.

types/block_test.go
```diff
@@ -167,10 +166,9 @@ func TestParanoidPanic(t *testing.T) {
 
 	paranoid = true
 
-	b1 := &Block{Nonce: 1}
+	b1 := &Block{}
 	b1.Cid()
 
-	b1.Nonce = 2
```
anorth: Similarly, you need to mutate the block to test this

anorth:
Sorry - I notice the WIP in the title, I was premature. Please use a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) when not yet ready for review. Possibly avoid setting reviewers until ready.

deaswang:
@anorth finished.

core/policy_test.go
```diff
@@ -216,7 +216,7 @@ func TestMessageQueuePolicy(t *testing.T) {
 			b.SetTicket([]byte{1})
 			b.SetTimestamp(0) // Tweak if necessary to force CID ordering opposite ticket ordering.
 		})
-		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) > 0)
+		assert.True(t, bytes.Compare(b1.Cid().Bytes(), b2.Cid().Bytes()) != 0)
```
ZenGround0: Blocking: the intent of this check is to ensure we exercise a pair of blocks where cid and ticket order are inverted to prevent regression.  If the cid orders have flipped without the nonce then we need to tweak the blocks to give us back this property.

cc @anorth.

mining/worker_test.go
```diff
@@ -280,28 +279,20 @@ func TestGenerateMultiBlockTipSet(t *testing.T) {
...
+	assert.Equal(t, 2, baseTipset.Len())
+
+	blk, err := worker.Generate(ctx, baseTipset, nil, types.PoStProof{}, 0)
 	assert.NoError(t, err)
 
 	assert.Equal(t, types.EmptyMessagesCID, blk.Messages)
 	assert.Equal(t, types.EmptyReceiptsCID, blk.MessageReceipts)
 	assert.Equal(t, types.Uint64(101), blk.Height)
-	assert.Equal(t, types.Uint64(1020), blk.ParentWeight)
+	assert.Equal(t, types.Uint64(120), blk.ParentWeight)
```
ZenGround0: I would slightly prefer if you set the baseTipset's weight after constructing with the builder, this way the internals of the builder's testing weight function are not needed to understand the test.

anorth: If that involved mutating the blocks after construction, it's probably a step backwards. I think this is ok for now. A configurable weight function in the builder would be better, but probably not worth it here.

protocol/hello/hello_test.go
```diff
@@ -199,18 +203,14 @@ func TestHelloMultiBlock(t *testing.T) {
...
-	)
-	heavy2 := th.RequireNewTipSet(t,
-		&types.Block{Nonce: 1001, Height: 3},
-		&types.Block{Nonce: 1003, Height: 3},
-		&types.Block{Nonce: 1005, Height: 3},
-	)
+	builder := chain.NewBuilder(t, address.Undef)
+
+	genesisA := builder.AppendBlockOn(types.UndefTipSet)
+	genesisTipset := th.RequireNewTipSet(t, genesisA)
```
anorth: Replace with `builder.NewGenesis()`

ZenGround0:
@deaswang because of the code freeze for release 0.4.x the conflicts are going to be a little insane for a while.  Try rebasing this and pinging me and @anorth immediately when you read it.  It might take a few tries though.

deaswang:
@ZenGround0 @anorth rebased and test passed.

PR #3252, opened 2019-08-16
misc: download parameters only for 1kb sector size

### LINKED ISSUE #3152: modify the `deps` command such that it downloads only Groth parameters and keys for 1KiB sectors
### Description

Original issue [here](https://app.zenhub.com/workspaces/cryptolabfilecoin-integration-5d31e8d6a02eb00001aa3df7/issues/filecoin-project/go-filecoin/3072):

Quoting @anorth 

> An acceptable state to me would be that deps only downloads verifying params and the 1KiB proof param, and no more. Miners must explicitly invoke paramfetch to fetch more. This is predicated on the go-filecoin daemon clearly alerting the user to a missing parameter as soon as they mining start (not waiting until a seal or PoSt is scheduled).

### Acceptance criteria

- `deps` uses `paramfetch` to download parameters for 1KiB sectors
- `mining start` produces an error message informing user if they're missing Groth parameters for their configured sector size

### Risks + pitfalls

The second of the two criteria, above, is blocked on [this rust-fil-sector-builder](https://github.com/filecoin-project/rust-fil-sector-builder/issues/22) issue.

### Where to begin

- [`SectorBuilder#init_with_metadata`](https://github.com/filecoin-project/rust-fil-sector-builder/blob/master/sector-builder/src/builder.rs#L38) could be a place to put the cache-checking call
- [`paramcache`](https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/bin/paramcache.rs#L21) contains the enumeration of circuits which seal and PoSt-generation are currently configured to use

icorderi:
Closes #3152

### DIFF
--- a/scripts/install-filecoin-parameters.sh
+++ b/scripts/install-filecoin-parameters.sh
@@ -3,11 +3,11 @@
 set -Eeo pipefail
 
 fetch_params() {
-  ./proofs/bin/paramfetch --all --verbose --json=./proofs/misc/parameters.json || true
+  ./proofs/bin/paramfetch -z 1024 --verbose --json=./proofs/misc/parameters.json || true
 }
 
 generate_params() {
-  RUST_LOG=info ./proofs/bin/paramcache
+  RUST_LOG=info ./proofs/bin/paramcache --test-only
 }
 
 fetch_params


laser:
> Does this still fetch all *verification* parameters? I think it should.

Yep!

> does the miner currently fail appropriately early if the required proving parameters are not present?

Not yet. [This PR adds that functionality](https://github.com/filecoin-project/go-filecoin/pull/3254).

E

PR #3256, opened 2019-08-19
Factor out shared chain traversal when inbox and outbox handle new head.

### LINKED ISSUE #3052: [bug] Same actor and nonce but different cid

anorth:
This is a step toward restoring lost messages to the outbox for #3052.


chain/traversal.go
```diff
@@ -87,3 +87,33 @@ func TipSetProviderFromBlocks(ctx context.Context, blocks BlockProvider) TipSetP
 func (p *tipsetFromBlockProvider) GetTipSet(tsKey types.TipSetKey) (types.TipSet, error) {
 	return LoadTipSetBlocks(p.ctx, p.blocks, tsKey)
 }
+
+// CollectTipsToCommonAncestor traverses chains from two tipsets (called old and new) until their common
+// ancestor, collecting all tipsets that are in one chain but not the other.
+// The resulting lists of tipsets are ordered by decreasing height.
+func CollectTipsToCommonAncestor(ctx context.Context, store TipSetProvider, oldHead, newHead types.TipSet) (oldTips, newTips []types.TipSet, err error) {
```
frrist: Mixed feelings the named returns :thinking:, I suppose this does improve readability.

core/outbox.go
```diff
@@ -35,8 +36,8 @@ type Outbox struct {
 }
 
 type outboxChainProvider interface {
```
frrist: This is identical to the `inboxChainProvider` interface. Seeing as they both live in the same package, is there any reason to separate them like this?

core/policy.go
```diff
@@ -81,14 +77,16 @@ func (p *DefaultQueuePolicy) HandleNewHead(ctx context.Context, target PolicyTar
...
-			log.Errorf("Outbound message %v expired un-mined after %d rounds", msg, p.maxAgeRounds)
+	if len(newTips) > 0 {
+		height, err := newTips[0].Height()
+		if err != nil {
+			return err
+		}
+		if height >= p.maxAgeRounds { // avoid uint subtraction overflow
+			expired := target.ExpireBefore(ctx, (height - p.maxAgeRounds))
+			for _, msg := range expired {
+				log.Errorf("Outbound message %v expired un-mined after %d rounds", msg, p.maxAgeRounds)
```
frrist: (I know you haven't changed this here, but..) Does this warrant an error message -- could this be a warning instead?

go.sum
```diff
@@ -1100,6 +1100,8 @@ golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 h1:LepdCS8Gf/MVejFIt8lsiexZA
 golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20190726091711-fc99dfbffb4e h1:D5TXcfTk7xF7hvieo4QErS3qqCB4teTffacDWr7CI+0=
 golang.org/x/sys v0.0.0-20190726091711-fc99dfbffb4e/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3 h1:4y9KwBHBgBNwDbtu44R5o1fdOCQUEXhbk/P4A9WmJq0=
```
frrist: I think these could be removed.

chain/traversal.go
shannonwells: Named returns are a good idea if you are returning multiples of the same type, that's my 2 centavos.

node/node.go
```diff
@@ -732,20 +732,31 @@ func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet
 				log.Error("tipset of size 0 published on heaviest tipset channel. ignoring and waiting for a new heaviest tipset.")
 				continue
 			}
-
-			if err := node.Outbox.HandleNewHead(ctx, prevHead, newHead); err != nil {
-				log.Error("updating outbound message queue for new tipset", err)
+			if newHead.Equals(prevHead) {
+				log.Errorf("non-new head published on heaviest tipset channel, ignoring %s", newHead.Key())
+				continue
```
shannonwells: Similar question to above, should this be an error or warning?

PR #3258, opened 2019-08-20
Track block, message and tipset size when processing.

frrist:
This PR adds metrics to track the block, message, and tipset sizes when processing as well as tipset size and duration when fetching.

This PR allows for the amount of data fetched to be compared with the amount of data processed. These metrics can be used when profiling chain syncing and processing times.

### DIFF
--- a/consensus/processor.go
+++ b/consensus/processor.go
@@ -26,6 +26,12 @@ var (
 	// Timers
 	amTimer = metrics.NewTimerMs("consensus/apply_message", "Duration of message application in milliseconds", msgMethodKey)
 	pbTimer = metrics.NewTimerMs("consensus/process_block", "Duration of block processing in milliseconds")
+	ptTimer = metrics.NewTimerMs("consensus/process_tipset", "Duration of tipset processing in milliseconds")
+
+	// Sizes
+	pbSize = metrics.NewInt64ByteBucket("consensus/block_size", "The size in bytes of a block")
+	amSize = metrics.NewInt64ByteBucket("consensus/message_size", "The size in bytes of a message", msgMethodKey)
+	tsSize = metrics.NewInt64ByteBucket("consensus/tipset_size", "The size in bytes of a tipset")
 )
 
 // BlockRewarder applies all rewards due to the miner's owner for processing a block including block reward and gas
@@ -111,6 +117,12 @@ func (p *DefaultProcessor) ProcessBlock(ctx context.Context, st state.Tree, vms 
 	span.AddAttributes(trace.StringAttribute("block", blk.Cid().String()))
 	defer tracing.AddErrorEndSpan(ctx, span, &err)
 
+	blkSize, err := blk.ToNode().Size()
+	if err != nil {
+		return nil, err
+	}
+	pbSize.Add(ctx, int64(blkSize))
+
 	pbsw := pbTimer.Start(ctx)
 	defer pbsw.Stop(ctx)
 
@@ -149,6 +161,10 @@ func (p *DefaultProcessor) ProcessTipSet(ctx context.Context, st state.Tree, vms
 	ctx, span := trace.StartSpan(ctx, "DefaultProcessor.ProcessTipSet")
 	span.AddAttributes(trace.StringAttribute("tipset", ts.String()))
 	defer tracing.AddErrorEndSpan(ctx, span, &err)
+
+	tsSize.Add(ctx, int64(ts.Size()))
+	sw := ptTimer.Start(ctx)
+	defer sw.Stop(ctx)
 
 	h, err := ts.Height()
 	if err != nil {
@@ -293,6 +309,16 @@ func (p *DefaultProcessor) ApplyMessage(ctx context.Context, st state.Tree, vms 
 	if err != nil {
 		log.Debugf("failed to insert tag for message method: %s", err.Error())
 	}
+
+	msgNode, err := msg.ToNode()
+	if err != nil {
+		return nil, err
+	}
+	msgSize, err := msgNode.Size()
+	if err != nil {
+		return nil, err
+	}
+	amSize.Add(ctx, int64(msgSize))
 
 	ctx, span := trace.StartSpan(ctx, "DefaultProcessor.ApplyMessage")
 	span.AddAttributes(trace.StringAttribute("message", msgCid.String()))
--- a/metrics/bucket.go
+++ b/metrics/bucket.go
@@ -1,0 +1,51 @@
+package metrics
+
+import (
+	"context"
+
+	"go.opencensus.io/stats"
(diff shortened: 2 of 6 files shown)


tools/fast/environment/environment_memory_genesis.go
```diff
@@ -220,6 +221,7 @@ func (e *MemoryGenesis) startGenesisServer() error {
 	}
 
 	e.genesisServerAddr = ln.Addr().String()
+	e.Log().Infof("Genesus URL: http://%s%s", e.genesisServerAddr, handlePath)
```
frrist: This is a one off change unrelated to the rest of the PR, added via separate commit.

metrics/bucket.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Float64Counter wraps an opencensus float64 measure that is uses as a counter.
+type Float64Counter struct {
+	measureCt *stats.Float64Measure
+	view      *view.View
+}
+
+// Add adds `m` to the values tracked by Float64Counter
+func (c *Float64Counter) Add(ctx context.Context, m uint64) {
```
frrist: It would be _**really**_ nice to have structured logging here (and in other metrics) it could tell us the size of each block, tipset, and message referenced by CID. In the Timer metrics, it could tell us how long each tipset, block and message took to processes. As this exists now, these metrics only tell half the story without adequate logging.

consensus/processor.go
```diff
@@ -26,6 +26,12 @@ var (
 	// Timers
 	amTimer = metrics.NewTimerMs("consensus/apply_message", "Duration of message application in milliseconds", msgMethodKey)
 	pbTimer = metrics.NewTimerMs("consensus/process_block", "Duration of block processing in milliseconds")
+	ptTimer = metrics.NewTimerMs("consusnsu/process_tipset", "Duration of tipset processing in milliseconds")
```
ZenGround0: nit: spelling.  Though I'd like to consider `consusnsu` as an `flarp` alternative cc @laser

metrics/bucket.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// NewBucketCounter creates a Float64Counter with units of bytes and a default set of aggregation
+// bounds from .25 kilobytes to 10 megabytes.
+func NewBucketCounter(name, desc string, tagKeys ...tag.Key) *Float64Counter {
+	defaultBounds := []float64{250, 500, 1000, 5000, 10000, 50000, 100000, 500000, 1000000, 5000000, 10000000}
+	return NewBucketCounterWithBuckets(name, desc, stats.UnitBytes, defaultBounds, tagKeys...)
+}
+
+// NewBucketCounterWithBuckets creates a Float64Counter wrapping an opencensus int64 measurement.
```
ZenGround0: nit: it appears to be wrapping an opencensus float64 measurement

metrics/bucket.go
```diff
@@ -0,0 +1,51 @@
...
+		Aggregation: view.Distribution(bounds...),
+	}
+	if err := view.Register(iView); err != nil {
+		// a panic here indicates a developer error when creating a view.
+		// Since this method is called in init() methods, this panic when hit
+		// will cause running the program to fail immediately.
+		panic(err)
+	}
+
+	return &Float64Counter{
```
ZenGround0: Dumb question: why do we need float64s to track number of bytes in open census?

consensus/processor.go
laser: "con-suu-snuuu"

metrics/bucket.go
frrist: Not a dumb question, this is wrong and I have revised its definition in the recent changes
(same part of my vacation brain that wrote "con-snuu-suu" wrote this)

consensus/processor.go
```diff
@@ -26,6 +26,12 @@ var (
 	// Timers
 	amTimer = metrics.NewTimerMs("consensus/apply_message", "Duration of message application in milliseconds", msgMethodKey)
 	pbTimer = metrics.NewTimerMs("consensus/process_block", "Duration of block processing in milliseconds")
+	ptTimer = metrics.NewTimerMs("consensus/process_tipset", "Duration of tipset processing in milliseconds")
+
+	// Sizes
+	pbSize = metrics.NewInt64ByteBucket("consensus/block_size", "The size in bytes of a block")
+	amSize = metrics.NewInt64ByteBucket("consensus/message_size", "The size in bytes of a message", msgMethodKey)
+	tsSize = metrics.NewInt64ByteBucket("consensus/tipset_size", "The size in bytes of a tipset")
```
anorth: IMO these names are too brief now there are so many.

metrics/bucket.go
```diff
@@ -0,0 +1,51 @@
...
+// NewInt64ByteBucket creates a Int64Bucket with units of bytes and a default set of aggregation
+// bounds from .25 kilobytes to 10 megabytes.
+func NewInt64ByteBucket(name, desc string, tagKeys ...tag.Key) *Int64Bucket {
+	defaultBounds := []float64{250, 500, 1000, 5000, 10000, 50000, 100000, 500000, 1000000, 5000000, 10000000}
+	return NewInt64BucketWithBounds(name, desc, stats.UnitBytes, defaultBounds, tagKeys...)
+}
+
+// NewInt64BucketWithBounds creates a Int64Bucket wrapping an opencensus int64 measurement.
+func NewInt64BucketWithBounds(name, desc, unit string, bounds []float64, tagKeys ...tag.Key) *Int64Bucket {
+	log.Infof("registering avg counter: %s - %s", name, desc)
```
anorth: Wrong comment. Maybe it's possible to use reflection to produce a comment that can be safely copy/pasted?

metrics/bucket.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+// NewInt64ByteBucket creates a Int64Bucket with units of bytes and a default set of aggregation
+// bounds from .25 kilobytes to 10 megabytes.
+func NewInt64ByteBucket(name, desc string, tagKeys ...tag.Key) *Int64Bucket {
```
anorth: "Bucket" is not a term widely used for metrics AFAIK. Use of the verb `Add` as a method below suggests that  that this is functioning as a *counter*, counting all the bytes. An alternative we could use is a *histogram* if we wanted to understand the distribution of values (but then the method should be `Record` or similar). Please use one of those words.

We already have an `Int64Counter`, so if that's the intention we might just need some more construction methods that supply different units and bounds. We don't already have `Int64Histogram`, but this could be it.

frrist:
This PR is trying to do too many things, will reopen a more focused set of changes and incorporate above comments.

PR #3261, opened 2019-08-20
Feat/set miner worker 2959

shannonwells:
Part of #2959 

Creates two commands, 
`miner worker` which displays the miner worker address, and already had a porcelain API function associated with it, requiring only the hookup to the CLI.

`miner set-worker <new-address>` which sets the new worker address. 

### miner worker
**$ go-filecoin  miner worker --help**
```
USAGE
  ./go-filecoin miner worker - Show the address of the miner worker

SYNOPSIS
  ./go-filecoin miner worker

DESCRIPTION

  Show the address of the miner worker
```

**$ go-filecoin miner worker**
```
t1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy
```

### miner set-worker
**$ go-filecoin miner set-worker --help**

```
USAGE
  ./go-filecoin miner set-worker <new-address> - Set the address of the miner worker

SYNOPSIS
  ./go-filecoin miner set-worker [--gas-price=<gas-price>] [--gas-limit=<gas-limit>] [--] <new-address>

ARGUMENTS

  <new-address> - The address of the new miner worker.

OPTIONS

  --gas-price string - Price (FIL e.g. 0.00013) to pay for each GasUnit consumed mining this message.
  --gas-limit uint64 - Maximum GasUnits this message is allowed to consume.

DESCRIPTION

  Set the address of the miner worker to the provided address. When a miner is created, this address defaults to the miner owner. Use this command to change the default.
```

**$ go-filecoin miner set-worker t1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy --gas-price=1.0 --gas-limit=300**

```
Changed worker to 't1nhudgskmseowv7rp4e6scxsmlt3qoysvpn73tuy'
```

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -123,7 +123,7 @@ type Ask struct {
 // State is the miner actors storage.
 type State struct {
 	// Owner is the address of the account that owns this miner. Income and returned
-	// collateral are paid to this address. This address is also allowed to change the
+	// collateral are paid to this address. This address is also allowed todd change the
 	// worker address for the miner.
 	Owner address.Address
 
--- a/commands/miner_daemon_test.go
+++ b/commands/miner_daemon_test.go
@@ -3,6 +3,7 @@ package commands_test
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -24,6 +25,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/gengen/util"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/tools/fast"
+	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
@@ -69,6 +72,20 @@ func TestMinerHelp(t *testing.T) {
 		}
 
 		result := runHelpSuccess(t, "miner", "create", "--help")
+		for _, elem := range expected {
+			assert.Contains(t, result, elem)
+		}
+	})
+	t.Run("worker --help shows worker help", func(t *testing.T) {
+		expected := []string{
+			"go-filecoin miner worker <newAddress> - Set the address of the miner worker",
+			"go-filecoin miner worker [--gas-price=<gas-price>] [--gas-limit=<gas-limit>] [--] <newAddress>",
+			"<newAddress> - The address of the new miner worker.",
+			"--gas-price string - Price (FIL e.g. 0.00013) to pay for each GasUnits consumed mining this message.",
+			" --gas-limit uint64 - Maximum number of GasUnits this message is allowed to consume.",
+			" Sets the address of the miner worker to the provided address. The new address must be that of a worker that has already been created. When a miner is created, this address defaults to the miner owner. Use this command to change the default.",
+		}
+		result := runHelpSuccess(t, "miner", "worker", "--help")
 		for _, elem := range expected {
 			assert.Contains(t, result, elem)
 		}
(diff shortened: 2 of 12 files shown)


actor/builtin/miner/miner.go
```diff
@@ -123,7 +123,7 @@ type Ask struct {
 // State is the miner actors storage.
 type State struct {
 	// Owner is the address of the account that owns this miner. Income and returned
-	// collateral are paid to this address. This address is also allowed to change the
+	// collateral are paid to this address. This address is also allowed todd change the
```
shannonwells: I debated calling the function to change the worker `Todd`, but wiser head prevailed.

commands/miner_daemon_test.go
```diff
@@ -209,23 +229,33 @@ func TestMinerSetPrice(t *testing.T) {
...
-	d1.ConnectSuccess(d)
-	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		miner := d.RunSuccess("miner", "create", "--from", fixtures.TestAddresses[2], "--gas-price", "1", "--gas-limit", "100", "200")
-		addr, err := address.NewFromString(strings.Trim(miner.ReadStdout(), "\n"))
-		assert.NoError(t, err)
-		assert.NotEqual(t, addr, address.Undef)
-		wg.Done()
+	ctx, env := fastesting.NewTestEnvironment(context.Background(), t, fast.FilecoinOpts{})
```
shannonwells: In the process of debugging a miner creation failure, I converted this to a FAST test.

porcelain/miner_test.go
```diff
@@ -36,13 +37,13 @@ type minerCreate struct {
 }
```
shannonwells: A bunch of the little changes in this file are to appease the linter deities -- variable/package name conflicts, error messages that are capitalized, etc.

porcelain/miner_test.go
```diff
@@ -599,3 +600,97 @@ func TestMinerGetLastCommittedSectorID(t *testing.T) {
...
+			ownerAddr:  minerOwner,
+			minerAddr:  minerAddr,
+		}
+
+		err := MinerSetWorkerAddress(context.Background(), plumbing, workerAddr, gprice, glimit)
+		assert.NoError(t, err)
+		assert.Equal(t, workerAddr.String(), plumbing.workerAddr.String())
+	})
+
+	testCases := []struct {
```
shannonwells: this setup & test plumbing with failure flags actually caught a missed error check early on.

commands/miner_daemon_test.go
ZenGround0: god's work 🙏

commands/miner_daemon_test.go
```diff
@@ -415,3 +445,81 @@ var testConfig = &gengen.GenesisCfg{
...
+	require.NoError(t, env.GenesisMiner.MiningStart(ctx))
+	defer func() {
+		require.NoError(t, env.GenesisMiner.MiningStop(ctx))
+	}()
+	env.RunAsyncMiner()
+
+	minerNode := env.RequireNewNodeWithFunds(1000)
+
+	t.Run("fails if there is no miner worker", func(t *testing.T) {
+		res, err := minerNode.MinerWorker(ctx)
```
ZenGround0: Blocking: this looks like a repeat of the "if there is no miner worker..." test of the `miner worker` command above.  Shouldn't this be testing miner worker setting?

ZenGround0: (or just be deleted if there is no analogue in the "set" case)

porcelain/miner.go
```diff
@@ -509,3 +509,57 @@ func MinerGetCollateral(ctx context.Context, plumbing mgaAPI, minerAddr address.
 	}
 	return types.NewAttoFILFromBytes(rets[0]), nil
 }
+
+// mwapi is the subset of the plumbing.API that Miner(Set|Get)WorkerAddress use.
```
ZenGround0: I would either move `MinerGetWorker` to use this interface, or else remove the reference to Get here.

porcelain/miner.go
```diff
@@ -509,3 +509,57 @@ func MinerGetCollateral(ctx context.Context, plumbing mgaAPI, minerAddr address.
...
+		types.ZeroAttoFIL,
+		gasPrice,
+		gasLimit,
+		"changeWorker",
+		workerAddr)
+	if err != nil {
+		return err
+	}
+
+	return plumbing.MessageWait(ctx, smsgCid, func(blk *types.Block, smsg *types.SignedMessage, receipt *types.MessageReceipt) error {
```
ZenGround0: Musing beyond the scope of this PR: it feels like we don't have a strong convention between returning the result of a message wait and returning a message cid that the caller can wait on.  I would like us to come up with a good one so it is obvious which approach is right given a new porcelain call like this.

commands/miner_daemon_test.go
shannonwells: whoops! thanks for catching that. I'll update the tests.

porcelain/miner.go
shannonwells: I can't really think of a satisfying answer.  I think I'd prefer always returning a CID and forcing the user to have to MessageWait every time. This is unsatisfying but it gives consistency and also the most freedom to not have to wait around for the command to finish.

shannonwells:
@ZenGround0 I realized I hadn't fulfilled all the criteria after the fact. I updated the PR description. I do think that at least item 4 should be done.

protocol/storage/fault_slasher_test.go
```diff
@@ -50,18 +54,23 @@ func TestFaultSlasher_Slash(t *testing.T) {
...
 		queryer := makeQueryer([][]byte{data})
-		minerOwnerAddr := signer.Addresses[0]
 
 		ob := outbox{}
-		fm := NewFaultSlasher(&slasherPlumbing{false, false, queryer}, &ob, minerOwnerAddr, DefaultFaultSlasherGasPrice, DefaultFaultSlasherGasLimit)
+		sp := slasherPlumbing{
+			Queryer:    queryer,
+			minerAddr:  signer.Addresses[0],
+			workerAddr: signer.Addresses[1],
+		}
```
anorth: Factor out this oft-duplicated block?

shannonwells:
> Thanks for fixing up the slasher's use of the address. In case you go on vacation before following through with the block producer, could you please hand off to @ingar so we don't get stuck in the inconsistent state Zen describes?

Will do.

protocol/storage/fault_slasher_test.go
shannonwells: I just tried to factor this and there are just enough small differences in various that it doesn't really help much.

PR #3263, opened 2019-08-21
Add "mining add-piece" command

### LINKED ISSUE #3044: Sealing data without a deal
### Description
The deal protocol is more or less independent of proving storage, and a miner logically can commit and prove storage without deals. Go-filecoin doesn't provide any interface to do that, so miners have to go through unnecessary self-dealing in order to prove storage.

We should add a CLI command that directly stages a piece of data for sealing, bypassing the deal flow.

### Acceptance criteria
- A miner can seal arbitrary miner-provided data without doing a storage deal
- CLI commands enable the above

### Where to begin

cc @acruikshank @laser

ingar:
## Summary
To allow miners to have more control over self-sealed data, this PR breaks apart the `mining seal-now` command into two parts:
1. `mining add-piece <file>` - adds arbitrary data to a staged sector
1. `mining seal-now` - forces the sealing of any staged sectors

Previously, `seal-now` would generate a staged sector if one didn't exist; now the only way a miner can create a staged sector without accepting a deal is to use `add-piece`.

resolves #3044

### DIFF
--- a/porcelain/sectorbuilder.go
+++ b/porcelain/sectorbuilder.go
@@ -67,7 +67,7 @@ func AddPiece(ctx context.Context, plumbing sbPlumbing, pieceReader io.Reader) (
 	return sectorID, nil
 }
 
-// SealNow forces the sectorbuilder to either seal the staged sectors it has or create a new one and seal it immediately
+// SealNow forces the sectorbuilder to seal the staged sectors it has, if any
 func SealNow(ctx context.Context, plumbing sbPlumbing) error {
 	if plumbing.SectorBuilder() == nil {
 		return errors.New("must be mining to seal sectors")
@@ -78,7 +78,7 @@ func SealNow(ctx context.Context, plumbing sbPlumbing) error {
 		return errors.Wrap(err, "could not retrieved staged sectors")
 	}
 
-	// if no sectors are staged, nothing to do
+	// nothing to do if no sectors are staged
 	if len(stagedSectors) == 0 {
 		return nil
 	}
--- a/porcelain/sectorbuilder.go
+++ b/porcelain/sectorbuilder.go
@@ -3,18 +3,24 @@ package porcelain
 import (
 	"bytes"
 	"context"
+	"io"
+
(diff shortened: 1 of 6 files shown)


porcelain/sectorbuilder.go
```diff
@@ -45,19 +78,9 @@ func SealNow(ctx context.Context, plumbing sbPlumbing) error {
...
-		if err != nil {
-			return errors.Wrap(err, "could not create cid for piece")
-		}
-		pieceRef := cid.NewCidV1(cid.DagCBOR, hash)
-		_, err = plumbing.SectorBuilder().AddPiece(ctx, pieceRef, 1, bytes.NewReader(data))
-		if err != nil {
-			return errors.Wrap(err, "could not add piece to trigger sealing")
-		}
-
+		return nil
```
anorth: IMO this is unnecessary. If the sector builder can't handle it, leave a comment explaining that and file an issue on it.

acruikshank: This should really error out here. If the miner has asked to seal and we can't because we have no staged sectors, it should not fail silently.

anorth:
We've hit a philosophical disagreement, I'm sorry it's in your PR @ingar.

I agree we're working around a poor sector-builder API. Please file issue an issue on https://github.com/filecoin-project/rust-fil-sector-builder asking for the ability to stage a sector without any piece data. A `stage-sector` command would then be great. We can't pretend to implement that by adding a 1-byte piece since the second such call will not stage a new sector, but add another byte to the first.

I oppose adding "business logic" that is not required by a minimal default mode of operation and that can easily be achieved by scripting against the CLI/API. We should not have that complexity in go-filecoin. We should work to expand the set of operations that can easily be achieved by such scripting. I consider a `--stage-if-empty` to be in that category.

This is still a vast improvement over self-dealing to stage sectors. It only requires a single Filecoin node and two CLI invocations per sector.

(cc @travisperson and thank you for recently expressing such clear ideas along those lines to me)

travisperson:
Filed the issue in #3344.

PR #3264, opened 2019-08-21
Enhanced Graphsync Mocker

hannahhoward:
# Goals

Build a more flexibile mock graphsync that is easy to construct responses against, and interoperates well with a chain

# Implementation

The basic idea here is this -- and I want to check that this is ok -- actually execute IPLD selector queries, backed by a CID loader, which by default just loads from the specified chain builder (but obviously with no network involved). The loader can also be overloaded to error on some cids, or to return undecodable blocks on some cids

# Discussion

Admittedly, this moves us just a bit a way from a unit test in the sense that we are actually counting on selector logic. The upside to this is:
1. It makes sure IPLD selector queries are constructed correctly (which to me makes the test more meaningful)
2. It makes for a very succinct mocking syntax (net more deletions than additions than deletions in this PR!)

Open to lots of revision feedback but would appreciate a quick up or down on whether this general approach is ok before I write a bunch of tests against it.

### DIFF
--- a/net/graphsync_fetcher_test.go
+++ b/net/graphsync_fetcher_test.go
@@ -16,7 +16,6 @@ import (
 	gsnet "github.com/ipfs/go-graphsync/network"
 	gsstoreutil "github.com/ipfs/go-graphsync/storeutil"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
-	cbor "github.com/ipfs/go-ipld-cbor"
 	format "github.com/ipfs/go-ipld-format"
 	"github.com/ipld/go-ipld-prime"
 	ipldfree "github.com/ipld/go-ipld-prime/impl/free"
@@ -71,21 +70,6 @@ func TestGraphsyncFetcher(t *testing.T) {
 	pid1 := th.RequireIntPeerID(t, 1)
 	pid2 := th.RequireIntPeerID(t, 2)
 
-	// Returns an array of IPLD nodes for the headers, messages, and receipts of some blocks.
-	ipldBlocks := func(bs ...*types.Block) []format.Node {
-		nodes := make([]format.Node, 3*len(bs))
-		for i, b := range bs {
-			nodes[3*i] = b.ToNode()
-			m, err := builder.LoadMessages(ctx, b.Messages)
-			require.NoError(t, err)
-			nodes[3*i+1] = types.MessageCollection(m).ToNode()
-			r, err := builder.LoadReceipts(ctx, b.MessageReceipts)
-			require.NoError(t, err)
-			nodes[3*i+2] = types.ReceiptCollection(r).ToNode()
-		}
-		return nodes
-	}
-
 	t.Run("happy path returns correct tipsets", func(t *testing.T) {
 		gen := builder.NewGenesis()
 		final := builder.BuildOn(gen, 3, func(b *chain.BlockBuilder, i int) {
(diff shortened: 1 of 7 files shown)


net/graphsync_fetcher_test.go
```diff
@@ -150,41 +116,14 @@ func TestGraphsyncFetcher(t *testing.T) {
...
-					makeGsResponse("parents", final.At(0).Cid()),
-					makeGsResponse("parents/0", gen.At(0).Cid()),
-				},
-				blks: ipldBlocks(final.At(0), gen.At(0)),
-			}},
-		}
-		mgs := &mockableGraphsync{stubs: stubs, t: t, store: bs}
+		mgs := newMockableGraphsync(ctx, builder, bs, t)
+		mgs.stubResponseWithSuccess(pid0, final.At(0).Cid(), layer1Selector)
+		mgs.stubResponseWithErrorOnCids(pid0, final.At(1).Cid(), layer1Selector, []cid.Cid{final.At(1).Cid()})
```
anorth: Consider varargs to reduce the slice construction noise.

net/graphsync_fetcher_test.go
```diff
@@ -369,32 +252,19 @@ func TestGraphsyncFetcher(t *testing.T) {
...
-			{
-				fakeRequest{pid0, cidlink.Link{Cid: notABlockCid}, layer1Selector},
-				fakeResponse{blks: []format.Node{notABlockObj}},
-			},
-		}
-		mgs := &mockableGraphsync{stubs: stubs, t: t, store: bs}
-
+		gen := builder.NewGenesis()
+		mgs := newMockableGraphsync(ctx, builder, bs, t)
+		mgs.stubResponseWithUndecodableBlocks(pid0, gen.At(0).Cid(), layer1Selector, []cid.Cid{gen.At(0).Cid()})
```
anorth: In this case I think it would be slightly better to be able to specify the response node explicitly. That would make it flexible for other responses too, like a block that does decode but fails syntactic or semantic validation.

net/graphsync_fetcher_test.go
```diff
@@ -522,12 +392,108 @@ type requestResponse struct {
 }
 
 type mockableGraphsync struct {
+	ctx              context.Context
+	builder          *chain.Builder
```
anorth: Consume an interface (extend `MessageStore`+`BlockProvider`) interface rather than this concrete type (in `tryBlockMessageReceiptNode` too)

net/graphsync_fetcher_test.go
```diff
@@ -522,12 +392,108 @@ type requestResponse struct {
...
+		ctx:     ctx,
+		builder: builder,
+		store:   store,
+		t:       t,
+	}
+}
+
+type mockGraphsyncLoader func(cid.Cid) (format.Node, error)
+
+func (mgs *mockableGraphsync) stubResponseWithLoader(pid peer.ID, c cid.Cid, s selector.Selector, loader mockGraphsyncLoader) {
```
anorth: nit: move this to the bottom, after its callers. Document it.

net/graphsync_fetcher_test.go
```diff
@@ -522,12 +392,108 @@ type requestResponse struct {
...
+	}
+	err = ipldbridge.TraversalProgress{
+		Cfg: &ipldbridge.TraversalConfig{
+			Ctx:        mgs.ctx,
+			LinkLoader: linkLoader,
+		},
+	}.TraverseInformatively(node, s, visitor)
+	mgs.stubs = append(mgs.stubs, requestResponse{
+		fakeRequest{pid, root, s},
+		fakeResponse{responses, []error{err}, blks},
```
anorth: It looks like this returns a slice with a single nil error in non-error cases. Is this intended? Documentation on the `fakeResponse` could help clarify.

net/graphsync_fetcher_test.go
```diff
@@ -491,15 +361,15 @@ func TestRealWorldGraphsyncFetchAcrossNetwork(t *testing.T) {
 	}
 }
 
-func tryBlockMessageReceipt(ctx context.Context, f *chain.Builder, c cid.Cid) ([]byte, error) {
+func tryBlockMessageReceiptNode(ctx context.Context, f *chain.Builder, c cid.Cid) (format.Node, error) {
```
anorth: Nit: I think we could simplify this name to `loadNode`.

Issue #2992, opened 2019-06-24
Rename core and cleanup


ZenGround0:
### Description
The name `core` has long outlived its usefulness.  It has become a home for the outbound message queue and the message pool and should be split into a messaging package, or better yet a message-queue and message-pool package.

The functions that remain in testing.go all need a new home.  The `Must` methods should likely be changed to `Require` methods too because we don't typically panic during test failure.

### Acceptance criteria
core is no more
### Risks + pitfalls
Users of testing.go are a bit spread across the codebase.  Likely they can all be moved to either a messaging specific testing.go or to testhelpers/core.go, consensus.go

### Where to begin
Trace down all users of core testing.

anorth:
The `core` package is gone (most of it lives now in the `message` package). As part of that, I split up its `testing.go` and sent the pieces somewhere sensible (or deleted duplicates).

PR #3265, opened 2019-08-21
Factor a NewHeadHandler from code in Node, add integration tests for message handling.

anorth:
This allows demonstration of the bug reported in #3052.

This PR is in two commits. The first is just noise moving out some test support code for reuse.

### DIFF
--- a/core/handler.go
+++ b/core/handler.go
@@ -1,0 +1,51 @@
+package core
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// NewHeadHandler wires up new head tipset handling to the message inbox and outbox.
+type NewHeadHandler struct {
+	// Inbox and outbox exported for testing.
+	Inbox  *Inbox
+	Outbox *Outbox
+	chain  chainProvider
+
+	prevHead types.TipSet
+}
+
+// NewHandler build a new new-head handler.
+func NewHandler(inbox *Inbox, outbox *Outbox, chain chainProvider, head types.TipSet) *NewHeadHandler {
+	return &NewHeadHandler{inbox, outbox, chain, head}
+}
+
+// HandleNewHead computes the chain delta implied by a new head and updates the inbox and outbox.
+func (h *NewHeadHandler) HandleNewHead(ctx context.Context, newHead types.TipSet) error {
+	if !newHead.Defined() {
+		log.Warning("received empty tipset, ignoring")
+		return nil
+	}
+	if newHead.Equals(h.prevHead) {
+		log.Warningf("received non-new head tipset, ignoring %s", newHead.Key())
+		return nil
+	}
+
+	oldTips, newTips, err := chain.CollectTipsToCommonAncestor(ctx, h.chain, h.prevHead, newHead)
+	if err != nil {
+		return errors.Errorf("traversing chain with new head %s, prev %s: %s", newHead.Key(), h.prevHead.Key(), err)
+	}
+	if err := h.Outbox.HandleNewHead(ctx, oldTips, newTips); err != nil {
+		log.Errorf("updating outbound message queue for tipset %s, prev %s: %s", newHead.Key(), h.prevHead.Key(), err)
+	}
+	if err := h.Inbox.HandleNewHead(ctx, oldTips, newTips); err != nil {
+		log.Errorf("updating message pool for tipset %s, prev %s: %s", newHead.Key(), h.prevHead.Key(), err)
+	}
+
+	h.prevHead = newHead
+	return nil
+}
--- a/core/handler_integration_test.go
+++ b/core/handler_integration_test.go
@@ -1,0 +1,119 @@
+package core_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-filecoin/actor/builtin/account"
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/config"
+	"github.com/filecoin-project/go-filecoin/core"
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// TestNewHeadHandlerIntegration tests inbox and outbox policy consistency.
+func TestNewHeadHandlerIntegration(t *testing.T) {
+	tf.UnitTest(t)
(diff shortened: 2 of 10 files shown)


core/handler_integration_test.go
```diff
@@ -0,0 +1,118 @@
...
+		inbox := core.NewInbox(mpool, maxAge, provider, provider)
+		queue := core.NewMessageQueue()
+		publisher := core.NewDefaultMessagePublisher(&core.MockNetworkPublisher{}, "Topic", mpool)
+		policy := core.NewMessageQueuePolicy(provider, maxAge)
+		outbox := core.NewOutbox(signer, &core.FakeValidator{}, queue, publisher, policy, provider, provider)
+
+		return core.NewHandler(inbox, outbox, provider, root)
+	}
+
+	t.Run("test send after reverted message", func(t *testing.T) {
```
acruikshank: ❤️

core/handler_integration_test.go
```diff
@@ -0,0 +1,118 @@
...
+		_, found = inbox.Pool().Get(mid1)
+		assert.True(t, found) // Message returns to pool to be mined again.
+
+		// Send another message from the same account.
+		// First, send a message and expect to find it in the message queue and pool.
+		_, err = outbox.Send(ctx, sender, dest, types.ZeroAttoFIL, gasPrice, gasUnits, true, "method2")
+		// This case causes the nonce to be wrongly calculated, since the first, now-unmined message
+		// is not in the outbox, and actor state has not updated, but the message pool already has
+		// a message with the same nonce.
+		assert.Error(t, err, "Fix this in issue #3052")
```
acruikshank: This is confusing me. The "Fix this..." message will appear in the test output if there is not an error. It appears that this test is reproducing #3052. Maybe move the "Fix this..." to a comment?

acruikshank: Also consider asserting the error is a "same actor and nonce" error to demonstrate it's actually reproducing the problem.

core/publisher.go
```diff
@@ -1,33 +1,33 @@
-package node
+package core
```
acruikshank: Love that this is leaving `node`. Wish it were going someplace more specific than `core`. I think we have more than enough functionality to justify a `message` or `messaging` package.

node/node.go
```diff
@@ -728,35 +729,16 @@ func (node *Node) handleNewChainHeads(ctx context.Context, prevHead types.TipSet
...
-			oldTips, newTips, err := chain.CollectTipsToCommonAncestor(ctx, node.ChainReader, prevHead, newHead)
-			if err == nil {
-				if err := node.Outbox.HandleNewHead(ctx, oldTips, newTips); err != nil {
-					log.Errorf("updating outbound message queue for tipset %s, prev %s: %s", newHead.Key(), prevHead.Key(), err)
-				}
-				if err := node.Inbox.HandleNewHead(ctx, oldTips, newTips); err != nil {
-					log.Errorf("updating message pool for tipset %s, prev %s: %s", newHead.Key(), prevHead.Key(), err)
-				}
-			} else {
-				log.Errorf("traversing chain with new head %s, prev %s: %s", newHead.Key(), prevHead.Key(), err)
```
acruikshank: ❤️

core/publisher.go
ZenGround0: @acruikshank At this point `core` is exclusively used for messaging stuff and should be renamed to `message` or `messaging`: #2992

core/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	return &NewHeadHandler{inbox, outbox, chain, head}
+}
+
+// HandleNewHead computes the chain delta implied by a new head and updates the inbox and outbox.
+func (h *NewHeadHandler) HandleNewHead(ctx context.Context, newHead types.TipSet) error {
+	if !newHead.Defined() {
+		return errors.New("received empty tipset, ignoring")
+	}
+	if newHead.Equals(h.prevHead) {
+		return errors.Errorf("received non-new head tipset, ignoring %s", newHead.Key())
```
shannonwells: Are these both really errors? I could see the first one but maybe this one isn't.  If it's determined the head isn't new at all it seems to me it's still handled -- it could log this message then return nil.

core/handler_integration_test.go
```diff
@@ -0,0 +1,118 @@
...
+		//require.NoError(t, err)
```
shannonwells: I take it this is the follow-up test code once a fix is in?

core/message_pool_test.go
```diff
@@ -226,3 +226,11 @@ func mustResignMessage(signer types.Signer, message *types.SignedMessage, f func
 func signMessage(signer types.Signer, message types.Message) (*types.SignedMessage, error) {
 	return types.NewSignedMessage(message, signer, types.NewGasPrice(0), types.NewGasUnits(0))
 }
+
+func mustAdd(t *testing.T, p *core.MessagePool, height uint64, msgs ...*types.SignedMessage) {
```
shannonwells: rename to requireAdd per convention now that it doesn't panic

core/policy.go
```diff
@@ -48,8 +48,8 @@ type DefaultQueuePolicy struct {
 
 // NewMessageQueuePolicy returns a new policy which removes mined messages from the queue and expires
 // messages older than `maxAgeTipsets` rounds.
-func NewMessageQueuePolicy(messages MessageProvider, maxAge uint64) *DefaultQueuePolicy {
-	return &DefaultQueuePolicy{messages, maxAge}
+func NewMessageQueuePolicy(messages MessageProvider, maxAge uint) *DefaultQueuePolicy {
```
shannonwells: why was this converted to a uint when it gets cast immediately to uint64?

core/publisher_test.go
```diff
@@ -1,4 +1,4 @@
-package node
+package core_test
```
shannonwells: 🥇 good to take tests out of the main package

core/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// NewHeadHandler wires up new head tipset handling to the message inbox and outbox.
+type NewHeadHandler struct {
```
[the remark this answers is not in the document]
anorth: I totally agree with all your comments about the package name. I've been putting off reorganising packages, but I'll do this one shortly.

core/handler_integration_test.go
anorth: I changed the message to `"#3052 is fixed!"`

core/message_pool_test.go
anorth: It collides with the same name in `inbox_test.go`, but renamed to `reqAdd`.

core/policy.go
anorth: The appropriate interface type is uint, since it's never going to be large. This is to match the inbox parameter that does the same thing. As an implementation detail, this one converts to uint64 for arithmetic later on.

Issue #3179, opened 2019-08-02
Deconstructing a filecoin miner


travisperson:
This relates closely to #2968

----------------------------------------

Currently when we talk about mining we are referring to three things.

1) Gaining block rewards (`block miner`)
2) Create `commitSector` and `submitPoSt` messages (`power miner`)
3) Accepting storage deals (`storage miner`)

While these three things are related, not all of them depend on each other. Being a `block
miner` requires one to be a `power miner`, but there is no actual requirement in the other
direction. A `power miner` is not required to be a `block miner` (it's just rational to do so).

There is also no requirement that a `power miner` accept storage deals and become a `storage
miner`. Accepting deals and sealing pieces of data into sectors is just a utility of the PoSt
proofs.

Currently we require a miner to be all three. This causes a few issues in test, and adds some
additional overhead that is not required.

1) The only way to gain power on the network at the moment is through storage deals.
   This adds unneed overhead for a `power miner`. A Piece-Inclusion-Proof must be generated
   in this process, though I believe it's not required at all for power generation. Piece-Inclusion-Proof
   are primarly so a `storage client` can have proof that a `storage miner` has included a piece of data
   is in a particular sector.
   
2) To accept storage deals, a `storage miner` must also be a `block miner`.
   This imposes a challenge in testing primarly at the moment, as pointed out in #2968.
   This is purely due to that fact that the only way to register our storage deal protocols
   is through starting the `block miner` scheduler (the _mining start_ command).
   
3) A `power miner` currently must be a `block miner`, or they cannot do PoSt proofs.
   This is due to the fact that the `storage miner` (as in the miner side of the storage protocol)
   is  responsible for the scheduling of PoSt proof generation. So this ties everything together.

anorth:
Thanks @travisperson !

This is an area of code that is not in good shape, and also a pain for miner operator customization #3140.

taoshengshi:
fantastic! thanks @travisperson ,and I can participate the design and implementation.

taylorshuang:
The PoST module has a big problem. I don't know if the next version will improve. No matter how much data I upload to the miners, dashboard will not change anything.

taylorshuang:
![image](https://user-images.githubusercontent.com/1488931/62508992-22a07380-b83c-11e9-9f54-94ce148c57d0.png)
It's been two weeks.

travisperson:
@taylorshuang you are probably running into one of two bug

[...]

Issue #3140, opened 2019-07-29
Clean up node construction and initialization for easier customization


anorth:
### Description
The go-filecoin code implements a relatively simple single-homed miner architecture. Miner operators may wish to create relatively complex and distributed architectures. We don't intend to implement much in the way of distributed storage architecture in go-filecoin (at least in the near term), but we should make it easier for miner operators to customise this code for their use.

The initialization code in `node.go` is a bit of a mess, and in particular includes a bunch of mining worker functionality. Construction/initialization and worker routines are probably just what operators want to customize. This mess makes makes it difficult for an operator to write their own node initialization and/or worker and keep it in sync with the rest of the implementation.

Let's clean this up.

### Acceptance criteria
- worker modules (block production, sector committing & proving, etc) are clearly isolated and can be instantiated in isolation
- initialization code in `node.go` is straightforward wiring up of components according to configuration

PR #3271, opened 2019-08-22
Remove mining starts from FAST integration tests

### LINKED ISSUE #2968: Add ability to propose storage deal with miner without running the mining scheduler

acruikshank:
closes #2968

### Problem

We have a lot of flaky FAST tests. A prime candidate for this flakiness is that we have a lot of extraneous processes running in the background that get triggered by `MiningStart`. We have to run mining start because it sets up some key infrastructure for storage mining. We also use mining start so that a process will mine message sent by blocking commands. We would like to not trigger mining at all when we just need to prepare the miner for accepting deals, and only use `MiningOnce` when we need to mine messages.

### Solution

The solution is multi-faceted:

1. Create a new `mining setup` command (with associated plumbing and functionality in node) to perform just the miner initialization without starting any schedulers. This command will only instantiate the miner functionality if it does not already exist, and gets called automatically by `MiningStart` and `MiningOnce`. Because it is a command it can be called independently by FAST tests.
2. Introduce `setup.CtxMiningNext` that will mine some number of messages from a blocking command that requires mining to halt. This command can be called prior to a blocking command to ensure its messages will be mined.
3. Fix `mpool ls --wait-for-count` and add it to FAST. The message waiting functionality is what allows us to ensure the message to be mined lands in the message pool prior to mining. This functionality was broken because it was using a libp2p subscription to wait for messages on the same topic as the subscription to put the message in the pool. The ordering in which these handlers fire is arbitrary, and the waiter was checking before the message landed, then waiting for the next message before checking again (which was just hanging). I simplified this to a simple polling mechanism with a short sleep.
4. Un-skip the skipped flaky FAST tests, remove many of the mining starts, add `CtxMiningOnce` and `CtxMiningNext` where necessary.

### Considerations

This solution is not guaranteed to de-flakify our tests. It's a reasonable assumption that it will help. If it doesn't we will need to revisit this, but we should at least have better tools at our disposal.

`CtxMiningNext` is a little tricky to work with. Its advantage is that we get precise control over how many times mining once is called with some guarantees that mining won't race message creation. The disadvantage is that it requires the test developer to understand how many messages will need to be mined by the following blocking command. 

If `CtxMiningNext` is called and the blocking experiences an expected error that causes it to exit early, the test will fail when the daemon running the mpool wait is shut down. I've taken a stab at some patterns for making this sane, but there is no argument that this is more complicated than firing up `mining start` at the beginning of the test. At best, we will need to wait an see if the extra determinism is worth it.

### DIFF
--- a/commands/mining_daemon_test.go
+++ b/commands/mining_daemon_test.go
@@ -69,6 +69,9 @@ func TestMiningSealNow(t *testing.T) {
 
 	sinfo := pparams.SupportedSectors[0]
 
+	// mine the create storage message, then mine the set ask message
+	series.CtxMiningNext(ctx, 2)
+
 	_, err = series.CreateStorageMinerWithAsk(ctx, minerNode, big.NewInt(500), big.NewFloat(0.0001), big.NewInt(3000), sinfo.Size)
 	require.NoError(t, err)
 
@@ -82,7 +85,8 @@ func TestMiningSealNow(t *testing.T) {
 		require.NoError(t, minerNode.MiningStop(ctx))
 	}()
 
-	// Mine the commitSector and the submitPoSt
+	// Since the miner does not yet have power, we still need the genesis node to mine
+	// the miner's commitSector and the submitPoSt messages
 	series.CtxMiningNext(ctx, 2)
 
 	// start sealing
--- a/commands/mining_daemon_test.go
+++ b/commands/mining_daemon_test.go
@@ -52,16 +52,11 @@ func TestMiningSealNow(t *testing.T) {
 		InitOpts:   []fast.ProcessInitOption{fast.POAutoSealIntervalSeconds(1)},
 		DaemonOpts: []fast.ProcessDaemonOption{fast.POBlockTime(50 * time.Millisecond)},
 	})
-	env.RunAsyncMiner()
 	defer func() {
 		require.NoError(t, env.Teardown(ctx))
 	}()
 
 	genesisNode := env.GenesisMiner
-	require.NoError(t, genesisNode.MiningStart(ctx))
-	defer func() {
-		require.NoError(t, genesisNode.MiningStop(ctx))
-	}()
 
 	minerNode := env.RequireNewNodeWithFunds(1000)
 
@@ -80,6 +75,15 @@ func TestMiningSealNow(t *testing.T) {
 	// get address of miner so we can check power
 	miningAddress, err := minerNode.MiningAddress(ctx)
 	require.NoError(t, err)
+
+	// start mining for miner node to seal and schedule PoSting
+	require.NoError(t, minerNode.MiningStart(ctx))
+	defer func() {
+		require.NoError(t, minerNode.MiningStop(ctx))
+	}()
+
+	// Mine the commitSector and the submitPoSt
+	series.CtxMiningNext(ctx, 2)
 
 	// start sealing
 	err = minerNode.SealNow(ctx)
--- a/porcelain/mpool.go
+++ b/porcelain/mpool.go
@@ -2,37 +2,23 @@ package porcelain
 
 import (
 	"context"
+	"time"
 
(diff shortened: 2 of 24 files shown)


protocol/storage/client.go
```diff
@@ -189,12 +189,7 @@ func (smc *Client) ProposeDeal(ctx context.Context, miner address.Address, data
 	// create payment information
 	totalCost := price.MulBigInt(big.NewInt(int64(pieceSize * duration)))
 	if totalCost.GreaterThan(types.ZeroAttoFIL) {
-		// The payment setup requires that the payment is mined into a block, currently we
-		// will wait for at most 5 blocks to be mined before giving up
-		ctxPaymentSetup, cancel := context.WithTimeout(ctx, 5*smc.api.BlockTime())
```
acruikshank: I don't know where this came from, but it's not a good idea. This is actually timing out after 5 rounds, not 5 blocks. 5 consecutive null blocks won't be that uncommon. Also when testing with a short block time, this is a very short duration that can contribute to flaky tests.

commands/mining_daemon_test.go
```diff
@@ -81,6 +76,15 @@ func TestMiningSealNow(t *testing.T) {
...
 	require.NoError(t, err)
 
+	// start mining for miner node to seal and schedule PoSting
+	require.NoError(t, minerNode.MiningStart(ctx))
+	defer func() {
+		require.NoError(t, minerNode.MiningStop(ctx))
+	}()
+
+	// Mine the commitSector and the submitPoSt
+	series.CtxMiningNext(ctx, 2)
```
anorth: Since this test still uses `MiningStart`, this is confusing at best.

I *think* what's happening is that the genesis node is not mining, and this `MiningNext` is an instruction to it. But the node under test `minderNode` is mining. But it has no power yet and so won't produce blocks? If that's write, please describe this situation in a comment.

Later, I think we want to separate block production from storage-power-mining (and both from deal-making), after which this could be more clear, and we could prevent the node under test from attempting to produce blocks. Re #3179, #3140

porcelain/mpool.go
```diff
@@ -2,37 +2,23 @@ package porcelain
...
 
-	"github.com/filecoin-project/go-filecoin/net"
-	"github.com/filecoin-project/go-filecoin/net/pubsub"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 // The subset of plumbing used by MessagePoolWait
 type mpwPlumbing interface {
 	MessagePoolPending() []*types.SignedMessage
-	PubSubSubscribe(topic string) (pubsub.Subscription, error)
```
anorth: 💯

PR #3272, opened 2019-08-22
Ticket array

### LINKED ISSUE #2222: EC Upgrade 1: ticket arrays

ZenGround0:
Closes #2222 This is just the datastructure change from a single ticket per block to an array of tickets.  There are no updates to the way tickets are created and used here.

### DIFF
--- a/consensus/block_validation.go
+++ b/consensus/block_validation.go
@@ -86,8 +86,13 @@ func (dv *DefaultBlockValidator) ValidateSyntax(ctx context.Context, blk *types.
 	if blk.Miner.Empty() {
 		return fmt.Errorf("block %s has nil miner address", blk.Cid().String())
 	}
-	if len(blk.Ticket) == 0 {
-		return fmt.Errorf("block %s has nil ticket", blk.Cid().String())
+	if len(blk.Tickets) == 0 {
+		return fmt.Errorf("block %s has no tickets", blk.Cid().String())
+	}
+	for _, ticket := range blk.Tickets {
+		if len(ticket.VRFProof) == 0 {
+			return fmt.Errorf("block %s has nil ticket", blk.Cid().String())
+		}
 	}
 	// TODO validate block signature: 1054
 	return nil
--- a/mining/worker.go
+++ b/mining/worker.go
@@ -182,13 +182,14 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
 
 	challenge, err := consensus.CreateChallengeSeed(base, uint64(nullBlkCount))
 	if err != nil {
+		log.Warningf("Worker.Mine couldn't create challenge seed: %s", err.Error())
 		outCh <- Output{Err: err}
 		return false
 	}
 	prCh := createProof(challenge, w.createPoSTFunc)
 
 	var proof types.PoStProof
-	var ticket []byte
+	var ticket types.Ticket
 	select {
 	case <-ctx.Done():
 		log.Infof("Mining run on base %s with %d null blocks canceled.", base.String(), nullBlkCount)
--- a/testhelpers/consensus.go
+++ b/testhelpers/consensus.go
@@ -77,14 +77,14 @@ func NewValidTestBlockFromTipSet(baseTipSet types.TipSet, stateRootCid cid.Cid, 
 	ticket, _ := consensus.CreateTicket(poStProof, minerWorker, signer)
 
 	return &types.Block{
-		Miner:        minerAddr,
-		Ticket:       ticket,
-		Parents:      baseTipSet.Key(),
-		ParentWeight: types.Uint64(10000 * height),
-		Height:       types.Uint64(height),
-		Nonce:        types.Uint64(height),
-		StateRoot:    stateRootCid,
-		Proof:        poStProof,
+		Miner:         minerAddr,
+		Tickets:       []types.Ticket{ticket},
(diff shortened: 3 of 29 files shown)


ZenGround0:
cc @shannonwells

consensus/block_validation.go
```diff
@@ -86,8 +86,8 @@ func (dv *DefaultBlockValidator) ValidateSyntax(ctx context.Context, blk *types.
 	if blk.Miner.Empty() {
 		return fmt.Errorf("block %s has nil miner address", blk.Cid().String())
 	}
-	if len(blk.Ticket) == 0 {
-		return fmt.Errorf("block %s has nil ticket", blk.Cid().String())
+	if len(blk.Tickets) == 0 {
+		return fmt.Errorf("block %s has no tickets", blk.Cid().String())
```
anorth: This should be extended to validate each ticket, so that each election proof is non-empty too.

mining/worker.go
```diff
@@ -182,13 +182,14 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
 
 	challenge, err := consensus.CreateChallengeSeed(base, uint64(nullBlkCount))
 	if err != nil {
+		log.Warningf("Worker.Mine couldn't create challenge seed")
```
anorth: Include `err`

node/node_test.go
```diff
@@ -138,6 +138,8 @@ func TestNodeStartMining(t *testing.T) {
 	seed := node.MakeChainSeed(t, node.TestGenCfg)
 	minerNode := node.MakeNodeWithChainSeed(t, seed, []node.ConfigOpt{}, node.PeerKeyOpt(node.PeerKeys[0]), node.AutoSealIntervalSecondsOpt(1))
 
+	fmt.Printf("made miner node\n")
```
anorth: Remove the prints

testhelpers/consensus.go
```diff
@@ -77,14 +77,14 @@ func NewValidTestBlockFromTipSet(baseTipSet types.TipSet, stateRootCid cid.Cid,
...
-		Miner:        minerAddr,
-		Ticket:       ticket,
-		Parents:      baseTipSet.Key(),
-		ParentWeight: types.Uint64(10000 * height),
-		Height:       types.Uint64(height),
-		Nonce:        types.Uint64(height),
-		StateRoot:    stateRootCid,
-		Proof:        poStProof,
+		Miner:         minerAddr,
+		Tickets:       []types.Ticket{ticket},
```
anorth: Spacing is off

types/ticket.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Data derived by running a VDF on VRFProof
+	VDFResult VDFY
+
+	// A proof of delay during computation of VDFResult
+	VDFProof VDFPi
+}
+
+// Bytes returns the canonical byte ordering of the ticket
+func (t Ticket) Bytes() []byte {
```
anorth: This could have a better name. `SortKey()`?

types/tipset.go
```diff
@@ -27,6 +27,8 @@ var (
 	errNoBlocks = errors.New("no blocks for tipset")
 	// errUndefTipSet is returned from tipset methods invoked on an undefined tipset.
 	errUndefTipSet = errors.New("undefined tipset")
+	// errNoTickets is returned from tipset methods invoked on tipsets with malformed ticket arrays.
+	errNoTickets = errors.New("Uncomparible block with empty ticket array in tipset")
```
anorth: *Incomparable*

types/tipset.go
```diff
@@ -110,11 +116,14 @@ func (ts TipSet) ToSlice() []*Block {
 }
 
 // MinTicket returns the smallest ticket of all blocks in the tipset.
-func (ts TipSet) MinTicket() (Signature, error) {
+func (ts TipSet) MinTicket() (Ticket, error) {
 	if len(ts.blocks) == 0 {
-		return nil, errUndefTipSet
+		return Ticket{}, errUndefTipSet
+	}
+	if len(ts.blocks[0].Tickets) == 0 {
```
anorth: Add a TODO here with issue reference for doing this properly.

types/tipset.go
```diff
@@ -110,11 +116,14 @@ func (ts TipSet) ToSlice() []*Block {
...
+		return Ticket{}, errNoTickets
 	}
-	return ts.blocks[0].Ticket, nil
+	return ts.blocks[0].Tickets[0], nil
```
anorth: Add a TODO here with issue reference for doing this properly.

consensus/block_validation.go
anorth: I think I meant VRFProof. In any case, the old code checked that the byte array was not empty, but this has been lost. Validate that the byte arrays in each ticket are not empty (better: are the expected length).

PR #3275, opened 2019-08-22
register block syntax validation on pubsub

### LINKED ISSUE #2783: Prevent propagation of invalid blocks with block pubsub topic validator

frrist:
### What
This PR does the following things: 
- adds a block syntax validator to the node pubsub and closes #2783. I have created #3277 as a placeholder for bringing block validation inline with the spec.
- also adds a mockable block validator, the topic validator uses this in its unit test.
- adds an integration that could be improved in #3285 
- adds metrics to track the number of decode and syntax failures.

### DIFF
--- a/net/validators.go
+++ b/net/validators.go
@@ -8,10 +8,13 @@ import (
 	"github.com/libp2p/go-libp2p-pubsub"
 
 	"github.com/filecoin-project/go-filecoin/consensus"
+	"github.com/filecoin-project/go-filecoin/metrics"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
 var blockTopicLogger = logging.Logger("net/block_validator")
+var mDecodeBlkFail = metrics.NewInt64Counter("net/pubsub_block_decode_failure", "Number of blocks that fail to decode seen on BlockTopic pubsub channel")
+var mInvalidBlk = metrics.NewInt64Counter("net/pubsub_invalid_block", "Number of blocks that fail syntax validation seen on BlockTopic pubsub channel")
 
 // BlockTopicValidator may be registered on go-libp2p-pubsub to validate pubsub messages on the
 // BlockTopic.
@@ -28,10 +31,12 @@ func NewBlockTopicValidator(bv consensus.BlockSyntaxValidator, opts ...pubsub.Va
 			blk, err := types.DecodeBlock(msg.GetData())
 			if err != nil {
 				blockTopicLogger.Debugf("block from peer: %s failed to decode: %s", p.String(), err.Error())
+				mDecodeBlkFail.Inc(ctx, 1)
 				return false
 			}
 			if err := bv.ValidateSyntax(ctx, blk); err != nil {
 				blockTopicLogger.Debugf("block: %s from peer: %s failed to validate: %s", blk.Cid().String(), p.String(), err.Error())
+				mInvalidBlk.Inc(ctx, 1)
 				return false
 			}
 			return true
--- a/net/validators.go
+++ b/net/validators.go
@@ -13,18 +13,17 @@ import (
 
 var blockTopicLogger = logging.Logger("net/block_validator")
 
-// BlockTopicValidator may be registered on go-libp2p-pubsub to validate
-// pubsub messages on the BlockTopic.
+// BlockTopicValidator may be registered on go-libp2p-pubsub to validate pubsub messages on the
+// BlockTopic.
 type BlockTopicValidator struct {
 	validator pubsub.Validator
 	opts      []pubsub.ValidatorOpt
 }
 
-// NewBlockTopicValidator retruns a BlockTopicValidator using `bv` for
-// message validation
+// NewBlockTopicValidator retruns a BlockTopicValidator using `bv` for message validation
 func NewBlockTopicValidator(bv consensus.BlockSyntaxValidator, opts ...pubsub.ValidatorOpt) *BlockTopicValidator {
 	return &BlockTopicValidator{
-		opts: nil,
+		opts: opts,
 		validator: func(ctx context.Context, p peer.ID, msg *pubsub.Message) bool {
 			blk, err := types.DecodeBlock(msg.GetData())
 			if err != nil {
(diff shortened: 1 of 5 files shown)


net/validators.go
```diff
@@ -0,0 +1,58 @@
...
+// BlockTopicValidator may be registered on go-libp2p-pubsub to validate
+// pubsub messages on the BlockTopic.
+type BlockTopicValidator struct {
+	validator pubsub.Validator
+	opts      []pubsub.ValidatorOpt
+}
+
+// NewBlockTopicValidator retruns a BlockTopicValidator using `bv` for
+// message validation
+func NewBlockTopicValidator(bv consensus.BlockSyntaxValidator, opts ...pubsub.ValidatorOpt) *BlockTopicValidator {
```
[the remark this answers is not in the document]
frrist: opps missed that

net/validators_test.go
```diff
@@ -0,0 +1,128 @@
...
+	validator := tv.Validator()
+
+	assert.Equal(t, net.BlockTopic, tv.Topic())
+	assert.True(t, validator(ctx, pid1, blkToPubSub(goodBlk)))
+	assert.False(t, validator(ctx, pid1, blkToPubSub(badBlk)))
+	assert.False(t, validator(ctx, pid1, nonBlkPubSubMsg()))
+}
+
+func TestBlockPubSubValidation(t *testing.T) {
+	tf.IntegrationTest(t)
```
frrist: TODO before merging file an issues as this test has some issues since pubsub messages are not guaranteed to arrive in order.

net/validators_test.go
```diff
@@ -0,0 +1,148 @@
...
+		StateRoot: types.NewCidForTestGetter()(),
+		Miner:     miner,
+		// TODO validate ticket arrays #2222
+		Ticket: types.Signature([]byte("1")),
+	}
+	// publish the invalid block
+	err = fsub1.Publish(btv.Topic(), validBlk.ToNode().RawData())
+	assert.NoError(t, err)
+
+	// FIXME: #3285
```
frrist: @anorth here's an integration test and an issue to make it better

net/validators.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var blockTopicLogger = logging.Logger("net/block_validator")
+
+// BlockTopicValidator may be registered on go-libp2p-pubsub to validate
+// pubsub messages on the BlockTopic.
+type BlockTopicValidator struct {
+	validator pubsub.Validator
+	opts      []pubsub.ValidatorOpt
+}
```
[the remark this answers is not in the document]
frrist: Not sure I can do that here. libp2p accepts 3 parameters, a topic string, a validator function and validator opts.
https://github.com/libp2p/go-libp2p-pubsub/blob/97e63e477ee7ac22e92d7e8232230586d0d60504/pubsub.go#L787-L807 

Here's where we plug em in: https://github.com/filecoin-project/go-filecoin/pull/3275/files#diff-380e79adccbaf1276bf6c91cc551cd7fR447

net/validators_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// setup a block validator and a topic validator
+	bv := consensus.NewDefaultBlockValidator(blocktime, mclock)
+	btv := net.NewBlockTopicValidator(bv)
+
+	// setup a floodsub instance on the host and register the topic validator
+	fsub1, err := pubsub.NewFloodSub(ctx, host1, pubsub.WithMessageSigning(false))
+	require.NoError(t, err)
+	err = fsub1.RegisterTopicValidator(btv.Topic(), btv.Validator(), btv.Opts()...)
+	require.NoError(t, err)
```
anorth: Thanks for this test, it's a good start.

I was hoping to exercise the logic in the node that does this hookup (e.g. via `GenNode`), rather than replicate it here. Right now this test is checking that libp2p does what it promises, which is not without value, but also not testing much of our code. 

Let's leave this test here, but could you add to the comments that the goal is to instantiate a node and check that it receives but rejects the invalid block header. That will require some more precise control over the node than is currently easy (e.g. injecting the fake clock), but I'm working on it!

testhelpers/consensus.go
```diff
@@ -145,6 +145,43 @@ func (fbv *FakeBlockValidator) ValidateSyntax(ctx context.Context, blk *types.Bl
...
 }
 
+// MockBlockValidator is a mockable block validator.
+type MockBlockValidator struct {
+	syntaxStubs   map[cid.Cid]error
+	semanticStubs map[cid.Cid]error
+}
+
+// NewMockBlockValidator creates a MockBlockValidator that allows errors to configured
+// for blocks passed to the Validate* methods.
```
anorth: "Stub" or "Fake" would be a better name than "Mock" for this behaviour.

PR #3276, opened 2019-08-22
Cleanliness is next to godliness

ZenGround0:
This PR removes the last of the ugly store test setup in favor of the new chain builder.

While trying to close out #2222 I found myself making laborious fixes to this code and realizing that it was all scheduled to be erased.  It seemed like as good a time as any to do this.

#2128 was called out in the chain store test comments as tracking cleanup.  I don't think we should close that with this merging because there is some unresolved and healthy discussion there.

### DIFF
--- a/chain/store_test.go
+++ b/chain/store_test.go
@@ -19,16 +19,18 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// Chain diagram below.  Note that blocks in the same tipset are in parentheses.
+// Default Chain diagram below.  Note that blocks in the same tipset are in parentheses.
 //
 // genesis -> (link1blk1, link1blk2) -> (link2blk1, link2blk2, link2blk3) -> link3blk1 -> (null block) -> (null block) -> (link4blk1, link4blk2)
-func newChainStore2(r repo.Repo, genCid cid.Cid) *chain.Store {
+
+// newChainStore creates a new chain store for tests.
+func newChainStore(r repo.Repo, genCid cid.Cid) *chain.Store {
 	return chain.NewStore(r.Datastore(), hamt.NewCborStore(), &state.TreeStateLoader{}, genCid)
 }
 
-// requirePutTestChain2 puts the count tipsets preceding head in the source to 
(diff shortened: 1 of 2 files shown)


chain/store_test.go
```diff
@@ -406,94 +182,90 @@ func TestGetTipSetState(t *testing.T) {
...
+	builder := chain.NewBuilder(t, address.Undef)
+	genTS := builder.NewGenesis()
+	r := repo.NewInMemoryRepo()
+	cs := newChainStore(r, genTS.At(0).Cid())
+
+	// Construct test chain data
+	link1 := builder.AppendOn(genTS, 2)
+	link2 := builder.AppendOn(link1, 3)
+	link3 := builder.AppendOn(link2, 1)
+	link4 := builder.BuildOn(link3, 2, func(bb *chain.BlockBuilder, i int) { bb.IncHeight(2) })
```
ZenGround0: I could abstract out the frequently repeated lines from 187 to 196 if reviewers prefer.

chain/store_test.go
```diff
@@ -9,273 +9,37 @@ import (
...
 	"github.com/filecoin-project/go-filecoin/state"
-	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
-// This setup stuff is a horrendous legacy holdover from when these store tests shared a
-// common monolithic setup with the syncer tests. There's no good reason for it to still exist.
-// https://github.com/filecoin-project/go-filecoin/issues/2128
-type SyncerTestParams struct {
```
frrist: :heart: :1st_place_medal: :fireworks: YOU ROCK

shannonwells: ❤️ 🎧 🎸

chain/store_test.go
anorth: No need IMO, but you could pull a lot of these simpler tests into `t.Run()` sharing the same builder/links and chain.

PR #3284, opened 2019-08-23
remove proof parameter cache from most circleci jobs

ognots:
recent changes in https://github.com/filecoin-project/go-filecoin/pull/3252 removed downloading proof parameters in deps and smartdeps build stages.
this PR removes CircleCI cache for proof parameters cache in all steps except docker build, where it's required.
moves save_cache directive to tthe docker build job


.circleci/config.yml
```diff
@@ -145,21 +137,13 @@ jobs:
       # restore_cache step above can override the go-filecoin code with cached code
       - git_fetch_all_tags
       - checkout
-      - restore_cache:
```
laser: BLOCKING: If you eliminate this `restore_cache`, you will _always_ attempt to download (and, failing that, generate) Groth parameters and verifying keys during the `"deps"` command.

.circleci/config.yml
```diff
@@ -52,8 +52,6 @@ jobs:
...
```
laser: BLOCKING: See comment below. Removing this `restore_cache` will cause `"deps"` to download/generate Groth parameters.

ognots:
this approach is borked, opening a new PR in  it's place

PR #3286, opened 2019-08-23
Use miner worker address (#2959)

### LINKED ISSUE #2959: Implement command and porcelain to set miner.Worker

shannonwells:
## Summary
Use the miner worker address from the miner's state in various places: ticket signing, PoSt proving, accepting proposals, etc.

Companion to PR #3261, resolves #2959

### DIFF
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -53,9 +53,8 @@ const dealsAwatingSealDatastorePrefix = "dealsAwaitingSeal"
 
 // Miner represents a storage miner.
 type Miner struct {
-	minerAddr  address.Address
-	ownerAddr  address.Address
-	workerAddr address.Address
+	minerAddr address.Address
+	ownerAddr address.Address
 
 	dealsAwaitingSealDs repo.Datastore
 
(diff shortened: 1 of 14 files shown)


protocol/storage/miner.go
```diff
@@ -336,7 +335,12 @@ func (sm *Miner) acceptProposal(ctx context.Context, p *storagedeal.SignedPropos
 		},
 	}
 
-	if err = resp.Sign(sm.porcelainAPI, sm.workerAddr); err != nil {
+	workerAddr, err := sm.porcelainAPI.MinerGetWorkerAddress(ctx, sm.minerAddr)
```
acruikshank: This code needs to be extracted into a method that can lookup the worker address and use it to sign a response. It could be called 3 times.

ingar: Extracted...

Issue #2225, opened 2019-03-09
EC Upgrade 4: integrate randomness sampling into the leader election rules


ZenGround0:
### Description
Right now the EC leader election proof and block validation only refers to data in the block and its parent.  [The new spec](https://github.com/filecoin-project/specs/blob/master/expected-consensus.md#secret-leader-election) includes upgrades to the protocol which entail sampling tickets from further back in the chain to prove and validate a block's status as a leader.

This issue tracks all the work needed to get the implementation of leader election in line with the spec.

### Acceptance criteria
* Block mining now actually uses randomness sampling to get tickets from the past.  Block mining includes "scratching off" a past ticket (using VRF) and including the scratched ticket ( i.e. VRF proof ) as an election proof (see the spec)
* ~Block mining now updates the ticket array of the new block properly by sampling a past ticket and running a mock VDF on that ticket before including it in the new ticket array.~
* Block validation checks election proofs with respect to the ticket a lookback parameter away and whatever notion of mock VRF validity we are using.
* ~Block validation checks tickets with respect to the previous ticket in the chain and whatever mock VRF and VDF validity we are using.~

If any of that doesn't make sense go read the spec!  If it still doesn't make sense open a spec issue and don't stop until it makes sense or the spec has been updated with any corrections!

### Where to begin
Read the spec!

anorth:
Closed by #3810

Issue #2223, opened 2019-03-09
EC Upgrade 2: randomness sampling from ticket chain


ZenGround0:
### Description
In a world with ticket arrays, randomness is sampled along the *ticket chain*.  We need to change randomness sampling fundamentally.

### Acceptance criteria
(1) Small design doc for moving randomness sampling from blockchain to ticket chain.

(2) Sync with the spec.  If things are underspecified [in the spec](https://github.com/filecoin-project/specs/blob/master/expected-consensus.md) then we need to get these details in the spec.  It should be pretty good though.

(3) Implement
* ancestor extraction now extracts tickets instead of blocks
* randomness sampling from the ticket chain conforms to the all the edge cases that by now should be in the spec.
* during block mining miners can correctly sample randomness from the tickets they have created as part of mining null blocks.  This edge case is essential for guaranteeing progress.
* please extensively test edge cases

### Risks + pitfalls
Lots of tricky details, many of which might exist as bits in the heads of others.

### Where to begin
Read the spec!

ZenGround0:
This is now tracking semantic changes to the ticket array
- [ ] mining null blocks adds ticket to ticket array
- [ ] create leader election lookback (can set to 1 for now)
- [ ] sampling randomness is now greatly simplified to iterating over max(N, M) tickets.  N is proving period height + post look back.  M is leader election look back
- [ ] use ticket iterator abstraction to help with the above
- [ ] mining updates sampling array to use null tickets to ensure liveness
- [ ] reason about / spec cross reference / spec update edge cases concerning iteration over ticket chain in the case of multiple winners with null blocks

anorth:
@ZenGround0 is ticket sampling complete, or if not could you jot down what we still need?

ZenGround0:
Not yet complete.  We need to introduce lookback into leader election and election validation.  I'm on this.

PR #3287, opened 2019-08-23
Separate election from ticket generation (Part of #2225)

ZenGround0:
## Why do we need this PR

Right now ticket generation and election proof generation are one and the same.  To get closer to the planned consensus construction we need to separate those things out.  This PR does so while at the same time tying the ticket and election proof operations to objects to organize functionality together.

- [ ] TicketMachine and ElectionMachine refactor
- [ ] Refactor consensus.Expected and mining.Worker to use these dependencies instead of consensus free functions
- [ ] Fix mining.Worker to use separate ticket generation and election proving operations

### Non-goals
- Introduce look back parameter for election proofs to finish out #2225 (this would require fake work if done before #2223)
- Semantic ticket generation correctness, i.e. multiple tickets per block when mining null blocks.  As a stop gap this PR includes a null block count parameter in ticket validation and generation.


ZenGround0:
@anorth @frrist I've put in effort to make this PR's commit history helpful for review.  It is best reviewed one commit at a time.  

- The first commit introduces the fundamental refactor that puts election proof and ticket generation and validation on separate objects 
- The second commit is a prefactor that turned out to be necessary in order to maintain existing structure.  It imperfectly adds a call to determine the miner worker address to the power table.  While this is not very clean it only messes with an abstraction that is arguably not that clean already (the power table).
- The third commit integrates the new election and ticket validation logic into the consensus validation rules checking
- The fourth commit integrates this logic into the mining worker.  It removes the dependency-injection `doSomeWorkFunc` because now these dependencies can be injected using the ticket generation interface.  It also adds a clock as a worker dependency so that sleep calls here can be made using a system aware clock.
- The fifth commit integrates all of the above into the node.  A lot of work went into getting node tests to pass.  Mocking out fake tickets and election proofs was challenging enough that I opted to actually mine winning blocks in process to use for the tests.  We'll need something better soon.

consensus/expected.go
```diff
@@ -278,29 +283,47 @@ func (c *Expected) RunStateTransition(ctx context.Context, ts types.TipSet, tsMe
...
+		result, err = c.ValidateTicket(prevTicket, blk.Tickets[0], workerAddr, uint64(blk.Height)-prevHeight-1)
+		if err != nil {
+			return errors.Wrap(err, "failed checking ticket array")
+		}
+		if !result {
+			return errors.New("invalid ticket extension")
+		}
 
 	}
+	return nil
```
anorth: Commentary: this looks like stuff that should move to the BlockSemanticValidator soon.

testhelpers/consensus.go
```diff
@@ -203,6 +218,53 @@ func (mbv *StubBlockValidator) StubSemanticValidationForBlock(child *types.Block
...
+}
+
+// FailingElectionValidator marks all elections as invalid
+type FailingElectionValidator struct{}
+
+// IsElectionWinner always returns false
+func (fev *FailingElectionValidator) IsElectionWinner(ctx context.Context, bs blockstore.Blockstore, ptv consensus.PowerTableView, st state.Tree, ticket types.Ticket, electionProof types.VRFPi, signerAddr, minerAddr address.Address) (bool, error) {
+	return false, nil
+}
+
```
anorth: `testhelpers` is a dumping ground, I don't want new stuff here. Put test helpers alongside the relevant code, in `consensus/testing.go`, or `consensus/testing_xxx.go` to split things up.

mining/worker.go
```diff
@@ -166,65 +160,83 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
...
-
 	log.Debugf("Mining on tipset: %s, with %d null blocks.", base.String(), nullBlkCount)
 	if ctx.Err() != nil {
 		log.Warningf("Worker.Mine returning with ctx error %s", ctx.Err().Error())
 		return false
 	}
 
-	challenge, err := consensus.CreateChallengeSeed(base, uint64(nullBlkCount))
+	// Read uncached worker address
+	workerAddr, err := w.api.MinerGetWorkerAddress(ctx, w.minerAddr)
```
anorth: This is one of those calls that should be clearly parameterised by the tipset keying the state in which the address is looked up, #3022. Can you make that change in a prefactor?

mining/worker.go
```diff
@@ -166,65 +160,83 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
...
-			log.Errorf("Worker.Mine got zero value from channel prChRead")
-			return false
-		}
-		proof := append(types.PoStProof{}, prChRead[:]...)
-		ticket, err = consensus.CreateTicket(proof, workerAddr, w.workerSigner)
-		if err != nil {
-			log.Errorf("failed to create ticket: %s", err)
-			return false
-		}
+	case <-done: //fallthrough
```
anorth: Use the keyword: https://github.com/golang/go/wiki/Switch#fall-through

mining/worker_test.go
```diff
@@ -19,19 +20,32 @@ import (
...
 	"github.com/filecoin-project/go-filecoin/core"
+	"github.com/filecoin-project/go-filecoin/clock"	
 	"github.com/filecoin-project/go-filecoin/mining"
 	"github.com/filecoin-project/go-filecoin/state"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+type flagTicketGen struct {
```
anorth: This looks like a mock to me, the "flag" name is confusing.

It could be made stronger by recording some of the parameters to methods, not just a boolean.

node/block_propagate_test.go
```diff
@@ -30,18 +30,42 @@ func connect(t *testing.T, nd1, nd2 *Node) {
...
+	// mine once with 0 null blocks
+	out := make(chan mining.Output)
+	var wonElection bool
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		wonElection = worker.Mine(ctx, headTipSet, 0, out)
+		wg.Done()
+	}()
+	next := <-out
```
anorth: Why the goroutine and channel suff here?

wallet/wallet_test.go
```diff
@@ -210,17 +210,56 @@ func TestWallet_CreateTicket(t *testing.T) {
...
+	})
+
+	t.Run("Returns error and empty election proof when signer is invald", func(t *testing.T) {
+		parent := types.Ticket{VDFResult: types.VDFY{0xbb}}
+		badAddress := address.TestAddress
+		ep, err := consensus.ElectionMachine{}.RunElection(parent, badAddress, w)
+		assert.Error(t, err)
+		assert.Equal(t, types.VRFPi{}, ep)
+	})
+}
```
anorth: These tests do not seem to have much to do with the wallet at all. Are they just testing it can sign things? We should just test that the `Sign` method works here, and move the rest of the test to `consensus` package (with a 
minimal signer).

anorth:
My feedback is mostly minor: great work. The opaque hex strings are perhaps my biggest problem.

The factoring here definitely needs some work so that we can write much more thorough tests, but well done on improving the situation.

node/block_propagate_test.go
ZenGround0: Just to interact with the miner worker directly.  I didn't want to call `StartMining` which I think I'd need to do to properly use the block api's `MineOnce` function.

mining/worker.go
ZenGround0: #3401 addresses this.  It may merge first, but I think this can go in before that if it is ready to go first as the race #3401 prevents is rare and is unlikely to cause test flakiness / bad user behavior in a short amount of time.

mining/worker.go
```diff
@@ -166,65 +160,83 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
...
 	}
-	prCh := createProof(challenge, w.createPoSTFunc)
 
-	// Read uncached worker address
-	workerAddr, err := w.api.MinerGetWorkerAddress(ctx, w.minerAddr)
+	// Create the next ticket.
+	// Derive from mining base's min ticket
+	prevTicket, err := base.MinTicket()
+	if err != nil {
+		log.Warningf("Worker.Mine couldn't read parent ticket")
```
frrist: Be sure to include the error here.

mining/worker.go
```diff
@@ -166,65 +160,83 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
...
+		outCh <- Output{Err: err}
+		return false
+	}
+	nextTicket, err := w.ticketGen.NextTicket(prevTicket, workerAddr, w.workerSigner, uint64(nullBlkCount))
 	if err != nil {
+		log.Warningf("Worker.Mine couldn't generate next ticket")
```
frrist: and here

mining/worker.go
frrist: This is confusing, I don't think we want to fallthrough here. I believe we intend to break, which is implicit. If that's the case remove the comment.

node/block_propagate_test.go
```diff
@@ -2,20 +2,20 @@ package node
...
-	"github.com/filecoin-project/go-filecoin/consensus"
+	"github.com/filecoin-project/go-filecoin/mining"
 	"github.com/filecoin-project/go-filecoin/protocol/storage"
-	"github.com/filecoin-project/go-filecoin/state"
-	"github.com/filecoin-project/go-filecoin/testhelpers"
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/types"
+
+	"github.com/libp2p/go-libp2p-core/peer"
```
frrist: These import go above the filecoin ones.

Issue #2674, opened 2019-04-29
Implement principled peer set & chain head bootstrapping


anorth:
### Description
Before syncing the chain, a node must determine from peers what the right chain head to sync is. Finding a peer set and chain head to trust is referred to as bootstrapping (distinct from libp2p DHT bootstrapping).

The spec https://github.com/filecoin-project/specs/pull/196 now describes a method, which we can implement.

Note that this is necessary for us to consider chain sync secure, but its absence should not block other chain sync work (fetching or validation).

### Acceptance criteria
- New or restarted nodes determine a peer set and chain head in accordance with the spec
- The spec is updated with any changes necessary during implementation

### Where to begin
The spec looks solid enough that additional technical design is probably not required.

anorth:
Note: I think some of the linked PR has since been removed from the spec, leaving discretion to implementers.

PR #3303, opened 2019-08-27
Perform catch up sync before enabling pubsub

frrist:
### What
The PR extends the hello protocol such that the `SayHello` method may be called with a boolean parameter `response` that will cause the method to behave synchronously if true, false the behaviour will remain the same.
This new logic is then wired up to the nodes initial sync logic and is used to ensure the node has caught up enough to trusted peers before disabling trust in calls to handle new tipset and enabling the block pubsub topic.

### What's left
- [ ] Get approval on (temporary?) hello protocol modification
  - If this is permanent then file an issue in the spec, else an issue here to remove it and reference #2674 
- [ ] Set the `trusted` flag to false where needed.
- [ ] Fix the Miner tests if this is well received.
- [ ] Implementing integration test

### DIFF
--- a/net/peer_tracker.go
+++ b/net/peer_tracker.go
@@ -18,14 +18,14 @@ var logPeerTracker = logging.Logger("peer-tracker")
 type PeerTracker struct {
 	// mu protects peers
 	mu sync.RWMutex
-	// peers maps stringified peer.IDs to info about their chains
-	peers map[string]*types.ChainInfo
+	// peers maps peer.IDs to info about their chains
+	peers map[peer.ID]*types.ChainInfo
 }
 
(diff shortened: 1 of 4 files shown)


net/peer_tracker.go
```diff
@@ -19,13 +19,13 @@ type PeerTracker struct {
 	// mu protects peers
 	mu sync.RWMutex
 	// peers maps stringified peer.IDs to info about their chains
-	peers map[string]*types.ChainInfo
+	peers map[peer.ID]*types.ChainInfo
```
anorth: Good, but remove "stringified" from the comment.

frrist:
Will revisit this after #3319 has been implemented, closing.

Issue #3107, opened 2019-07-23
Expose chain syncing information


travisperson:
### Description

Currently the only way to understand what is happening during chain syncing is to read through log messages.

This is difficult and not ideal for users, it's also not convenient for testing as log messages may change frequently and makes testing for certain information fragile.

We should provide a cli command for users that will provide information to users so they can understand the current state of their chain.

There is already work being done to improve the chain syncing of the daemon at the moment. This work would tail off of the improvements being made.

There are two main pieces of information we want to know

1) Is chain syncing running?
2) What are we syncing?
3) Current progress?

To answer these questions a first proposal would to be expose the start / end time of the last / current call to `HandleNewTipset`, along with the tipset key.

```
{
   "mode": "",
   "start": "",
   "end": "",
   "tipset": [
      ""
   ],
   "delta": 0
}
```
_key names are completely up for change_

### Acceptance criteria

- CLI command to expose chain syncing status
- Result contains the `start` to the last call of `HandleNewTipset`
- Result contains the `end` to the last call of `HandleNewTipset` or null if it's running
- Result contains the `mode` of the chain syncer (`Syncing` or `CaughtUp`).
- Result contains the `tipset` passed to `HandleNewTipset`.
- Result contains the `delta`, which is the block difference between the last tipset fetched and our chain height.

--------------------

I defer strongly to suggests from @frrist who is currently leading the chain syncing rework.

anorth:
Thanks for filing this. I think we can expose some great info about both downloading and validation, more than you've suggested, in a `chain status` command. @frrist can work out the details after we've finished the present refactors for graphsync and mode.

anorth:
Moar status! 

- Since this structure is intended for external consumption, we should decouple it from the internal ChainInfo (also, please extract the trusted flag from that).
- For long fetches, include information about how much has been fetched so far (a tool could then indicate progress-%).

```
type SyncStatus struct {
  // The heaviest that has been fully validated
  ValidatedHead types.TipSetKey
  // The height of ValidatedHead
  ValidatedHeadHeight uint64

  // They key of the head of the chain currently being fetched/validator, or undef if none
  SyncingTip types.TipSetKey
  // The height of SyncingTIp
  SyncingHeight uint64
  // Whether SyncingTip is trusted as a head far away from the validated head
  SyncingTrusted boolean
  // The lowest height fetched for SyncingTip
  SyncingFetchedToHeight uint64
[...]

PR #3305, opened 2019-08-27
add Status command to syncer

frrist:
### What
This PR address #3107 by adding a status value to the chain syncer. Its value is updated every time `HandleNewTipSet` runs. The syncer has been added to the plumbing API so that its status may be queried via the CLI.

### What's Left
- [x] Testing Status command
- [x] Fix linting once we decide on a name for the plumbing that holds the syncer


commands/chain.go
```diff
@@ -19,12 +19,31 @@ var chainCmd = &cmds.Command{
 		Tagline: "Inspect the filecoin blockchain",
 	},
 	Subcommands: map[string]*cmds.Command{
-		"head": chainHeadCmd,
-		"ls":   chainLsCmd,
+		"store": storeCmd,
+		"sync":  syncCmd,
```
anorth: Too complicated UX. Please flatten these out into a single `chain` subcommand.

commands/chain.go
```diff
@@ -108,3 +127,16 @@ var chainLsCmd = &cmds.Command{
 		}),
 	},
 }
+
+var syncStatusCmd = &cmds.Command{
```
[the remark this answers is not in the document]
anorth: I err against it without a concrete use case. We can add it alongside the FAST test (for example) that uses it.

plumbing/cst/chain_sync.go
```diff
@@ -0,0 +1,31 @@
...
+type chainSyncer interface {
+	HandleNewTipSet(context.Context, *types.ChainInfo, bool) error
+	Status() *chain.SyncerStatus
+}
+
+type ChainSyncReceiver struct {
+	syncer chainSyncer
+}
+
+func NewChainSyncReceiver(chainSyncer chainSyncer) *ChainSyncReceiver {
```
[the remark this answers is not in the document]
anorth: How about "ChainSyncProvider"

plumbing/api.go
```diff
@@ -64,6 +65,7 @@ type API struct {
 type APIDeps struct {
 	Bitswap       exchange.Interface
 	Chain         *cst.ChainStateProvider
+	Syncer        *cst.ChainSyncReceiver
```
anorth:
```suggestion
	Sync        *cst.ChainSyncProvider
```

chain/store_test.go
```diff
@@ -207,7 +208,7 @@ func initSyncTest(t *testing.T,
 
 	fetcher := th.NewTestFetcher()
 	fetcher.AddSourceBlocks(calcGenBlk)
-	syncer := chain.NewSyncer(con, chainStore, messageStore, fetcher) // note we use same cst for on and offline for tests
+	syncer := chain.NewSyncer(con, chainStore, messageStore, fetcher, th.NewFakeSystemClock(time.Now())) // note we use same cst for on and offline for tests
```
anorth: Don't use `time.Now()` in tests (or ever!), use a constant time value so tests are fully deterministic.

chain/syncer.go
```diff
@@ -99,10 +102,14 @@ type Syncer struct {
 	chainStore syncerChainReaderWriter
 	// Provides message collections given cids
 	messageProvider MessageProvider
+
+	statusMu sync.Mutex
```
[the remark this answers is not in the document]
anorth: The clock should be reentrant, so move it above the mutex to avoid implying that it's protected by it.

If the status value is to be mutated in place (as currently in endRun) then the mutex is most appropriate.

chain/syncer.go
```diff
@@ -308,6 +324,53 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
...
+
+func (s *SyncerStatus) String() string {
+	return fmt.Sprintf("running: %t, syncing: %s, trusted: %t, started: %s, head: %s, height: %d",
+		s.Running, s.NewChain.String(), s.Started, s.Trusted, s.CurrentHead, s.CurrentHeight)
+}
+
+// Status returns the syncers current status, this includes whether or not the syncer is currently
+// running, the chain its processsing/last processed, and the time it started processing said chain.
+func (syncer *Syncer) Status() *SyncerStatus {
+	return syncer.status
```
anorth: Blocking: this should return a copy, not a pointer to internal data.

chain/syncer.go
```diff
@@ -308,6 +324,53 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
...
+}
+
+func (syncer *Syncer) startRun(curHead types.TipSetKey, curHeight uint64, trusted bool, newChain *types.ChainInfo) {
```
anorth: The `start` and `end` run names sound like they do a lot more than they really do. I suggest something like `setRunStatus`.

net/fetcher.go
```diff
@@ -19,6 +20,7 @@ type Fetcher interface {
 	// this includes the provided `ts`. The TipSet that evaluates to true when
 	// passed to `done` will be in the returned slice. The returns slice of TipSets is in Traversal order.
 	FetchTipSets(ctx context.Context, tsKey types.TipSetKey, from peer.ID, done func(ts types.TipSet) (bool, error)) ([]types.TipSet, error)
+	Notifications() *pubsub.PubSub
```
frrist: I don't expect to land this with the package `pubsub` as the interface definition. This was the easy thing to do until we decide on a pull or push based system for this information.

chain/syncer_status.go
```diff
@@ -0,0 +1,173 @@
...
+		SyncingFetchComplete:   false,
+		SyncStarted:            s.clock.Now(),
+	}
+	s.statusMu.Unlock()
+	go func() {
+		fetchSub := s.fetchNotifi.Notifications().Sub(net.FetchTsTopic)
+		defer s.fetchNotifi.Notifications().Unsub(fetchSub, net.FetchTsTopic)
+		for {
+			select {
+			case raw := <-fetchSub:
```
frrist: An alternative to this could be a ticker that queries a method on the fetcher for its current highest fetched tipset.

chain/syncer_status.go
```diff
@@ -0,0 +1,173 @@
...
+				// update the status as the fetcher fetches things.
+				switch raw := raw.(type) {
+				case types.TipSet:
+					h, err := raw.Height()
+					if err != nil {
+						panic(err)
```
frrist: oops, will replace with a log message

chain/syncer.go
```diff
@@ -111,6 +116,21 @@ func NewSyncer(e syncStateEvaluator, s syncerChainReaderWriter, m MessageProvide
 		stateEvaluator:  e,
 		chainStore:      s,
 		messageProvider: m,
+		clock:           c,
+		status: &SyncerStatus{
+			ValidatedHead:       types.UndefTipSet.Key(),
+			ValidatedHeadHeight: 0,
```
anorth: I think you can read the above values from the store right away.

chain/syncer.go
```diff
@@ -308,6 +333,29 @@ func (syncer *Syncer) widen(ctx context.Context, ts types.TipSet) (types.TipSet,
 	return wts, nil
 }
 
+// UpdateStatus updates the syncer status.
+func (syncer *Syncer) UpdateStatus(update ...StatusUpdates) {
```
anorth: Probably shouldn't be exported.

commands/chain.go
```diff
@@ -108,3 +112,54 @@ var chainLsCmd = &cmds.Command{
...
+	Helptext: cmdkit.HelpText{
+		Tagline: "Sync and validate a chain.",
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("peerid", true, false, "Base58-encoded libp2p peer ID that a chain will be synced from"),
+		cmdkit.StringArg("cid", true, false, "CID of the TipSet to sync"),
+	},
+	Options: []cmdkit.Option{
+		cmdkit.BoolOption("trusted", "Trust the TipSet when syncing").WithDefault(true),
+		cmdkit.Uint64Option("height", "Height of the TipSet to sync").WithDefault(0),
```
anorth: This is a bit weird. If we had a serious need for this functionality, I think we would want to avoid the user needing to specify the height, since we can fetch the tipset and inspect its height. The HandleNewTipSet method signature is the right shape for our internal use, but not for external consumption. I'd rather we do this properly, exposing an ergonomic endpoint, or not at all until we're prepared to do that work.

frrist: A user only needs to specify a PeerID and CID, these other 2 are optional. But I hear ya, this is not really made for users. I personally would like to keep it as I'll probably use it occasionally, but if you have strong feelings about removing it I'll oblige.

chain/syncer.go
frrist: Not quite, the `chainStore` doesn't know what its head is when we initialize the syncer.

We call NewSyncer in the nodes `Build()` method:
https://github.com/filecoin-project/go-filecoin/blob/83becde01402803454b43ea75c77364d13e28564/node/node.go#L458
When it asks the `chainStore` for its head via `GetHead` it returns an empty tipset because we don't set the `chainStore` head until `Load` is called in node `Start()`:
https://github.com/filecoin-project/go-filecoin/blob/83becde01402803454b43ea75c77364d13e28564/node/node.go#L536

I could plug the genesisCID and height in here since the chainStore can tell us that at node build time, but that will be outdated once the chainStore loads its state.

frrist: I would prefer to leave the values undefined and wait for them to be updated once the syncer encounters a new tipset, what do you think?

anorth:
Ok

chain/syncer.go
hannahhoward: I wonder if since this is a somewhat global decoupled object if it should initialized and owned by the syncer? Perhaps it gets initialized on its own (looks like based on the values it wouldn't need any dependencies to write a constructor), and then the ChainReader gets a reference too that it can modify when Load gets called?

frrist: That sounds like a good idea, will push something up shortly.

frrist: pushed in https://github.com/filecoin-project/go-filecoin/pull/3305/commits/73b03f6b95ed6c1ccefcf996c9e9e77209d854bf

consensus/genesis.go
```diff
@@ -186,7 +187,6 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
 
 		genesis := &types.Block{
 			StateRoot:       c,
```
ZenGround0: rebase weirdness?

mining/worker_test.go
```diff
@@ -280,30 +279,20 @@ func TestGenerateMultiBlockTipSet(t *testing.T) {
...
+	assert.Equal(t, 2, baseTipset.Len())
+
+	blk, err := worker.Generate(ctx, baseTipset, types.Ticket{VRFProof: []byte{2}}, types.PoStProof{}, 0)
 	assert.NoError(t, err)
 
 	assert.Equal(t, types.EmptyMessagesCID, blk.Messages)
 	assert.Equal(t, types.EmptyReceiptsCID, blk.MessageReceipts)
 	assert.Equal(t, types.Uint64(101), blk.Height)
-	assert.Equal(t, types.Uint64(1020), blk.ParentWeight)
+	assert.Equal(t, types.Uint64(120), blk.ParentWeight)
```
ZenGround0: Is this also rebase weirdness?

net/validators.go
```diff
@@ -0,0 +1,60 @@
+package net
```
ZenGround0: I remember seeing this before too.  Is this more rebase weirdness or did this PR get closed down and now these commits are coming in?

If its the second one then please put these commits in another PR as they seem very extractable and orthogonal to the status command.

plumbing/cst/chain_sync.go
```diff
@@ -0,0 +1,39 @@
+package cst
+
+import (
```
ZenGround0: About time!

net/validators.go
ZenGround0: Definitely in #3305, what's up with these changes showing up here too?

chain/status.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// UpdateStatus updates the status heald by StatusReporter.
+func (sr *StatusReporter) UpdateStatus(update ...StatusUpdates) {
+	sr.statusMu.Lock()
+	defer sr.statusMu.Unlock()
+	for _, u := range update {
+		u(sr.status)
+	}
+	logSyncer.Infof("syncing status: %s", sr.status.String())
```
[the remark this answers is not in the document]
frrist: We log everytime the status _updates_, not every time it's queried via the CLI (that's the below method). I agree that info is too high, will create a new logger and log debug on that.

chain/status.go
```diff
@@ -0,0 +1,165 @@
...
+	// Whether SyncingHead has been validated.
+	SyncingComplete bool
+	// Whether SyncingHead has been fetched.
+	SyncingFetchComplete bool
+
+	// The key of the tipset currently being fetched
+	FetchingHead types.TipSetKey
+	// The height of FetchingHead
+	FetchingHeight uint64
+}
```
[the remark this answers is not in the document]
frrist: These are a slightly modified version of what was proposed here: https://github.com/filecoin-project/go-filecoin/issues/3107#issuecomment-525986096

I realize now that `ValidatedStarted` should instead be `SyncingStarted`, will correct that. The rest seem good for now, we can rework in a follow on.

chain/status.go
```diff
@@ -0,0 +1,165 @@
...
+		s.SyncingHead, s.SyncingHeight, s.SyncingTrusted, s.SyncingComplete, s.SyncingFetchComplete,
+		s.FetchingHead, s.FetchingHeight)
+}
+
+// StatusUpdates defines a type for updateing syncer status.
+type StatusUpdates func(*Status)
+
+//
+// Validation Updates
+//
```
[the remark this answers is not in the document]
frrist: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

chain/store.go
```diff
@@ -364,3 +373,8 @@ func (store *Store) GenesisCid() cid.Cid {
 func (store *Store) Stop() {
 	store.headEvents.Shutdown()
 }
+
+// Status returns the current chain status.
```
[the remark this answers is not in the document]
frrist: The Status method exists on the store for testing the store sets the status correctly on `SetHead` calls.

> The syncer is the only caller of SetHead

The chain store `Load()` method also calls `SetHead()`, and since the Syncer is instantiated before the store is loaded we needed a way to update the status once the store was done loading. The calls in the syncer code were extraneous, will remove those.

chain/syncer.go
```diff
@@ -338,16 +351,29 @@ func (syncer *Syncer) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo,
...
+		syncHead(ci.Head), syncHeight(ci.Height), syncTrusted(trusted), syncComplete(false))
+	defer syncer.reporter.UpdateStatus(syncComplete(true))
+
 	// If we do not trust the peer head check finality
 	if !trusted && syncer.exceedsFinalityLimit(curHeight, ci.Height) {
 		return ErrNewChainTooLong
 	}
 
+	syncer.reporter.UpdateStatus(syncFetchComplete(false))
+	defer syncer.reporter.UpdateStatus(syncFetchComplete(true))
```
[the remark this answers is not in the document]
frrist: True, I didn't like plugging this in before the error check, but since UpdateStatus doesn't I'll just do that.

chain/status.go
```diff
@@ -0,0 +1,169 @@
...
+		status: newDefaultChainStatus(),
+	}
+}
+
+// Status defines a structure used to represent the state of a chain store and syncer.
+type Status struct {
+	// The heaviest TipSet that has been fully validated.
+	ValidatedHead types.TipSetKey
+	// The height of ValidatedHead.
+	ValidatedHeadHeight uint64
```
frrist: Current value of the chain store head and its height.

chain/status.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// They head of the chain currently being fetched/validated, or undef if none.
+	SyncingHead types.TipSetKey
+	// The height of SyncingHead.
+	SyncingHeight uint64
+	// Whether SyncingTip is trusted as a head far away from the validated head.
+	SyncingTrusted bool
+	// Unix time at which syncing of chain at SyncingHead began, zero if valdation hasn't started.
+	SyncingStarted int64
```
frrist: The head passed to HandleNewTipSet, its height, whether we trust it, and when we started processing it.

chain/status.go
```diff
@@ -0,0 +1,169 @@
...
+	// Whether SyncingHead has been validated.
+	SyncingComplete bool
+	// Whether SyncingHead has been fetched.
+	SyncingFetchComplete bool
```
frrist: when we completed processing `SyncingHead` and when we completed fetching `SyncingHead`

chain/status.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The key of the tipset currently being fetched
+	FetchingHead types.TipSetKey
+	// The height of FetchingHead
+	FetchingHeight uint64
```
frrist: The head and height of the tipeset the fetcher is current;y fetching, these will change as the fetcher traverses from `SyncingHead` back until its done callback is satisfied.

chain/store.go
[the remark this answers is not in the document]
frrist: @ZenGround0 are you alright leaving this here given the above?

chain/syncer.go
```diff
@@ -338,18 +346,30 @@ func (syncer *Syncer) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo,
 		return err
 	}
 
+	syncer.reporter.UpdateStatus(syncingStarted(syncer.clock.Now().Unix()), syncHead(ci.Head), syncHeight(ci.Height), syncTrusted(trusted), syncComplete(false))
+	defer syncer.reporter.UpdateStatus(syncComplete(true))
+
```
frrist: update that we are trying to sync a new head, when we started, and if we trusts it.
when this method exists we are done syncing the new head, it may or may not be our new validated head.

chain/syncer.go
```diff
@@ -338,18 +346,30 @@ func (syncer *Syncer) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo,
...
 		if err != nil {
 			return true, err
 		}
+		height, err := t.Height()
+		if err != nil {
+			return false, err
+		}
+
+		// update status with latest fetched head and height
+		syncer.reporter.UpdateStatus(fetchHead(t.Key()), fetchHeight(height))
```
frrist: update status with the last head we fetched.

frrist:
@ZenGround0 I have fix the discrepancies you pointed out, let me know if there is anything else that's unclear.

chain/store.go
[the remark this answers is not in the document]
ZenGround0: The blocking stands. I would prefer an undef tipset in heaviest head until there is a sync over adding an exported `Status` method to the store for the reasons stated above.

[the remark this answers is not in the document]
frrist: An earlier implementation of this did just that (set values to undef), reviewers commented (https://github.com/filecoin-project/go-filecoin/pull/3305#discussion_r319777238) that we could get this information from the store. The way to address this was to share status between the syncer and the store. 

Here are the options I think we have to remove this blocking comment:
- accept this exported method as a necessary evil of testing
- create a new test file in the chain package (not in chain_test)
- define an exported method that accepts a chain store and returns its reporter
- instantiate a syncer in each store test and inspect its status return
- not test the store updates its status

I would also like to point out that `Status` has not been added to the store interface definitions, it's a method on the struct, nothing that accepts a store has access to it.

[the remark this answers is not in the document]
ZenGround0: Thanks for this context.

My takeaway from @anorth's comment is "if its easy to set the head at setup, then do it", not "the head must be set"

I do not accept that the existing approach is a necessary evil of testing.  Convenient coupling of the syncer and the store and adding to the store's exports is often attractive in the short term but in my experience it is the wrong thing to do.  

In my view we have two better different paths forward, preferable to this coupling.  
(A) we go back to your earlier implementation and leave the head value unset
(B) we introduce a `Start`/`StartStatus` function into the syncer that simply sets the Status head/head-height to a value determined by making a call to the chain store post load in `node.Start()`

> I would also like to point out that Status has not been added to the store interface definitions, it's a method on the struct, nothing that accepts a store has access to it.

This is true, thanks for flagging my imprecise language.  I am talking about the store's exported function set which I believe should follow guidelines for good API design even though go makes a point of taking selective subsets of it.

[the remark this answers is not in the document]
frrist: > (A) we go back to your earlier implementation and leave the head value unset

We can do that, but it feels like a step backwards.

> (B) we introduce a Start/StartStatus function into the syncer that simply sets the Status head/head-height to a value determined by making a call to the chain store post load in node.Start()

I don't see how adding a new exported method to the syncer in place of one on the store is any simpler.

To clarify, is this blocking comment about the Store having an exported `Status` method or about the Store and Syncer sharing a status reporter?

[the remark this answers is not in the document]
ZenGround0: > I don't see how adding a new exported method to the syncer in place of one on the store is any simpler.

From my perspective it is a great deal simpler.  If we remove the shared dependency we remove all consideration of the syncer's status reporter when trying to understand the chain store.  This keeps the chain store's definition better contained; now we don't need to understand the syncing status code and node building/starting code to understand why the chain store is the way it is.  We also address the original concern of not having more methods that do the same thing exported from the store.

Yes we would be adding a method to the syncer in proposal B, and it is a well-scoped and simple method that does what we want without relying on sharing state with separate parts of the system.  It also does a new unique thing for the syncer that its other methods don't already do in some way.

> To clarify, is this blocking comment about the Store having an exported Status method or about the Store and Syncer sharing a status reporter?

At this point they seem roughly equivalent.  I originally blocked because the `Status` method seems like a poor fit for the store.  It has become clearer to me that you needed to add this method to the store as a result of adding functionality to the store that I believe the syncer/its caller should perform instead.  This strengthens my intuition that the method doesn't belong and suggests that the right fix is to remove the shared state.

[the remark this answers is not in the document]
frrist: I don't like the idea of allowing code external to the syncer to influence it's state, from my understanding option B allows that.

In the interest of getting this merged I will revert the changes that allow the store to modify status and propose that in a follow on.

Issue #3180, opened 2019-08-02
In-process integration test harness & patterns


anorth:
### Description
At present, our tests mostly take one of two forms:
- unit tests, exercising a single or small group of components in isolation
- daemon/FAST tests, which orchestrate multiple nodes in separate processes to exercise long sequences

Multi-process-style integration tests are absolutely necessary for high level functional tests, and FAST is a great framework for allowing us to run such tests in different environments, including devnets. However, orchestrating multiple isolated processes is cumbersome and brings a number of drawbacks, such as requiring non-trivial sealing and block propagation latencies, difficult configuration, poor control over time and event interleaving; such tests are almost necessarily non-deterministic. We are currently experiencing a [bout of flaky tests](https://github.com/filecoin-project/go-filecoin/issues?q=is%3Aissue+is%3Aopen+label%3AC-transient-test-failure), most of which are attempting to exercise a relatively small part of the code, but being forced to bring in much more complexity.

We have a need for something in the middle, which I think should be met by a in-process integration test pattern. An in-process integration test could/would:
- run multiple node logical processes within a single OS process
- manipulate those nodes through direct code APIs (e.g. plumbing/porcelain) rather than the exposed API, thereby bypassing CLI overheads
- inspect node state through similar direct APIs
-  replace some components by fakes and/or mocks manipulable by the test code: network, sealing, proving and verification 
- control work loops (block production, proving) explicitly
- enable precise event ordering through standard Go synchronization primitives
- use an entirely artificial concept of time, allowing tests to run as fast as possible with no delays or sleeps
- when event ordering and time are appropriately controlled, be entirely deterministic

I have mentioned this idea in #2194 and many other places. We actually have a stub to work from here, but at the time it was created it was architecturally impossible to control the nodes to the extent necessary.

I think that more than half of the tests that currently run as daemon tests could be migrated to such a pattern, becoming faster, reliable, more clear and with wider coverage in the process.

### Acceptance criteria
- An in-process integration test pattern is established from which most daemon/FAST tests can be ported
- Some daemon tests are ported as examples

### Risks + pitfalls
- Achieving this may depend on refactors to the node and other components, potentially gnarly ones
- There could be some static/global/singleton state that's difficult to work around

### Where to begin
At present, @anorth intends to experiment and demonstrate to establish some starting points.

PR #3310, opened 2019-08-28
Move the Output structure out of the TestDaemon code for re-use by in-process CLI tests

anorth:
I also defined and cleaned up the semantics of the `CmdOutput` type a bit. The move makes it difficult to see the diff, but probably best to review this as new code now that it's intended for re-use.

This supports re-use of the same output-processing logic for tests run against Filecoin nodes that may be remote, local, or even in-process (I have demonstrated the latter locally). A logical next step might be to factor out the command-issuing parts of TestDaemon and maybe also FAST so that command-execution and output evaluation can all be shared.

This PR also renames testhelpers/commands.go to test_daemon.go after its primary contents (in second commit).

For #3180, based on PR #3309.


testhelpers/output.go
```diff
@@ -0,0 +1,124 @@
...
+
+// ReadStdoutTrimNewlines returns a string representation of stdout,
+// with trailing line breaks removed.
+func (o *CmdOutput) ReadStdoutTrimNewlines() string {
+	// TODO: handle non unix line breaks
+	return strings.Trim(o.ReadStdout(), "\n")
+}
+
+// AssertSuccess asserts that the output represents a successful execution.
+func (o *CmdOutput) AssertSuccess() *CmdOutput {
```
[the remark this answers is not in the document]
anorth: I changed the calls inside to be `assert` instead of `require`.

testhelpers/output.go
```diff
@@ -0,0 +1,125 @@
...
+	tb testing.TB
+}
+
+// ReadOutput reads the `stdout` and `stderr` streams completely and returns a new Output object.
+func ReadOutput(tb testing.TB, args []string, stdout io.ReadCloser, stderr io.ReadCloser) *CmdOutput {
+	stdoutBytes, err := ioutil.ReadAll(stdout)
+	require.NoError(tb, err)
+
+	stderrBytes, err := ioutil.ReadAll(stderr)
+	require.NoError(tb, err)
```
travisperson: I don't think we will run into this issue, but depending on the size of the buffers used we may deadlock around a write to stderr (logs) as we are producing / reading from stdout.

I don't think we need to fix this right now but wanted to point it out so it's in the back of a few peoples heads just in case we encounter this.

anorth: Thanks. That seems like a difficult-to-diagnose issue if it were to happen, so I've addressed it now by reading the streams in parallel (potentially consuming lots of memory).

PR #3317, opened 2019-08-29
Implement test.NodeBuilder and test.NodeAPI for in-process integration tests

anorth:
This PR introduces and demonstrates a couple of things.

`test.NodeBuilder` creates nodes configured for in-memory use. There is existing code in `node/testing.go` to create nodes, primarily `GenNode` and other methods that call into it. My goal is to replace all uses of that with this builder and extensions to it.

`test.NodeAPI` runs a real command API around an in-process node, that listens on a TCP socket. It also provides an in-process client that sends commands to that socket over the cmdkit RPC. That is, it provides both the daemon and client processes, in-process. If using the TestDaemon (deprecated) or FAST, both of those would involve forking processes, and the daemon would use an on-disk repo. The resulting tests run in ⅓ or less of the time and provide direct access to the node under test for configuration and inspection (things with otherwise have to be plumbed through the CLI).

I've ported two very simple tests to demonstrate this. The protocol daemon test is a direct replacement, exercising the cmdkit commands, and is testing the complete code path but much faster. The ping daemon test I've added an alternative version that demonstrates testing directly against the porcelain API, skipping over the cmdkit API entirely. This shows a path to testing things without forcing us to write CLI commands for them (though I'm not saying we won't want to expose this plumbing/porcelain over an RPC one day anyway). This one is waaay faster, ~500ms down from 2.2s (and probably be faster still with libp2p mocknet, which is now possible to use).

On request, I could change this ping test to also exercise the cmdkit I/O, but I am inclined not to. I intend to delete the daemon test version before merging.

For #3180

### DIFF
--- a/commands/ping_daemon_test.go
+++ b/commands/ping_daemon_test.go
@@ -1,34 +1,40 @@
 package commands_test
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
-	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	"github.com/filecoin-project/go-filecoin/node"
+	"github.com/filecoin-project/go-filecoin/node/test"
 	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 )
 
 func TestPing2Nodes(t *testing.T) {
 	tf.IntegrationTest(t)
 
-	d1 := th.NewDaemon(t).Start()
-	defer d1.ShutdownSuccess()
-	d2 := th.NewDaemon(t).Start()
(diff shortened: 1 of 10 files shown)


commands/ping_daemon_test.go
```diff
@@ -32,3 +35,31 @@ func TestPing2Nodes(t *testing.T) {
 	t.Log("[success] 2 -> 1")
 	assert.Contains(t, ping2.ReadStdout(), "Pong received")
 }
+
+func TestPing2NodesPorcelain(t *testing.T) {
```
anorth: The old version of this test is still present above, but I plan to delete it.

PR #3321, opened 2019-08-29
update megascript to use "show messages" command to look for commitSector and submitPoSt

laser:
## Why does this PR exist?

After messages were removed from the `chain ls` output, the megascript broke.

## What's in this PR?

This changeset modifies the megascript to instead use `show messages` to look for `commitSector` and `submitPoSt`.

### DIFF
--- a/functional-tests/lib/helpers.bash
+++ b/functional-tests/lib/helpers.bash
@@ -156,7 +156,9 @@ function wait_for_message_in_chain_by_method_and_sender {
         exit 1
     fi
 
-    __hodl=$(echo "$(chain_ls "$3")" | jq ".[] | select(.messages != null) | .messages[].meteredMessage.message | select(.method == \"$1\").from | select(. == \"$2\")" 2>/dev/null | head -n 1 || true)
+    __hodl=$(echo "$(chain_ls $3)" \
+        | jq -r '.[].messages["/"]' | while read -r cid; do ./go-filecoin show messages $cid --enc=json --repodir=$3; done | jq -s 'add' \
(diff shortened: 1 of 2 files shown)


functional-tests/lib/helpers.bash
```diff
@@ -156,7 +156,9 @@ function wait_for_message_in_chain_by_method_and_sender {
         exit 1
     fi
 
-    __hodl=$(echo "$(chain_ls "$3")" | jq ".[] | select(.messages != null) | .messages[].meteredMessage.message | select(.method == \"$1\").from | select(. == \"$2\")" 2>/dev/null | head -n 1 || true)
+    __hodl=$(echo "$(chain_ls $3)" \
+        | jq -r '.[].messages["/"]' | while read -r cid; do ./go-filecoin show messages $cid --enc=json --repodir=$3; done | jq -s 'add' \
+        | jq ".[] | select(.meteredMessage != null) | .meteredMessage.message | select(.method == \"$1\").from | select(. == \"$2\")" 2>/dev/null | head -n 1 || true)
```
ingar: I'm sure you have a reason, but why sometimes double-quoting (and then `\"`) the jq expression and sometimes single-quoting?

laser: > I'm sure you have a reason

Good question. [Quoth the Bash Reference](https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html):

> Enclosing characters in single quotes (‘'’) preserves the literal value of each character within the quotes.

I use single quotes when I don't need to expand variables, and double quotes when I do.

Issue #3319, opened 2019-08-29
PeerTracker uses hello protocol as heartbeat-like service


frrist:
### Description
This issue proposes a partial solution to #2674 until a full design is agreed upon.
___

We have opted to disable the block pubsub topic until a node has synced its chain to a fully caught-up state. Due to this decision we must rely on the messages we receive from the hello protocol to know the current state of the network. Since there is no guarantee that these messages will continue to arrive as the network grows we need a way to trigger nodes to send a hello message upon our request. 

Ways to accomplish this without changing the bits of the protocol message:
1. Bounce our libp2p connection with a peer (i.e. disconnect & connect), this works because of the [notify interface](https://github.com/filecoin-project/go-filecoin/blob/master/protocol/hello/hello.go#L174).
2. Ensure we have sent the same number of hello messages to a peer as we have received from them.

Both of these methods are an abuse of the hello protocol. I would argue that 2. is the lesser of the two and the one we should implement. 

The PeerTracker tracker currently tracks peers a node has received a hello message from via a hello protocol callback. The tracker should be extended to periodically issue hello messages to its peers to keep its state up to date. We can then compare the best head the tracker has with our own to determine if we are in a caught up state.

### Acceptance criteria
- The PeerTracker is capable of sending a hello message to some peer.
- The PeerTracker ensures it has sent the same number of hello messages it has received from each peer.
- The PeerTracker has a method that returns the highest head it knows about.

### Risks + pitfalls
- Avoid entering a hello loop

anorth:
This sounds good. Optionally, the set of peers that a repeated hello message is sent to could be limited to a configured trusted set (i.e. those nodes already specified in the config file).

anorth:
On more reflection (and current devnet woes) I think the re-hello *should* be limited to trusted peer set for now. We should set the trust flag directly from membership in that set.

frrist:
addressed via #3327

Issue #3268, opened 2019-08-21
Enable block pubsub topic once caught-up to "trusted" peers and distrust blocks received.


frrist:
### Description
Currently we consider the initial bootstrapping of a chain to be done after the syncer has synchronously processed the chain up to the head reported by the first peer to respond to a hello message. Right now this peer is one of the bootstrap nodes held in the nodes config. For now we may consider heads published by these nodes to be "trusted". (using "" since this issue will be a precursor to #2674 and we are moving away from the idea of "trusted peer").

The syncers `HandleNewTipSet` method has a boolean parameter `trusted` that when set to false, will cause the syncer to reject tipsets exceeding the finality limit:
https://github.com/filecoin-project/go-filecoin/blob/e71515a4ca441d70bb78664bf3dc74b14872297a/chain/syncer.go#L341-L344 
Currently `trusted` is set to true everywhere, this issue aims to lay to foundation for allowing it to be set to false.

When a filecoin node joins the network after the chain has become very long (relative to its current head) it will take a long time for the node to process the chain up to the head it receives from a "trusted node". Because of this we will need to sync with a trusted node until our height is within ~finality of theirs. After this we should then enable the block pubsub topic and treat all blocks we receive on it as untrusted. 

### Acceptance criteria
- Define a means for requesting the latest head from a "trusted" peer.
  - Currently it is unclear if the hello protocol should be modified to allow this or if something new is required.
- Modify the node Start procedure to sync with trusted peers until it is "sufficiently caught up" to their heads.
- Enable the block pubsub topic once caught up.

ZenGround0:
Some opinions on a quick path forward.  I'm still new to thinking about secure bootstrapping so take with a grain of salt.  This is just one route, it might contradict ones you already have in mind which might turn out better.

The key and kind of subtle point I want to make is that syncing is inherently content based, not peer based (potentially many peers are talked to but only one chain is synced per invocation), so we shouldn't be altering the behavior of syncing invocations based on how much we trust the peer that told us about a head.  Instead we should be altering the the behavior based on whether the chain is too long to rely on untrusted nodes.  If none of our trusted bootstrappers have the risky chain then we don't fetch it.

I think the `trusted` boolean passed to `HandleNewTipSet` should be called `requiresTrust` and set to `ci.Height - headHeight > UntrustedChainLimit`.  Alternatively `noTrust` and `ci.Height - headHeight <= UntrustedChainLimit`.

[...]

PR #3327, opened 2019-09-01
SayHello returns hello message from peer

frrist:
### What
This PR lays the foundation for completing #3319 and later #3268 by modifying the `SayHello` method to receive a response. The method has also been exported so that it may be called externally to receive an updated chain head from a peer.

This is not a permanent change to the hello protocol and will likely be reverted once https://github.com/filecoin-project/go-filecoin/issues/2674 is implemented.

### DIFF
--- a/protocol/hello/hello.go
+++ b/protocol/hello/hello.go
@@ -84,17 +84,10 @@ func New(h host.Host, gen cid.Cid, helloCallback helloCallback, getHeaviestTipSe
 
 func (h *Handler) handleNewStream(s net.Stream) {
 	defer s.Close() // nolint: errcheck
-
-	msg, err := h.getOurHelloMessage()
-	if err != nil {
-		log.Error(err)
-		return
-	}
-	if err := cbu.NewMsgWriter(s).WriteMsg(&msg); err != nil {
-		log.Error(err)
+	if err := h.sendHello(s); err != nil {
+		log.Errorf("failed to send hello message:%s", err)
 	}
 	return
-
 }
 
 // ErrBadGenesis is the error returned when a mismatch in genesis blocks happens.
@@ -132,8 +125,8 @@ func (h *Handler) getOurHelloMessage() (*Message, error) {
 	}, nil
 }
 
-// SayHello receives a hello message from peer `p` and returns it.
-func (h *Handler) SayHello(ctx context.Context, p peer.ID) (*Message, error) {
+// ReceiveHello receives a hello message from peer `p` and returns it.
+func (h *Handler) ReceiveHello(ctx context.Context, p peer.ID) (*Message, error) {
 	s, err := h.host.NewStream(ctx, p, protocol)
 	if err != nil {
 		return nil, err
@@ -146,6 +139,15 @@ func (h *Handler) SayHello(ctx context.Context, p peer.ID) (*Message, error) {
 		return nil, err
 	}
 	return &hello, nil
+}
+
+// sendHello send a hello message on stream `s`.
+func (h *Handler) sendHello(s net.Stream) error {
+	msg, err := h.getOurHelloMessage()
+	if err != nil {
+		return err
+	}
+	return cbu.NewMsgWriter(s).WriteMsg(&msg)
 }
 
 // New peer connection notifications
@@ -168,10 +170,10 @@ func (hn *helloNotify) Connected(n net.Network, c net.Conn) {
 
 		// receive the hello message
 		from := c.RemotePeer()
-		hello, err := hn.hello().SayHello(ctx, from)
+		hello, err := hn.hello().ReceiveHello(ctx, from)
 		if err != nil {
 			log.Warningf("failed to receive hello handshake from peer %s: %s", from, err)
-			c.Close() //nolint: errcheck
+			_ = c.Close()
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


protocol/hello/hello.go
```diff
@@ -176,7 +198,7 @@ func (hn *helloNotify) Connected(n net.Network, c net.Conn) {
 		ctx, cancel := context.WithTimeout(context.Background(), helloTimeout)
 		defer cancel()
 		p := c.RemotePeer()
-		if err := hn.hello().sayHello(ctx, p); err != nil {
+		if _, err := hn.hello().SayHello(ctx, p); err != nil {
```
frrist: Sending a hello message on this call and then again in the handler is not optimal, but it's easy, doesn't require changing the protocol bits, and doesn't rewrite much of the hello code. If this extra sending is not acceptable a proposed alternative could be to:
- modify the method `handleNewStream` such that it only sends a hello message.
- modify `SayHello` to read a message from a steam
- modify `Connect` to send a hello message, receive a response, process it and trigger the callback.

protocol/hello/hello.go
```diff
@@ -131,34 +141,46 @@ func (h *Handler) processHelloMessage(from peer.ID, msg *Message) error {
 	return nil
 }
 
-func (h *Handler) getOurHelloMessage() *Message {
+func (h *Handler) getOurHelloMessage() (*Message, error) {
 	heaviest, err := h.getHeaviestTipSet()
 	if err != nil {
-		panic("cannot fetch chain head")
+		return nil, err
```
ZenGround0: nice

protocol/hello/hello.go
ZenGround0: I'm fine with this for now.  Please put up a TODO with a follow on issue for investigating making this simpler.

frrist: implemented here: https://github.com/filecoin-project/go-filecoin/pull/3327/commits/06dbf50930399ea4c028944adf47e9c5fe8a1776, turned out to be cleaner than anticipated.

protocol/hello/hello.go
```diff
@@ -85,31 +85,16 @@ func New(h host.Host, gen cid.Cid, helloCallback helloCallback, getHeaviestTipSe
...
-		s.Conn().Close() // nolint: errcheck
-		return
-	case ErrWrongVersion:
-		log.Debugf("code not at same version: peer has version %s, daemon has version %s, disconnecting from peer: %s", hello.CommitSha, h.commitSha, from)
-		versionErrCt.Inc(context.TODO(), 1)
-		s.Conn().Close() // nolint: errcheck
-		return
-	case nil: // ok, noop
-	default:
+	if err := cbu.NewMsgWriter(s).WriteMsg(&msg); err != nil {
```
anorth: Factor the above found lines out into a method `SendHello`.

protocol/hello/hello.go
```diff
@@ -118,47 +103,49 @@ var ErrBadGenesis = fmt.Errorf("bad genesis block")
...
 		HeaviestTipSetCids:   heaviest.Key(),
 		HeaviestTipSetHeight: height,
 		CommitSha:            h.commitSha,
-	}
+	}, nil
 }
 
-func (h *Handler) sayHello(ctx context.Context, p peer.ID) error {
+// SayHello receives a hello message from peer `p` and returns it.
+func (h *Handler) SayHello(ctx context.Context, p peer.ID) (*Message, error) {
```
anorth: Blocking: this name is totally wrong now. Rename to `ReceiveHello`.

protocol/hello/hello.go
```diff
@@ -171,13 +158,38 @@ func (hn *helloNotify) hello() *Handler {
...
-		p := c.RemotePeer()
-		if err := hn.hello().sayHello(ctx, p); err != nil {
-			log.Warningf("failed to send hello handshake to peer %s: %s", p, err)
+
+		// receive the hello message
+		from := c.RemotePeer()
+		hello, err := hn.hello().SayHello(ctx, from)
+		if err != nil {
+			log.Warningf("failed to receive hello handshake from peer %s: %s", from, err)
+			c.Close() //nolint: errcheck
```
anorth: Replace this idiom with `_ := c.Close()`

protocol/hello/hello_test.go
```diff
@@ -229,3 +229,47 @@ func TestHelloMultiBlock(t *testing.T) {
...
+	mn, err := mocknet.FullMeshConnected(ctx, 2)
+	assert.NoError(t, err)
+
+	a := mn.Hosts()[0]
+	b := mn.Hosts()[1]
+
+	builder := chain.NewBuilder(t, address.Undef)
+
+	genesisTipset := builder.NewGenesis()
+	assert.Equal(t, 1, genesisTipset.Len())
```
anorth: You don't need to check this

frrist: :man_facepalming:

whyrusleeping:
heads up: this makes go-filecoin no longer in compliance with the spec for the hello protocol.

anorth:
@whyrusleeping I was going to write something about this in https://github.com/filecoin-project/specs/issues/478 (we're trying to work around the limitations of the defined protocol here) but upon closer inspection I decided that this was in fact in compliance. Certainly backwards compatible, but likely to be fully so.

The spec says:
> Whenever a node gets a new connection, it opens a new stream on that connection and "says hello" ... finally, closing the stream.
> ...
> Upon receiving a "hello" stream from another node, you should [read and verify]

This change just opens the hello stream more than once. A compliant node should respond every time. The spec would have to have extra words saying "and if another hello stream is opened you should reject it" in order to make this change incompatible. The spec seems to be leaving just enough room for implementers to decide when and how often to query peers.

We chose this in favour of fully disconnecting the libp2p connection with the peer, which would be more disruptive, but compatible even with the additional only-once restriction.

PR #3330, opened 2019-09-02
PeerTracker has trusted set and SelectHead logic

frrist:
### What
This PR extends the PeerTracker to have:
- A set of trusted peers that it my be initialized with
- An update function that may be applied to tracked peers, currently is is only exposed via `UpdateTrusted`.
- A method for setting an update function.
- A method for selecting the head with the greatest height from all tracked peers or just the trusted peers.
- Testing for the above
-  An initial set of trusted node which for now are the bootstrap nodes.

### Goal
This PR makes progress towards implementing chain catch-up logic. In a follow on the trusted peers and select head logic will be used to fully sync a chain before proceeding with caught up chain sync. 

### Status
This PR is a follow on to #3327

### DIFF
--- a/net/peer_tracker.go
+++ b/net/peer_tracker.go
@@ -1,11 +1,15 @@
 package net
 
 import (
+	"context"
+	"sort"
 	"sync"
 
 	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/filecoin-project/go-filecoin/types"
 )
@@ -18,19 +22,51 @@ var logPeerTracker = logging.Logger("peer-tracker")
 type PeerTracker struct {
 	// mu protects peers
 	mu sync.RWMutex
-	// peers maps stringified peer.IDs to info about their chains
-	peers map[string]*types.ChainInfo
 
 	// self tracks the ID of the peer tracker's owner
 	self peer.ID
+
+	// peers maps peer.IDs to info about their chains
+	peers    map[peer.ID]*types.ChainInfo
+	trusted  map[peer.ID]struct{}
+	updateFn updatePeerFn
 }
 
+type updatePeerFn func(ctx context.Context, p peer.ID) (*types.ChainInfo, error)
+
 // NewPeerTracker creates a peer tracker.
-func NewPeerTracker(self peer.ID) *PeerTracker {
-	return &PeerTracker{
-		peers: make(map[string]*types.ChainInfo),
-		self:  self,
+func NewPeerTracker(self peer.ID, trust ...peer.ID) *PeerTracker {
+	trustedSet := make(map[peer.ID]struct{}, len(trust))
+	for _, t := range trust {
+		trustedSet[t] = struct{}{}
 	}
+	return &PeerTracker{
+		peers:   make(map[peer.ID]*types.ChainInfo),
+		trusted: trustedSet,
+		self:    self,
+	}
+}
+
+// SetUpdateFn sets the update function `f` on the peer tracker. This function is a prerequisite
+// to the UpdateTrusted logic.
+func (tracker *PeerTracker) SetUpdateFn(f updatePeerFn) {
+	tracker.updateFn = f
+}
+
+// SelectHead returns the chain info from trusted peers with the greatest height.
+// An error is returned if no peers are in the tracker.
+func (tracker *PeerTracker) SelectHead() (*types.ChainInfo, error) {
+	heads := tracker.listTrusted()
+	if len(heads) == 0 {
+		return nil, errors.New("no peers tracked")
+	}
+	sort.Slice(heads, func(i, j int) bool { return heads[i].Height > heads[j].Height })
+	return heads[0], nil
+}
+
+// UpdateTrusted updates ChainInfo for all trusted peers.
+func (tracker *PeerTracker) UpdateTrusted(ctx context.Context) error {
+	return tracker.updatePeers(ctx, tracker.trustedPeers()...)
 }
 
 // Track adds information about a given peer.ID
@@ -38,10 +74,10 @@ func (tracker *PeerTracker) Track(ci *types.ChainInfo) {
 	tracker.mu.Lock()
 	defer tracker.mu.Unlock()
 
-	pidKey := ci.Peer.Pretty()
-	_, tracking := tracker.peers[pidKey]
-	tracker.peers[pidKey] = ci
-	logPeerTracker.Infof("Tracking %s, new=%t, count=%d", ci, !tracking, len(tracker.peers))
+	_, tracking := tracker.peers[ci.Peer]
+	_, trusted := tracker.trusted[ci.Peer]
+	tracker.peers[ci.Peer] = ci
+	logPeerTracker.Infof("Tracking %s, new=%t, count=%d trusted=%t", ci, !tracking, len(tracker.peers), trusted)
 }
 
 // Self returns the peer tracker's owner ID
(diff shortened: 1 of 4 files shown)


net/peer_tracker.go
```diff
@@ -57,15 +127,55 @@ func (tracker *PeerTracker) List() []*types.ChainInfo {
 	return out
 }
 
+// ListTrusted returns the chain info of the trusted tracked peers. The info tracked by the tracker can
```
frrist: Unsure if the below methods `ListTrusted` and `TrustedPeers` should exists or if `Peers` and `List` should accept parameters, thoughts?

net/peer_tracker.go
```diff
@@ -18,32 +22,98 @@ var logPeerTracker = logging.Logger("peer-tracker")
...
 	return &PeerTracker{
-		peers: make(map[string]*types.ChainInfo),
+		peers:   make(map[peer.ID]*types.ChainInfo),
+		trusted: trustedSet,
+	}
+}
+
+// SetUpdateFn sets the update function `f` on the peer tracker. This function is a prerequisite
+// to the UpdateTrusted logic.
+func (tracker *PeerTracker) SetUpdateFn(f updatePeerFn) {
```
frrist: This is pretty clunky. It exists because the `PeerTracker` is instantiated when the node is built, the hello protocol in instantiated when the node is started, and the `SayHello` method is used as the callback.

net/peer_tracker.go
```diff
@@ -18,32 +22,98 @@ var logPeerTracker = logging.Logger("peer-tracker")
...
+	tracker.updateFn = f
+}
+
+// SelectHead returns the chain info with the greatest height being tracked. If `trust` is true then
+// only the trusted tracked nodes will be considered. An error is returned if no peers are in the tracker.
+func (tracker *PeerTracker) SelectHead(trust bool) (types.ChainInfo, error) {
```
frrist: Moving forward I see this method feeding the syncer during the initial sync, usage would be something like:
```go
for NotCaughtUp() {
  peertracerk.UpdatePeers()
  head := peertracker.SelectHead()
  syncer.HandleNewTipSet(head)
}
```

net/peer_tracker.go
```diff
@@ -18,32 +22,98 @@ var logPeerTracker = logging.Logger("peer-tracker")
...
+	}
+	if len(heads) == 0 {
+		return types.ChainInfo{}, errors.New("no peers tracked")
 	}
+	sort.Slice(heads, func(i, j int) bool { return heads[i].Height > heads[j].Height })
+	return *heads[0], nil
+}
+
+// UpdateTrusted updates ChainInfo for all trusted peers.
+func (tracker *PeerTracker) UpdateTrusted(ctx context.Context) error {
```
frrist: Similar vein to my comments on `ListTrusted` and `TrustedPeers`, unsure if this should be an `Update` method that accepts parameters or explicitly named like this.

net/peer_tracker.go
```diff
@@ -18,32 +22,99 @@ var logPeerTracker = logging.Logger("peer-tracker")
...
+// the peers in `ps` fails then the first non-nil error is returned.
+func (tracker *PeerTracker) updatePeers(ctx context.Context, ps ...peer.ID) error {
+	if tracker.updateFn == nil {
+		// programmer error
+		panic("canot call PeerTracker peer update logic without setting an update function")
+	}
+	grp, ctx := errgroup.WithContext(ctx)
+	for _, p := range ps {
+		peer := p
+		grp.Go(func() error {
```
frrist: I expect updating the PeerTrackers to require going to the network, because of this I decided to perform this operation concurrently.

net/peer_tracker.go
anorth: I don't see any reason for those methods to exist, except as internal helpers, with or without a parameter. They are not "interface" methods. 

Don't test them directly, just test the behaviour we really care about - that the non-trusted peers are not included in the trusted head calculation.

net/peer_tracker.go
frrist: Sounds good to me, currently `SelectHead` will return an error if at least one update fails, I will modify s.t. it only errors if all fail.

net/peer_tracker.go
```diff
@@ -18,32 +22,92 @@ var logPeerTracker = logging.Logger("peer-tracker")
...
+func (tracker *PeerTracker) UpdateTrusted(ctx context.Context) error {
+	return tracker.updatePeers(ctx, tracker.TrustedPeers()...)
+}
+
+// updatePeers will run the trackers updateFn on each peer in `ps` in parallel, if updating any of
+// the peers in `ps` fails then the first non-nil error is returned.
+func (tracker *PeerTracker) updatePeers(ctx context.Context, ps ...peer.ID) error {
+	if tracker.updateFn == nil {
+		// programmer error
+		panic("canot call PeerTracker peer update logic without setting an update function")
```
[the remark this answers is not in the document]
frrist: The case I had in mind for this: a programmer wrote code to instantiate a peertracker, didn't set an update function, and called update. 
Until we can instantiate the peer tracker with an update method I would prefer to panic.

net/peer_tracker.go
```diff
@@ -89,3 +141,66 @@ func TrackerRegisterDisconnect(ntwk network.Network, tracker *PeerTracker) {
...
+	}
+	// check if anyone failed to update
+	if err := grp.Wait(); err != nil {
+		// full failure return an error
+		if len(updateErr) == len(ps) {
+			logPeerTracker.Errorf("failed to update all %d peers:%v", len(ps), updateErr)
+			return errors.New("all peers failed to update")
+		}
+		// partial failure
+		logPeerTracker.Warningf("failed to update %d of %d peers:%v", len(updateErr), len(ps), updateErr)
```
anorth: INFO is sufficient

net/peer_tracker.go
```diff
@@ -89,3 +141,66 @@ func TrackerRegisterDisconnect(ntwk network.Network, tracker *PeerTracker) {
...
+		}
+	}
+	out := make([]*types.ChainInfo, len(tracked))
+	copy(out, tracked)
+	return out
+}
+
+// updatePeers will run the trackers updateFn on each peer in `ps` in parallel, iff all updates fail
+// is an error is returned, a partial update is considered successful.
+func (tracker *PeerTracker) updatePeers(ctx context.Context, ps ...peer.ID) error {
```
anorth: It's probably worth an INFO log if there are no trusted peers.

net/peer_tracker.go
```diff
@@ -57,15 +121,55 @@ func (tracker *PeerTracker) List() []*types.ChainInfo {
...
 	tracker.mu.Lock()
 	defer tracker.mu.Unlock()
 
-	pidKey := pid.Pretty()
-	if _, tracking := tracker.peers[pidKey]; tracking {
-		logPeerTracker.Infof("Dropping peer %s", pidKey)
-		delete(tracker.peers, pidKey)
+	if _, tracking := tracker.peers[pid]; tracking {
+		logPeerTracker.Infof("Dropping peer %s", pid.Pretty())
+		delete(tracker.peers, pid)
```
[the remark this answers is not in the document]
anorth: Sorry, I take it back. Leave the `trusted` value there in case the peer is added back again later.

net/peer_tracker.go
```diff
@@ -89,3 +140,70 @@ func TrackerRegisterDisconnect(ntwk network.Network, tracker *PeerTracker) {
...
+	if tracker.updateFn == nil {
+		// programmer error
+		panic("canot call PeerTracker peer update logic without setting an update function")
+	}
+	if len(ps) == 0 {
+		logPeerTracker.Info("update peers aborting: no peers to update")
+		return nil
+	}
+	var updateErr []error
+	grp, ctx := errgroup.WithContext(ctx)
```
ZenGround0: woahdude this thing is great!

net/peer_tracker.go
```diff
@@ -18,42 +22,72 @@ var logPeerTracker = logging.Logger("peer-tracker")
 type PeerTracker struct {
 	// mu protects peers
 	mu sync.RWMutex
-	// peers maps stringified peer.IDs to info about their chains
-	peers map[string]*types.ChainInfo
+	// peers maps peer.IDs to info about their chains
+	peers    map[peer.ID]*types.ChainInfo
+	trusted  map[peer.ID]bool
```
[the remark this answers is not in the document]
frrist: Good idea, although but it does make conditionals _slightly_ more verbose- will change.

net/peer_tracker.go
```diff
@@ -89,3 +140,70 @@ func TrackerRegisterDisconnect(ntwk network.Network, tracker *PeerTracker) {
...
+	copy(out, tracked)
+	return out
+}
+
+// updatePeers will run the trackers updateFn on each peer in `ps` in parallel, iff all updates fail
+// is an error is returned, a partial update is considered successful.
+func (tracker *PeerTracker) updatePeers(ctx context.Context, ps ...peer.ID) error {
+	if tracker.updateFn == nil {
+		// programmer error
+		panic("canot call PeerTracker peer update logic without setting an update function")
```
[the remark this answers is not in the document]
frrist: @anorth has already made this point and I pushed back, given that I am now a minority on this view I will return an error here.

PR #3333, opened 2019-09-02
Prevent storage fault slasher from slashing a late miner repeatedly.

### LINKED ISSUE #3332: Storage fault slasher slashes tardy miner repeatedly
**Describe the bug**:
The storage fault slasher will create a SlashStorageFault message for a tardy miner *on every tipset validated*. This floods the node's message pool, and could lead to other messages being rejected. If the node wins a block, it will waste a bunch of gas on all messages except the first.

Stats from our bootstrap nodes confirm that message pools are much more full in the 0.4 release than previously (though not approaching capacity).

The error logs are also swamped with `error in handler github.com/filecoin-project/go-filecoin/node.(*Node).processMessage-fm for topic /fil/msgs/devnet-4: validation error adding message to pool: message pool contains message with same actor and nonce but different cid`

Also some `slashStorageFault message failed: failed to add message to message pool: validation error adding message to pool: message nonce (130) is too much greater than actor nonce (27)`

Both of these are explainable by the node putting multiple slashing messages for the same actor in its pool, in response to receiving a new tipset over pubsub. Since those messages do *not* go in the outbound queue, they will (incorrectly) share the same nonce (see #3052). And after too many are created, that nonce will rise above the messages pool's sanity check limit.

Note that these slash messages are never broadcast, so we can only blame a node's own operation for these.

**Expected behavior**:
A node constructs only a single slashing message per miner and proving period missed.

anorth:
Fixes #3332.

FYI @travisperson this may be worth a release. I'm not sure if the problem is the direct/sole cause of the devnet's current problems, but it's probably one.


protocol/storage/fault_slasher.go
```diff
@@ -92,6 +97,18 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
...
 	for minerStr, state := range *lms {
+		if _, ok := sfm.slashed[minerStr]; ok {
+			// Skip slashed miner.
+			// This logic is not perfect. A miner that appears to be late could redeem itself
+			// (e.g. due to a chain re-org) such that a submitted slashing message fails, but
+			// this slasher will remember that it has attempted to slash that miner, and won't
+			// re-attempt to do so in the future.
+			// The `slashed` cache key should include the proving period for which the attempt was
+			// made so that if a miner does enter another period, it becomes slashable again.
+			// Getting that information requires painful plumbing of data through actor methods
```
acruikshank: This actually only requires including MessageQuery, calling the `getProvingPeriod` method on the miner actor, and parsing the result. Still possibly overkill given the likelihood of a problem.

anorth:
FYI @ZenGround0 for potential cherry-pick

PR #3339, opened 2019-09-03
add clock interface to node

### LINKED ISSUE #3337: The node structure should contain the clock interface
### Description
The node structure should be modified to contain the [clock interface](https://github.com/filecoin-project/go-filecoin/blob/989cb180995fe637d31bef6100c4d33be5a90877/clock/clock.go#L9). When starting a real filecoin node the value may contain the [`SystemClock`](https://github.com/filecoin-project/go-filecoin/blob/989cb180995fe637d31bef6100c4d33be5a90877/clock/clock.go#L17), when running a node in process the value may contain the [`FakeClock`](https://github.com/filecoin-project/go-filecoin/blob/master/testhelpers/clock.go#L15), or any other yet-to-come interface implementation. This will provide more control over how the node behaves wrt time during in process testing.

### Acceptance criteria
- Node structure contains a clock interface.
- Filecoin nodes started normally us the SystemClock.

### Related
This does not invalidate https://github.com/filecoin-project/go-filecoin/issues/2961 since some things are on both the node and porcelain (e.g. wallet).

The need for this surfaced during work for #3330 which will need to perform some [polling on the peer trackers](https://github.com/filecoin-project/go-filecoin/pull/3330/files#diff-380e79adccbaf1276bf6c91cc551cd7fR614) head in order to determine chain syncing state.

frrist:
### What
This PR modifies the node structure to contain the clock interface and closes #3337

### DIFF
--- a/testhelpers/clock.go
+++ b/testhelpers/clock.go
@@ -19,11 +19,16 @@ func NewFakeSystemClock(n time.Time) *FakeSystemClock {
 }
 
 // Now returns the current value of the FakeSystemClock.
-func (mc *FakeSystemClock) Now() time.Time {
-	return mc.now
+func (fc *FakeSystemClock) Now() time.Time {
+	return fc.now
+}
+
+// Sleep returns immediately.
+func (fc *FakeSystemClock) Sleep(d time.Duration) {
+	return
 }
 
 // Set sets the current time value of the FakeSystemClock.
-func (mc *FakeSystemClock) Set(t time.Time) {
-	mc.now = t
+func (fc *FakeSystemClock) Set(t time.Time) {
+	fc.now = t
 }
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -8,6 +8,7 @@ import (
 // time module.
 type Clock interface {
 	Now() time.Time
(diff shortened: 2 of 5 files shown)


testhelpers/clock.go
```diff
@@ -19,11 +19,16 @@ func NewFakeSystemClock(n time.Time) *FakeSystemClock {
...
 // Now returns the current value of the FakeSystemClock.
-func (mc *FakeSystemClock) Now() time.Time {
-	return mc.now
+func (fc *FakeSystemClock) Now() time.Time {
+	return fc.now
+}
+
+// Sleep returns immediately.
+func (fc *FakeSystemClock) Sleep(d time.Duration) {
+	return
```
anorth: This is not a useful implementation. Please follow-up to implement this so that it behaves like a real clock: a call to `Sleep` does not return until the time reported by the clock is `d` greater than the time at which `Sleep` is invoked. This will mean accumulating a list of waiting calls (with a channel or Latch) and checking them all in `Set`.

frrist: :+1: will merge this with #3355 as a follow on to address this.

PR #3340, opened 2019-09-03
move ChainSynced Latch to node as exported type

### LINKED ISSUE #3338: The ChainSynced latch should be accessible externally from tests.
### Description
During node startup we use a [`Latch`](https://github.com/filecoin-project/go-filecoin/blob/master/node/node.go#L604) to avoid performing some operations until the nodes chain has reach a caught-up state with its peers. Some tests, such as [`TestBlockPropsManyNodes`](https://github.com/filecoin-project/go-filecoin/blob/6aa421d0bda4f65fe57ecbc83c9c62d75c71f21f/node/block_propagate_test.go#L33), do not require a node to be in a caught-up state and will fail if the node remains out of that state during testing.

Tests should instead be able to:
- Wait on the node to reach a caught-up state.
- Change the state of the node such that the actions waiting on the latch may run.

### Acceptance criteria
- Expose the ChainSynced latch in tests.

frrist:
- closes #3338


anorth:
The PR description doesn't motivate this change. It's fairly innocuous so I'll approve anyway, but suspect that whatever this is ultimately enabling could be achieved another way too.

anorth:
Sorry, I missed the reference to #3388

PR #3343, opened 2019-09-03
refactor: extract NodeBuilder out of node.go

icorderi:
This PR moves the `Config` struct and all the `ConfigOpt` builder methods along with the node building methods (`New()` and `Build()`) out of the `node.go` file into their own `./node/builder.go` file.


icorderi:
@anorth @acruikshank I left some notes to reviewers flagged as `Review` and a couple `TODO`s I would like to address on a followup PR if we think it's the right path. 

I will cleanup and adjust the TODO's based on your comments before merging this in.

node/node.go
```diff
@@ -291,237 +186,6 @@ func readGenesisCid(ds datastore.Datastore) (cid.Cid, error) {
 	return c, nil
 }
 
-// buildHost determines if we are publically dialable.  If so use public
-// Address, if not configure node to announce relay address.
-func (nc *Config) buildHost(ctx context.Context, makeDHT func(host host.Host) (routing.Routing, error)) (host.Host, error) {
```
acruikshank: This is too much networking logic for a builder. It should be a free function in a different file that takes `isRelay`, `publicRelayAddress`, and `libP2POptions` as parameters.

This could wait for later though.

node/node.go
```diff
@@ -291,237 +186,6 @@ func readGenesisCid(ds datastore.Datastore) (cid.Cid, error) {
...
-	validator := blankValidator{}
-
-	var peerHost host.Host
-	var router routing.Routing
-
-	bandwidthTracker := p2pmetrics.NewBandwidthCounter()
-	nc.Libp2pOpts = append(nc.Libp2pOpts, libp2p.BandwidthReporter(bandwidthTracker))
-
-	if !nc.OfflineMode {
-		makeDHT := func(h host.Host) (routing.Routing, error) {
```
acruikshank: This should be moved into buildHost.

It can also wait for another PR.

node/builder.go
```diff
@@ -0,0 +1,364 @@
+package node
+
+// Review: is a builder pattern with dot-chaining idiomatic in go?
```
anorth: It's not nearly as common as in other languages, but I am also implementing one for the node builder for tests (possibly I should rename or we can converge).

The [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) is often used for similar goals. I don't like it as much for the mutability, but it is canonical. Either is fine here.

node/builder.go
```diff
@@ -0,0 +1,364 @@
...
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	dhtopts "github.com/libp2p/go-libp2p-kad-dht/opts"
+	libp2pps "github.com/libp2p/go-libp2p-pubsub"
+	rhost "github.com/libp2p/go-libp2p/p2p/host/routed"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+	ma "github.com/multiformats/go-multiaddr"
+	"github.com/pkg/errors"
+)
+
+// TODO: Create a `Builder` struct to keep track of the configuration
```
anorth: We have a policy against merging new `TODO`s unless they reference GitHub issues. Your options are:
- just delete the comment if you'll follow up with it
- remove the word TODO and leave commentary, e.g. "an alternative way of doing this is ...."
- create an issue and reference the issue in the TODO

node/builder.go
```diff
@@ -0,0 +1,364 @@
...
+
+// Config is a helper to aid in the construction of a filecoin node.
+// TODO: rename to `BuilderConfig`
```
anorth: IMO this config object *is* the builder; I don't think we need another layer on top of it - replace this one. A new name would help avoid confusion with the `config.Config`.

Nit: I think all the fields on this struct should be non-exported.

node/builder.go
```diff
@@ -0,0 +1,364 @@
...
+	Rewarder    consensus.BlockRewarder
+	Repo        repo.Repo
+	IsRelay     bool
+}
+
+// ConfigOpt is a configuration option for a filecoin node.
+type ConfigOpt func(*Config) error
+
+// New creates a new node.
+// TODO: rename to `(*NodeBuilder) Build() (*Node, error)`
```
anorth: This one is a free function, not a method, so it would be something like `node.Build(ctx, opts...)`.

icorderi: Yes, you are correct. It currently is a free function. The signature was in line with the Review comment asking if dot chaining was idiomatic, in that case the last step on the chain would be: `.Build()`.

node/builder.go
icorderi: Yes, it should be an opaque structure, and a name change to avoid config.Config confusion will be a next pass. 

The breakup of the config structures from the Builder was more on a followup pass to enable breaking up the Node. I'm not certain all config options present are needed in all combinations of subsystems we might want to start at a given time. But this can be postponed.

node/builder.go
icorderi: I will read it up, in any case, it was more to decide what to do with a follow up PR. I will clean up the review and TODO and we can deal with it afterwards.

node/node.go
icorderi: Yes, it will likely be moved to a `host.go` on a followup PR.

PR #3345, opened 2019-09-03
Update graphsync

hannahhoward:
Graphsync was recently updated to address a bug that likely affects filecoin:
- There was no limit on size for outgoing messages
- There was a 4MB limit on incoming messages
- A slow network connection and very large blocks could result in messages sent that exceeded 4MB, which mean they would not decode and the query would fail

This has been addressed (with no breaking changes) and this PR updates our Graphsync dependency to the new version.


anorth:
@hannahhoward as a committer, do please tag @acruikshank or myself for review up front if you're fairly confident about a PR - no need to wait for a first round of feedback most times.

PR #3353, opened 2019-09-05
add leb128 cli

### LINKED ISSUE #3030: UX: leb128 cli enc/decode
### Description
Our blockchain datastructures are full of leb128 encoded integers.  Sometimes manually interpreting the raw bytes of actor state dags, or internal HAMT bitfields or block headers is important, even if there are pretty print functionalities that will interpret this data for you.  For this reason we should have a CLI command on go-filecoin for at least reading (and probably writing) leb128.  

### Acceptance criteria
`go-filecoin leb128 read "A=="`
`go-filecoin leb128 write 1000`

### Risks + pitfalls
Should be very straigtforward.

### Where to begin
Try avoiding porcelain altogether and just using the Uint64 type in a commands/encodings.go file or something like that.

deaswang:
fixes #3030

### DIFF
--- a/commands/leb128.go
+++ b/commands/leb128.go
@@ -1,0 +1,69 @@
+package commands
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+
+	"github.com/filecoin-project/go-leb128"
+	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
+	cmds "github.com/ipfs/go-ipfs-cmds"
+)
+
+var leb128Cmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Leb128 cli enc/decode",
+		ShortDescription: `Reading and writing leb128.`,
+	},
+	Subcommands: map[string]*cmds.Command{
+		"read":  readLeb128Cmd,
+		"write": writeLeb128Cmd,
+	},
+}
+
+var readLeb128Cmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "read leb128",
+		ShortDescription: `Read leb128 to decode`,
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("text", true, false, `The leb128 encoded text`),
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		text := req.Arguments[0]
+		val := leb128.ToUInt64([]byte(text))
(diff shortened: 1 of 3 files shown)


commands/leb128.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/filecoin-project/go-leb128"
+	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
+	cmds "github.com/ipfs/go-ipfs-cmds"
+)
+
+var leb128Cmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline:          "Leb128 cli enc/decode",
+		ShortDescription: `Reading and writing leb128.`,
```
ZenGround0: nit: say leb128 uint64s.  Then we can updated this when we add big ints to this command using a flag/subcommand.

commands/leb128.go
```diff
@@ -0,0 +1,69 @@
...
+	},
+	Subcommands: map[string]*cmds.Command{
+		"read":  readLeb128Cmd,
+		"write": writeLeb128Cmd,
```
ZenGround0: nit: I know you're just going off of my ideas in #3030 but in retrospect I'd slightly prefer `encode` and `decode` here.

commands/leb128.go
```diff
@@ -0,0 +1,69 @@
...
+	},
+	Run: func(req *cmds.Request, res cmds.ResponseEmitter, env cmds.Environment) error {
+		num, err := strconv.ParseUint(req.Arguments[0], 10, 64)
+		if err != nil {
+			return err
+		}
+		out := leb128.FromUInt64(num)
+		return cmds.EmitOnce(res, out)
+	},
+	Type: []byte{},
```
ZenGround0: Blocking: the write subcommand should print out the leb128 in the canonical string encoding used elsewhere.  For example `go-filecoin leb128 write 65` should print out `A==`, not [65]

commands/main.go
```diff
@@ -170,6 +171,7 @@ var rootSubcmdsDaemon = map[string]*cmds.Command{
 	"dht":              dhtCmd,
 	"id":               idCmd,
 	"inspect":          inspectCmd,
+	"leb128":           leb128Cmd,
```
anorth: I have just realised this should be in the "local" list above. It doesn't require a daemon.

deaswang: if need to use local cmd. should merge pr #3382 first.

anorth:
I have verified this locally at 2395b4a. Invoking moderator privilege to merge.

Issue #3302, opened 2019-08-27
Protocol upgrade schedule


acruikshank:
### Description

In order to determine how to create or validate chain data (block headers, message, actor state), we need to know which protocol version it belongs to. Protocol versions will begin at specific chain heights on a particular network, so we need a schedule to describe these transitions. 

### Acceptance criteria

1. A schedule must be checked in with the go-filecoin source code (it can be Go code). At some point we may want to be able to override the schedule with local data, but this is not part of this issue.
2. The schedule will be a table that contains network name, protocol version, and the chain height at which the protocol version goes into effect.
3. The schedule must be reasonably easy to read and update with protocol upgrades to all known networks.
4. A data structure must exist to store all schedule info. When given a network name and chain height, it must efficiently find the last protocol version for the network whose chain height is at or below the given chain height.
5. The data structure will be initialized by node, and dependency injected into any code that requires it (currently only plumbing).
6. The protocol version will be retrieved by the `protocol` command and displayed in its output.

acruikshank:
This should have been closed by #3354.

PR #3354, opened 2019-09-05
Protocol Upgrade Table

acruikshank:
### Problem

In order to maintain a durable network, we will need to carefully manage changes to protocol affecting code. Part of this is ensuring the entire network transitions to the code at the same block height. In order to coordinate this, we will need to:

* schedule these upgrades in advance
* distribute this schedule with source code and binaries, so the code can transition automatically
* specify a different schedule for every network in operation since their block heights will differ

See #3302 

### Solution

This is advance work to establish the mechanism for specifying the upgrade schedule without taking advantage of it in any way. The work encompasses the following elements:

1. Create a `ProtocolUpgradeTable` that can be configured with a network and the protocol version and effective at block height for every upgrade relevant to a network. Once configured, it can return the active protocol version for any block height.
2. Create a `ConfigureProtocolVersions` that populates the protocol upgrade table with all known upgrades.
3. Have the node lookup the network name configured in the genesis block so that it can use it to configure the protocol upgrade table.
4. Instantiate the protocol upgrade table and configure it during node intialization. In future work, this PUT will be injected into other components so that they can make decisions about which code to use based on the active protocol version.


acruikshank:
There are a lot of choices that went into this PR. Some were reasoned out in the [design doc](https://docs.google.com/document/d/17VsfFQk1mZKJj9gkXgSzIPWZAcVeL7VyNjP-is576e4/edit#heading=h.63z8wdxpvyeu), others are just the simplest thing I could think of at the time. Some will only come into effect with the first PR that attempts to use the PUT. Some won't matter until we actually attempt to upgrade a running network, and some will take longer still to play out. I'm not wedded to anything, but I think this is a good basis for moving forward with the upgrade mechanism.

cc @anorth @ingar

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NewProtocolUpgradeTable creates a new ProtocolUpgradeTable that only tracks upgrades for the given network
+func NewProtocolUpgradeTable(network string) *ProtocolUpgradeTable {
+	return &ProtocolUpgradeTable{
+		Network: network,
+	}
+}
+
+// Add configures an upgrade for a network. If the network doesn't match the PUT's network, this upgrade will be ignored.
+func (put *ProtocolUpgradeTable) Add(network string, version uint64, effectiveAt *types.BlockHeight) {
```
anorth: There's no need for this table to be mutable - AFAICT it will always be fully known at construction. Given that, I would much prefer it were immutable, and constructed full-formed by constructor function or builder pattern. That will allow easier enforcing of invariants (like a mandatory height-zero version) and simplify the logic of inserting into the slice down to just sorting a slice once.

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,68 @@
...
+	// ignore upgrade if not part of our network
+	if network != put.Network {
+		return
+	}
```
anorth: I think I see how this is working with the static upgrade definitions, but it's obscure. The network name is known when the upgrade table is configured and a node only ever needs one table: the one for its network name. Hence the network name should not required as a property on this object at all.

`protocol_upgrades.go` may have a map of network names to lists of protocol versions, but only one entry of that map need be looked up when this table is built.

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,68 @@
...
+// protocolUpgrade specifies that a particular protocol version goes into effect at a particular block height
+type protocolUpgrade struct {
+	Version     uint64
+	EffectiveAt *types.BlockHeight
+}
+
+// ProtocolUpgradeTable is a data structure capable of specifying which protocol versions are active at which block heights
+type ProtocolUpgradeTable struct {
+	Network  string
+	upgrades []protocolUpgrade
```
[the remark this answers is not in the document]
acruikshank: This is now documented here on this struct and enforced in the builder's `Build` method.

consensus/protocol_upgrade_table.go
acruikshank: The network name is now only configured in the builder. The table only contains the upgrades.

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,103 @@
+package consensus
```
anorth: How about a `version` subpackage?

consensus/protocol_upgrade_table.go
anorth: This silent ignoring of the network name still puts me off. I think it would be cleaner if the network parameter was removed.

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,103 @@
...
+	if network != putb.network {
+		return putb
+	}
+
+	upgrade := protocolUpgrade{
+		Version:     version,
+		EffectiveAt: effectiveAt,
+	}
+
+	// insert upgrade sorted by effective at
```
anorth: Remove comment

consensus/protocol_upgrade_table.go
```diff
@@ -0,0 +1,103 @@
...
+	upgrades := make([]protocolUpgrade, len(putb.upgrades))
+	copy(upgrades, putb.upgrades)
+
+	// enforce that the current network has an entry at block height zero
+	if len(upgrades) == 0 {
+		return nil, errors.Errorf("no protocol versions specified for network %s", putb.network)
+	}
+	if !upgrades[0].EffectiveAt.Equal(types.NewBlockHeight(0)) {
+		return nil, errors.Errorf("no protocol version at genesis for network %s", putb.network)
+	}
```
anorth: Validate that there are not two upgrades for the same version number

consensus/protocol_versions.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+// ALPHA1 is the first alpha network
+const ALPHA1 = "alpha1"
+
+// DEVNET4 is the network name of devnet
+const DEVNET4 = "devnet"
```
anorth: `"devnet4"` too

version/protocol_version_table.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// ProtocolVersionTable is a data structure capable of specifying which protocol versions are active at which block heights.
+// It must be constructed with the ProtocolVersionTableBuilder which enforces that the table has at least one
+// entry at block height zero and that all the versions are sorted.
+type ProtocolVersionTable struct {
+	versions []protocolVersion
+}
+
+// VersionAt returns the protocol versions at the given block height for this PUT's network.
```
ingar: Super nit: "block height for this **PVT**'s network" now

PR #3355, opened 2019-09-05
replace clock impl with wrapper around clockwork

frrist:
### What
As a follow on to #3339, this PR replaces the existing clock implementation with a wrapper around [github.com/jonboulle/clockwork](https://github.com/jonboulle/clockwork). The package clockwork was chosen based on the interface's it provides and its frequent usage in the go ecosystem, a full list of importers can be seen [here](https://godoc.org/github.com/jonboulle/clockwork?importers). I wrapped this implementation instead of using it directly to make code changes easier if we decided to use something different, or write our own.

Clock interface:
```go
type Clock interface {
	After(d time.Duration) <-chan time.Time
	Sleep(d time.Duration)
	Now() time.Time
	Since(t time.Time) time.Duration
	NewTicker(d time.Duration) Ticker
}
```
FakeClock interface:
```go
type FakeClock interface {
	Clock
	// Advance advances the FakeClock to a new point in time, ensuring any existing
	// sleepers are notified appropriately before returning
	Advance(d time.Duration)
	// BlockUntil will block until the FakeClock has the given number of
	// sleepers (callers of Sleep or After)
	BlockUntil(n int)
}
```

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -1,31 +1,25 @@
 package clock
 
 import (
-	"time"
+	"github.com/jonboulle/clockwork"
 )
 
-// Clock defines an interface for fetching time that may be used instead of the
-// time module.
+// Clock provides an interface that packages can use instead of directly
+// using the time module, so that chronology-related behavior can be tested.
 type Clock interface {
-	Now() time.Time
-	Sleep(d time.Duration)
+	clockwork.Clock
 }
 
-// SystemClock delegates calls to the time package.
-type SystemClock struct{}
+// SystemClock is a wrapper around the jonboulle/clockwork.Clock interface
+// Clock provides an interface that packages can use instead of directly
+// using the time module, so that chronology-related behavior can be tested.
+type SystemClock struct {
+	clockwork.Clock
+}
 
-// NewSystemClock returns a SystemClock that delegates calls to the time package.
+// NewSystemClock returns a SystemClock that delegates calls to the jonboulee/clockwork package.
 func NewSystemClock() *SystemClock {
-	return &SystemClock{}
-}
-
-// Now returns the current local time.
-func (sc *SystemClock) Now() time.Time {
-	return time.Now()
-}
-
-// Sleep pauses the current goroutine for at least the duration d.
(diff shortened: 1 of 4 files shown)


clock/clock.go
```diff
@@ -1,31 +1,25 @@
...
-
-// Sleep pauses the current goroutine for at least the duration d.
-// A negative or zero duration causes Sleep to return immediately.
-func (sc *SystemClock) Sleep(d time.Duration) {
-	time.Sleep(d)
+// NewSystemClock returns a SystemClock that delegates calls to the jonboulee/clockwork package.
+func NewSystemClock() *SystemClock {
+	return &SystemClock{
+		Clock: clockwork.NewRealClock(),
+	}
```
anorth: Since `clockwork.NewRealClock` satisfies the `Clock` interface directly, there's no value in wrapping it. Change the return type to be the interface and return `NewRealClock()`.

testhelpers/clock.go
```diff
@@ -1,34 +1,20 @@
...
 )
 
-// FakeSystemClock returns a mocked clock implementation that may be manually
-// set for testing things related to time.
+// FakeSystemClock wrapps the jonboulle/clockwork.FakeClock interface.
+// FakeSystemClock provides an interface for a clock which can be
+// manually advanced through time.
 type FakeSystemClock struct {
-	now time.Time
+	clockwork.FakeClock
```
anorth: I don't think the wrapper adds much value. Test code can just call `clockwork.NewFakeClockAt` directly.

anorth:
Thanks for the references. I'm wary of external deps, but this one seems good. We'll wait for security audit.

testhelpers/clock.go
frrist: In the event we decide to replace the `clockwork` dep with our own implementation I'd like to avoid re-writing imports in tests that use it. Any objections to:
```go
type FakeClock interface {
        clockwork.FakeClock
}

// NewFakeClock returns a FakeClock initialised at the given time.Time.
func NewFakeClock(n time.Time) FakeClock {
        return clockwork.NewFakeClockAt(n)
}
```

PR #3356, opened 2019-09-05
rename exceedsFinalityLimit to indicate trust

### LINKED ISSUE #3130: Correct inaccurate "finality" name
### Description
Our syncer won't sync a chain that is too far ahead under certain circumstances.  Our code calls this concept finality which veers from the well known concept of finality.  We should give `finality` a snappy, accurate name.  

### Acceptance criteria
Goal: change the misleading name
Non-goal: evaluate if it is safe to do block chains that are too long or if we should actually be implementing finality checks.

### Where to begin
First think of a good name, then change [this function](https://github.com/filecoin-project/go-filecoin/blob/master/chain/syncer.go#L481) and all consumers.

frrist:
closes #3130

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -24,9 +24,9 @@ func init() {
 	reorgCnt = metrics.NewInt64Counter("chain/reorg_count", "The number of reorgs that have occurred.")
 }
 
-// FinalityLimit is the maximum number of blocks ahead of the current consensus
-// chain height to accept once in caught up mode
-var FinalityLimit = 600
+// UntrustedChainHeightLimit is the maximum number of blocks ahead of the current consensus
+// chain height to accept if syncing without trust.
+var UntrustedChainHeightLimit = 600
 var (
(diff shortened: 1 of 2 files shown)


chain/syncer.go
```diff
@@ -425,7 +425,9 @@ func (syncer *Syncer) Status() Status {
 	return syncer.reporter.Status()
 }
 
-func (syncer *Syncer) exceedsFinalityLimit(curHeight, newHeight uint64) bool {
-	finalityHeight := curHeight + uint64(FinalityLimit)
+// ExceedsUntrustedChainLength returns true if the delta between curHeight and newHeight
+// exceeds the maximum number of blocks to accept if syncing without trust, false otherwise.
+func ExceedsUntrustedChainLength(curHeight, newHeight uint64) bool {
+	finalityHeight := curHeight + uint64(UntrustedChainHeightLimit)
```
ZenGround0: nit `finalityHeight` here should go too to be consistent

PR #3362, opened 2019-09-06
Add Has to Peertracker, filter pubsub blocks

ZenGround0:
This is a simpler version of a change that has improved devnet forking situation.  It prevents syncing of pubsub messages that come from peers that a node does not have a connection with.

If these messages are valid heads coming in from a new peer then the parallel hello protocol will serve this head to the syncer anyway.  If that hello protocol fails then this change ensures that we don't end up blocking for a 60 second timeout on a peer that we can't connect to.

### DIFF
--- a/node/block.go
+++ b/node/block.go
@@ -52,6 +52,13 @@ func (node *Node) processBlock(ctx context.Context, pubSubMsg pubsub.Message) (e
 	if err != nil {
 		return errors.Wrap(err, "got bad block data")
 	}
+
+	// ignore messages from peers not yet tracked
+	if !node.PeerTracker.Has(from) {
+		log.Debugf("ignoring new block cid: %s from unconnected peer: %s", blk.Cid().String(), from.Pretty())
+		return nil
+	}
+
 	span.AddAttributes(trace.StringAttribute("block", blk.Cid().String()))
 
 	log.Infof("Received new block from network cid: %s", blk.Cid().String())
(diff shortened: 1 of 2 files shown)


node/block.go
```diff
@@ -45,6 +45,12 @@ func (node *Node) processBlock(ctx context.Context, pubSubMsg pubsub.Message) (e
 		return nil
 	}
 
+	// ignore messages from peers not yet tracked
+	if !node.PeerTracker.Has(from) {
+		log.Debugf("Ignoring new block from %s because we don't know this peer", from.Pretty())
```
frrist: nit: should be "ignoring new block from unconnected peer %s" or similar- say something about not being connected to them. Also consider putting the block CID in here, might be helpful for debugging later.

anorth:
This change makes me very uneasy! It's possible that more justification could quell my fears, but that needs to go into the code.

A node will be connected to only a small subset of peers. Most new block messages will thus originate from peers to which the node is not connected. The Hello protocol will not be connecting to all the peers and this appears to me to just be ignoring valid blocks.

The libp2p code documentation is weak, but I interpret the `From` as being the message originator, rather than the pubsub peer from which the message was directly received (if the latter case, of course we'd be connected already).

Later, when we implement the block protocol properly, a node will source almost all the messages locally from its message pool and there will be no graphsync request for the block header anyway.

I have filed #3368 to track reverting this.

Issue #3347, opened 2019-09-03
Support building via Makefile


frrist:
### Description
go-filecoin currently supports building via the [`build/`](https://github.com/filecoin-project/go-filecoin/tree/master/build) package. It's written in go and basically a wrapper around the `os/exec` package. The goal of this build system is to make development easier on all operating system since go is widely supported.

I would like to discuss adding support for the `make` build tool as a way to increase developer productivity and UX. Advantages of make:
- Build targets(!!!): stop building things that haven't changed.
- Consolidated deps process: currently in addition to running deps you must [first run a submodule command](https://github.com/filecoin-project/go-filecoin/issues/3211), this could be avoided with make.

### Acceptance criteria
- We decide if we are willing to support building via Make.

anorth:
I have no objection in principle, but can't judge this until seeing a proof-of-concept implemented. I don't think build is a major problem right now (though we should add the submodule update to `deps`.)

PR #3363, opened 2019-09-06
POC Makefile

frrist:
### What
This is a POC for #3347. It adds a basic Makefile for every binary that may be built in the go-filecoin repo and allows the parent Makefile to call them.


anorth:
Can you delete the go build files too so we see a comparison?

This needs advice from more than the usual two reviewers - please announce somewhere.

PR #3366, opened 2019-09-06
Implement chain catchup syncer

frrist:
### What does this PR do
This PR adds a chain catchup syncer. The job of the catchup syncer is to ensure a filecoin node is caught up to trusted peers. Once caught up a node may sync all heads it receives over pubsub with the `trusted` flag set to false.

#### How should I review this PR?
I have broken these changes up into separate commits, please review each one individually. Some commits will not make much sense initial as their usage comes in later commits. I did this to ensure CI passed for each one.

##### Clean up code and extend methods:
- 341d5f5427e699f1ac160c5eee3c62ad99a14762: alias a our clock.Ticker
- 38526c1d4a3bcb51572c7f815aefec946d16ffb6: SetupGenesisNode accepts a variadic parameter that allows for modification to a genesis nodes config before its started (cc @travisperson)
- a818bfc83150f923a403e3ad93793a9984d295be: fix some import path ordering.

##### New config options added for syncing:
- 5b4217a872db80e7b24b8133c4e0572c214f7931: adds a `SyncConfig`, the values used in this control how a filecoin node syncs (it will be used in following commits). See comments on config for details.
- 6c0006be9ba77a439c148b6013e8986ba9f19f30: adds an init flag `CatchupSyncTestMode`, when true the `SyncConfig` will be modified s.t. `TrustAllPeers` is true and `CatchupSyncerPeriod` is 500ms. This is required because TestDaemon and iptbwrapper tests do not allow their config to be modified before starting, and the default values for `SyncConfig` are too slow for tests.
- fa2f7f6498174bd615c662ea4ea90aa59798f386: does two things. (1) passes `--catchup-sync-test-mode` to daemon and iptbwrapper tests. (2) modifies FAST nodes config with a testing `SyncConfig`. (cc @travisperson)

##### peer tracker modification for trust and selection threshold.
- 0123fb61f395dcc5f24d0290e8099565511d959c: adds a new field and method to the peer tracker.
  - `trustAllPeers` when true will cause `SelectHead` to consider the heads of all peers instead of just trusted peers. This is helpful when writing tests that do not want to depend on chain catchup logic.
  - `Trust(p peer.ID)` adds a peer to the peer trackers trusted list, this method is helpful for unit testing when perIDs are not deterministic and cannot be placed in a config before starting.

##### catchup syncer implementation and tests
- 49802229473c38ef47f6e1ff723096b611cebbb2: adds the catchup syncer to the chain package along with unit and integration tests. Wires up the catchup syncer to node. And mofifies `block_propagate_test.go` to skip catchup syncing. 
- 5cca6823b7f1a52afabc8385a1012437b2629715: adds a functional test using FAST lib.


chain/catchup.go
```diff
@@ -0,0 +1,123 @@
...
+	return &CatcherUpper{
+		tracker: t,
+		syncer:  sy,
+		store:   st,
+		done:    done,
+	}
+}
+
+// Start starts a go routine to performe the catchup process.
+func (cu *CatcherUpper) Start(ctx context.Context, tick clockwork.Ticker) <-chan error {
```
frrist: Accepting a ticker with the a goal of making testing easier using the new clock we have added. Using the ticker we should be able to control when this loop executes.

chain/catchup.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (cu *CatcherUpper) tryCatchUp(ctx context.Context) (bool, error) {
+	// It bothers me that tick could have a shorter duration than the timeout...
+	updtCtx, cancel := context.WithTimeout(ctx, time.Second*15)
+	defer cancel()
+
+	// if this failes it means we are tracking peers, but none of them replied.
+	if err := cu.tracker.UpdateTrusted(updtCtx); err != nil {
+		return false, retryableError{err}
```
frrist: Borrowed this error wrapping from [os.errors](https://github.com/golang/go/blob/master/src/os/error.go#L110). Any errors that come from the tracker should be retried. Errors from the store and syncer indicate a more serious problem.

anorth: I think it would be simpler to just return three values: success, retry, error

config/config.go
```diff
@@ -229,6 +230,26 @@ func newDefaultSectorbaseConfig() *SectorBaseConfig {
 	}
 }
 
+// SyncConfig holds all values that control how and when a filecoin node syncs its chain.
+type SyncConfig struct {
+	// CatchupSyncerPeriod represents how frequently a node will attempt to catch up when performing
+	// catchup sync.
+	CatchupSyncerPeriod string `json:"catchupSyncerPeriod"`
+	// TrustedAddress represents trusted peers to sync from.
+	TrustedAddresses []string `json:"trustedPeers"`
```
ZenGround0: Is it weird to have both `TrustedAddresses` here and `BootstrapAddresses` in the Bootstrap config?  It's not clear to me if these two fields are the same thing or not.

commands/init.go
```diff
@@ -40,6 +40,7 @@ var initCmd = &cmds.Command{
 		cmdkit.BoolOption(DevnetStaging, "when set, populates config bootstrap addrs with the dns multiaddrs of the staging devnet and other staging devnet specific bootstrap parameters."),
 		cmdkit.BoolOption(DevnetNightly, "when set, populates config bootstrap addrs with the dns multiaddrs of the nightly devnet and other nightly devnet specific bootstrap parameters"),
 		cmdkit.BoolOption(DevnetUser, "when set, populates config bootstrap addrs with the dns multiaddrs of the user devnet and other user devnet specific bootstrap parameters"),
+		cmdkit.BoolOption(CatchupSyncTestMode, "when set, the daemon is places in a special mode for chain catchup operations").WithDefault(false),
```
ZenGround0: nit: "placed"

also what does it mean placed in a "special mode" it would help to be more specific here.

commands/main.go
```diff
@@ -81,6 +81,10 @@ const (
 	// IsRelay when set causes the the daemon to provide libp2p relay
 	// services allowing other filecoin nodes behind NATs to talk directly.
 	IsRelay = "is-relay"
+
+	// CatchupSyncTestMode places the daemon in a spcial catchup mode for testing
```
ZenGround0: nit: special, also file an issue and add that issue number to the TODO

commands/init.go
ZenGround0: forget about the "be more specific" idea as its clear from below and a temporary option

config/config.go
frrist: Yes a bit weird as this change is premature. Currently the role of bootstrap nodes is overloaded, they mine blocks in addition to bootstrapping nodes to the network. In the future I believe we plan to separate these roles- bootstrappers and miners. Once that happens the addresses in this field will be different than the addresses in the bootstrap filed.

commands/init.go
frrist: I'll will adjust the flag description to make it more clear that this is temporary .

frrist:
Closing until this becomes a priority.

Issue #2962, opened 2019-06-19
Fragile and wasteful fetch/validation of blocks received by pubsub


anorth:
### Description
The messages received on the block pubsub are full blocks (later: headers), not CIDs. The pubsub listener `node.processBlock` decodes the block but then drops it on the floor, passing only the CID to `Syncer.HandleNewTipSet`. The syncer then fetches the block again from another peer.

This is obviously wasteful, but less obviously is critical to block syntax validation (incl timestamp). Blocks are only validated by the fetcher, so if we adjusted HandleNewTipSet to accept the decoded blocks, they would skip validation. We could validate them there too, of course, but such a pattern of requiring validation before each caller of `HandleNewTipSet` is fragile.

### Acceptance criteria
- Blocks (headers) received via pubsub are not re-fetched from another peer
- There is a clear single entry point to chain sync where new blocks are validated

This is closely related to #2128. FYI @frrist

anorth:
Implementing a pubsub topic validator would probably address this #2783. We should be very clear about whether a block header passed to the syncer is expected to be already validated, though.

anorth:
@frrist with messages separated from block headers, and pubsub validators, we might be able to fix this now.

frrist:
Agreed, landing this will require re-working part of how the syncer processes potential new heads. Currently, it only works with TipSets and has some tangled (https://github.com/filecoin-project/go-filecoin/issues/2128) assumptions of what's in the store. I'll need to sketch something out before providing a more concrete proposal on the changes required.

anorth:
@ZenGround0  did more recent chain sync work resolve this?

ZenGround0:
No.  This is one of the first performance improvements we need to make once interop is working.

PR #3374, opened 2019-09-09
Remove from peer.ID in fetcher

### LINKED ISSUE #3191: Remove Bitswap Fetcher once Graphsync Fetcher is battle tested
### Description

#3190 makes Graphsync the default fetcher for Blockchains in Filecoin. Once this goes in production in 0.4, and any issues are worked out, there no need to keep the Bitswap Fetcher around, and the code can be removed.

### Acceptance criteria

Bitswap Fetcher is not in Filecoin codebase

### Risks + pitfalls

Again, Graphsync Fetcher is not battle tested yet, so this shouldn't be done until we're sure we won't need the Bitswap Fetcher

### Where to begin

n/a

### LINKED ISSUE #3368: Stop ignoring heads from disconnected peers
### Description
#3362 introduced a change to ignore blocks received over pubsub that originate from peers a node is not directly connected to.

A node will be connected to only a small subset of peers. Most new block messages will thus originate from peers to which the node is not connected. The Hello protocol will not be connecting to all the peers and this appears to me to just be ignoring valid blocks.

### Acceptance criteria
- Blocks received over pubsub are immediately considered for validation (without re-introducing the timeout-related problems that prompted the change).

FYI @ZenGround0

ZenGround0:
closes #3368 
closes #3191 too


net/graphsync_fetcher_test.go
```diff
@@ -571,6 +578,21 @@ type fakePeerTracker struct {
...
+
 func (fpt *fakePeerTracker) List() []*types.ChainInfo {
 	return fpt.peers
 }
+
+func (fpt *fakePeerTracker) Self() peer.ID {
+	return peer.ID("")
+}
+
+func requireBlockStorePut(t *testing.T, bs bstore.Blockstore, data format.Node) {
```
frrist: Copy from `net/fetcher_test.go`? Not seeing where this is used.

net/graphsync_fetcher.go
```diff
@@ -30,6 +30,14 @@ const (
 	requestTimeout = 60 * time.Second
 )
 
+// Fetcher defines an interface that may be used to fetch data from the network.
+type Fetcher interface {
+	// FetchTipSets will only fetch TipSets that evaluate to `false` when passed to `done`,
+	// this includes the provided `ts`. The TipSet that evaluates to true when
+	// passed to `done` will be in the returned slice. The returns slice of TipSets is in Traversal order.
+	FetchTipSets(context.Context, types.TipSetKey, peer.ID, func(types.TipSet) (bool, error)) ([]types.TipSet, error)
```
frrist: Thinking out loud here:

I believe the only reason we need `peer.ID` as a parameter to this method is because we are not persisting blocks from pubsub (instead we are refetching them). 
If nodes persist the blocks they received over pubsub then I think we can guarantee peers we share a connection with (i.e. that are in the peer tracker) will always have the blocks we are fetching (else how would have we gotten the pubsub message). I imagine the concept of an `initialPeer` can go away once #2962 lands since the peers in the tracker will have the block.
Although I don't think that covers the case of fetching blocks our peers have deemed semantically invalid and thus not stored..

net/graphsync_fetcher.go
```diff
@@ -340,3 +367,20 @@ func (pri *requestPeerFinder) FindNextPeer() error {
 	}
 	return fmt.Errorf("Unable to find any untried peers")
 }
+
+func sanitizeBlocks(ctx context.Context, unsanitized []blocks.Block, validator consensus.BlockSyntaxValidator) ([]*types.Block, error) {
```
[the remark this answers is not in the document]
ZenGround0: yup, putting here after I removed `net/fetcher.go`

net/graphsync_fetcher_test.go
ZenGround0: yes, used in `graphsync_fetcher_test.go`

net/graphsync_fetcher.go
ZenGround0: > I believe the only reason we need peer.ID as a parameter to this method is because we are not persisting blocks from pubsub

I think this is a little less guaranteed then what you are claiming.  There is a race between (1)  the other peer sending us the pubsub message and us asking for data from that peer and (2) the other peer forwarding the pubsub message and then requesting and receiving the data from one of its peers.

However the way that we use `initialPeer` here is really only to check if this node was the one that mined the block so that we can send a request to ourselves over graphsync in this (and only this case).  The reason for doing this is unfortunately complicated -- requesting for data from ourselves *when we have it* does not block on graph sync, but requesting for data from ourselves when we don't have it *does block* so we can't simply include our peer id in the tracker.

net/graphsync_fetcher.go
```diff
@@ -319,10 +331,25 @@ type requestPeerFinder struct {
...
-	pri.triedPeers[initialPeer] = struct{}{}
-	return pri
+func newRequestPeerFinder(peerTracker graphsyncFallbackPeerTracker, initialPeer peer.ID) (*requestPeerFinder, error) {
+	pri := &requestPeerFinder{
+		peerTracker: peerTracker,
+		triedPeers:  make(map[peer.ID]struct{}),
+	}
+
+	// Use the initialPeer's ID if it is our own
+	if peerTracker.Self() == initialPeer {
```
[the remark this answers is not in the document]
anorth: Please leave some more comments here, explicitly calling out that `initialPeer` is ignored if not self (and why). It should probably be renamed to be descriptive rather than prescriptive - maybe `originatingPeer`? Is that information received over libp2p pubsub even verifiable? Do we require signed messages?

If all we need is a boolean here, it would be even simpler to just pass a boolean.

net/graphsync_fetcher.go
frrist: In the event that our node is the miner of the block we should avoid going to graphsync all together since we have the block (and its parents) locally. Given the wiring of the syncing code this isn't an option. I am thinking that it should be, and that #2962 moves us in that direction. Do you have similar thoughts on this?

net/graphsync_fetcher.go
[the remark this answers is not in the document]
ZenGround0: I like boolean and will do.

net/graphsync_fetcher.go
ZenGround0: 100% agreed.  This made more sense with the bitswap fetcher and now is just causing more friction.  I also agree that #2962 is a great place to do this.  Once we have code to handle complete blocks arriving over pubsub that short circuits fetching then we can quickly use it for the case of blocks arriving locally from the mining subsystem.

PR #3378, opened 2019-09-10
Re-queue messages from the losing chain segment in a re-org so that nonces are consistent with the pool.

### LINKED ISSUE #3052: [bug] Same actor and nonce but different cid

anorth:
Fixes #3052 (see discussion there for discussion of alternatives).

### DIFF
--- a/core/policy.go
+++ b/core/policy.go
@@ -30,6 +30,7 @@ type QueuePolicy interface {
 // PolicyTarget is outbound queue object on which the policy acts.
 type PolicyTarget interface {
 	RemoveNext(ctx context.Context, sender address.Address, expectedNonce uint64) (msg *types.SignedMessage, found bool, err error)
+	Requeue(ctx context.Context, msg *types.SignedMessage, stamp uint64) error
 	ExpireBefore(ctx context.Context, stamp uint64) map[address.Address][]*types.SignedMessage
 }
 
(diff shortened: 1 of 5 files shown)


icorderi:
Is the solution of resubmitting the messages on the "alternative" head sound?
Since there is a new state, some of the messages might no longer be applicable, isn't safer to reject all messages that are not present on the new chain?

core/policy.go
```diff
@@ -54,7 +55,22 @@ func NewMessageQueuePolicy(messages MessageProvider, maxAge uint) *DefaultQueueP
...
 func (p *DefaultQueuePolicy) HandleNewHead(ctx context.Context, target PolicyTarget, oldTips, newTips []types.TipSet) error {
-	// Rearrange the tipsets into increasing height order so messages are discovered in nonce order.
+	chainHeight := uint64(0)
+	var err error
+	if len(newTips) > 0 {
+		chainHeight, err = newTips[0].Height()
+		if err != nil {
+			return err
+		}
+	} else if len(oldTips) > 0 { // A pure rewind is unlikely in practice.
```
acruikshank: This is a little weird. Seems like the only consequence of this clause is that the message requeued will be stamped with the maximum old chain height. It would be more accurate to use the minimum old chain height. Like you say, this is unlikely (impossible?) and the consequences seem small, so seems fine.

anorth: I opted for the maximum as a "safe" option, to prevent premature expiration.

anorth:
@icorderi the goal is to match the behaviour of the message pool (with which this may one day be merged). The pool definitely needs to make un-mined messages available for inclusion in new blocks else such messages would be lost entirely.

In practise, this works very well nearly all the time: the message nonce means messages must apply to a single actor's state in the order they were generated. If multiple messages with the same actor and nonce are generated (on different nodes) then it's possible for a subsequent message to have differing assumptions about the prior state, but this can't generally be avoided in such a distributed system, except by waiting for messages to apply or expire. It's generally considered much more useful to keep trying to apply messages (and have message semantics that make this ok) than to bump the problem back up to the user or system that wants to send a message. We borrow these ideas from prior art like Ethereum.

PR #3379, opened 2019-09-10
add power table cli

### LINKED ISSUE #3028: UX: power table visualization
### Description
Searching the output of `actor ls` and then running `miner power` is something but we should have an easier way for a node operator to easily look at the power table.

### Acceptance criteria
A CLI command that helps users visualize the power command

### Risks + pitfalls
Our CLI top level is getting messy.  This should not make the mess worse.

### Where to begin
Maybe a `go-filecoin storage-market` or `go-filecoin market` under `ACTOR COMMANDS` is the right place to put this?

deaswang:
fixes #3028
use cli `go-filecoin actor power-table`.

### DIFF
--- a/commands/actor.go
+++ b/commands/actor.go
@@ -13,8 +13,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/actor/builtin/miner"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/paymentbroker"
 	"github.com/filecoin-project/go-filecoin/actor/builtin/storagemarket"
-	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/exec"
+	"github.com/filecoin-project/go-filecoin/state"
 	"github.com/filecoin-project/go-filecoin/types"
 
 	"github.com/ipfs/go-cid"
@@ -31,13 +31,6 @@ type ActorView struct {
 	Balance   types.AttoFIL   `json:"balance"`
 	Exports   readableExports `json:"exports"`
 	Head      cid.Cid         `json:"head,omitempty"`
-}
-
-// PowerTable represents entry of miner power table
-type PowerTable struct {
-	Address string
-	Power   types.BytesAmount
-	Total   types.BytesAmount
 }
 
 // readableFunctionSignature is a representation of an actors function signature,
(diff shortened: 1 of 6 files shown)


commands/actor.go
```diff
@@ -47,7 +49,8 @@ var actorCmd = &cmds.Command{
 		Tagline: "Interact with actors. Actors are built-in smart contracts.",
 	},
 	Subcommands: map[string]*cmds.Command{
-		"ls": actorLsCmd,
+		"ls":          actorLsCmd,
+		"power-table": actorPowerCmd,
```
anorth: Only miner actors can have power, and it would be more findable if this command were with the existing `miner power` command. Please move this to the `miner` command. Move the tests appropriately.

commands/actor_daemon_test.go
```diff
@@ -48,3 +53,40 @@ func TestActorDaemon(t *testing.T) {
...
+		consensus.MinerActor(minerAddr2, addr2, th.RequireRandomPeerID(t), types.NewAttoFILFromFIL(10), types.OneKiBSectorSize),
+		consensus.MinerActor(minerAddr3, addr3, th.RequireRandomPeerID(t), types.NewAttoFILFromFIL(100), types.OneKiBSectorSize),
+	)
+
+	// Create node (it's not necessary to start it).
+	b := test.NewNodeBuilder(t)
+	node := b.WithGenesisInit(testGen).
+		WithConfig(func(c *config.Config) {
+			c.Mining.AutoSealIntervalSeconds = 120
+		}).
```
anorth: Unnecessary config

commands/actor_daemon_test.go
```diff
@@ -48,3 +53,40 @@ func TestActorDaemon(t *testing.T) {
...
+	require.NoError(t, node.ChainReader.Load(ctx))
+
+	// Run the command API.
+	cmd, stop := test.RunNodeAPI(ctx, node, t)
+	defer stop()
+
+	out := cmd.RunSuccess(ctx, "actor", "power-table").ReadStdout()
+	assert.Contains(t, out, minerAddr1.String())
+	assert.Contains(t, out, minerAddr2.String())
+	assert.Contains(t, out, minerAddr3.String())
```
anorth: Please assert the entire output. I hope that the order is deterministic.

anorth:
I am on the fence about accepting this. See my comment https://github.com/filecoin-project/go-filecoin/issues/3028#issuecomment-532976757

The one reason I'd accept it is in recognition that @deaswang you stepped up to take on an issue that we had filed (although categorized as icebox) and we really value your engagement. Nevertheless, I think this code will be deleted as we migrate to a new API in the future.

@ZenGround0

ZenGround0:
I'll echo @anorth in saying that your work is high quality and valuable for the project @deaswang.  From the above it sounds like the right call is to close this down and close the issue too.  

@anorth, @deaswang  how do you think we can prevent mixups like this in the future.  Should we ask contributors to comment below issues they want to take on so that we can catch things we've subsequently decided against before contributors put in work?  Should we be asking contributors to only consider candidate issues (might require zenhub which contributors can't access) or issues with a certain label, i.e. `help-wanted`?  Are issues like #3028 rare enough in our backlog that this isn't a problem worth addressing right now.

anorth:
I do think contributors should post a comment on issues they are intending to work on, and be then assigned to such issue to indicate that status.

ZenHub *is* accessible to all contributors: it's free for open source projects. It's documented and linked from our contributing guide. I would expect regular contributors to consult it.

PR #3380, opened 2019-09-10
Fix syncer.syncOne twice bug

ridewindx:
In `syncer.HandleNewTipSet`, if the first tipset is widened and `syncOne`ed, the original first tipset is `syncOne`ed again, but it's lighter and will not be set to head. It is a waste of CPU, since the `widen` of  tipsets happens very often.
The PR wants to avoid the useless second `syncOne` execution of the first tipset, and it also consider that the `chain` must have lenght of 1. It should not impact the correctness of expected consensus.


ridewindx:
@travisperson @anorth
Why the parameter files downloading in `ci/circleci deps_linux` failed?

anorth:
I don't know why that is failing all of a sudden. It's a huge issue and we'll fix it.

chain/syncer.go
```diff
@@ -403,14 +404,18 @@ func (syncer *Syncer) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo,
...
 		}
-		if err = syncer.syncOne(ctx, parent, ts); err != nil {
-			// While `syncOne` can indeed fail for reasons other than consensus,
-			// adding to the badTipSets at this point is the simplest, since we
-			// have access to the chain. If syncOne fails for non-consensus reasons,
-			// there is no assumption that the running node's data is valid at all,
-			// so we don't really lose anything with this simplification.
-			syncer.badTipSets.AddChain(chain[i:])
-			return err
+		// If first tipset is widened, only `syncOne` it when length of `chain` is greater than 1.
```
ZenGround0: nit: it woule be good to add to this comment by describing why do it this way.  How about this:
"If the chain has length greater than 1 then we need to sync each tipset in the chain in order to process the chain fully, including the non-widened first tipset"
"If the the chan has length == 1 we can avoid processing the non-widened tipset as a performance optimization because this tipset cannot be heavier than the widened first tipset."

chain/syncer_test.go
```diff
@@ -70,7 +70,7 @@ func TestTipSetIncremental(t *testing.T) {
 	verifyHead(t, store, t1)
 
 	assert.NoError(t, syncer.HandleNewTipSet(ctx, types.NewChainInfo(peer.ID(""), t2.Key(), heightFromTip(t, t2)), true))
-	verifyTip(t, store, t2, builder.StateForKey(t2.Key()))
+	// verifyTip(t, store, t2, builder.StateForKey(t2.Key()))
```
ZenGround0: nit: remove the commented out line here and line 151.

idea: you could also verify that `t2` is NOT in the store here.

travisperson:
@ridewindx parameter fetching is currently an issue we are trying to address. Unfortunately it disproportionately effects outside contributors without write access more than those with due to the security around CircleCI cache. Every CI run for users without write access must fetch the parameters.

PR #3384 is working to address this issue in the short term.

chain/syncer.go
ridewindx: OK, added comments.

chain/syncer_test.go
ridewindx: Test is also updated.

ridewindx:
@ZenGround0 Yes, I was supposed to open this PR with an issue.

PR #3381, opened 2019-09-10
Disable fetching of test groth parameters in favour of local generation

anorth:
On my MacBook, the one-time generation (whenever new parameters come in upstream) takes about a minute. It's my understanding they're usually cached on CI.

Please land this for me when it gets two ✅.


anorth:
Closing in favour of #3384

PR #3382, opened 2019-09-10
fix local cmd not success

deaswang:
be able to use local cmds.


anorth:
I don't quite understand this. Can you explain what was broken?

deaswang:
> I don't quite understand this. Can you explain what was broken?

Old code: it use Command pointer to compare, but if `go-filecoin leb128 encode 65`, the req.Command will be `encodeLeb128Cmd`, but cmd is `leb128Cmd`, which is value of `rootSubcmdsLocal`, `encodeLeb128Cmd` should not subcmds of root, and I can`t get parent cmd of `encodeLeb128Cmd` too. so it will not equal which should equal.
New code: It compare key of `rootSubcmdsLocal` and first arg of req.Path. so the subcmd of subcmd of root still be the subcmd of root.

commands/main.go
```diff
@@ -306,8 +306,8 @@ func getAPIAddress(req *cmds.Request) (string, error) {
 }
 
 func requiresDaemon(req *cmds.Request) bool {
-	for _, cmd := range rootSubcmdsLocal {
-		if req.Command == cmd {
+	for cmd := range rootSubcmdsLocal {
+		if len(req.Path) > 0 && req.Path[0] == cmd {
```
frrist: The value of `req.Path[0]` here is going to be the first argument supplied to `go-filecoin`, for example:
- `go-filecoin daemon` req.Path[0] == "daemon"
- `go-filecoin version` req.Path[0] == "version"
- `go-filecoin chain ls` req.Path[0] == "chain"

The modification to the loop will cause `cmd` to be one of the below keys
https://github.com/filecoin-project/go-filecoin/blob/0ed37844d780810b6733680190fb21eccdecc991/commands/main.go#L154-L159

The length check on `req.Path` is to avoid panics when we don't pass arguments, is this all correct based on your understanding @deaswang?

commands/main_test.go
```diff
@@ -12,12 +12,15 @@ import (
 func TestRequiresDaemon(t *testing.T) {
 	tf.UnitTest(t)
 
-	reqWithDaemon, err := cmds.NewRequest(context.Background(), []string{}, nil, []string{"chain"}, nil, chainCmd)
+	reqWithDaemon, err := cmds.NewRequest(context.Background(), []string{"chain", "head"}, nil, []string{}, nil, rootCmdDaemon)
```
frrist: These changes are required because previously we were passing `chain", "head` as arguments, which this are not, they go in path ([relevant method signature](https://github.com/ipfs/go-ipfs-cmds/blob/746d639be4ddb96c76ad8a0020466b069edb5e90/request.go#L27)). Your modification addresses this. Is that correct?

PR #3375, opened 2019-09-09
use paramcache only, and create a circleci cache for large params

ognots:
disables the use of paramfetch as work continues to make downloading
more consistent and reliable.

paramfetch is now used to exclusively generate Groth proof parameters
during CircleCI build.

In CircleCI docker build, a new cache is created for the larger groth
parameters required to operate docker image in devnets. Larger
parameters are no longer generated by build and testing stages

PR #3384, opened 2019-09-10
disable paramfetch until optimizations are made

ognots:
paramfetch can sometimes timeout, causing builds to fail and blocking general development pipeline.
problems are especially prevelant for outside committers during CircleCI builds, as they are unable to use the CircleCI cache storing the parameters.

### DIFF
--- a/scripts/install-filecoin-parameters.sh
+++ b/scripts/install-filecoin-parameters.sh
@@ -10,5 +10,11 @@ generate_params() {
   RUST_LOG=info ./proofs/bin/paramcache --test-only
 }
 
-fetch_params
+# Disabling paramfetch until optimizations can be completed
+# Fetching parameters, even small ones, has a long history of being slow and unreliable.
+# It's disabled now in favour of local generation of the small parameter files needed
+# for tests.
+# On a 2015 MacBook pro, generation takes about a minute, one-time.
+#
+# fetch_params
 generate_params


scripts/install-filecoin-parameters.sh
```diff
@@ -10,5 +10,6 @@ generate_params() {
   RUST_LOG=info ./proofs/bin/paramcache --test-only
 }
 
-fetch_params
+# Disabling paramfetch until optimizations can be completed
```
laser: @ognots - `paramcache --test-only` will generate Groth parameters and verifying keys for 1024-byte sectors only. This means that, if this PR is merged, developers (i.e. someone running `deps`) will not be able to verify proofs generated over 256MiB sectors.

I strongly suggest that if you disable `fetch_params` that you remove the `--test-only` flag.

ognots: @laser I believe there was already a decision made to _omit_ the large proof parameters from the `deps` function, as they are not needed to run tests, which is the primary use case. @anorth @travisperson probably have more context on this.

For anyone wanting to prove large sector size, the Wiki's _Getting Started_ page will instruct users to invoke `paramcache` with no args, as you suggest.

ognots: also see https://github.com/filecoin-project/go-filecoin/pull/3375

travisperson: @ognots I think the issue is that we want verifying keys for large sectors as well, and paramcache does not support a way to generate just the verifying keys.

laser: @travisperson 

> [..] we want verifying keys for large sectors as well, and paramcache does not support a way to generate just the verifying keys

Yep, that's correct.

anorth:
For unit tests, we only need the test parameters. It would be good to have the test parameters thus always generated as this PR.

However, for nodes on a devnet, they need the real parameters/keys, and unless we want to fetch them (which we eventually have to b/c trusted setup) we should generate them. Some nodes will never join a devnet, so those nodes should not be forced to generate unwanted parameters.

If the go-filecoin node could abort immediately on startup if the keys are missing, we can push the generation (or fetch) decision to the user at the point that they wish to join a devnet, with instructions in the Wiki. This is similar to other requests for the sector builder to fail early if initialised without the required parameters. WDYT @laser?

@ognots could you pull some of the commentary I added in my similar PR #3381

anorth:
Closing in favour of #3404

PR #3386, opened 2019-09-10
use network name to namespace dht

### LINKED ISSUE #2509: Separate devnet DHTs

acruikshank:
closes #2509

### Problem

We use a DHT to discover peers on the network. The DHT is self healing by design, so it will attempt to merge with any other DHT of the same protocol that it finds. Unfortunately, this means that the DHTs of Filecoin nodes on different networks could end up merging and the nodes will end up trying to connect to each other (unsuccessfully since they'll have different genesis nodes).

### Solution

Append the network name to the DHT protocol to prevent merging.

### DIFF
--- a/net/protocols.go
+++ b/net/protocols.go
@@ -1,8 +1,12 @@
 package net
 
 import (
+	"fmt"
+
 	"github.com/libp2p/go-libp2p-core/protocol"
 )
 
-// FilecoinDHT is the protocol of the filecoin DHT.
-const FilecoinDHT protocol.ID = "/fil/kad/devnet-4"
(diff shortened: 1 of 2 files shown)


acruikshank:
@anorth since it was ready, I went ahead and merged this. I created a new, in progress issue (#3396) for the rest of the work.

PR #3393, opened 2019-09-11
Rename core package to message. Rename types to avoid redundancy.

anorth:
By request!

There are two commits here: the first moves things out of the `core` package that don't belong, the second has the renames.


message/testing.go
```diff
@@ -0,0 +1,114 @@
+package message
```
acruikshank: This appears to be a move from core/testing.go to message/testing.go but it's showing up as a new file. Did you use `git mv`? It would be nice if we could keep the history.

PR #3400, opened 2019-09-11
refactor: rename node.Config to node.Builder

icorderi:


node/builder.go
```diff
@@ -63,36 +62,36 @@ type Config struct {
...
-// ConfigOpt is a configuration option for a filecoin node.
-type ConfigOpt func(*Config) error
+// BuilderOpt is an option for building a filecoin node.
+type BuilderOpt func(*Builder) error
 
 // OfflineMode enables or disables offline mode.
-func OfflineMode(offlineMode bool) ConfigOpt {
-	return func(c *Config) error {
+func OfflineMode(offlineMode bool) BuilderOpt {
+	return func(c *Builder) error {
```
ingar: Nit: A little bit weird to have these receivers named `c` and `nc` now.

icorderi: Yea... I guess if you are gonna go for single char variable names might as well just call it all `x`..

node/message_propagate_test.go
```diff
@@ -37,7 +37,7 @@ func TestMessagePropagation(t *testing.T) {
 	// Initialize the first node to be the message sender.
 	senderNodeOpts := TestNodeOptions{
 		GenesisFunc: genesis,
-		ConfigOpts:  DefaultTestingConfig(),
+		BuilderOpts: DefaultTestingConfig(),
```
anorth: Can you rename `DefaultTestingConfig` in a follow-up?

PR #3402, opened 2019-09-11
Update localnet to start mining after miner create

travisperson:
Creating a miner with series.CreateMinerWithAsk no longer will start the mining scheduler with a call to `mining start`.

### DIFF
--- a/tools/fast/bin/localnet/main.go
+++ b/tools/fast/bin/localnet/main.go
@@ -287,6 +287,11 @@ func main() {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


travisperson:
Sure, we should discuss what you are looking for in localnet or something similar. Currently localnet is designed to mimic the current deployment process being executed on the nightly / staging / user devnets. This PR is just fixing localnet after it was broken when the behavior of `series.CreateMinerWithAsk` was changed.

PR #3404, opened 2019-09-11
feat(replace paramfetch with paramcache)

laser:
## Why does this PR exist?

The `rust-fil-proofs` submodule has been made redundant now that the `rust-fil-sector-builder` release contains the parameter-obtaining binary. Also, paramfetch isn't working as well as we need it to.

## What's in this PR?

1. Drop the `proofs/rust-fil-proofs` submodule
1. Modify `install-filecoin-parameters.sh` to always generate Groth parameters and keys (for 1KiB sectors)
1. Modify release-packaging script to include `paramcache` instead of `paramfetch` (and `parameters.json`)


anorth:
Thanks!

taoshengshi:
good work to remote redundant `rust-fil-proof` !  The proof interface is more clear for go programers but the down side is hide the `rust-fil-proof` behind the go-sector-build

PR #3405, opened 2019-09-11
FAST Devnet environment uses configuration object

travisperson:
Passing in predefined configuration values for external resources such as the genesis and faucet will allow for supporting a wider range of network deployments.


tools/fast/environment/environment_devnet.go
```diff
@@ -36,9 +36,30 @@ type Devnet struct {
...
+		return config, nil
+	}
+
+	return DevnetConfig{}, fmt.Errorf("failed to look up config for network %s", name)
+}
+
 // NewDevnet builds an environment that uses deployed infrastructure to
 // the kittyhawk devnets.
-func NewDevnet(network, location string) (Environment, error) {
+func NewDevnet(network DevnetConfig, location string) (Environment, error) {
```
frrist: ditto the above statement about calling this `config`

PR #3408, opened 2019-09-11
add peer tracker to porcelain with Trust method

frrist:
### What
This adds the PeerTracker to porcelain. We need this for catch up syncing daemon tests as we must instruct daemons to trust each other in order for them to sync their chains initially. Without this daemons will be unable to move out of the catch up sync phase.

### DIFF
--- a/net/peer_tracker.go
+++ b/net/peer_tracker.go
@@ -69,6 +69,14 @@ func (tracker *PeerTracker) UpdateTrusted(ctx context.Context) error {
 	return tracker.updatePeers(ctx, tracker.trustedPeers()...)
 }
 
+// Trust adds `pid` to the peer trackers trusted node set.
+func (tracker *PeerTracker) Trust(pid peer.ID) {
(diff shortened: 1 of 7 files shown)


net/peer_tracker.go
```diff
@@ -69,6 +69,14 @@ func (tracker *PeerTracker) UpdateTrusted(ctx context.Context) error {
 	return tracker.updatePeers(ctx, tracker.trustedPeers()...)
 }
 
+// Trust adds `pid` to the peer trackers trusted node set.
+func (tracker *PeerTracker) Trust(pid peer.ID) {
```
ZenGround0: Would it make sense to include an `Untrust` as well? If I were playing around with this on the CLI I wouldn't want to make a change I couldn't undo.

frrist:
Realizing this might now be ideal as the peers passed to peer tracker as "trusted" are not persisted to disk. Going to revisit at a later point.

PR #3409, opened 2019-09-11
Remove requirement for code checked out into old GOPATH structure

travisperson:
Resolves #3407

PR #3410, opened 2019-09-11
Allow setting path to binary under test

travisperson:
To enable running the network deployment tests during network setup we can build the tests as an executable binary using the `-c` flag of `go test`. Currently all of our tests are forced into using a binary located at the project root. This flags allows overriding the binary that will be used for tests which fork `go-filecoin`.

### DIFF
--- a/testhelpers/util.go
+++ b/testhelpers/util.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"sync"
 	"time"
+
+	"github.com/filecoin-project/go-filecoin/testhelpers/testflags"
 )
 
 // GetFreePort gets a free port from the kernel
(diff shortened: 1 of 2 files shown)


testhelpers/util.go
```diff
@@ -54,13 +56,17 @@ func MustGetFilecoinBinary() string {
...
-	if err != nil {
-		return "", err
+	bin, provided := testflags.BinaryPath()
+	if !provided {
+		gopath, err := GetGoPath()
+		if err != nil {
+			return "", err
+		}
+
+		bin = filepath.Join(gopath, "/src/github.com/filecoin-project/go-filecoin/go-filecoin")
```
anorth: I think we should delete the fallback to GOPATH, but instead expect "./go-filecoin". Yes, this requires the right CWD, but the `binary-path` option supports other cases.

travisperson: See: https://github.com/filecoin-project/go-filecoin/pull/3409

PR #3411, opened 2019-09-11
feat(actor): update go-hamt-ipld library to latest

hannahhoward:
# Goals

Update Actor code to the latest go-hamt-ipld library, which supports fast CBOR decode/encode and also supports alternate bitwidths on HAMT. 

# Implementation

- The main change from the previous version of go-ipld-hamt is the deferral of deserializing actual values stored in the HAMT. This means that when you lookup values, you need to pass a pointer to a value where raw CBOR is decoded into. Also, when iterating the values in a HAMT, it's neccesary to use a callback and a type so that deserialization happens properly. 
- go-hamt-ipld also now includes cbor-gen, which allows you to define fast-path deserialization/serialization for specific types. For the moment, it only applies to HAMT stuff but we could generate fast-path code for actor types as well.
- This does not actually alter the bitwidth for the HAMT fanout -- that
will be a future PR.

# For discussion

I've changed a couple methods that didn't have a lot of individual test coverage to begin with -- should that be added to get this merged?

See comment on lines below for the MakeGenisisFunc function -- I don't understand how this didn't fail the waiter_test in plumbing/msg before?


consensus/genesis.go
```diff
@@ -169,6 +169,9 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
 			if err = s.Commit(scid, a.Head); err != nil {
 				return nil, err
 			}
+			if err := st.SetActor(ctx, addr, a); err != nil {
```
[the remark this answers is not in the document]
acruikshank: Presumably the problem is that the actor's Head is not getting updated in the state tree without this line. If this is the case, is line 160 still necessary? If so, we need a comment explaining why we're setting the actor twice.

state/tree.go
```diff
@@ -150,38 +148,17 @@ func (t *tree) GetBuiltinActorCode(codePointer cid.Cid) (exec.ExecutableActor, e
...
 
-	var act actor.Actor
-	if err := hackTransferObject(data, &act); err != nil {
-		return nil, err
-	}
-
 	return &act, nil
 }
 
-func hackTransferObject(from, to interface{}) error {
```
acruikshank: 🎉

vm/storage.go
```diff
@@ -109,8 +111,14 @@ func (s Storage) Put(v interface{}) (cid.Cid, error) {
 	if blk, ok := v.(blocks.Block); ok {
 		// optimize putting blocks
 		nd, err = cbor.DecodeBlock(blk)
-	} else if bytes, ok := v.([]byte); ok {
-		nd, err = cbor.Decode(bytes, types.DefaultHashFunction, -1)
+	} else if raw, ok := v.([]byte); ok {
+		nd, err = cbor.Decode(raw, types.DefaultHashFunction, -1)
+	} else if cm, ok := v.(cbg.CBORMarshaler); ok {
```
acruikshank: Are we putting `CBORMarshaler`s into storage? Is this new functionality? If so, and we aren't yet using it, we should wait to introduce this until it is exercised by code (and tests).

frrist:
@hannahhoward I will defer to whatever AC & AN suggest here. My feedback is primarily nits: combine `if err :=`, no unnecessary import renames, and a slight preference for `select` over `else if` in `vm/storage.go`.

hannahhoward:
> @hannahhoward your editor seems to be inserting unnecessary package aliases. I'd appreciate you figuring out how to avoid that.

Hey so I finally tracked this down to not by editor but goimports, which does autoimporting, and specifically, this issue: https://github.com/golang/go/issues/28428 & https://go-review.googlesource.com/c/tools/+/152000

They actually chose to add this feature and I honestly understand their reasoning -- the thought is even if the underlying package name is correct, if there's a mismatch with the filename there's no way to determine that it is without digging into the underlying source code of the file, so having the rename allows the imports of the file to be validated just by looking at the file itself.

Anyway, I only say this cause goimports is a pretty standard go tool, though I guess I am the only one who uses it on this project. But I'll turn it off and just do manual imports.

vm/storage.go
hannahhoward: so a CBORMarshaler just means an object that conforms to that interface, meaning it has methods to fast path it's own serialization / deserialization -- this covers all hamt nodes, and in fact, they need to be fastpathed to encode properly. WrapObject won't work here because of the cbg.Deferred object. The actual solution is that go-ipld-cbor should check for fast path in its own calls to marshal/unmarshal, which there is an unmerged PR for: https://github.com/ipfs/go-ipld-cbor/pull/64 . Absent that getting merged, this is neccesary for a couple direct calls to storage.Put

hannahhoward: ^^ @acruikshank

hannahhoward: will add a comment to that effect.

PR #3412, opened 2019-09-11
Add BlockSig field to block

ZenGround0:
This is the bare minimum we need to avoid a data-structure based protocol upgrade post alpha net

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -52,6 +52,9 @@ type Block struct {
 	// The timestamp, in seconds since the Unix epoch, at which this block was created.
 	Timestamp Uint64 `json:"timestamp"`
 
+	// The signature of the miner's worker key over the block
+	BlockSig Signature
+
 	cachedCid cid.Cid
 
 	cachedBytes []byte


commands/schema/filecoin_block.schema.json
```diff
@@ -52,6 +52,12 @@
         },
         "timestamp": {
           "type": "string"
+        },
```
ZenGround0: I know this is going away soon but it was easier to add this than remove the whole thing to get CI green

anorth: Please do the honours for us in a different PR

types/block.go
```diff
@@ -52,6 +52,9 @@ type Block struct {
 	// The timestamp, in seconds since the Unix epoch, at which this block was created.
 	Timestamp Uint64 `json:"timestamp"`
 
+	// The signature of the miner's worker key over the block
+	BlockSig Signature `json:"blocksig"`
```
anorth: I'll try to get the spec changed to avoid this clunky name.

PR #3413, opened 2019-09-12
enforce rational post timing

acruikshank:
### Problem

The new rational PoSt construction requires a narrow window around when a miner can submit a PoSt. The start of the window is defined by the availability of chain randomness from the start of the challenge time (proving period end minus the PoSt challenge time), and the end is still the predetermined proving period end. If a miner submits a late PoSt, they must submit it between the next challenge time and the next proving period end.

### Solution

This PR implements the construction described above. In particular it:

1. Defines POST_CHALLENGE_TIME. The correct value for this is not currently understood, so it is set to half the proving period as a safe start.
2. Retrieves the challenge seed at either provingPeriodEnd-POST_CHALLENGE_TIME or nextProvingPeriodEnd-POST_CHALLENGE_TIME depending on whether the PoSt is late. This will fail if the tipset at that block height is not yet available.
3. Redefines the proving period start to be provingPeriodEnd-POST_CHALLENGE_TIME. This prevents the miner from submitting a PoSt until the challenge seed is ready.


actor/builtin/miner/miner.go
```diff
@@ -863,18 +858,20 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, fault
...
 
-		// Calcuate any penalties for late submission
-		generationAttackGracePeriod := GenerationAttackTime(state.SectorSize)
-		if chainHeight.GreaterThan(state.ProvingPeriodEnd.Add(generationAttackGracePeriod)) {
-			// The PoSt has been submitted after the generation attack time.
+		provingPeriodDuration := types.NewBlockHeight(ProvingPeriodDuration(state.SectorSize))
+		nextProvingPeriodEnd := state.ProvingPeriodEnd.Add(provingPeriodDuration)
+
+		// ensure PoSt is not too late entirely
+		if chainHeight.GreaterEqual(state.ProvingPeriodEnd.Add(nextProvingPeriodEnd)) {
```
anorth:
```suggestion
		if chainHeight.GreaterEqual(nextProvingPeriodEnd) {
```

actor/builtin/miner/miner.go
```diff
@@ -905,9 +902,22 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, fault
...
-				return nil, errors.RevertErrorWrap(err, "failed to sample chain for challenge seed")
+			// calculate challenge seed now. If PoSt is too early, this will fail.
+			poStChallengeTime := types.NewBlockHeight(PoStChallengeTimeBlocks)
+
+			var seed types.PoStChallengeSeed
+			if chainHeight.LessThan(state.ProvingPeriodEnd) {
+				// proof is in time, sample at start of post challenge time
+				seed, err = getPoStChallengeSeed(ctx, state, state.ProvingPeriodEnd.Sub(poStChallengeTime))
+				if err != nil {
+					return nil, errors.RevertErrorWrap(err, "failed to sample chain for challenge seed, PoSt is probably early")
```
anorth: Include the window open height, window end, and current chain height in the message.

actor/builtin/miner/miner.go
```diff
@@ -905,9 +902,22 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, fault
...
+				}
+			} else {
+				// proof is late, sample at start of next post challenge time
+				seed, err = getPoStChallengeSeed(ctx, state, nextProvingPeriodEnd.Sub(poStChallengeTime))
+				if err != nil {
+					return nil, errors.RevertErrorWrap(err, "failed to sample chain for challenge seed, PoSt is probably early")
```
anorth: Ditto include all the relevant heights instead of the "probably..."

actor/builtin/miner/miner.go
```diff
@@ -34,15 +34,10 @@ func init() {
...
-//
-// TODO: If the following PR is merged - and the network doesn't define a
-// largest sector size - this constant and consensus.AncestorRoundsNeeded will
-// need to be reconsidered.
-// https://github.com/filecoin-project/specs/pull/318
-const LargestSectorGenerationAttackThresholdBlocks = 100
+// PoStChallengeTimeBlocks defines the blocks time prior to the proving
+// period end at which the PoSt challenge seed is chosen. This dictates the
+// earliest point at which a PoSt may be submitted.
+const PoStChallengeTimeBlocks = 500
```
anorth: I think `PoStChallengePeriodBlocks` is slightly better. Or "window" instead of "period".

commands/mining_daemon_test.go
```diff
@@ -90,17 +93,13 @@ func TestMiningAddPieceAndSealNow(t *testing.T) {
...
-	series.CtxMiningNext(ctx, 2)
-
 	// start sealing
 	err = minerNode.SealNow(ctx)
 	require.NoError(t, err)
 
 	// We know the miner has sealed and committed a sector if their power increases on chain.
-	// Wait up to 3 minutes for that to happen.
-	for i := 0; i < 180; i++ {
+	// Wait up to 5 minutes for that to happen.
```
anorth: "minutes" looks incorrect, the sleep below is 1 second. "rounds" might be less brittle.

PR #3414, opened 2019-09-12
Reduce functional test parallelism to 2. Hold Codecov reports until all builds complete.

anorth:
We're still seeing some coverage reports that appear incomplete. My hypothesis is that some splits of the functional tests sometimes complete ahead of integration tests, and the coverage is reported too early.

E.g. https://codecov.io/gh/filecoin-project/go-filecoin/pull/3410/changes, https://circleci.com/workflow-run/f43cc8b9-84aa-4dcf-ac45-1e18c484f821

Functional test parallelism was too high: one of the containers ran zero tests.


anorth:
The coverage reported above does not inspire confidence. But I'm not sure if we need the changes to be on the base branch, master, to make the difference? Anyway, this is less wrong.

Issue #3293, opened 2019-08-26
Sign blocks with miner worker key


frrist:
### Description
As defined in the [data-structures](https://github.com/filecoin-project/specs/blob/master/data-structures.md#block) portion of the spec, blocks should contain a signature calculated over the hash of the entire block using the miners worker key to ensure that the block is not tampered with after creation.

See https://github.com/filecoin-project/specs/issues/457

### Acceptance criteria
- blocks are signed using the miners worker key. The signature is calculated over the hash of the entire block.

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/master/types/block.go

anorth:
FYI @ZenGround0 this is probably part of your upcoming work.

anorth:
Done

Issue #1609, opened 2019-01-15
Verifying a block verifies the blocks signature


frrist:
### Description
Requires getting the public keys for each message in the block.
Details to be added when they surface.

### Acceptance criteria
TODO

shannonwells:
Related to #1010 Signature validation in runStateTransition, blocked by needing signer public key in block (#1054 part 3)

Issue #1010, opened 2018-09-28
Signature Validation in consensus.RunStateTransition


ZenGround0:
### Description

We still don't validate the signature on the block using the miner's public key during block processing.

I think that because of the weirdness of our signature scheme we don't need to check message signatures, but this should be explicitly evaluated and stated in the code.  If this assumption is wrong we need to check those signatures too.

mishmosh:
@shannonwells is this ready to be closed?

PR #3419, opened 2019-09-12
Feat/validate block sig

ZenGround0:
Miners sign blocks and validation includes signature verification.

This Closes out various issues.

https://github.com/filecoin-project/go-filecoin/issues/3293
https://github.com/filecoin-project/go-filecoin/issues/1609
https://github.com/filecoin-project/go-filecoin/issues/1010

Note that previous issues describing this work + inline TODOs dictated performing the signature validation check as part of syntactic block validation but this PR puts the validation within the `validateMining` call of the expected consensus module.  This is necessary in our current validation architecture because the miner worker address is tracked in the state machine and header (syntactic) validation happens at header receive time which comes before there is any guarantee of being able to read the parent state.

There are tons of directions we could (and in some cases we should) go with our validation architecture to validate data more aggressively.  For example we could opportunistically check if we have the parent state and if so filter blocks by (1) miner's membership in the power table (2) valid signature.

### DIFF
--- a/chain/syncer_integration_test.go
+++ b/chain/syncer_integration_test.go
@@ -125,12 +125,6 @@ func TestTipSetWeightDeep(t *testing.T) {
 
 	ctx := context.Background()
 
-	mockSigner, ki := types.NewMockSignersAndKeyInfo(3)
-	minerWorker1, err := ki[0].Address()
-	require.NoError(t, err)
-	minerWorker2, err := ki[1].Address()
-	require.NoError(t, err)
-
 	// set up dstP.genesis block with power
 	genCfg := &gengen.GenesisCfg{
 		ProofsMode: types.TestProofsMode,
@@ -159,6 +153,12 @@ func TestTipSetWeightDeep(t *testing.T) {
 	info, err := gengen.GenGen(ctx, genCfg, cst, bs, 0)
 	require.NoError(t, err)
 
+	minerWorker1, err := info.Keys[0].Address()
+	require.NoError(t, err)
+	minerWorker2 := minerWorker1
+	// Include gengen miner worker key so that we can correctly sign blocks
+	mockSigner := types.NewMockSigner([]types.KeyInfo{*info.Keys[0]})
+
 	var calcGenBlk types.Block
 	require.NoError(t, cst.Get(ctx, info.GenesisCid, &calcGenBlk))
 
@@ -168,8 +168,6 @@ func TestTipSetWeightDeep(t *testing.T) {
 	require.NoError(t, err)
 	emptyReceiptsCid, err := messageStore.StoreReceipts(ctx, []*types.MessageReceipt{})
 	require.NoError(t, err)
-
-	con := consensus.NewExpected(cst, bs, th.NewTestProcessor(), th.NewFakeBlockValidator(), &th.TestView{}, calcGenBlk.Cid(), th.BlockTimeTest, &consensus.FakeElectionMachine{}, &consensus.FakeTicketMachine{})
 
 	// Initialize stores to contain dstP.genesis block and state
 	calcGenTS := th.RequireNewTipSet(t, &calcGenBlk)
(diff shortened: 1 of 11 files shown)


types/block_test.go
```diff
@@ -229,3 +233,164 @@ func TestBlockJsonMarshal(t *testing.T) {
...
+		
+		cpy := b.Miner
+		defer func(){ b.Miner = cpy}()
+		
+		b.Miner = diff.Miner
+		after := b.DataToSign()
+		assert.False(t, bytes.Equal(before, after))
+	}()
+
+	func() {
```
ZenGround0: It felt nice to separate each case into a function call and use defer to reset data each time.  Might be more verbose than necessary 🤷‍♂

types/block.go
```diff
@@ -142,3 +142,23 @@ func (b *Block) Score() uint64 {
 func (b *Block) Equals(other *Block) bool {
 	return b.Cid().Equals(other.Cid())
 }
+
+// DataToSign returns the block's bytes without the blocksig for signature
+// creating and verification
+func (b *Block) DataToSign() []byte {
```
frrist: nit: Since we also pass this to verification methods, maybe call this `SignatureData`? A similar pattern is used for payment vouchers.

types/block_test.go
frrist: I think you could also use `t.Run()`

types/block.go
frrist: Alternatively, something like the below would also be nice. 
```go
func (b *Block) Sign(address address.Address, signer Signer) error {}
func (b *Block) ValidSignature(addr address.Address, verifier Recoverer) bool {}
```

frrist:
tests coverage results seem wrong (or concerning) looks like a fluke?

chain/syncer_integration_test.go
```diff
@@ -159,6 +153,15 @@ func TestTipSetWeightDeep(t *testing.T) {
...
 	require.NoError(t, err)
 
+	// Include gengen miner worker key so that we can correctly sign blocks
+	mockSigner, _ := types.NewMockSignersAndKeyInfo(1)
+	minerWorker1, err := info.Keys[0].Address()
+	require.NoError(t, err)
+	minerWorker2 := minerWorker1
+	mockSigner.Addresses = append(mockSigner.Addresses, minerWorker1)
+	mockSigner.PubKeys = append(mockSigner.PubKeys, info.Keys[0].PublicKey())
+	mockSigner.AddrKeyInfo[minerWorker1] = *(info.Keys[0])
```
anorth: Add a method to `MockSigner` that does these three things in one.

consensus/block_validation.go
```diff
@@ -59,6 +59,8 @@ func NewDefaultBlockValidator(blkTime time.Duration, c clock.Clock) *DefaultBloc
 
 // ValidateSemantic validates a block is correctly derived from its parent.
 func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *types.Block, parents *types.TipSet) error {
+	// Validate signature
```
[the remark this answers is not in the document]
anorth: This is where I would expect the signature validation to go (along with the rest of `validateMining`, if it depends on some information outside of the block header itself. Additional parameters or other re-organisation to make that possible is fine.

consensus/expected_test.go
```diff
@@ -164,6 +175,32 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
...
 	})
+
+	t.Run("returns nil + mining error when signature is invalid", func(t *testing.T) {
+		pTipSet := types.RequireNewTipSet(t, genesisBlock)
+		stateTree, err := state.LoadStateTree(ctx, cistore, genesisBlock.StateRoot, builtin.Actors)
+		require.NoError(t, err)
+		vms := vm.NewStorageMap(bstore)
+
+		blocks, minerToWorker := requireMakeBlocks(ctx, t, pTipSet, stateTree, vms)
+		blocks[0].BlockSig = blocks[1].BlockSig
```
anorth: Add a comment highlighting this line - it took me ages to find it.

types/block.go
anorth: Prefer the former: `Block` shouldn't depend on `Signer` and `Recoverer`.

consensus/block_validation.go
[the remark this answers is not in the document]
ZenGround0: @anorth take a look at the [opening comment](https://github.com/filecoin-project/go-filecoin/pull/3419#issue-317115086) for a bit of discussion on this particular issue.  Sounds like we are on the same page.

Issue #3417, opened 2019-09-12
De-obscure election tests 


ZenGround0:
### Description
Get rid of opaque hex strings and show how we generate those tickets in test.
https://github.com/filecoin-project/go-filecoin/pull/3287#discussion_r323047866

PR #3424, opened 2019-09-13
Generate one ticket per null block

ZenGround0:
Closing out #2223


ZenGround0:
cc @anorth this is about halfway from where I want it (missing randomness sampling updates and of course tests).  Depending on how much further I get we might consider adding this into 0.5.

ZenGround0:
Testing TODO
- [ ] mining scheduler runs for the exact number of null blocks when passed a seed ticket constructed with new `SeedWinnerInNRounds` util
- [ ] test for new min ticket behavior in tipset with multiple tickets in array
- [ ] test that tipset creation fails with inconsistent ticket array lengths
- [ ] validation fails if ticket array length doesn't match height
- [ ] validation fails if ticket inside array is not generated from parent ticket inside array
- [ ] validation fails with inconsistent ticket arrays

consensus/testing.go
```diff
@@ -142,3 +152,63 @@ func MakeFakeElectionProofForTest() []byte {
...
+//
+// Note that this is a deterministic function of the inputs as we return the
+// first ticket that seeds a winner in N rounds given the inputs starting from
+// MakeFakeTicketForTest().
+//
+// Note that there are no guarantees that this function will terminate on new
+// inputs as miner power might be so low that winning a ticket is very
+// unlikely.  However runtime is deterministic so if it runs fast once on
+// given inputs is safe to use in tests.
+func SeedFirstWinnerInNRounds(t *testing.T, n int, ki *types.KeyInfo, minerPower, totalPower uint64) types.Ticket {
```
ZenGround0: Note: this is going to be used to address the election test tech debt in #3417.

mining/scheduler_test.go
```diff
@@ -34,17 +41,83 @@ func TestMineOnce(t *testing.T) {
...
+
+// TestMineOnce10Null calls mine once off of a base tipset with a ticket that
+// will only win 10 rounds and verifies that the output has 10 tickets and a
+// +10 height.
+func TestMineOnce10Null(t *testing.T) {
+	mockSigner, kis := types.NewMockSignersAndKeyInfo(5)
+	ki := &(kis[0])
+	addr, err := ki.Address()
+	require.NoError(t, err)
+	baseTicket := consensus.SeedFirstWinnerInNRounds(t, 10, ki, 1, 10)
```
anorth: Use larger numbers for the miner/total power so these parameters are less opaque to read.

mining/scheduler_test.go
```diff
@@ -34,17 +41,83 @@ func TestMineOnce(t *testing.T) {
...
+}
+
+func (fwp *fakeWorkerPorcelain) BlockTime() time.Duration {
+	return time.Millisecond
+}
+
+// TestMineOnce10Null calls mine once off of a base tipset with a ticket that
+// will only win 10 rounds and verifies that the output has 10 tickets and a
+// +10 height.
+func TestMineOnce10Null(t *testing.T) {
```
anorth: Set the test flags. I'm torn, but since it uses `Sleep` rather than a fake clock, it's probably an integration test.

types/tipset.go
```diff
@@ -120,11 +124,11 @@ func (ts TipSet) MinTicket() (Ticket, error) {
 	if len(ts.blocks) == 0 {
 		return Ticket{}, errUndefTipSet
 	}
-	// TODO #2223 fix this to properly handle multiple tickts per block
-	if len(ts.blocks[0].Tickets) == 0 {
+	numTickets := len(ts.blocks[0].Tickets)
+	if numTickets == 0 {
```
anorth: We can now catch this in the constructor and prevent the creation of such a tipset (though that doesn't win us much since there's still an error return value here).

ZenGround0:
> I am wary about including such a significant change in the 0.5 release, but defer the decision to you (merge if you want).

Me too but I would prefer to debug issues live sooner over later and charge ahead.

> How will bugs in this system appear in a running network? Will they be easy to diagnose?

Hard to say what exactly they will look like.  My instinct is that they will be easier to deal with than the worst kind of bugs we see so far (like interaction between proof submission and forking) because the problems will show up directly during mining and validating blocks instead of popping up in the state machine / protocol processes.  Of course I can't be sure what these bugs will look like as I have written the code believing there are no such bugs.

consensus/expected.go
```diff
@@ -83,7 +83,7 @@ type Processor interface {
 
 // TicketValidator validates that an input ticket is valid.
 type TicketValidator interface {
-	ValidateTicket(types.Ticket, types.Ticket, address.Address, uint64) (bool, error)
+	ValidateTicket(types.Ticket, types.Ticket, address.Address) (bool, error)
```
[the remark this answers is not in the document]
ZenGround0: This would violate the convention set in this file through the `Processor` interface so I am hesitant.  We really need a project wide convention so that we can do the right thing without thinking about it.

If you declare this convention here and now across the whole project I'll make this change, spread the word and open a tech debt issue for bringing code in line with this convention.

types/tipset.go
```diff
@@ -53,6 +54,9 @@ func NewTipSet(blocks ...*Block) (TipSet, error) {
 			if blk.Height != height {
 				return UndefTipSet, errors.Errorf("Inconsistent block heights %d and %d", height, blk.Height)
 			}
+			if len(blk.Tickets) != numTickets {
+				return UndefTipSet, errors.Errorf("Inconsistent ticket array length %d and %d", len(blk.Tickets), numTickets)
```
frrist: I think you might have what's being printed here flipped, should print the info about the block last, similar to the other errors around this.

types/tipset.go
frrist: Strong preference for the former, reasoning about the errors these methods return is easier (imo) when there is only one type of error.

ZenGround0: SGTM.  Since we already check for this in the constructor I'll just kill this extraneous check

mining/scheduler_test.go
ZenGround0: I'm not sure I understand why we would add a clock to the scheduler.  Right now the worker does all of the waiting around and the scheduler helplessly blocks on serial worker calls.  I'm guessing I'm missing something, please fill me in if you see this.

mining/testing.go
```diff
@@ -42,37 +42,37 @@ func (s *MockScheduler) IsStarted() bool {
...
+func (w *TestWorker) Mine(ctx context.Context, ts types.TipSet, ticketArray []types.Ticket, outCh chan<- Output) (bool, types.Ticket) {
 	if w.WorkFunc == nil {
 		panic("must set MutableTestWorker's WorkFunc before calling Work")
 	}
-	return w.WorkFunc(ctx, ts, nullBlkCount, outCh)
+	return w.WorkFunc(ctx, ts, ticketArray, outCh)
 }
 
-// NewTestWorkerWithDeps creates a worker that calls the provided input
+// NewTestWorkerWithDeps crates a worker that calls the provided input
```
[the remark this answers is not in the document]
ZenGround0: 🗃

mining/worker.go
```diff
@@ -211,45 +218,46 @@ func (w *DefaultWorker) Mine(ctx context.Context, base types.TipSet, nullBlkCoun
 
 	select {
 	case <-ctx.Done():
-		log.Infof("Mining run on base %s with %d null blocks canceled.", base.String(), nullBlkCount)
-		return false
+		log.Infof("Mining run on base %s with %d null blocks canceled.", base.String(), len(ticketArray))
+		return
 	case <-done:
 	case err := <-errCh:
 		log.Infof("Error notarizing time on ticket")
```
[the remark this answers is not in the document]
ZenGround0: It was just a side effect of copy pasta though if we ever do add a VDF it will be good to log the error

mining/scheduler_test.go
frrist: We would replace this sleep call in the scheduler with a call to our own clock: https://github.com/filecoin-project/go-filecoin/blob/bcbd054bdf113a5ade5652c51a2d9249a0295983/mining/scheduler.go#L108-L109

When writing a unit test we can swap it out for a fake clock and avoid depending on time.

mining/worker.go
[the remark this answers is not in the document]
frrist: In that case, would you mind preemptively logging the error message in this change set?

mining/scheduler_test.go
ZenGround0: Aha! I missed that sleep.  All clear thx.

Issue #3430, opened 2019-09-16
actor get by address


GXKe:
### Description
go-filecoin actor ls too much output data. We need a CLI command for get out a actor by address.

### Acceptance criteria
go-filecoin actor get address

Issue #1587, opened 2019-01-10
go-filecoin show and go-filecoin chain should be combined


rosalinekarr:
### Description

The following commands have similar purposes but different and confusing names:

```
go-filecoin show block
go-filecoin chain ls
go-filecoin chain head
```

We should reduce them into the following commands and add support for verbose, human-readable output to all of them:

```
go-filecoin chain ls [optional cids for showing specific blocks]
go-filecoin chain head
```

### Acceptance criteria
- wiki must be updated to replace any instances of the old commands
- readme must be scrubbed for any instances of the old commands

deaswang:
How about update like this:

`go-filecoin chain block [-m] <CID>`
because `show block` and `chain ls` call different Porcelain APIs, different logic process.

add optional `-b, --begin num` and `-e, --end num` to control the block height for `go-filecoin chain ls`, without end mean from head tipset, without begin mean to genesis.

deaswang:
how to commit Wiki repo? how about submit one patch file at here?

deaswang:
@mishmosh how is the proposal?

deaswang:
here is WIki patch file
[0001-update-wiki-for-combine-chain-and-show-1587.zip](https://github.com/filecoin-project/go-filecoin/files/3421400/0001-update-wiki-for-combine-chain-and-show-1587.zip)

eshon:
We were thinking of moving the go-filecoin wiki to [filecoin-project/docs/](https://github.com/filecoin-project/docs/tree/master/go-filecoin-tutorial) to enable community PRs but not sure if this has been discussed with the go-filecoin dev team yet.

Also this update would depend on merging the PR and adding it to a later release. We can apply the wiki patch later on if that happens.

deaswang:
#3196 
Should this issue necessary now?
@rosalinekarr @mishmosh

anorth:
Yes, this issue remains to clean up the organisation of the CLI actions, such that they read like `chain show`, `messages show` rather than `show block`. This will require moving much of that code around.

Maybe something like:
- `chain show [--full] <cid>...`: prints block headers / full blocks
- `chain ls`: prints block header cids for the whole chain
- `chain head`: prints head tipset CIDs

I don't think I've made a complete proposal - I haven't addresses messages/receipts.

Regarding your proposal for arguments, I think we should defer that to a separate PR from this reorganisation, but I like the idea in general.

PR #3431, opened 2019-09-16
add command: actor get by address

GXKe:
#3430


commands/actor.go
```diff
@@ -107,6 +110,62 @@ var actorLsCmd = &cmds.Command{
...
+		result, err := GetPorcelainAPI(env).ActorGet(req.Context, addr)
+		if err != nil {
+			return err
+		}
+
+		addrStr := addr.String()
+		var output *ActorView
+		switch {
+		case result.Empty(): // empty (balance only) actors have no Code.
+			output = makeActorView(result, addrStr, nil)
```
ingar: As this code is essentially the same as the code for `actor ls`, let's move it to a private helper for forward-looking maintainability.

commands/actor.go
```diff
@@ -107,6 +110,62 @@ var actorLsCmd = &cmds.Command{
...
+		}
+		if err := re.Emit(output); err != nil {
+			return err
+		}
+		return nil
+	},
+	Type: &ActorView{},
+	Encoders: cmds.EncoderMap{
+		cmds.JSON: cmds.MakeTypedEncoder(func(req *cmds.Request, w io.Writer, a *ActorView) error {
```
ingar: Same thing here for this encoder, as it's the same as for `actor ls`

commands/actor.go
GXKe: go-filecoin actor ls --address <actor address>
good idea

GXKe:
> Thank you for the PR.
> 
> Couple of things:
> 
> * I have a couple of comments about reducing the duplication in the file.
> * If there ends up being two actor commands, `ls` and `get`, some `Helptext` would be nice for each command.
> * Along those lines, this could be written with an argument to `ls`, as in `go-filecoin actor ls --address <actor address>`.  I'm unclear if it would be cleaner or not.
> * Tests need to be added to `commands/actor_daemon_test.go`

Has been resolved

frrist:
Thanks @GXKe! @ingar and @ZenGround0 what do you think of putting this on the `show` command? It would then look similar to:
```
USAGE
  go-filecoin show - Get human-readable representations of filecoin objects

SYNOPSIS
  go-filecoin show

SUBCOMMANDS
  go-filecoin show block <cid>    - Show a full filecoin block by its header CID
  go-filecoin show header <cid>   - Show a filecoin block header by its CID
  go-filecoin show messages <cid> - Show a filecoin message collection by its CID
  go-filecoin show receipts <cid> - Show a filecoin receipt collection by its CID
  go-filecoin show actor <address> - Show a filecoin actor by its Address.
```

commands/actor.go
```diff
@@ -52,10 +55,35 @@ var actorCmd = &cmds.Command{
...
-		if err != nil {
-			return err
+		addrString, _ := req.Options["address"].(string)
+		var results <-chan state.GetAllActorsResult
+		if len(addrString) > 0 {
+			addr, err := address.NewFromString(addrString)
+			if err != nil {
+				return err
+			}
+			act, err := GetPorcelainAPI(env).ActorGet(req.Context, addr)
```
frrist: My preference would be to make this a separate command, either on `actor` or `show` as stated elsewhere. I don't think we need to be using a channel for this and adding it as a separate command would clean up that logic.

anorth:
`show` should be disbanded as a top-level command. Don't put more stuff under it. See #1587

ZenGround0:
This may be worth using in the future after scoping CLI redesign work and checking that it is in scope.  Closing for now.

Issue #3433, opened 2019-09-16
Complete consensus protocol: randomness is sampled per spec


ZenGround0:
### Description
This is a spin off of #2223 which was a bit too overscoped for implementation.  This story handles the randomness sampling upgrade.  At a high level instead of sampling one unit of randomness from one tipset the protocol now calls for sampling one unit of randomness from one round.  This means breaking open the ticket array of the minimum ticket block in the tipset.

### Acceptance criteria
Move to new randomness sampling procedure. 
- create ticket iterator
- update ancestor sampling to only take randomness, not tipsets

### Risks + pitfalls
-- correctly identifying a ticket that falls in the interior of a tipset's ticket array (null round ticket)

### Where to begin
make chain/traversal.go ticket iterator
get_ancestors gets ticket chain in syncer and block generator

PR #3437, opened 2019-09-16
Feat/3433 prep randomness sampling to current spec: ticket iterator

ZenGround0:
This PR updates randomness sampling to make use of all ticket array values per the spec.  Closes out #3433, don't review until #2223 is merged.

This is a protocol breaking change.  If we can get it in for 0.5.x awesome, if not we need to make an upgrade.

### DIFF
--- a/chain/traversal.go
+++ b/chain/traversal.go
@@ -2,6 +2,7 @@ package chain
 
 import (
 	"context"
+	"errors"
 
 	"github.com/ipfs/go-cid"
 
@@ -87,7 +88,7 @@ func (it *TicketIterator) Complete() bool {
 // https://github.com/filecoin-project/specs/blob/master/definitions.md#ticket-chain
 func (it *TicketIterator) Next() error {
 	if it.tips.Complete() {
-		return errors.NewError("calling next on complete ticket iterator")
+		return errors.New("calling next on complete ticket iterator")
 	}
 
 	// Advance internal tipset iterator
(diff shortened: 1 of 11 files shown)


chain/traversal.go
```diff
@@ -53,6 +53,63 @@ func (it *TipsetIterator) Next() error {
...
+func (it *TicketIterator) Complete() bool {
+	return it.tips.Complete()
+}
+
+// Next advances the iterator to the next ticket. Tickets are taken from the
+// block with the minimum final ticket in its ticket array
+// https://github.com/filecoin-project/specs/blob/master/definitions.md#ticket-chain
+func (it *TicketIterator) Next() error {
+	if it.tips.Complete() {
+		return nil
```
anorth: Probably a bug so return an error.

chain/traversal.go
```diff
@@ -53,6 +53,63 @@ func (it *TipsetIterator) Next() error {
...
+	}
+
+	// Advance internal tipset iterator
+	if it.idx == 0 {
+		if err := it.tips.Next(); err != nil {
```
anorth: After `Next` you must check `Complete`.

chain/traversal.go
```diff
@@ -53,6 +53,63 @@ func (it *TipsetIterator) Next() error {
...
+			return err
+		}
+		it.idx = len(it.tips.Value().At(0).Tickets) - 1
+		if it.tips.Value().Equals(types.UndefTipSet) {
```
anorth: The line above will panic if this comparison is true.

ZenGround0:
This is low priority to merge because it won't make it into the node's live code path until we launch an associated protocol change.  It is also possible that ticket sampling is going to change significantly in the spec between now and then in which case we'll get rid of this anyway.

I'm not including the integration step here because that should be done separately with a protocol breaking change.  I will file a quick follow on draft PR demonstrating how that will look.

anorth:
@zen if the PR as is constitutes a protocol change, please close it for now. Please do put up any parts of it that we can land without a protocol upgrade.

PR #3438, opened 2019-09-17
Change HAMT bitwidth to 5 for both global state tree and actor state

hannahhoward:
# Goals

Lower the fanout on our HAMTs to increase performance of state transitions

# Implementation

- Simply use an alternate bitwidth for when creating and loading nodes in our HAMT.
- Based off testing of HAMT performance in https://github.com/ipfs/go-hamt-ipld/pull/33 - bitwidth 4 (fanout = 16) shows  very strong performance. Bitwidth 3 shows slightly stronger performance but approaches diminishing returns and my anxiety is lower than is commonly used in creating HAMTS
- Based of @acruikshank 's testing of Payment Channels however, bitwidth 5 shows the strongest performance in real world tests, so that is what we are using until we have better HAMT tests.

# For Discussion

This is a PROTOCOL BREAKING CHANGE because the difference in storage means there will be different state CIDs. We won't be able to change it again after 0.5 without breaking protocol, so it's not clear whether we want to do more extensive testing before we throw it in.

### DIFF
--- a/actor/storage.go
+++ b/actor/storage.go
@@ -17,7 +17,7 @@ import (
 const (
 	// TreeBitWidth is the bit width of the HAMT used to by an actor to
 	// store its state
-	TreeBitWidth = 4
+	TreeBitWidth = 5
 )
 
 // MarshalStorage encodes the passed in data into bytes.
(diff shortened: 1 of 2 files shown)


whyrusleeping:
SGTM, interesting to see 4 show better perf than 5. Looking forward to seeing the followup chain benchmarks

PR #3441, opened 2019-09-17
do not slash miners with the same worker address

### LINKED ISSUE #3439: Prevent storage fault slasher from slashing self
### Description
The storage fault slasher will produce and mine a message slashing the miner actor for the node on which it runs. While not incorrect, it's not rational. It also leads to a situation where the only or last mining node on a network can slash itself, leading to total loss of power and network halt. If a miner refrained from slashing itself, it would at least require another node with power in order to mine a block slashing the faulty one.

See also #3436. This issue might be sufficient and less invasive to implement than more special-casing for the bootstrap though.

### Acceptance criteria
- Storage fault slasher does not slash miners with the same worker address (or a whitelist of worker addresses)

frrist:
- closes #3439


protocol/storage/fault_slasher.go
```diff
@@ -122,11 +122,17 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
...
 			continue
 		}
+
 		minerAddr, err := address.NewFromString(minerStr)
 		if err != nil {
 			return errors.Wrap(err, "could not create minerAddr string")
 		}
 
+		if minerAddr == workerAddr {
+			sfm.log.Warning("skip slashing own worker address")
```
anorth: Include both the miner and worker addresses in the message.

protocol/storage/fault_slasher_test.go
```diff
@@ -143,6 +143,30 @@ func TestFaultSlasher_Slash(t *testing.T) {
 		assert.Equal(t, 2, ob.msgCount) // A new slashing message.
 	})
 
+	t.Run("does not slash own worker address", func(t *testing.T) {
+		getf := address.NewForTestGetter()
```
anorth: Please improve this whole test file by distinguishing the miner actor from account actor (worker) addresses much more clearly. The miner actor addresses should not come from a signer, while the worker generally should. Change the `slasherPlumbing` to actually check the parameter to `MinerGetWorkerAddress`.

protocol/storage/fault_slasher_test.go
```diff
@@ -143,6 +143,30 @@ func TestFaultSlasher_Slash(t *testing.T) {
...
+			addr1.String(): miner.PoStStateUnrecoverable,
+		})
+		require.NoError(t, err)
+
+		queryer := makeQueryer([][]byte{data1})
+		ob := outbox{}
+		plumbing := slasherPlumbing{
+			Queryer:    queryer,
+			minerAddr:  signer.Addresses[0],
+			workerAddr: addr1,
```
anorth: These two addresses are the wrong way round (after fixing the code).

protocol/storage/fault_slasher.go
```diff
@@ -122,20 +122,26 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
...
 		}
-		minerAddr, err := address.NewFromString(minerStr)
+
+		laterMinerActorAddr, err := address.NewFromString(lateMinerActor)
 		if err != nil {
 			return errors.Wrap(err, "could not create minerAddr string")
 		}
 
+		if laterMinerActorAddr == myMinerActorAddr {
+			sfm.log.Warningf("skip slashing self %s", laterMinerActorAddr.String())
```
anorth: `.String()` is unnecessary when formatted with `%s`

protocol/storage/fault_slasher_test.go
```diff
@@ -62,11 +67,15 @@ func TestFaultSlasher_Slash(t *testing.T) {
 		require.NoError(t, err)
 		queryer := makeQueryer([][]byte{data})
 
+		getf := address.NewForTestGetter()
+		ownMiner := getf()
+		ownWorker := signer.Addresses[0]
```
anorth: It's not necessary to redeclare these in each `Run`.

protocol/storage/fault_slasher.go
```diff
@@ -108,8 +108,8 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
 	sfm.log.Debugf("there are %d late miners\n", len(*lms))
 
 	// Slash late miners.
-	for minerStr, state := range *lms {
-		if _, ok := sfm.slashed[minerStr]; ok {
+	for lateMinerActor, state := range *lms {
+		if _, ok := sfm.slashed[lateMinerActor]; ok {
```
ZenGround0: nit: renaming ok to `alreadySlashed` or something like that would make this already clear code crystal clear.

protocol/storage/fault_slasher.go
```diff
@@ -108,8 +108,8 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
 	sfm.log.Debugf("there are %d late miners\n", len(*lms))
 
 	// Slash late miners.
-	for minerStr, state := range *lms {
-		if _, ok := sfm.slashed[minerStr]; ok {
+	for lateMinerActor, state := range *lms {
```
ZenGround0: Great renames here.

protocol/storage/fault_slasher.go
```diff
@@ -122,20 +122,26 @@ func (sfm *FaultSlasher) Slash(ctx context.Context, currentHeight *types.BlockHe
 			// Alternatives are discussed in #3358
 			continue
 		}
-		minerAddr, err := address.NewFromString(minerStr)
+
+		laterMinerActorAddr, err := address.NewFromString(lateMinerActor)
```
ZenGround0: out of curiosity why "later"?  It looks like it would make more sense to use "lateMinerActorStr" and then "lateMinerActor"?

frrist: :open_hands: later because I have fat fingers, will correct it to `late`

protocol/storage/fault_slasher.go
frrist: imo the past tense of slash -> slashed is already clear, if this is a strong opinion let me know and I'll rename, else im going to leave it.

PR #3444, opened 2019-09-17
expose set head and sync over cli

### LINKED ISSUE #3442: Expose CLI for explicit chain sync and head manipulation
### Description
While debugging staging-0.5.0 I added some debug statements to the storage miner actor code. The only way to trigger these was to destroy my repo and then re-sync the chain, including re-fetching it from remote peers. Now that a new release has been deployed, I can't do that.

It would be more efficient and direct to instead support explicit rewinding of chain head back to the genesis, and then sync of the desired head tipset. This would cause the chain to be re-validated (fetching all the required data from the local store).

Chain import export #3443 could be another solution to the general problem.

### Acceptance criteria
- `chain set-head <cid>....` command replaces the chain store's head with the provided tipset key cids (which must be present in the store)
- `chain sync <cid>...` invokes `HandleNewTipset` with the provided tipset key, revalidating that part of the chain subsequent to wherever the new head diverges from the prior chain.

### Risks + pitfalls
These commands allow an unwitting user to screw up their local chain state. I don't think this should be a dominating concern during the present stage of development.

FYI @ZenGround0 @frrist

frrist:
This PR closes #3414
Of the options discussed here: https://github.com/filecoin-project/go-filecoin/issues/3442#issuecomment-532071672, we are going with C: bouncing the node after a call to SetHead to trigger chain reprocessing.

### DIFF
--- a/commands/chain.go
+++ b/commands/chain.go
@@ -10,6 +10,7 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit"
 	"github.com/ipfs/go-ipfs-cmds"
+	"github.com/libp2p/go-libp2p-core/peer"
 
 	"github.com/filecoin-project/go-filecoin/types"
 )
@@ -19,9 +20,11 @@ var chainCmd = &cmds.Command{
 		Tagline: "Inspect the filecoin blockchain",
 	},
 	Subcommands: map[string]*cmds.Command{
-		"head":   storeHeadCmd,
-		"ls":     storeLsCmd,
-		"status": storeStatusCmd,
+		"head":     storeHeadCmd,
+		"ls":       storeLsCmd,
+		"status":   storeStatusCmd,
+		"set-head": storeSetHeadCmd,
+		"sync":     storeSyncCmd,
 	},
 }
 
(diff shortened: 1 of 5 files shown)


plumbing/api.go
```diff
@@ -161,6 +161,11 @@ func (api *API) ChainHeadKey() types.TipSetKey {
 	return api.chain.Head()
 }
 
+// ChainSetHead sets `ts` as the new head of this chain iff it exists in the nodes chain store.
```
ZenGround0: This comment is untrue and needs to be changed or the behavior of `SetHead` needs to be updated to check for membership in the store.

commands/chain.go
```diff
@@ -122,3 +125,62 @@ var storeStatusCmd = &cmds.Command{
...
+	},
+}
+
+var storeSyncCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Instruct the chain syncer to sync a specific chain head, going to network if required.",
+	},
+	Arguments: []cmdkit.Argument{
+		cmdkit.StringArg("peerid", true, false, "Base58-encoded libp2p peer ID to sync from"),
+		cmdkit.StringArg("cids", true, true, "CID's of the tipset to set the chain head to."),
```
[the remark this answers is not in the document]
frrist: Copy-pasta, woops.

commands/chain.go
```diff
@@ -122,3 +125,62 @@ var storeStatusCmd = &cmds.Command{
...
+		for i, arg := range req.Arguments {
+			c, err := cid.Decode(arg)
+			if err != nil {
+				return err
+			}
+			headCids[i] = c
+		}
+		headKey := types.NewTipSetKey(headCids...)
+
+		headTs, err := GetPorcelainAPI(env).ChainTipSet(headKey)
```
anorth: This logic, while small, belongs behind the porcelain API, not in front of it. Change `ChainSetHead` to take a `TipSetKey` rather than tipset.

commands/chain.go
```diff
@@ -122,3 +125,62 @@ var storeStatusCmd = &cmds.Command{
...
+		}
+
+		syncCids := make([]cid.Cid, len(req.Arguments[1:]))
+		for i, arg := range req.Arguments[1:] {
+			c, err := cid.Decode(arg)
+			if err != nil {
+				return err
+			}
+			syncCids[i] = c
+		}
```
anorth: Factor out the duplicated decode-a-list-of-cids

plumbing/cst/chain_state.go
```diff
@@ -24,10 +24,15 @@ type chainReader interface {
 	GetTipSetState(context.Context, types.TipSetKey) (state.Tree, error)
 }
 
+type chainWriter interface {
+	SetHead(ctx context.Context, ts types.TipSet) error
+}
+
 // ChainStateProvider composes a chain and a state store to provide access to
```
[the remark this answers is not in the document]
frrist: How about ChainStateReadWriter?

commands/chain.go
```diff
@@ -122,3 +125,62 @@ var storeStatusCmd = &cmds.Command{
...
+		syncKey := types.NewTipSetKey(syncCids...)
+		ci := &types.ChainInfo{
+			Peer:   syncPid,
+			Height: 0, // because this is only checked when trusted is true.
```
anorth: But trusted *is* set to true below.... I think you meant "is not checked".

plumbing/api.go
anorth: I think that functionality should be changed to require the head to exist, i.e. have been validated as part of a valid chain at some point. Changing the parameter here to a `TipSetKey` is consistent with that.

plumbing/cst/chain_state.go
```diff
@@ -24,10 +24,15 @@ type chainReader interface {
 	GetTipSetState(context.Context, types.TipSetKey) (state.Tree, error)
 }
 
+type chainWriter interface {
+	SetHead(ctx context.Context, ts types.TipSet) error
+}
```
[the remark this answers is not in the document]
frrist: I think it provides _some_ clarity/separation of which methods do which things. I agree the area of plumbing is in need of a refactor and I agree that refactoring it is outside the scope of this PR. But, creating something like:
```go
type chainReadWriter interface {
// code stuff
}
```
isn't much better, except that it saves us 2 lines of code.

commands/chain.go
frrist: Opps, thanks.

commands/chain.go
```diff
@@ -122,3 +125,50 @@ var storeStatusCmd = &cmds.Command{
 		return nil
 	},
 }
+
+var storeSetHeadCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Modify the head of the current chain.",
```
anorth: "Set the chain head to a specific tipset key"

PR #3448, opened 2019-09-17
change lookback parameter to 15

acruikshank:
### Problem

Storage miners generate PoSt proofs at the very beginning of their proving window. The chain randomness they get had been 3 tip sets before this (defined by the `LookbackParameter`. It turns out 3 tipsets is too little because reorgs of 3 or more tipsets are fairly common.

### Solution

This is a short term solution to increase the lookback parameter to 15. The Rational PoSt spec does note define a LookbackInterval, so a better (but more time consuming) solution would be to remove the LookbackParameter altogether and define a safety interval (also 15) for the miner to wait after the proving window start to avoid reorgs.

### DIFF
--- a/sampling/chain_randomness_test.go
+++ b/sampling/chain_randomness_test.go
@@ -18,21 +18,21 @@ func TestSamplingChainRandomness(t *testing.T) {
 	tf.UnitTest(t)
 
 	// set a tripwire
-	require.Equal(t, sampling.LookbackParameter, 3, "these tests assume LookbackParameter=3")
+	require.Equal(t, sampling.LookbackParameter, 15, "these tests assume LookbackParameter=3")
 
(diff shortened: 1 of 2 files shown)


sampling/chain_randomness_test.go
```diff
@@ -18,21 +18,21 @@ func TestSamplingChainRandomness(t *testing.T) {
 	tf.UnitTest(t)
 
 	// set a tripwire
-	require.Equal(t, sampling.LookbackParameter, 3, "these tests assume LookbackParameter=3")
+	require.Equal(t, sampling.LookbackParameter, 15, "these tests assume LookbackParameter=3")
```
anorth: Fix the message too

acruikshank:
Looks like the better solution has been merged, so closing this.

PR #3449, opened 2019-09-17
shorten proving period and proving window

acruikshank:
### Problem

Our proving period is longer than it needs to be. It is currently undefined in the spec. Shortening it and the associated PoSt challenge time will provide a better experience for miners and more feedback for developers.

### Solution

Lower the `LargestSectorSizeProvingPeriodBlocks` to 300 and the `PoStChallengeWindowBlocks` to 150.


actor/builtin/miner/miner.go
```diff
@@ -33,12 +33,12 @@ func init() {
 // largest sector size - this constant and consensus.AncestorRoundsNeeded will
 // need to be reconsidered.
 // https://github.com/filecoin-project/specs/pull/318
-const LargestSectorSizeProvingPeriodBlocks = 1000
+const LargestSectorSizeProvingPeriodBlocks = 300
```
ZenGround0: why not make even smaller? (50, 100)

anorth: A single powerful core takes ~40 minutes (80 blocks) to compute a PoSt, so these values give about double that time, leaving room for various delays, reorgs etc. Multi-core machines can compute the PoSt much faster, but not assuming that seems good for now so we don't surprise miners too much.

Issue #3429, opened 2019-09-16
PoSts are rejected on-chain (challenge seed sampled at wrong height?)


anorth:
### Description
After the Rational PoSt integration #3318, #3413 some PoSts submitted by miners (`prover.go`) are being rejected by the `SubmitPoSt` actor method. This leads to  subsequent loss of power as reported in #3377.

One likely suspect is the challenge seed sampling, which changed from the previous proof construction, and for which logic is duplicated between the actor and prover, and prover logic spread across multiple modules. I think we lack a test that checks the prover's sample height selection or an integration tests that verifies the same selection from both prover and actor.

The invalid proofs from staging-0.5.0 are captured in #3428.

### Acceptance criteria
- PoSts computed by the prover are valid

### Risks + pitfalls
It's possible that the failures are due to chain re-orgs, which the prover does not handle yet at all. Once a challenge is available, the prover begins calculation and is unresponsive to a subsequent re-org. However, since we have a lookback parameter of 3 tips on the chain sampling code (which we will have to remove), this seems unlikely.

### Where to begin
- Improve `prover_test` to verify the sample height selected
- Maybe a prover integration test that tests the prover is compatible is miner actor?

anorth:
In addition to #3450 I have WIP for the miner detecting and recovering from a re-org that does change the challenge seed.

PR #3450, opened 2019-09-18
Storage miner worker waits 15 rounds before beginning PoSt computation.

anorth:
This 15 is hard-coded in the name of expediency, but should be pulled out
into a configuration value, along with other policy constants.

The latch returned from OnNewHeaviestTipSet reduces the miner_test latency
from some 8 seconds to 20ms.

This is a partial solution to #3429.

### DIFF
--- a/protocol/storage/miner.go
+++ b/protocol/storage/miner.go
@@ -34,6 +34,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/repo"
 	"github.com/filecoin-project/go-filecoin/types"
 	"github.com/filecoin-project/go-filecoin/util/convert"
+	"github.com/filecoin-project/go-filecoin/util/moresync"
 	"github.com/filecoin-project/go-sectorbuilder"
 )
 
@@ -47,6 +48,11 @@ const (
 	submitPostGasPrice = 1
 
 	waitForPaymentChannelDuration = 2 * time.Minute
+
+	// Number of rounds to wait after the challenge window opens before sampling the chain for
+	// challenge seed and beginning PoSt computation.
+	// Larger values are reduce fragility to re-orgs changing the challenge seed.
+	challengeDelayRounds = 15
 )
 
 const dealsAwatingSealDatastorePrefix = "dealsAwaitingSeal"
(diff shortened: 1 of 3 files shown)


anorth:
@acruikshank I am merging this before your approval, based on prior discussion, for inclusion in 0.5.3. Please review subsequently.

protocol/storage/miner.go
```diff
@@ -733,23 +739,26 @@ func (sm *Miner) handleQueryDeal(s inet.Stream) {
...
 	if err != nil {
-		return errors.Errorf("could not determine if actor created for bootstrapping: %s", err)
+		return doneLatch, errors.Errorf("could not determine if actor created for bootstrapping: %s", err)
 	}
 
 	if isBootstrapMinerActor {
 		// this is not an error condition, so log quietly
 		log.Info("bootstrap miner actor skips PoSt-generation flow")
-		return nil
+		return doneLatch, nil
```
acruikshank: Here and below. There is no feedback to the caller that the latch returned here will never be done. It's only used in testing where this isn't a problem, but it could cause problems in the future.

A quick solution would to return a nil latch if no PoSt is started. This method could also return an additional bool parameter to indicate whether or not a PoSt is started. I'm not sure if the latch would work properly if you call done on it before returning, but, if so, that could be done here and at the 3 other non error exit points.

protocol/storage/miner.go
```diff
@@ -772,43 +781,47 @@ func (sm *Miner) OnNewHeaviestTipSet(ts types.TipSet) error {
...
-		} else {
-			// we are too late
-			// TODO: figure out faults and payments here #3406
-			return errors.Errorf("too late start=%s  end=%s current=%s", provingWindowStart, provingWindowEnd, h)
+			postLatch := moresync.NewLatch(1)
+			go func() {
+				sm.submitPoSt(ctx, provingWindowStart, provingWindowEnd, inputs)
+				postLatch.Done()
+			}()
+			return postLatch, nil
```
ZenGround0: This is awesome.  Looking forward to seeing this get used by the node.

protocol/storage/miner.go
anorth: The `doneLatch` (with count of zero) is immediately done/open. A call to `Wait` will not block.

Issue #3025, opened 2019-07-08
Simplify access to chain history for sampling


anorth:
### Description
Chain sampling is used to generate a PoSt challenge seed, which is just the min ticket from a block a fixed number of rounds prior to the proving period.

The chain randomness sampling code traverses an ordered list of tipsets looking for one of a particular height. This list of tipsets is constructed with many complex parameters in `chain.GetAncestors`, called every block by `syncer.syncOne`, and passed through a deep call hierarchy. All the eventual caller wants is the ticket from a tipset at (or prior to) some height. Building this list every time only to sample a single tipset from it is wasteful and obscures the simplicity of what is actually needed.

Rather than compute and pass this big list around, some component (the tip index) should just index tipsets by height and provide a direct lookup (by binary search).

https://github.com/filecoin-project/go-filecoin/blob/0201cdfcc690c261b1b98d79b98474c54cec7ee8/chain/get_ancestors.go#L61

### Acceptance criteria
- Sampling the chain at a particular height is performed by an indexed lookup

### Risks + pitfalls
The tip index already has an index that includes height in the key (but also some CIDs). It may be reasonable to split this index up, or may require another one. The tip index is a major consumer of memory; consider holding pointers rather than tipset values (which are only slices anyway).

### Where to begin
Remove SampleChainRandomness and give callers a function to lookup a tipset with height less-or-equal to some sample height.

ZenGround0:
Some complications to keep in mind when setting out to do this:

The ugly list exists to circumvent a dependency cycle: `chain ==> consensus ==> chain`.  Allowing vm code to directly access the tips index will require non-trivial design as it violates the existing dependency direction in a big way.

Once we reorganize our dependency tangle the tip index needs work to support the desired feature.  Looking up by height in the tip index will currently provide all valid tipsets seen by a daemon at the requested height, including tipsets on forks.  The tip index needs to offer a good interface for looking up blocks in a direct ancestral subtree for this refactor to land.

ZenGround0:
Also please keep in mind another significant change already tracked that needs to land in the same area and will likely have a big impact on the design of this solution: https://github.com/filecoin-project/go-filecoin/issues/2223

anorth:
Closed by #3810

Issue #1872, opened 2019-02-12
Bootstrapping chain randomness


ZenGround0:
### Description
Randomness for leader election and post generation needs to be taken from "far enough back in the chain".  When the chain has fewer tipsets than the lookback parameter we use to sample randomness it is undefined how we get this randomness.  The protocol needs to handle this edge case.  Currently we use the genesis ticket (0 bytes) as the chain randomness for all blocks fewer than lookback tipsets away from genesis.

### Acceptance criteria
The spec has a well motivated way to handle this and the implementation is in sync with the spec.

rosalinekarr:
@whyrusleeping, do you think this should this have a separate task for `specs` and `go-filecoin` or is the one task good enough?

anorth:
See also #2251

anorth:
I expect the new spec to resolve this issue, and we should wait for that.

PR #3453, opened 2019-09-18
Remove lookback from chain sampling for PoSt challenge seed.

anorth:
Rational PoSt has no lookback, the chain is sampled at the opening of the proving window.
The spec is unclear about what to do if this is a null tipset, this PR takes the previous
non-empty tipset. This is intended to make sense along with #3450, which waits for the
chain to settle before beginning a PoSt.

This removes PoSt challenge generation from the scope of #1872.

The code that gets the blocks for sampling is all waaay more complicated than necessary now (#3025).

### DIFF
--- a/chain/get_ancestors_test.go
+++ b/chain/get_ancestors_test.go
@@ -99,7 +99,7 @@ func TestCollectTipSetsPastHeight(t *testing.T) {
 
 	stopHeight := types.NewBlockHeight(uint64(4))
 	iterator := chain.IterAncestors(ctx, builder, head)
-	tipsets, err := chain.CollectTipSetsPastHeight(ctx, iterator, stopHeight)
+	tipsets, err := chain.CollectTipSetsPastHeight(iterator, stopHeight)
 	assert.NoError(t, err)
 	latestHeight, err := tipsets[0].Height()
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


chain/get_ancestors_test.go
```diff
@@ -278,22 +269,22 @@ func TestGetRecentAncestorsStartingEpochIsNull(t *testing.T) {
...
 	require.NoError(t, err)
-	ancestors, err := chain.GetRecentAncestors(ctx, head, builder, types.NewBlockHeight(h+uint64(1)), types.NewBlockHeight(epochs), uint(lookback))
+	sampleHeight := types.NewBlockHeight(h).Sub(types.NewBlockHeight(epochs))
+	ancestors, err := chain.GetRecentAncestors(ctx, head, builder, sampleHeight)
 	require.NoError(t, err)
 
-	// We expect to see 20+1 blocks include less than earliest block in the first 28 epochs and an additional 6 for the lookback parameter
-	assert.Equal(t, len2+lookback+2, len(ancestors))
+	// We expect to see 20+1 tips: the head of the chain plus one after the nulls.
+	assert.Equal(t, len2+2, len(ancestors))
```
frrist: slight preference for less naked numbers, i.e give `2` a name.

anorth:
@ZenGround0 @acruikshank I'm landing this before your review for inclusion in a 0.5.3 release candidate. Please review, and revert and/or fix if I've done something wrong.

consensus/expected.go
```diff
@@ -69,6 +69,7 @@ const ECPrM uint64 = 100
 // TODO: If the following PR is merged - and the network doesn't define a
 // largest sector size - this constant will need to be reconsidered.
 // https://github.com/filecoin-project/specs/pull/318
+// NOTE(anorth): This height is excessive, but safe, with the Rational PoSt construction.
```
ZenGround0: yup should be `miner.PoStChallengeWindowBlocks`

PR #3455, opened 2019-09-18
Log PoSt computation and verification parameters

anorth:
These are the logs that @ZenGround0 had in 0.5.2. I'm preserving them to master for now, though we may rework the logs later.

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -6,6 +6,7 @@ import (
 	"github.com/filecoin-project/go-sectorbuilder"
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
+	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-core/peer"
 	xerrors "github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


actor/builtin/miner/miner.go
```diff
@@ -950,6 +953,11 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, fault
 			}
 
 			sortedSectorInfo := go_sectorbuilder.NewSortedSectorInfo(sectorInfos...)
+			log.Infof("Verifying post for addr %s -- end: %s -- seed: %x \n", ctx.Message().To.String(), state.ProvingPeriodEnd.String(), seed)
+			for i, ssi := range sortedSectorInfo.Values() {
+				log.Infof("ssi %d: sector id %d -- commR %x\n", i, ssi.SectorID, ssi.CommR)
```
frrist: no need for `\n` in log messages, see elsewhere too.

actor/builtin/miner/miner.go
```diff
@@ -950,6 +953,11 @@ func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, fault
 			}
 
 			sortedSectorInfo := go_sectorbuilder.NewSortedSectorInfo(sectorInfos...)
+			log.Infof("Verifying post for addr %s -- end: %s -- seed: %x \n", ctx.Message().To.String(), state.ProvingPeriodEnd.String(), seed)
```
frrist: `String()` is not needed here.

PR #3459, opened 2019-09-18
refactor: cleanup node internals 

icorderi:
**Remaining work:**
- [x] ~I will proceed to group up the extracted pieces by submodules "block production, sector committing & proving, etc" to then rebuild the construction/cleanup based on the pieces.~
- [x] ~finish finding new homes for things in `ToSplitOrNotToSplitNode`~

**Following PRs:**
- cleanup initialization for each submodule.
- split start/stop code into pieces and move them to each submodule. 
- move submodules to their own package to force visibility issues

### DIFF
--- a/commands/daemon.go
+++ b/commands/daemon.go
@@ -10,12 +10,12 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/ipfs/go-ipfs-cmdkit"
-	"github.com/ipfs/go-ipfs-cmds"
+	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
+	cmds "github.com/ipfs/go-ipfs-cmds"
 	cmdhttp "github.com/ipfs/go-ipfs-cmds/http"
 	writer "github.com/ipfs/go-log/writer"
 	ma "github.com/multiformats/go-multiaddr"
-	"github.com/multiformats/go-multiaddr-net"
+	manet "github.com/multiformats/go-multiaddr-net"
 	"github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/clock"
@@ -153,12 +153,12 @@ func getRepo(req *cmds.Request) (repo.Repo, error) {
 // A message sent to or closure of the `terminate` channel signals the server to stop.
 func RunAPIAndWait(ctx context.Context, nd *node.Node, config *config.APIConfig, ready chan interface{}, terminate chan os.Signal) error {
 	servenv := &Env{
-		blockMiningAPI: nd.BlockMiningAPI,
+		blockMiningAPI: nd.BlockMining.BlockMiningAPI,
 		ctx:            ctx,
 		inspectorAPI:   NewInspectorAPI(nd.Repo),
 		porcelainAPI:   nd.PorcelainAPI,
-		retrievalAPI:   nd.RetrievalAPI,
-		storageAPI:     nd.StorageAPI,
+		retrievalAPI:   nd.RetrievalProtocol.RetrievalAPI,
+		storageAPI:     nd.StorageProtocol.StorageAPI,
 	}
 
 	cfg := cmdhttp.NewServerConfig()
--- a/commands/protocol_cmd_test.go
+++ b/commands/protocol_cmd_test.go
@@ -23,7 +23,7 @@ func TestProtocol(t *testing.T) {
 			c.Mining.AutoSealIntervalSeconds = 120
 		}).
 		Build(ctx)
-	require.NoError(t, node.ChainReader.Load(ctx))
+	require.NoError(t, node.Chain.ChainReader.Load(ctx))
 
 	// Run the command API.
 	cmd, stop := test.RunNodeAPI(ctx, node, t)
--- a/node/block_mining_submodule.go
+++ b/node/block_mining_submodule.go
@@ -1,0 +1,26 @@
+package node
+
+import (
+	"context"
+	"sync"
+
+	"github.com/filecoin-project/go-filecoin/mining"
+	"github.com/filecoin-project/go-filecoin/protocol/block"
+)
+
+// BlockMiningSubmodule enhances the `Node` with block mining capabilities.
+type BlockMiningSubmodule struct {
+	BlockMiningAPI *block.MiningAPI
+
+	// Mining stuff.
+	AddNewlyMinedBlock newBlockFunc
+	// cancelMining cancels the context for block production and sector commitments.
+	cancelMining    context.CancelFunc
+	MiningWorker    mining.Worker
+	MiningScheduler mining.Scheduler
+	mining          struct {
+		sync.Mutex
+		isMining bool
+	}
+	miningDoneWg *sync.WaitGroup
+}
--- a/node/builder.go
+++ b/node/builder.go
@@ -272,28 +272,40 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
 	outbox := message.NewOutbox(fcWallet, consensus.NewOutboundMessageValidator(), msgQueue, msgPublisher, outboxPolicy, chainStore, chainState)
 
 	nd := &Node{
-		blockservice: bservice,
-		Blockstore:   bs,
-		cborStore:    &ipldCborStore,
-		Clock:        nc.Clock,
-		Consensus:    nodeConsensus,
-		ChainReader:  chainStore,
-		ChainSynced:  moresync.NewLatch(1),
-		MessageStore: messageStore,
-		Syncer:       chainSyncer,
-		PowerTable:   powerTable,
-		PeerTracker:  peerTracker,
-		Fetcher:      fetcher,
-		Exchange:     bswap,
-		host:         peerHost,
-		Inbox:        inbox,
-		OfflineMode:  nc.OfflineMode,
-		Outbox:       outbox,
-		NetworkName:  network,
-		PeerHost:     peerHost,
-		Repo:         nc.Repo,
-		Wallet:       fcWallet,
-		Router:       router,
+		Clock:       nc.Clock,
+		OfflineMode: nc.OfflineMode,
+		Repo:        nc.Repo,
+		Network: NetworkSubmodule{
+			host:        peerHost,
+			PeerHost:    peerHost,
+			NetworkName: network,
+			PeerTracker: peerTracker,
+			Router:      router,
+		},
+		Wallet: WalletSubmodule{
+			Wallet: fcWallet,
+		},
+		Messaging: MessagingSubmodule{
+			Inbox:  inbox,
+			Outbox: outbox,
+		},
+		Blockstore: BlockstoreSubmodule{
+			blockservice: bservice,
+			Blockstore:   bs,
+			cborStore:    &ipldCborStore,
+		},
+		StorageNetworking: StorageNetworkingSubmodule{
+			Exchange: bswap,
+		},
+		Chain: ChainSubmodule{
+			Fetcher:      fetcher,
+			Consensus:    nodeConsensus,
+			ChainReader:  chainStore,
+			ChainSynced:  moresync.NewLatch(1),
+			MessageStore: messageStore,
+			Syncer:       chainSyncer,
+			PowerTable:   powerTable,
+		},
 	}
 
 	nd.PorcelainAPI = porcelain.New(plumbing.New(&plumbing.APIDeps{
@@ -328,7 +340,7 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
 		return nil, errors.Wrapf(err, "couldn't parse bootstrap addresses [%s]", ba)
 	}
 	minPeerThreshold := nd.Repo.Config().Bootstrap.MinPeerThreshold
-	nd.Bootstrapper = net.NewBootstrapper(bpi, nd.Host(), nd.Host().Network(), nd.Router, minPeerThreshold, period)
+	nd.Network.Bootstrapper = net.NewBootstrapper(bpi, nd.Host(), nd.Host().Network(), nd.Network.Router, minPeerThreshold, period)
 
 	return nd, nil
 }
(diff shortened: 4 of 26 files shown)


anorth:
Thanks for the WIP. It's not really clear to me what the decision procedure is for what goes on the node directly. I guess "things that are used by the higher level services", but that implies a single flat layer that may not really exist. Further work to break down some submodules may clear it up.

An alternative approach that might make this a little more clear is to pick a single submodule to cleanly extract first, leaving the final re-org until after most submodules are more clearly defined.

In person I expressed a desire to land stages of big refactors where possible (i.e. non-broken, tests pass etc). This isn't there yet in terms of being clearly the right direction.

anorth:
One boundary that would be clear is between things which consume the plumbing or porcelain API, and things which provide part of it. There are limitations on layering that this arrangement introduces, which I think we may want to break, but it's nevertheless a clear line at present.

icorderi:
@anorth the state at the time you looked was just trying to get a rough cut on anything non-networking separated to see the damage and get a start point for the submodules.
The current state already has the a lot of submodules broken up if you want to take another look.
The remainder of things needing a new home are in `ToSplitOrNotToSplitNode`.

node/chain_submodule.go
```diff
@@ -0,0 +1,33 @@
...
+	HeaviestTipSetCh chan interface{}
+	// cancelChainSync cancels the context for chain sync subscriptions and handlers.
+	cancelChainSync context.CancelFunc
+	// ChainSynced is a latch that releases when a nodes chain reaches a caught-up state.
+	// It serves as a barrier to be released when the initial chain sync has completed.
+	// Services which depend on a more-or-less synced chain can wait for this before starting up.
+	ChainSynced *moresync.Latch
+	// Review: should we have this on chain or is a bit more generic?
+	// Fetcher is the interface for fetching data from nodes.
+	Fetcher net.Fetcher
```
anorth: This is specific to chain sync - it belongs here but need not be exported.

node/helpers.go
```diff
@@ -0,0 +1,57 @@
...
+		return cid.Undef, errors.Wrap(err, "failed to read genesisKey")
+	}
+
+	var c cid.Cid
+	err = json.Unmarshal(bb, &c)
+	if err != nil {
+		return cid.Undef, errors.Wrap(err, "failed to cast genesisCid")
+	}
+	return c, nil
+}
```
anorth: This last function belongs with chain store.

node/sector_builder_submodule.go
```diff
@@ -0,0 +1,9 @@
+package node
+
+import "github.com/filecoin-project/go-filecoin/proofs/sectorbuilder"
+
+// SectorBuilderSubmodule enhances the `Node` with sector storage capabilities.
+type SectorBuilderSubmodule struct {
+	// SectorBuilder is used by the miner to fill and seal sectors.
+	sectorBuilder sectorbuilder.SectorBuilder
+}
```
anorth: This should probably go with the storage protocol, though I'm not 100% sure about that.

node/theotherguy.go
```diff
@@ -0,0 +1,61 @@
...
+	blockservice bserv.BlockService
+
+	// Review: where do we place this guy?
+	// Blockstore is the un-networked blocks interface
+	Blockstore bstore.Blockstore
+
+	// Review: should we move this to the chain submodule?
+	// TODO: cache for chain and data shared, wrapper for blockstore
+	// CborStore is a temporary interface for interacting with IPLD objects.
+	cborStore *hamt.CborIpldStore
```
anorth: Yes it's for chain and chain state. Possibly shared with piece data exchange for deals at the moment, but we should separate them. cc @ingar

node/theotherguy.go
```diff
@@ -0,0 +1,61 @@
...
+	// Review: candidate to be moved back to Node
+	Repo repo.Repo
+
+	// Review: I need to better understand this guy..
+	// Incoming messages for block mining.
+	Inbox *message.Inbox
+	// Messages sent and not yet mined.
+	Outbox *message.Outbox
+
+	// Review: is this its own submodule?
```
anorth: Yes. Possibly to simple to be worth the bother.

node/theotherguy.go
```diff
@@ -0,0 +1,61 @@
...
+	Wallet *wallet.Wallet
+
+	// Review: is this only used for storage deals or retrieval too?
+	//         if its only storage -> move to StorageSubmodule
+	//         if its both, move to SectorBuilderSubmodule?
+	// Exchange is the interface for fetching data from other nodes.
+	Exchange exchange.Interface
```
anorth: Both deals and retrieval. But not related to sector builder. We can either gather all the storage protocol stuff into one component (including deals, retrieval, sector builder), or separate deals and retrieval from the core storage mining.

node/theotherguy.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// Review: where do we place this guy?
+	// Blockservice is a higher level interface for fetching data
+	blockservice bserv.BlockService
+
+	// Review: where do we place this guy?
+	// Blockstore is the un-networked blocks interface
+	Blockstore bstore.Blockstore
```
anorth: It's hard to split out right now since it's shared by a couple of components.

node/theotherguy.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// Review: should we move this to the chain submodule?
+	// TODO: cache for chain and data shared, wrapper for blockstore
+	// CborStore is a temporary interface for interacting with IPLD objects.
+	cborStore *hamt.CborIpldStore
+}
```
anorth: My recommendations for now:
- inbox/outbox: a message exchange component
- the rest still on node (but some can move after more fundamental refactors)

node/sector_builder_submodule.go
icorderi: I left it out of it because it will be required by the retrieval protocol too. 
It is also possible that we might have a node running the storage protocol but not being responsible for actually storing the data locally, in which case they might have a different implementation of this submodule which talks with sector builder nodes.

node/helpers.go
icorderi: Yes, I'll move all the functions floating around node, helper, builder, etc. closer to each submodule on another PR. 
I want to keep this one to just the new submodule structures and the attributes shuffling.

node/chain_submodule.go
icorderi: I'll reduce the visibility of quite a few things on a future PR, there are other things that could be private too.

node/builder.go
```diff
@@ -272,31 +272,37 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
...
+			Wallet:       fcWallet,
+		},
+		Chain3140: ChainSubmodule{
+			Fetcher:      fetcher,
+			Consensus:    nodeConsensus,
+			ChainReader:  chainStore,
+			ChainSynced:  moresync.NewLatch(1),
+			MessageStore: messageStore,
+			Syncer:       chainSyncer,
+			PowerTable:   powerTable,
```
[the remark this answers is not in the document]
icorderi: Sounds good, left a TODO

node/theotherguy.go
icorderi: I guess you could start an instance that only has the SectorBuilder and not the bitswap Piece Exchange running. Maybe for running some local analysis or maybe just for running some proofs (PIP, PoST).
I think they are tightly related though, but I can separate it as a `StorageNetworkingSubmodule`.

node/network_submodule.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/libp2p/go-libp2p-core/routing"
+)
+
+// NetworkSubmodule enhances the `Node` with networking capabilities.
+type NetworkSubmodule struct {
+	NetworkName string
+
+	host host.Host
+
+	PeerHost host.Host
```
frrist: Having 2 hosts fields seems weird to me. I understand its not the intent of this PR to address this so I have filed an issue to discuss it: https://github.com/filecoin-project/go-filecoin/issues/3477. Would you mind linking to it with a TODO?

icorderi: sure, linked

icorderi:
@frrist, @anorth, @acruikshank the PR is ready if you guys want to take a final look.

node/theotherguy.go
taoshengshi: @anorth @icorderi I suggest one basic principle is to separate the data from :

- onchain(chain and chain state, which is broadcast and replicated across global network) 

and 

- offchain(deals,miners,which act as meta data to retrieve data pieces，payment channel with miners,... ).

commands/daemon.go
```diff
@@ -153,12 +153,12 @@ func getRepo(req *cmds.Request) (repo.Repo, error) {
...
 	servenv := &Env{
-		blockMiningAPI: nd.BlockMiningAPI,
+		blockMiningAPI: nd.BlockMining.BlockMiningAPI,
 		ctx:            ctx,
 		inspectorAPI:   NewInspectorAPI(nd.Repo),
 		porcelainAPI:   nd.PorcelainAPI,
-		retrievalAPI:   nd.RetrievalAPI,
-		storageAPI:     nd.StorageAPI,
+		retrievalAPI:   nd.RetrievalProtocol.RetrievalAPI,
+		storageAPI:     nd.StorageProtocol.StorageAPI,
```
anorth: These names stutter now. It might not immediately make sense, but we should move toward the exported references being like `nd.StorageProtocol.API`.

commands/protocol_cmd_test.go
```diff
@@ -23,7 +23,7 @@ func TestProtocol(t *testing.T) {
 			c.Mining.AutoSealIntervalSeconds = 120
 		}).
 		Build(ctx)
-	require.NoError(t, node.ChainReader.Load(ctx))
+	require.NoError(t, node.Chain.ChainReader.Load(ctx))
```
anorth: Similarly, `node.Chain.Reader`?

node/block_mining_submodule.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/filecoin-project/go-filecoin/net/pubsub"
+	"github.com/filecoin-project/go-filecoin/protocol/block"
+)
+
+// BlockMiningSubmodule enhances the `Node` with block mining capabilities.
+type BlockMiningSubmodule struct {
+	BlockMiningAPI *block.MiningAPI
+
+	// Mining stuff.
+	AddNewlyMinedBlock newBlockFunc
```
anorth: Move `newBlockFunc` to this file

node/block_mining_submodule.go
```diff
@@ -0,0 +1,30 @@
...
+	MiningWorker    mining.Worker
+	MiningScheduler mining.Scheduler
+	mining          struct {
+		sync.Mutex
+		isMining bool
+	}
+	miningDoneWg *sync.WaitGroup
+
+	// Network Fields
+	MessageSub pubsub.Subscription
```
anorth: This belongs more on the messaging submodule. The block mining code does not receive these messages directly, but fetches them from the message pool (inbox).

node/builder.go
```diff
@@ -272,28 +272,40 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
...
+		},
+		Chain: ChainSubmodule{
+			Fetcher:      fetcher,
+			Consensus:    nodeConsensus,
+			ChainReader:  chainStore,
+			ChainSynced:  moresync.NewLatch(1),
+			MessageStore: messageStore,
+			Syncer:       chainSyncer,
+			PowerTable:   powerTable,
+		},
```
anorth: `BlockMiningSubmodule` is missing here. I see that, unlike other submodules, it's not instantiated until started. That at least deserves a clear comment here. I hope that future work will resolve that so that it can be instantiated without starting and we can avoid the nil references.

node/chain_submodule.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/filecoin-project/go-filecoin/util/moresync"
+)
+
+// ChainSubmodule enhances the `Node` with chain capabilities.
+type ChainSubmodule struct {
+	BlockSub     pubsub.Subscription
+	Consensus    consensus.Protocol
+	ChainReader  nodeChainReader
+	MessageStore *chain.MessageStore
+	Syncer       nodeChainSyncer
```
anorth: Move `nodeChainReader` and `nodeChainReader` to this file.

node/sector_builder_submodule.go
anorth: Right we need to split the StorageProtocol up. It must be possible to have a storage miner (committing sectors, submitting PoSt) without any deal protocol. The latter is built on the former.

node/storage_networking_submodule.go
```diff
@@ -0,0 +1,11 @@
+package node
+
+import exchange "github.com/ipfs/go-ipfs-exchange-interface"
+
+// StorageNetworkingSubmodule enhances the `Node` with data transfer capabilities.
+type StorageNetworkingSubmodule struct {
+	// Exchange is the interface for fetching data from other nodes.
+	//
+	// TODO: move to a `StorageNetworkingSubmodule`
```
anorth: Remove TODO

node/block_mining_submodule.go
icorderi: moved.

node/builder.go
icorderi: Correct. The next PR add the call to initialize all the nil submodules, and attempts to move some of the initialization code back here. It might take a few PR's to fully separate initialization from start.

node/chain_submodule.go
icorderi: If it's all the same with you, I rather move code closer to where it belongs on a followup PR.

commands/daemon.go
icorderi: Yes, agreed. I'll rename things on a separate PR.

PR #3460, opened 2019-09-18
Better shorter timeouts on Graphsync Requests

### LINKED ISSUE #3371: Fetching timeout update
### Description
Graphsync requests for 64 blocks should have a longer timeout than requests for 1 block.  We should scale timeout based on the graphsync fetcher's `recursionDepth` parameter.

Furthermore we need to keep this timeout as small as is reasonable without losing data.  Currently timeouts of ~ 2 block times blocks the node from making progress and effectively partitions the network causing forking.

### Acceptance criteria
- Measurements for estimating good defaults for a fetching a number of blocks.  Should probably do this across big geographic distances to get a more realistic idea
- Make fetching timeouts a function of recursionDepth

### Risks + pitfalls
- May be hard to measure
- If we push timeouts too low it may negatively interfere with syncing performance

### Where to begin
- Talk to @frrist and @anorth about measurement setup (they've done something like this before).
- net/graphsync_fetcher.go

hannahhoward:
# Goals

When a peer becomes unresponsive to Graphsync Fetcher requests, fail more quickly and move on to other peers

# Implementation

- Timeout for a graphsync request is now based on unresponsiveness, rather than total time. IOW -- if a request stop sending data for a specific time, cancel it. 
- Because the timeout is based on unresponsiveness rather than total time, use a shorter timeout than before (10 seconds)
- Also provides variadic options struct to the fetcher to override the
unresponsiveness timeout
- Add tests of timeout behavior to verify behavior when various requests simply hangup with no further response

fix #3371

### DIFF
--- a/net/graphsync_fetcher.go
+++ b/net/graphsync_fetcher.go
@@ -24,10 +24,10 @@ import (
 var logGraphsyncFetcher = logging.Logger("net.graphsync_fetcher")
 
 const (
-	// Timeout for a single graphsync request (which may be for many blocks).
-	// We might prefer this timeout to scale with the number of blocks expected in the fetch,
-	// when that number is large.
-	requestTimeout = 60 * time.Second
+	// Timeout for a single graphsync request getting "stuck"
+	// -- if no more responses are received for a period greater than this,
+	// we will assume the request has hung-up and cancel it
+	unresponsiveTimeout = 10 * time.Second
 )
 
 // Fetcher defines an interface that may be used to fetch data from the network.
@@ -55,23 +55,41 @@ type graphsyncFallbackPeerTracker interface {
 // GraphSyncFetcher is used to fetch data over the network.  It is implemented
 // using a Graphsync exchange to fetch tipsets recursively
 type GraphSyncFetcher struct {
-	exchange    GraphExchange
-	validator   consensus.SyntaxValidator
-	store       bstore.Blockstore
-	ssb         selectorbuilder.SelectorSpecBuilder
-	peerTracker graphsyncFallbackPeerTracker
+	exchange            GraphExchange
+	validator           consensus.SyntaxValidator
+	store               bstore.Blockstore
+	ssb                 selectorbuilder.SelectorSpecBuilder
+	peerTracker         graphsyncFallbackPeerTracker
+	unresponsiveTimeout time.Duration
+}
+
+// GraphsyncFetcherOption is function that configures graphsync. It should not
+// be created directly but should instead generated through an option function like
+// UseFetcherTimeout
+type GraphsyncFetcherOption func(*GraphSyncFetcher)
+
+// UseUnresponsiveTimeout sets up the GraphsyncFetcher with a different
+// unresponsiveness timeout than the default
+func UseUnresponsiveTimeout(timeout time.Duration) GraphsyncFetcherOption {
+	return func(gsf *GraphSyncFetcher) {
+		gsf.unresponsiveTimeout = timeout
+	}
 }
 
 // NewGraphSyncFetcher returns a GraphsyncFetcher wired up to the input Graphsync exchange and
 // attached local blockservice for reloading blocks in memory once they are returned
 func NewGraphSyncFetcher(ctx context.Context, exchange GraphExchange, blockstore bstore.Blockstore,
-	bv consensus.SyntaxValidator, pt graphsyncFallbackPeerTracker) *GraphSyncFetcher {
+	bv consensus.SyntaxValidator, pt graphsyncFallbackPeerTracker, options ...GraphsyncFetcherOption) *GraphSyncFetcher {
 	gsf := &GraphSyncFetcher{
-		store:       blockstore,
-		validator:   bv,
-		exchange:    exchange,
-		ssb:         selectorbuilder.NewSelectorSpecBuilder(ipldfree.NodeBuilder()),
-		peerTracker: pt,
+		store:               blockstore,
+		validator:           bv,
+		exchange:            exchange,
+		ssb:                 selectorbuilder.NewSelectorSpecBuilder(ipldfree.NodeBuilder()),
+		peerTracker:         pt,
+		unresponsiveTimeout: unresponsiveTimeout,
+	}
+	for _, option := range options {
+		option(gsf)
 	}
 	return gsf
 }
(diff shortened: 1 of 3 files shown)


net/graphsync_fetcher.go
```diff
@@ -24,10 +24,10 @@ import (
...
 
 const (
-	// Timeout for a single graphsync request (which may be for many blocks).
-	// We might prefer this timeout to scale with the number of blocks expected in the fetch,
-	// when that number is large.
-	requestTimeout = 60 * time.Second
+	// Timeout for a single graphsync request getting "stuck"
+	// -- if no more responses are received for a period greater than this,
+	// we will assume the request has hung-up and cancel it
+	unresponsiveTimeout = 10 * time.Second
```
ZenGround0: I'm curious -- do we have information on the upper bound of the delay we would expect with high probability from a peer with no network issues?  My intuition is that we want to set this as low as we can reasonably get away with before we start killing productive connections.  My uninformed intuition is also that 10 seconds is probably higher than we need and I'd love to know if this is wrong and 10 seconds is already pushing the limit.

hannahhoward: The short version is I honestly don't know :( I would probably prefer to wait on dropping it lower till we are at least requesting two requests in parallel. Then if we get a false positive on believing ourselves unresponsive we can at least still pickup from the other request.

hannahhoward: another option would be to track actual latency and use a multiple of the average latency between progress steps

net/graphsync_fetcher.go
```diff
@@ -224,27 +242,57 @@ func (gsf *GraphSyncFetcher) fetchBlocks(ctx context.Context, cids []cid.Cid, ta
...
+		err := gsf.consumeResponse(requestChan, errChan, cancelFunc)
+		if err != nil {
 			anyError = err
 		}
 	}
 	return anyError
 }
 
+func (gsf *GraphSyncFetcher) consumeResponse(requestChan <-chan graphsync.ResponseProgress, errChan <-chan error, cancelFunc func()) error {
+	timer := time.NewTimer(gsf.unresponsiveTimeout)
```
frrist: :disappointed: our clock package doesn't have a timer, just a ticker, which means we can't mock this during testing (there is a [feature request](https://github.com/jonboulle/clockwork/issues/10) open, bit stale). This should be fine, but wanted to flag it as we should try and avoid any direct calls to the time package.

hannahhoward: got it! So noted... good to know about the no direct use of time thing -- that would make better sense here than the variadic options pattern... but also, a lot of things to change.

net/graphsync_fetcher.go
```diff
@@ -55,23 +55,41 @@ type graphsyncFallbackPeerTracker interface {
...
-	validator   consensus.SyntaxValidator
-	store       bstore.Blockstore
-	ssb         selectorbuilder.SelectorSpecBuilder
-	peerTracker graphsyncFallbackPeerTracker
+	exchange            GraphExchange
+	validator           consensus.SyntaxValidator
+	store               bstore.Blockstore
+	ssb                 selectorbuilder.SelectorSpecBuilder
+	peerTracker         graphsyncFallbackPeerTracker
+	unresponsiveTimeout time.Duration
```
anorth: `progressTimeout` would be a slightly better name.

net/graphsync_fetcher.go
anorth: `defaultProgressTimeout`

net/graphsync_fetcher.go
```diff
@@ -224,27 +242,57 @@ func (gsf *GraphSyncFetcher) fetchBlocks(ctx context.Context, cids []cid.Cid, ta
 		efsb.Insert("messageReceipts", gsf.ssb.Matcher())
 	}).Node()
 	errChans := make([]<-chan error, 0, len(cids))
-	requestCtx, requestCancel := context.WithTimeout(ctx, requestTimeout)
-	defer requestCancel()
+	requestChans := make([]<-chan graphsync.ResponseProgress, 0, len(cids))
+	cancelFuncs := make([]func(), 0, len(cids))
```
anorth: Nit: with so many parallel arrays it might be cleaner to create a single array of structs.

net/graphsync_fetcher.go
anorth: Blocking: no new dependencies on `time`. Two options for consideration:
- (ab)use a ticker instead that we can fake for testing
- factor out a "progress watchdog" that removes references to time completely while being implementable in terms of a `time.Timer` in production.

net/graphsync_fetcher_test.go
```diff
@@ -533,6 +536,180 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+		mgs.expectRequestToRespondWithHangupAfter(pid0, layer1Selector, loader, 0, final.At(1).Cid(), final.At(2).Cid())
+		mgs.expectRequestToRespondWithHangupAfter(pid1, layer1Selector, loader, 0, final.At(1).Cid(), final.At(2).Cid())
+		mgs.expectRequestToRespondWithHangupAfter(pid2, layer1Selector, loader, 0, final.At(1).Cid(), final.At(2).Cid())
+
+		fetcher := net.NewGraphSyncFetcher(ctx, mgs, bs, bv, pt, net.UseUnresponsiveTimeout(100*time.Millisecond))
+		done := doneAt(gen.Key())
+		ts, err := fetcher.FetchTipSets(ctx, final.Key(), pid0, done)
+		mgs.verifyReceivedRequestCount(7)
+		mgs.verifyExpectations()
+		require.Errorf(t, err, "Failed fetching tipset: %s", final.Key().String())
```
anorth: I don't think this does what you expect it to. I makes no assertion about the content of the error message. You need to check that more explicitly.

net/graphsync_fetcher_test.go
```diff
@@ -533,6 +536,180 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+
+		mgs := newMockableGraphsync(ctx, bs, t)
+		mgs.expectRequestToRespondWithLoader(pid0, layer1Selector, loader, final.At(0).Cid())
+		mgs.expectRequestToRespondWithHangupAfter(pid0, layer1Selector, loader, 0, final.At(1).Cid(), final.At(2).Cid())
+		mgs.expectRequestToRespondWithLoader(pid1, layer1Selector, loader, final.At(1).Cid())
+		mgs.expectRequestToRespondWithHangupAfter(pid1, layer1Selector, loader, 0, final.At(2).Cid())
+		mgs.expectRequestToRespondWithLoader(pid2, layer1Selector, loader, final.At(2).Cid())
+		mgs.expectRequestToRespondWithLoader(pid2, recursiveSelector(1), loader, final.At(0).Cid())
+
+		fetcher := net.NewGraphSyncFetcher(ctx, mgs, bs, bv, pt, net.UseUnresponsiveTimeout(100*time.Millisecond))
```
anorth: Blocking: any test that uses `time` is 1) automatically a `tf.IntegrationTest(t)` because it's slow, and 2) probably nondeterministic and flaky.

It may be ok to add some integration tests (in a new file) with very long timeouts to hope for non-flakiness, but in this case we can probably do much better by explicitly controlling time.

net/graphsync_fetcher_test.go
hannahhoward: I need to be using ErrorEqual I think :)

net/graphsync_fetcher_test.go
hannahhoward: I'm just gonna factor out time somehow :) No worries. Nice that we have this design decision already :)

net/graphsync_fetcher.go
hannahhoward: appreciate this decision and agree. will make changes.

net/graphsync_fetcher.go
ZenGround0: > another option would be to track actual latency

Not sure how it fits into priorities and I'm guessing this comes after other fires are put out but I love the idea of making this decision using observations.   I'm imagining gathering a distribution of latencies over different data (single block, long chains) separating into "healthy" and "unhealthy" connections and taking the cutoff point to be some threshold (95% of latencies grouped into "healthy connections" for example).

net/graphsync_fetcher.go
```diff
@@ -55,23 +55,41 @@ type graphsyncFallbackPeerTracker interface {
...
+	return func(gsf *GraphSyncFetcher) {
+		gsf.unresponsiveTimeout = timeout
+	}
 }
 
 // NewGraphSyncFetcher returns a GraphsyncFetcher wired up to the input Graphsync exchange and
 // attached local blockservice for reloading blocks in memory once they are returned
 func NewGraphSyncFetcher(ctx context.Context, exchange GraphExchange, blockstore bstore.Blockstore,
-	bv consensus.SyntaxValidator, pt graphsyncFallbackPeerTracker) *GraphSyncFetcher {
+	bv consensus.SyntaxValidator, pt graphsyncFallbackPeerTracker, options ...GraphsyncFetcherOption) *GraphSyncFetcher {
```
ZenGround0: It seems off to use the function options pattern for setting the timeout dependency while setting the other dependencies with explicit inputs.  I could be missing the differentiating factor though.

hannahhoward:
@anorth this should have issues resolved now

net/graphsync_fetcher.go
```diff
@@ -60,18 +62,20 @@ type GraphSyncFetcher struct {
...
 	ssb         selectorbuilder.SelectorSpecBuilder
 	peerTracker graphsyncFallbackPeerTracker
+	systemClock clock.Clock
 }
 
 // NewGraphSyncFetcher returns a GraphsyncFetcher wired up to the input Graphsync exchange and
 // attached local blockservice for reloading blocks in memory once they are returned
 func NewGraphSyncFetcher(ctx context.Context, exchange GraphExchange, blockstore bstore.Blockstore,
-	bv consensus.SyntaxValidator, pt graphsyncFallbackPeerTracker) *GraphSyncFetcher {
+	bv consensus.SyntaxValidator, systemClock clock.Clock, pt graphsyncFallbackPeerTracker) *GraphSyncFetcher {
```
anorth: Annoying package/var name collision. How about `clk` for the variable?

net/graphsync_fetcher.go
```diff
@@ -223,19 +227,48 @@ func (gsf *GraphSyncFetcher) fetchBlocks(ctx context.Context, cids []cid.Cid, ta
...
+			}
+		}(requestChan, errChan, requestCancel)
 	}
-	// Any of the multiple parallel requests might fail. Wait for all of them to complete, then
-	// return any error (in this case, the last one to be received).
+	wg.Wait()
+	return anyError
+}
+
+func (gsf *GraphSyncFetcher) consumeResponse(requestChan <-chan graphsync.ResponseProgress, errChan <-chan error, cancelFunc func()) error {
```
anorth: Please briefly describe this method's intent in a comment.

PR #3461, opened 2019-09-18
feat(proofscache): prepare go-filecoin for env variable change

laser:
As per https://github.com/filecoin-project/rust-fil-proofs/issues/741, which has recently been completed and is live in rust-fil-proofs master, the `FILECOIN_PARAMETER_CACHE` environment variable has been replaced with `FIL_PROOFS_PARAMETER_CACHE`.


laser:
@frrist 

> Any reason to leave the FILECOIN_PARAMETER_CACHE env variables? I don't believe they are used by the infra repo.

You should definitely leave the `FILECOIN_PARAMETER_CACHE` as it is until you update the go-sectorbuilder SHA to use the newest, latest, and most-greatest version. Once you _do_ change to the newest version, `FILECOIN_PARAMETER_CACHE` will be obsoleted by `FIL_PROOFS_PARAMETER_CACHE`.

Make sense?

anorth:
@laser assuming this is still relevant, could you please bring it in for landing?

anorth:
@laser Should we be landing this now, or only when next integrating the storage market?

laser:
> @laser Should we be landing this now, or only when next integrating the storage market?

You should integrate it now. The version of libfilecoin that go-filecoin currently uses respects the new (`FIL_PROOFS_PARAMETER_CACHE`) environment variable.

PR #3467, opened 2019-09-19
[WIP] refactor Queryer object, rework chain and consensus packages.

acruikshank:
description forthcoming...


consensus/queryer.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Abstracts over a store of blockchain state.
+type chainStateChainReader interface {
+	GetHead() types.TipSetKey
+	GetTipSetState(context.Context, types.TipSetKey) (state.Tree, error)
+	GetTipSet(types.TipSetKey) (types.TipSet, error)
+}
+
+// ChainState knows how to send read-only messages for querying actor state.
+type ChainState struct {
```
ZenGround0: I'm not 100% this is the right place for this (maybe it should be in its own package).  However this abstraction is sorely needed and has my vote.

consensus/queryer.go
```diff
@@ -0,0 +1,82 @@
...
+	// To get the head tipset state root.
+	chainReader chainStateChainReader
+	// To load the tree for the head tipset state root.
+	cst *hamt.CborIpldStore
```
ZenGround0: Why is this included?  I don't see it being used.

PR #3468, opened 2019-09-19
feat(clock): add timer functionality

hannahhoward:
# Goals

Adds timer functionality to clock and fake clock, so we can use timers but still test them easily

# Implementation

Based off https://github.com/jonboulle/clockwork/pull/22 -- merges code in from base library and PR, so that we are not dependent on https://github.com/jonboulle/clockwork which seems to be somewhat dead.

# For Discussion

This is a bunch of code but the vast majority of it has moved over with only minor modifications from the source library.
 
Here are the deltas from the source library:
- everything related to the fake clock is moved over to testhelpers
- I used the existing factory method names we were using in our code
- the tests have been updated to use testify
- I renamed "sleeper" to "timer" cause it make more sense to me
- I removed one test of a private function because we intentionally keep our tests in a different module than the code
- I rewrote one of the tests so that it tested more accurately the behavior of BlockUntil (it's still not perfect, but should be closer to capturing actual behavior)

Will be integrated into: https://github.com/filecoin-project/go-filecoin/pull/3460 once merged

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -2,11 +2,18 @@ package clock
 
 import (
 	"github.com/jonboulle/clockwork"
+	"time"
 )
 
 // Clock provides an interface that packages can use instead of directly
 // using the time module, so that chronology-related behavior can be tested.
 type Clock interface {
+	clockwork.Clock
+	NewTimer(d time.Duration) Timer
+	AfterFunc(d time.Duration, f func()) Timer
+}
+
+type realClock struct {
 	clockwork.Clock
 }
 
@@ -14,5 +21,13 @@ type Clock interface {
 // SystemClock is a Clock which simply delegates calls to the actual time
 // package; it should be used by packages in production.
 func NewSystemClock() Clock {
-	return clockwork.NewRealClock()
+	return &realClock{Clock: clockwork.NewRealClock()}
+}
+
+func (rc *realClock) NewTimer(d time.Duration) Timer {
+	return &realTimer{time.NewTimer(d)}
+}
+
+func (rc *realClock) AfterFunc(d time.Duration, f func()) Timer {
+	return &realTimer{time.AfterFunc(d, f)}
 }
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -1,27 +1,44 @@
 package clock
 
 import (
-	"github.com/jonboulle/clockwork"
 	"time"
 )
 
 // Clock provides an interface that packages can use instead of directly
-// using the time module, so that chronology-related behavior can be tested.
+// using the time module, so that chronology-related behavior can be tested
+// Adapted from: https://github.com/jonboulle/clockwork
 type Clock interface {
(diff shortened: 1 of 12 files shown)


hannahhoward:
@anorth btw my inclination is not to copy over the whole clockwork package ATM cause that seems like something of an undertaking.

hannahhoward:
@anorth follow-up, actually gonna go ahead and bring over that code. :)

testhelpers/clock_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	th "github.com/filecoin-project/go-filecoin/testhelpers"
+	tf "github.com/filecoin-project/go-filecoin/testhelpers/testflags"
+)
+
+func TestFakeClockAfter(t *testing.T) {
+	tf.UnitTest(t)
+	fc := th.NewFakeSystemClock(time.Now())
```
frrist: nit: It's probably worth using a constant time value here (and everywhere else we use `time.Now()` in this file) , e.g. `time.Unix(123456789,0)`

clock/clock.go
```diff
@@ -1,18 +1,57 @@
...
 
 import (
-	"github.com/jonboulle/clockwork"
+	"time"
 )
 
 // Clock provides an interface that packages can use instead of directly
-// using the time module, so that chronology-related behavior can be tested.
+// using the time module, so that chronology-related behavior can be tested
+// Adapted from: https://github.com/jonboulle/clockwork
```
anorth: @ianjdarrow Could you please confirm that this is all that's required here? I was expecting more, like a reference to the license under which we've copied and adapted this code, a copyright line, ...

clock/clock.go
```diff
@@ -1,18 +1,57 @@
...
+
+func (rc *realClock) NewTimer(d time.Duration) Timer {
+	return &realTimer{time.NewTimer(d)}
+}
+
+func (rc *realClock) AfterFunc(d time.Duration, f func()) Timer {
+	return &realTimer{time.AfterFunc(d, f)}
 }
 
 // NewSystemClock returns a SystemClock that delegates calls to the jonboulee/clockwork package.
```
anorth: Fix comment

testhelpers/clock.go
```diff
@@ -1,19 +1,291 @@
...
-	"github.com/jonboulle/clockwork"
+	"github.com/filecoin-project/go-filecoin/clock"
 )
 
-// FakeSystemClock wrapps the jonboulle/clockwork.FakeClock interface.
 // FakeSystemClock provides an interface for a clock which can be
-// manually advanced through time.
+// manually advanced through time
+// Adapted from: https://github.com/jonboulle/clockwork
 type FakeSystemClock interface {
```
anorth: Since you're making big changes here, rename this to `FakeClock`

PR #3469, opened 2019-09-19
Miner addFaults method

### LINKED ISSUE #3425: addFaults miner actor method
### Description
For Rational PoSt, the Storage miner actor needs to implement the [`addFaults`](https://github.com/filecoin-project/specs/blob/master/actors.md#addfaults) method.

This will include adding `currentFaultSet`, `nextFaultSet` to the miner state.

### Acceptance criteria
- `addFaults` method implemented with gas tested, as close as possible to the spec.
- new fields in Miner state.

ingar:
## Summary
Add the Miner [`addFaults`](https://github.com/filecoin-project/specs/blob/069882c3e0c61ff19512879a6ceb0a422835a523/actors.md#addfaults) method, and use these faults in `submitPoSt`.

## Notes
The code is based on the spec and is not validating that the faulted sector ids are in the current sector commitments.  I've confirmed that the sector builder can handle bogus sector ids, but it's easy enough to add the validation if that seems prudent.

Also, [`submitPoSt`](https://github.com/filecoin-project/specs/blob/069882c3e0c61ff19512879a6ceb0a422835a523/actors.md#submitpost) no longer needs the `faults` parameter but we [still have it in the codebase](https://github.com/filecoin-project/go-filecoin/blob/2879caa02d9d9be6dbb36cae05de5b447e92ca68/actor/builtin/miner/miner.go#L273).

resolves #3425

### DIFF
--- a/actor/builtin/miner/miner.go
+++ b/actor/builtin/miner/miner.go
@@ -269,6 +269,10 @@ var minerExports = exec.Exports{
 		Params: []abi.Type{},
 		Return: []abi.Type{abi.BytesAmount},
 	},
+	"addFaults": &exec.FunctionSignature{
+		Params: []abi.Type{abi.FaultSet},
+		Return: []abi.Type{},
+	},
 	"submitPoSt": &exec.FunctionSignature{
 		Params: []abi.Type{abi.PoStProof, abi.FaultSet, abi.IntSet},
 		Return: []abi.Type{},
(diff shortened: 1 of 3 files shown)


actor/builtin/miner/miner.go
```diff
@@ -854,6 +858,33 @@ func (ma *Actor) GetActiveCollateral(ctx exec.VMContext) (types.AttoFIL, uint8,
...
+	if err != nil {
+		return errors.CodeError(err), err
+	}
+
+	return 0, nil
+}
+
 // SubmitPoSt is used to submit a coalesced PoST to the chain to convince the chain
 // that you have been actually storing the files you claim to be.
 func (ma *Actor) SubmitPoSt(ctx exec.VMContext, poStProof types.PoStProof, faults types.FaultSet, done types.IntSet) (uint8, error) {
```
acruikshank: This seems a reasonable time to remove faults from the signature.

anorth:
I don't think we can merge this as-is without protocol upgrade infrastructure. I'm interested in landing all the bits that *don't* require that, but we can't change the signature or code of `SubmitPoSt`. I'm ok with adding an un-called method, but that might not be worthwhile without the rest of it.

@acruikshank this might need to go on hold until a protocol upgrade table allows us to ship code without this immediately going into effect on 0.5.

ingar:
@anorth I pruned down this PR to just the `addFaults` method.  I have the other changes to `submitPoSt` ready to go for the protocol upgrade path.

PR #3470, opened 2019-09-20
do not close connection if hello not supported

frrist:
This will allow filecoin nodes to continue to connect to the stats dashboard and other non-hello protocol speaking services.


protocol/hello/hello.go
```diff
@@ -165,7 +165,6 @@ func (hn *helloNotify) Connected(n net.Network, c net.Conn) {
 		hello, err := hn.hello().ReceiveHello(ctx, from)
 		if err != nil {
 			log.Debugf("failed to receive hello handshake from peer %s: %s", from, err)
-			_ = c.Close()
```
frrist: From the previous work done here, we never closed the _connection_ if the hello handshake failed: https://github.com/filecoin-project/go-filecoin/pull/3327/files#diff-a348bd8f91eeee23380b03033d246a96L180

PR #3471, opened 2019-09-20
fix a bug where CID != Seralized IPLD signed message

frrist:


frrist:
> Should we do this for metered message too in case we ever need those cids?

hmm, good question- i'm not sure. I don't think Metered-Messages should have a CID or ToNode method, at least I can't see a reason for it. 

Actually, this is making me question a lot of things about the design of the Message* types. I'm having a hard time coming up with a scenario where we would want the CID of a message or metered message since we deal almost exclusively with signed messages.

ZenGround0:
> I'm having a hard time coming up with a scenario where we would want the CID of a message or metered message since we deal almost exclusively with signed messages.

I agree with this, nothing jumps out at me.

>  I don't think Metered-Messages should have a CID or ToNode method

I agree with this.  But the metered messages do have the unintuitive `Cid` and `ToNode` methods that operate on the inner message data without GasPrice and GasLimit.  It might be worth adding these methods to metered message as a safeguard against unexpected future use cases that call these methods.

PR #3472, opened 2019-09-20
fixed race in peer update

frrist:
Wrote some racy code that flaked in CI, here's a fix for it. I would be great to run go's [race-detector](https://blog.golang.org/race-detector) as a part of CI.

### DIFF
--- a/net/peer_tracker_test.go
+++ b/net/peer_tracker_test.go
@@ -147,9 +147,6 @@ func TestUpdateWithErrors(t *testing.T) {
 	tracker.Track(ci2)
 	tracker.Track(ci3)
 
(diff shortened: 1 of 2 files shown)


net/peer_tracker_test.go
```diff
@@ -159,6 +156,9 @@ func TestUpdateWithErrors(t *testing.T) {
 	err := tracker.UpdateTrusted(context.Background())
 	assert.Error(t, err, "all updates failed")
 
+	updatedHead := types.NewTipSetKey(types.CidFromString(t, "UPDATE"))
```
frrist: This has nothing to do with the bug, just moving associated code closer together.

PR #3473, opened 2019-09-20
Export and Import chain to/from a .car file

frrist:
### Goals
Allow a go-filecoin node to export a chain to a [car file](https://github.com/ipfs/go-car) and import a chain from a car file.

### Non-Goals
Allow a go-filecoin node to initialize from a car file (filed #3554 to address this)

### Implementation
- A new file: `car.go`  was added to the chain/ package. This file contains all logic for importing and exporting a chain.
- Two new sub-commands were added to the `chain` command:
```
go-filecoin chain export <file> <TipSetKey...> - Export the chain store to a car file.
go-filecoin chain import <file> - Import the chain from a car file.
```
- Export is an operation on the chain store, it writes its internal store to a car file.
- Import is an operation on the chain store, it imports the contents of a car file into its blockstore.

### Follow-on issues:
#3554
#3555

### DIFF
--- a/chain/car.go
+++ b/chain/car.go
@@ -1,0 +1,102 @@
+package chain
+
+import (
+	"context"
+	"io"
+
+	blocks "github.com/ipfs/go-block-format"
+	car "github.com/ipfs/go-car"
+	carutil "github.com/ipfs/go-car/util"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	logging "github.com/ipfs/go-log"
+
+	"github.com/filecoin-project/go-filecoin/types"
+)
+
+var logCar = logging.Logger("chain/car")
+
+type carChainReader interface {
+	GetTipSet(types.TipSetKey) (types.TipSet, error)
+}
+type carMessageReader interface {
+	MessageProvider
+}
+
+// Export will export a chain (all blocks and their messages) to the writer `out`.
+func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carMessageReader, out io.Writer) (types.TipSetKey, error) {
+	// fail if headTS isn't in the store.
+	if _, err := cr.GetTipSet(headTS.Key()); err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+
+	// Write the car header
+	chb, err := cbor.DumpObject(car.CarHeader{
+		Roots:   headTS.Key().ToSlice(),
+		Version: 1,
+	})
+	if err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+
+	logCar.Debugf("car file chain head: %s", headTS.Key())
+	if err := carutil.LdWrite(out, chb); err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+
+	iter := IterAncestors(ctx, cr, headTS)
+	// accumulate TipSets in descending order.
+	for ; !iter.Complete(); err = iter.Next() {
+		if err != nil {
+			return types.UndefTipSet.Key(), err
+		}
+		tip := iter.Value()
+		// write blocks
+		for i := 0; i < tip.Len(); i++ {
+			hdr := tip.At(i)
+			logCar.Debugf("writing block: %s", hdr.Cid())
+			if err := carutil.LdWrite(out, hdr.Cid().Bytes(), hdr.ToNode().RawData()); err != nil {
+				return types.UndefTipSet.Key(), err
+			}
+
+			msgs, err := mr.LoadMessages(ctx, hdr.Messages)
+			if err != nil {
+				return types.UndefTipSet.Key(), err
+			}
+
+			if len(msgs) > 0 {
+				logCar.Debugf("writing message collection: %s", hdr.Messages)
+				if err := carutil.LdWrite(out, hdr.Messages.Bytes(), types.MessageCollection(msgs).ToNode().RawData()); err != nil {
+					return types.UndefTipSet.Key(), err
+				}
+			}
+
+			// TODO(#3473) we can remove MessageReceipts from the exported file once addressed.
+			rect, err := mr.LoadReceipts(ctx, hdr.MessageReceipts)
+			if err != nil {
+				return types.UndefTipSet.Key(), err
+			}
+
+			if len(rect) > 0 {
+				logCar.Debugf("writing message-receipt collection: %s", hdr.Messages)
+				if err := carutil.LdWrite(out, hdr.MessageReceipts.Bytes(), types.ReceiptCollection(rect).ToNode().RawData()); err != nil {
+					return types.UndefTipSet.Key(), err
+				}
+			}
+		}
+	}
+	return headTS.Key(), nil
+}
+
+type carStore interface {
+	Put(blocks.Block) error
+}
+
+// Import imports a chain from `in` to `bs`.
+func Import(ctx context.Context, cs carStore, in io.Reader) (types.TipSetKey, error) {
+	header, err := car.LoadCar(cs, in)
+	if err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+	headKey := types.NewTipSetKey(header.Roots...)
+	return headKey, nil
+}
--- a/chain/car.go
+++ b/chain/car.go
@@ -7,6 +7,7 @@ import (
 	blocks "github.com/ipfs/go-block-format"
 	car "github.com/ipfs/go-car"
 	carutil "github.com/ipfs/go-car/util"
+	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	logging "github.com/ipfs/go-log"
 
@@ -24,6 +25,9 @@ type carMessageReader interface {
 
 // Export will export a chain (all blocks and their messages) to the writer `out`.
 func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carMessageReader, out io.Writer) error {
+	// ensure we don't duplicate writes to the car file. // e.g. only write EmptyMessageCID once.
+	filter := make(map[cid.Cid]bool)
+
 	// fail if headTS isn't in the store.
 	if _, err := cr.GetTipSet(headTS.Key()); err != nil {
 		return err
@@ -54,8 +58,12 @@ func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carM
 		for i := 0; i < tip.Len(); i++ {
 			hdr := tip.At(i)
 			logCar.Debugf("writing block: %s", hdr.Cid())
-			if err := carutil.LdWrite(out, hdr.Cid().Bytes(), hdr.ToNode().RawData()); err != nil {
-				return err
+
+			if !filter[hdr.Cid()] {
+				if err := carutil.LdWrite(out, hdr.Cid().Bytes(), hdr.ToNode().RawData()); err != nil {
+					return err
+				}
+				filter[hdr.Cid()] = true
 			}
 
 			msgs, err := mr.LoadMessages(ctx, hdr.Messages)
@@ -63,11 +71,12 @@ func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carM
 				return err
 			}
 
-			if len(msgs) > 0 {
+			if !filter[hdr.Messages] {
 				logCar.Debugf("writing message collection: %s", hdr.Messages)
 				if err := carutil.LdWrite(out, hdr.Messages.Bytes(), types.MessageCollection(msgs).ToNode().RawData()); err != nil {
 					return err
 				}
+				filter[hdr.Messages] = true
 			}
 
 			// TODO(#3473) we can remove MessageReceipts from the exported file once addressed.
@@ -76,11 +85,12 @@ func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carM
 				return err
 			}
 
-			if len(rect) > 0 {
+			if !filter[hdr.MessageReceipts] {
 				logCar.Debugf("writing message-receipt collection: %s", hdr.Messages)
 				if err := carutil.LdWrite(out, hdr.MessageReceipts.Bytes(), types.ReceiptCollection(rect).ToNode().RawData()); err != nil {
 					return err
 				}
+				filter[hdr.MessageReceipts] = true
 			}
 		}
 	}
--- a/chain/car.go
+++ b/chain/car.go
@@ -23,10 +23,10 @@ type carMessageReader interface {
 }
 
 // Export will export a chain (all blocks and their messages) to the writer `out`.
-func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carMessageReader, out io.Writer) (types.TipSetKey, error) {
+func Export(ctx context.Context, headTS types.TipSet, cr carChainReader, mr carMessageReader, out io.Writer) error {
 	// fail if headTS isn't in the store.
 	if _, err := cr.GetTipSet(headTS.Key()); err != nil {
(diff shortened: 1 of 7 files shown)


node/builder.go
```diff
@@ -299,7 +299,7 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
 	nd.PorcelainAPI = porcelain.New(plumbing.New(&plumbing.APIDeps{
 		Bitswap:       bswap,
 		Chain:         chainState,
-		Sync:          cst.NewChainSyncProvider(chainSyncer),
+		Sync:          cst.NewChainSyncProvider(chainSyncer, bs, peerHost.ID()),
```
frrist: The addition of the blockstore will likely be the most controversial change in this PR, here's why we are doing it: https://github.com/filecoin-project/go-filecoin/pull/3473/files#r327391073

chain/car.go
```diff
@@ -0,0 +1,99 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// Import imports a chain from `in` to `bs`.
+func Import(ctx context.Context, bs blockstore.Blockstore, in io.Reader) (types.TipSetKey, error) {
+	header, err := car.LoadCar(bs, in)
```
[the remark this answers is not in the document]
frrist: PR on go-car to reduce interface requirements: https://github.com/ipfs/go-car/pull/15

commands/chain_daemon_test.go
```diff
@@ -127,3 +134,71 @@ func TestChainLs(t *testing.T) {
 		assert.Contains(t, chainLsResult, `"height":"1"`)
 	})
 }
+
+func TestChainImportExport(t *testing.T) {
```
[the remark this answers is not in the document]
travisperson: More blocks are being mined than you expect. `series.SendFilecoinDefaults` calls `series.SendFilecoinFromDefault` which will mine a block after each `MessageSend`. So in your for loop 40 blocks are being mined.

commands/chain_daemon_test.go
```diff
@@ -127,3 +134,71 @@ func TestChainLs(t *testing.T) {
...
+	series.CtxMiningOnce(ctx)
+
+	// create block with 1 message
+	require.NoError(t, series.SendFilecoinDefaults(ctx, genesisNode, exportNode, 10))
+	series.CtxMiningOnce(ctx)
+
+	// create block with many messages
+	for i := 0; i < 20; i++ {
+		require.NoError(t, series.SendFilecoinDefaults(ctx, genesisNode, exportNode, 10))
+		require.NoError(t, series.SendFilecoinDefaults(ctx, exportNode, genesisNode, 10))
```
frrist: Will change this to use the SendMessage action instead as these series mine blocks.

chain/car.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	iter := IterAncestors(ctx, cr, headTS)
+	// accumulate TipSets in descending order.
+	for ; !iter.Complete(); err = iter.Next() {
+		if err != nil {
+			return err
+		}
+		tip := iter.Value()
+		// write block
```
ZenGround0: super nit: "blocks"

chain/car.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			if len(msgs) > 0 {
+				logCar.Debugf("writing message collection: %s", hdr.Messages)
+				if err := carutil.LdWrite(out, hdr.Messages.Bytes(), types.MessageCollection(msgs).ToNode().RawData()); err != nil {
+					return err
+				}
+			}
+
+			// TODO(#3473) we can remove MessageReceipts from the exported file once addressed.
```
ZenGround0: 👍

plumbing/cst/chain_state.go
```diff
@@ -161,3 +162,8 @@ func (chn *ChainStateProvider) GetActorSignature(ctx context.Context, actorAddr
 
 	return export, nil
 }
+
+// ChainExport exports trhe nodes current chain to `out`
```
ZenGround0: super nit: "the"

plumbing/cst/chain_sync.go
```diff
@@ -37,3 +48,26 @@ func (chs *ChainSyncProvider) Status() chain.Status {
 func (chs *ChainSyncProvider) HandleNewTipSet(ctx context.Context, ci *types.ChainInfo, trusted bool) error {
 	return chs.sync.HandleNewTipSet(ctx, ci, trusted)
 }
+
+// ChainImport imports a chain with the same genesis block from `in`.
+func (chs *ChainSyncProvider) ChainImport(ctx context.Context, in io.Reader) error {
```
ZenGround0: In my view this does too much.  To keep separation of concerns and a predictable CLI I want `import` to wrap only the `chain.Import` call and return the headKey.  Then I want the `HandleNewTipSet` call to be the single target of a new `chain sync` command.

Ideally `chain export` would also take in a tipset key for exporting and users would run `chain head` first to get that key to get the behavior we have here.  This is less important to me than separating data IO from scheduling sync jobs.

anorth: I concur with all of these.

Importing is something one does with a chain store, not a syncer. It should be parallel with export. The `chain sync` command is then already sufficient to switch to the imported head if heavier.

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+	cb := chain.NewBuilder(t, address.Undef)
+
+	gene := cb.NewGenesis()
+	ts10 := cb.AppendManyOn(9, gene)
+
+	var buf bytes.Buffer
+	carW := bufio.NewWriter(&buf)
+
+	// export the car file to a buffer
+	exportedKey, err := chain.Export(ctx, ts10, cb, cb, carW)
```
anorth: There are too many too-short var names here, please expand some of them.

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+		mm.NewSignedMessage(alice, 3),
+		mm.NewSignedMessage(alice, 4),
+		mm.NewSignedMessage(alice, 5),
+	}
+	rcts := types.EmptyReceipts(5)
+	ts11 := cb.BuildOneOn(ts10, func(b *chain.BlockBuilder) {
+		b.AddMessages(
+			msgs,
+			rcts,
+		)
```
anorth: Unwrap

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+	})
+
+	var buf bytes.Buffer
+	carW := bufio.NewWriter(&buf)
+
+	// export the car file to a buffer
+	exportedKey, err := chain.Export(ctx, ts11, cb, cb, carW)
+	assert.NoError(t, err)
+	assert.Equal(t, ts11.Key(), exportedKey)
+	require.NoError(t, carW.Flush())
```
anorth: Factor out this repeated blob of code to an `exportToBuffer()` function.

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	mds := ds.NewMapDatastore()
+	bstore := blockstore.NewBlockstore(mds)
+
+	// import the car file from the buffer
+	carR := bufio.NewReader(&buf)
+	importedKey, err := chain.Import(ctx, bstore, carR)
+	assert.NoError(t, err)
+	assert.Equal(t, ts11.Key(), importedKey)
```
anorth: Factor out a helper

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+	assert.Equal(t, ts12.Key(), importedKey)
+
+	// walk the blockstore and assert it had all blocks imported
+	validateBlockstoreImport(t, ts12.Key(), gene.Key(), bstore)
+}
+
+func validateBlockstoreImport(t *testing.T, start, stop types.TipSetKey, bstore blockstore.Blockstore) {
+	// walk the blockstore and assert it had all blocks imported
+	cur := start
+	for !cur.Equals(stop) {
```
anorth: This fails to check that the genesis (`stop`) is included

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+		parents := cid.NewSet()
+		for _, c := range cur.ToSlice() {
+			bsBlk, err := bstore.Get(c)
+			assert.NoError(t, err)
+			blk, err := types.DecodeBlock(bsBlk.RawData())
+			assert.NoError(t, err)
+
+			if !blk.Messages.Equals(types.EmptyMessagesCID) {
```
anorth: The store should have an object corresponding to EmptyMessagesCID - make it so and remove the special case. Ditto for receipts.

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+			if !blk.MessageReceipts.Equals(types.EmptyReceiptsCID) {
+				bsRcts, err := bstore.Get(blk.MessageReceipts)
+				assert.NoError(t, err)
+				_, err = types.DecodeReceipts(bsRcts.RawData())
+				assert.NoError(t, err)
+			}
+
+			for _, p := range blk.Parents.ToSlice() {
+				parents.Add(p)
+			}
```
anorth: The CidSet is unnecessary, just use a slice. All blocks in a tipset should have the same parent tipset key (you can require this).

chain/car_test.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+func TestChainImportExportSimple(t *testing.T) {
+	tf.UnitTest(t)
+	ctx := context.Background()
+
+	cb := chain.NewBuilder(t, address.Undef)
+
+	gene := cb.NewGenesis()
+	ts10 := cb.AppendManyOn(9, gene)
```
anorth: 9 is an arbitrary number here; how is it better than 8?
Cases that are worth testing:
- just the genesis
- genesis plus singleton tip
- genesis plus two singleton tips
- genesis plus a two-block tip
- [gen, 2blocks, 1block]
- [gen, 2block, 2blocks]

Factor out helper methods aggressively so that these cases are described as succinctly as possible.

Some exercising of the messages/nomessages cases as you already have is good too.

plumbing/api.go
```diff
@@ -194,6 +194,16 @@ func (api *API) ChainSyncHandleNewTipSet(ctx context.Context, ci *types.ChainInf
 	return api.syncer.HandleNewTipSet(ctx, ci, trusted)
 }
 
+// ChainExport exports the chain store to `out`.
+func (api *API) ChainExport(ctx context.Context, head types.TipSetKey, out io.Writer) (types.TipSetKey, error) {
+	return api.chain.ChainExport(ctx, head, out)
+}
+
+// ChainImport imports the chain with the same genesis block from `in`.
```
anorth: I don't understand the "with the same genesis block" comment. Is that old?

plumbing/cst/chain_state.go
```diff
@@ -50,6 +68,7 @@ func NewChainStateReadWriter(crw chainReadWriter, messages chain.MessageProvider
 	return &ChainStateReadWriter{
 		readWriter:      crw,
 		cst:             cst,
+		store:           newCarStore(cst),
```
anorth: Create this on the fly rather than store it here - it's no actually new information nor expensive to create.

plumbing/cst/chain_state.go
```diff
@@ -171,3 +190,29 @@ func (chn *ChainStateReadWriter) SetHead(ctx context.Context, key types.TipSetKe
 	}
 	return chn.readWriter.SetHead(ctx, headTs)
 }
+
+// ChainExport exports nodes current chain to `out`
```
anorth:
```suggestion
// ChainExport exports the chain from `head` up to and including the genesis block to `out`
```

plumbing/cst/chain_state.go
```diff
@@ -171,3 +190,29 @@ func (chn *ChainStateReadWriter) SetHead(ctx context.Context, key types.TipSetKe
...
+func (chn *ChainStateReadWriter) ChainExport(ctx context.Context, head types.TipSetKey, out io.Writer) (types.TipSetKey, error) {
+	headTS, err := chn.GetTipSet(head)
+	if err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+	logStore.Info("starting CAR file export")
```
anorth: Include the head key

plumbing/cst/chain_state.go
```diff
@@ -171,3 +190,29 @@ func (chn *ChainStateReadWriter) SetHead(ctx context.Context, key types.TipSetKe
...
+	headKey, err := chain.Export(ctx, headTS, chn.readWriter, chn.messageProvider, out)
```
anorth: Maybe I misled with an earlier comment - there's no need to return the head key if it's already passed as an argument.

plumbing/cst/chain_state.go
```diff
@@ -171,3 +190,29 @@ func (chn *ChainStateReadWriter) SetHead(ctx context.Context, key types.TipSetKe
...
+	if err != nil {
+		return types.UndefTipSet.Key(), err
+	}
+	logStore.Infof("exported CAR file with head: %s", headKey)
+	return headKey, nil
+}
+
+// ChainImport imports a chain with the same genesis block from `in`.
```
anorth:
```suggestion
// ChainImport imports all blocks from CAR file `in` into the store. It returns the set of roots from the CAR as a TipSetKey.
```

plumbing/api.go
frrist: Old sorry. Import would fail when syncing the chain if the genesis block was different, since Import no longer syncs the imported chain this comment can be removed. Although (somewhat obviously) it still holds true that an imported chain with a different genesis block will fail to sync.

frrist:
Follow on issues:
https://github.com/filecoin-project/go-filecoin/issues/3554
https://github.com/filecoin-project/go-filecoin/issues/3555

chain/car_test.go
frrist: addressed.

chain/car_test.go
frrist: Having trouble groking this comment, EmptyMessage CID's aren't persisted to the store for each block if that's what you are getting at?

anorth: As discussed in person, I expected the CAR (content-addressed archive) not to store duplicate records with the same content, but I might be wrong. We want to store the empty-message-list object exactly once, but then there's no need for any special cases (for a case that isn't really very special).

frrist: >  I expected the CAR (content-addressed archive) not to store duplicate records with the same content

From my understanding this is unfortunately not the case. The method ([LDWrite](https://github.com/ipfs/go-car/blob/master/util/util.go#L69)) we are using to persist the chain doesn't deduplicate writes for the same content.

I can modify `chain/car.go` to persist the empty message/receipt collections, but that will mean duplicating their content in the car file for each block containing the empty set.

anorth: Duplicating the blocks is not a good option. Two ideas:
- Wrap the CAR writer with a stateful structure that records all the keys written and de-dups
- Write the empty collections once at the beginning of export, then retain the export-side special case to avoid duplicating it.

The former removes all the special case handling. The latter removes only the import-side special cases.

chain/car.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			secpMsgs, blsMsgs, err := mr.LoadMessages(ctx, hdr.Messages)
+			if err != nil {
+				return err
+			}
+
+			if !filter[hdr.Messages.SecpRoot] {
+				logCar.Debugf("writing message collection: %s", hdr.Messages)
+				if err := carutil.LdWrite(out, hdr.Messages.SecpRoot.Bytes(), types.MessageCollection(secpMsgs).ToNode().RawData()); err != nil {
```
anorth: FYI @acruikshank collides with #3562

PR #3482, opened 2019-09-23
Use Queryer in core services

acruikshank:
### Why this PR?

Like much of the code in core services, `CallQueryMethod` takes in a lot of state as parameters so that it can remain stateless, but still operate on chain state. As we move towards consensus upgrades, even more state would need to be plumbed through the app to call query methods which would mean even more initialization in tests. This PR aims to create a `Queryer` abstraction, so components in the core services layer do not need to manage so many naked dependencies.

The `Queryer` abstraction already exists in the plumbing layer. Unfortunately, this makes it unavailable to code that lies beneath it. This PR moves the `Queryer` interface down and ensures all non-test code uses it rather than call `CallQueryMethod` directly.

### What is in this PR?

This is a rather large refactor that touches a lot of core code. It is best approached through individual commits:

1. The first commit takes the Queryer out of plumbing and moves it down the dependency tree into `consensus`. It still exposes the interface through the plumbing API.
2. The second commit is primarily concerned with getting the PowerTableView to use the query interface. It also rernames `consensus.ChainState` to `consesnus.ActorState` to avoid confusion (sorry this wasn't a separate commit).  To make this work, we stop injecting `PowerTableView` as a dependency. Instead we inject an `ActorState` instance, then ask it for a `Queryer` that we can use to construct a `PowerTableView` that can only answer questions about a particular point in the chain. This breaks out responsibility into:
  *  `ActorState` maintains references to stores and can use them to produce a `Queryer` tied to a particular tip set.
  * `Queryer` maintains enough state to run queries at a particular tip set.
  * `PowerTableView` knows which queries to call to learn about miner power. It relies on it its internal `Queryer` to actually perform the queries.
3. Since PowerTableView is no longer dependency injected, this commit removes its interface. When we need need to test it, we create a test `Queryer` for an argument. 
4. The fourth commit simply converts the call to get network name to use the new interface. This is the last bit of production code that called `CallQueryMethod` directly.
5. While reviewing this PR, I noticed that `IsElectionWinner` no longer needs a blockstore. This commit eliminates it.


consensus/actor_state.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Query sends a read-only message against the state of the provided base tipset.
+func (q *ProcessorQueryer) Query(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
+	encodedParams, err := abi.ToEncodedValues(params...)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to encode message params")
+	}
+
+	r, ec, err := CallQueryMethod(ctx, q.st, q.vms, to, method, encodedParams, optFrom, q.height)
```
ZenGround0: Highly doubtful this belongs in this PR (or is even the right call) but it seems like putting `CallQueryMethod` on a processor object and having the *Processor*Queryer keep track of a processor internally is a good direction.

You might imagine the Processor keeps track of the history-dependent vms and new processing invocations take in the per-call state and height params which the `Queryer` could continue to track.

ZenGround0: More future design thoughts:

This also makes me think that using a variant of the `Queryer` (maybe even an unmodified `Queryer`) within the actual block processing code could make sense.  I.e. batched message application for processing blocks and tipsets could be an operation that is a method on a thing that takes in a `Queryer` and then uses the `Queryer` mutliple times to actually do the message application (along with block reward state mods etc).

This would potentially make the many layers of redirection within the processor `ProcessTipSet` --> `ProcessBlock` --> `ApplyMessagesAndPayRewards` --> `ApplyMessage` --> `attemptApplyMessage` more explicit and obvious by moving the redirection to specific types.  For example we could have a `BlockProcessor` that owns a queryer and logic for processing tipsets and blocks, a `Queryer` as defined here or slightly varied which manages running a single message through the state machine, and a `StateMachineProcessor`/`StateMachineRunner` that keeps the logic for `attemptApplyMessage` or that can be easily mocked and passed to the `Queryer`

ZenGround0: Also maybe the `Queryer` becomes the `StateMachineRunner` above, i.e. it makes more sense to group the `attemptApplyMessage` logic in directly and mock out the queryer than add another layer of indirection.  I think the clearest thing I can see is that running a message through the state machine and handling the block/tipset processing logic definitely deserve separate owning abstractions.

plumbing/api.go
```diff
@@ -235,11 +235,16 @@ func (api *API) MessagePreview(ctx context.Context, from, to address.Address, me
 	return api.msgPreviewer.Preview(ctx, from, to, method, params...)
 }
 
+// TODO: This should simply expose the queryer method and callers should then call Query
```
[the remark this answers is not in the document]
acruikshank: Good catch.

consensus/actor_state.go
```diff
@@ -0,0 +1,96 @@
...
+func NewActorState(chainReader chainStateChainReader, cst *hamt.CborIpldStore, bs bstore.Blockstore) *ActorState {
+	return &ActorState{chainReader, cst, bs}
+}
+
+type ActorStateQueryer interface {
+	Query(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error)
+}
+
+// Queryer returns a query interface to query the chain at a particular tipset
+func (cs ActorState) Queryer(ctx context.Context, baseKey types.TipSetKey) (ActorStateQueryer, error) {
```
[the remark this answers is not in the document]
acruikshank: Agreed that a concrete base class is more idiomatic. The problem is the components we want to test hold an ActorState, not a Queryer. If `ActorState` exposed a concrete class, we could create test code that produces a `ProcessorQueryer` (the concrete class) in a different way, but the `ProcessorQueryer` would behave the same way. This interface is what really allows us to simplify the testing of things that make queries.

One solution to this ugliness would be to break out things like `Expected` into the service that maintains the interface to actor state, and the specific actions that make queries (which would only depend on a queryer. These could be tested separately and wouldn't would eliminate the need for this interface. That's more of a refactor than I signed up for here.

consensus/actor_state.go
```diff
@@ -0,0 +1,96 @@
...
+	h, err := base.Height()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get the head tipset height")
+	}
+
+	return cs.stateTreeQueryer(st, types.NewBlockHeight(h)), nil
+}
+
+// stateTreeQueryer returns a query interface to query the chain for a particular state tree and optional block height
+func (cs ActorState) stateTreeQueryer(st state.Tree, bh *types.BlockHeight) ActorStateQueryer {
```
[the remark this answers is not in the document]
acruikshank: ditto on the response.

consensus/actor_state.go
```diff
@@ -0,0 +1,96 @@
+package consensus
```
ZenGround0: nit: I'd be slightly more inclined to call this file `queryer` as the actor state abstraction seems less essential than the `queryer` abstraction.

consensus/actor_state.go
acruikshank: Much of the the motivation of this PR comes from the fact that `CallQueryMethod` is a free function in the `consensus` package that just happens to live in the `processor.go` file. Otherwise, giving `ProcessorQueryer` a processor seems like a good idea.

My understanding of the current state is that `Expected` runs answers questions about chain state including chain weight and the validity of tip sets. `Processor` concerns itself with processing signed messages in various packages (blocks, slices, single) for state transitions or simply to query. The actual execution of a message is handled in the `vm` package. 

I think I sort of get where you're going with the refactoring ideas. Perhaps instead of a `Queryer`, we have a better `Processor` (Transitioner?) abstraction that is tied to some base state. The caller will be responsible for retrieving a `Processesor` with the correct state. They could then use the Processor to `play` messages in tip sets, blocks or individually. Conceivably the `Processor` would return a new `Processor` in addition to the results that would represent the new state after the messages had been played, so the processor's internal state tree would not need to mutate.

consensus/testing.go
```diff
@@ -33,28 +31,73 @@ type TestPowerTableView struct {
...
+	if method == "getTotalStorage" {
+		return [][]byte{tq.totalPower.Bytes()}, nil
+	} else if method == "getPower" {
+		// always return 1
+		return [][]byte{tq.minerPower.Bytes()}, nil
+	} else if method == "getWorker" {
+		if tq.minerToWorker != nil {
+			return [][]byte{tq.minerToWorker[to].Bytes()}, nil
+		}
+		// just return the miner address
```
ZenGround0: This behavior is a bit unexpected.  It would make this method more easy to predict (though a bit more burdensome on the user) if this special casing were not here.  If you keep this it would be good to call it out explicitly in a comment over the TestQuerier.

mining/block_generate.go
```diff
@@ -31,7 +31,12 @@ func (w *DefaultWorker) Generate(ctx context.Context,
 		return nil, errors.Wrap(err, "get state tree")
 	}
 
-	if !w.powerTable.HasPower(ctx, stateTree, w.blockstore, w.minerAddr) {
+	powerTable, err := w.getPowerTable(ctx, baseTipSet.Key())
+	if err != nil {
+		return nil, errors.Wrap(err, "get state tree")
```
ZenGround0: nit: this should say "get power table"

mining/worker.go
```diff
@@ -72,11 +72,12 @@ type MessageApplier interface {
 type workerPorcelainAPI interface {
 	BlockTime() time.Duration
 	MinerGetWorkerAddress(ctx context.Context, minerAddr address.Address, baseKey types.TipSetKey) (address.Address, error)
+	Queryer(ctx context.Context, baseKey types.TipSetKey) (consensus.ActorStateQueryer, error)
```
ZenGround0: Nice

plumbing/api.go
```diff
@@ -65,7 +65,7 @@ type API struct {
 type APIDeps struct {
 	Bitswap       exchange.Interface
 	Chain         *cst.ChainStateProvider
-	ChnState      *consensus.ChainState
+	ChnState      *consensus.ActorState
```
ZenGround0: nit: this should probably be renamed ActorState on the api

testhelpers/mining.go
```diff
@@ -33,6 +50,35 @@ func (t *TestWorkerPorcelainAPI) MinerGetWorkerAddress(_ context.Context, _ addr
 	return t.workerAddr, nil
 }
 
+// Queryer returns a queryer object for the given tipset
```
ZenGround0: This package imports consensus so why not resuse the test Queryer defined in `consensus/testing.go` which looks very similar?

consensus/power_table_view.go
```diff
@@ -3,83 +3,75 @@ package consensus
...
 	// WorkerAddr returns the address of the miner worker given the miner addr
-	WorkerAddr(context.Context, state.Tree, blockstore.Blockstore, address.Address) (address.Address, error)
+	WorkerAddr(context.Context, address.Address) (address.Address, error)
 }
 
 // MarketView is the power table view used for running expected consensus in
 // production.  It's methods use data from an input state's storage market to
 // determine power values in a chain.
-type MarketView struct{}
+type MarketView struct {
```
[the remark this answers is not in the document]
ZenGround0: Oooh nice looks like you do this in commit 3

consensus/expected_test.go
```diff
@@ -239,46 +239,30 @@ func setupCborBlockstore() (*hamt.CborIpldStore, blockstore.Blockstore) {
...
-	return tv.minerPower, errors.New("something went wrong with the miner power")
-}
-
-func (tv *FailingMinerTestPowerTableView) WorkerAddr(ctx context.Context, mAddr address.Address) (address.Address, error) {
-	return mAddr, nil
-}
-
-func (tv *FailingMinerTestPowerTableView) HasPower(ctx context.Context, mAddr address.Address) bool {
-	return true
+func (fq *FailingTestPowerTableViewQueryer) Query(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
```
ZenGround0: This is very similar to the `consensus/testing.go` `TestPowerTableViewQueryer`.  I wonder if adding the errors there and setting 0 power explicitly would help lower power table testing footprint.

consensus/power_table_view.go
```diff
@@ -13,40 +13,20 @@ import (
 
 // PowerTableView defines the set of functions used by the ChainManager to view
 // the power table encoded in the tipset's state tree
-type PowerTableView interface {
```
ZenGround0: 🎉 🎉 🌮

consensus/power_table_view.go
```diff
@@ -3,83 +3,75 @@ package consensus
...
-	}
-
 	return types.NewBytesAmountFromBytes(rets[0]), nil
 }
 
 // WorkerAddr returns the address of the miner worker given the miner address.
-func (v *MarketView) WorkerAddr(ctx context.Context, st state.Tree, bstore blockstore.Blockstore, mAddr address.Address) (address.Address, error) {
-	vms := vm.NewStorageMap(bstore)
-	rets, ec, err := CallQueryMethod(ctx, st, vms, mAddr, "getWorker", []byte{}, address.Undef, nil)
+func (v *MarketView) WorkerAddr(ctx context.Context, mAddr address.Address) (address.Address, error) {
```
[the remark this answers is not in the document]
acruikshank: I questioned the future of the PTV too, but in it's more limited role, I still like it. The responsibilities are: **Worker/Expected** have access to chain state and need to know about miner power in order to operate. **PTV** has limited access to chain state of a single tipset and knows how to ask chain state. It adds the logic to make specific queries about miner state.  **Queryer:** provices access to query chain state from a single tipset. I knows nothing about the content of these queries.

consensus/actor_state.go
acruikshank: I'm about 49% with you. What holds me back is that `ActorState` is the thing you actually construct and seems a more natural entry point. Breaking this up into multiple files might solve the problem, but I'd like to pollute the package directory as little as possible.

consensus/expected_test.go
acruikshank: Good call. Done

consensus/testing.go
acruikshank: Good call. It was inherited from the old test code, but was not actually being used. Deleted.

plumbing/api.go
acruikshank: Done.

testhelpers/mining.go
acruikshank: 👍

consensus/actor_state.go
```diff
@@ -0,0 +1,97 @@
...
```
anorth: This seems pretty independent of consensus. I think we need a package for the combination of chain and state primitives into things like this. I don't think the right package exists yet, so I'm flagging here that we want to make room for that. cc @icorderi

consensus/power_table_view_test.go
```diff
@@ -1,4 +1,4 @@
-package chain_test
+package consensus_test
```
anorth: Thanks :-/

consensus/testing.go
```diff
@@ -26,40 +21,63 @@ func RequireNewTipSet(require *require.Assertions, blks ...*types.Block) types.T
 	return ts
 }
 
-// TestPowerTableView is an implementation of the powertable view used for testing mining
-// wherein each miner has totalPower/minerPower power.
-type TestPowerTableView struct {
-	minerPower, totalPower *types.BytesAmount
-	minerToWorker          map[address.Address]address.Address
+// TestActorState provides a queryer that responds to power table view queries with the given parameters
+type TestActorState struct {
```
anorth: Nit: because `Test` is necessarily used as a verb in Go, I find structures called TestX to cause friction in my head. It's also a little bit too much on the side of a component telling the user how it should be used, rather than what it provides.

`FakeActorState` is unambiguous and unopinionated.

consensus/testing.go
```diff
@@ -26,40 +21,63 @@ func RequireNewTipSet(require *require.Assertions, blks ...*types.Block) types.T
...
+	minerPower    *types.BytesAmount
+	totalPower    *types.BytesAmount
+	minerToWorker map[address.Address]address.Address
 }
 
-// NewTestPowerTableView creates a test power view with the given total power
-func NewTestPowerTableView(minerPower *types.BytesAmount, totalPower *types.BytesAmount, minerToWorker map[address.Address]address.Address) *TestPowerTableView {
-	return &TestPowerTableView{minerPower: minerPower, totalPower: totalPower, minerToWorker: minerToWorker}
+// NewTestActorState creates an actor state that returns a test queryer such that PowerTableView queries return predefined results
```
anorth: Name confusion: I expect NewXxx to return an Xxx and the comment to match.

consensus/testing.go
```diff
@@ -26,40 +21,63 @@ func RequireNewTipSet(require *require.Assertions, blks ...*types.Block) types.T
...
+		totalPower:    t.totalPower,
+		minerToWorker: t.minerToWorker,
+	}
 }
 
-// Miner always returns value that was supplied to NewTestPowerTableView.
-func (tv *TestPowerTableView) Miner(ctx context.Context, st state.Tree, bstore blockstore.Blockstore, mAddr address.Address) (*types.BytesAmount, error) {
-	return tv.minerPower, nil
+// TestPowerTableViewQueryer returns a queryer that can be fed into a PowerTableView to produce specific values
+type TestPowerTableViewQueryer struct {
```
anorth: Ditto, sounds like a verb phrase.

plumbing/api.go
anorth: Or use a name like `ChainState` for the object.

Issue #3269, opened 2019-08-22
Persistent journal for event recording, inspection


anorth:
### Description
For a couple of different reasons now, there's a desire for a system for recording some event-type information for subsequent inspection:
- #2443 recording CLI commands to keep track of piece->miner mappings (and other info)
- #3230 recording messages that enter and exit the outbox (for testing and manual inspection)

In both cases:
- the events should be persisted to disk and survive node restarts
- the events carry structured information
- data size is fairly small

I think this calls for a *journal* of structured (but schema-free) data, persisted to the repo. I am avoiding calling this a log because, unlike application logs, we don't need levels, formatters, and all the configuration options thereby made possible.

I propose a journal facility for recording events:
- each event has a topic (string) and payload (struct serialised to JSON)
- each topic is persisted to a file in `repo/journal/<topic>`, one JSON object per line
- the journaller might add a `timestamp` to every record for convenience

In the near term, the interface to this can probably be a simple one-method interface, 
```
  Record(topic string, payload interface{})
```
Errors are unlikely to be actionable, so the journaller can log them but allow the caller to move on. If we later implement an event bus (#2309) that might be an even lighter coupling way of emitting events.

For FAST testing, we also need a mechanism to read the journal file (ideally, streaming like `tail -f`). The files would be human readable, so we probably don't need to add any CLI access.

@travisperson @frrist @acruikshank I'd welcome your thoughts and critique here.

@deaswang you have previously proposed something similar. After we get some feedback on this proposal, I'd welcome your contributions to the implementation.

### Acceptance criteria
TBC

### Risks + pitfalls
Do we really need this? Is there a better way to solve the motivating problems?

acruikshank:
In general, I think the format is good. I think adding a timestamp will be essential, since any type of forensics will require us to interleave topics to understand the sequence of events. I can make some arguments for putting all the topics in a single file, but, with timestamps, it becomes just a matter of having the right tools.

I don't think this is how we should handle piece->miner mappings for reasons described in #2443.

I do think the command history (hopefully command+API history soon) will be useful for troubleshooting, so my concerns about the above are somewhat irrelevant to this discussion.

I do like the idea of inspecting the message log (or the command log) for FAST testing, and having a well specified format, structured logs and some tooling around that should make that straightforward, so 👍.

frrist:
+1 to timestamps and ndjson.

[...]

PR #3485, opened 2019-09-24
Initial Journal Implementation

frrist:
### Goals
Make progress towards addressing #3269.

### Implementation
- Define a new package `journal/`.
- A journal is a single file that all topics and their associated operations and metadata are written to. And example of this file is shown by [this gist.](https://gist.github.com/frrist/b8014e16d117c854d3772bbf229805c8) The journal file will be kept in `$REPODIR/filecoin_journal.json.`
- Define a `journal.Writer` type, services that wish to keep a journal may accept a `journal.Writer` and use it for recording events. This writer has been added to the message outbox.


journal/journal.go
```diff
@@ -0,0 +1,51 @@
...
+	// Record records an operation and its metadata to a Journal accepting variadic key-value
+	// pairs.
+	Record(operation string, meta ...interface{})
+}
+
+// Builder defines a method for creating Journals with a topic.
+type Builder func(topic string) Journal
+
+// NewZapJournalBuilder returns a Builder backed by a zap logger.
+func NewZapJournalBuilder(filepath string) (Builder, error) {
```
anorth: I think the topic string should be provided from the outside rather than determined within a component. So some naming things:
- NewZapJournalBuilder -> New
- JournalBuilder -> Journal
- Journal -> Writer
- Record() -> Write()

Rather than a func, use an interface for the ~Builder~Journal.

In the node, we'll then have:
```
j := journal.New(path)
syncer := chain.NewSyncer(..., j.Topic("chain"))
```
and in a component
```
  self.journal.Write("event", "key", value")
```

journal/journal.go
```diff
@@ -0,0 +1,51 @@
...
+// ZapJournal implemented the Journal interface.
+type ZapJournal struct {
+	logger *zap.SugaredLogger
+}
+
+// Record records an operation and its metadata to a Journal accepting variadic key-value
+// pairs.
+func (zj *ZapJournal) Record(operation string, kv ...interface{}) {
+	zj.logger.Infow(operation, kv...)
+}
```
anorth: Move the Zap impl to `journal/zap.go` and implement an in-memory journal here (for tests).

message/pool.go
```diff
@@ -34,6 +35,8 @@ type Pool struct {
 	validator     PoolValidator
 	pending       map[cid.Cid]*timedmessage // all pending messages
 	addressNonces map[addressNonce]bool     // set of address nonce pairs used to efficiently validate duplicate nonces
+
+	journal journal.Journal
```
anorth: For now, don't journal this. It's massively higher traffic than the outbox and the need is less clear. I think we should do the command API before this, but in any case let's defer that discussion to new PRs.

node/builder.go
```diff
@@ -163,6 +165,15 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
 		return nil, err
 	}
 
+	repoPath, err := nc.Repo.Path()
+	if err != nil {
+		return nil, err
+	}
+
+	jBuilder, err := journal.NewZapJournalBuilder(fmt.Sprintf("%s/%s", repoPath, "journal.json"))
```
anorth: The repo should be in control of its internals. Ask the repo to construct the path.

commands/daemon.go
```diff
@@ -96,6 +97,12 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter) error {
 	opts = append(opts, node.BlockTime(blockTime))
 	opts = append(opts, node.ClockConfigOption(clock.NewSystemClock()))
 
+	journal, err := journal.NewZapJournal(rep.JournalPath())
+	if err != nil {
+		return err
+	}
+	opts = append(opts, node.JournalConfigOption(journal))
+
```
frrist: Only create a zap backed journal when something runs the daemon via the `daemon` command.

journal/journal.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Journal defines an interface for creating Journals with a topic.
+type Journal interface {
+	// Topic returns a Writer that records events for a topic.
+	Topic(topic string) Writer
+}
+
+// NewInMemoryJournal returns a journal backed by an in-memory map.
+func NewInMemoryJournal(clk clock.Clock) Journal {
```
frrist: Thinking it might make more sense for this to go in a testing package as that's all I image we'll use it for, thoughts?

repo/fsrepo.go
```diff
@@ -557,6 +557,11 @@ func (r *FSRepo) Path() (string, error) {
 	return r.path, nil
 }
 
+// JournalPath returns the path the journal is at.
+func (r *FSRepo) JournalPath() string {
+	return fmt.Sprintf("%s/filecoin_journal.json", r.path)
+}
+
```
frrist: This should probably be configurable, but doing so now feels out of scope. I propose that if we want a configurable path we do it in a follow on.

journal/journal.go
frrist: I would also be open to a `NoopJournal`, getting this to have the same behaviour as the `ZapJournal` isn't simple.

journal/journal.go
```diff
@@ -0,0 +1,120 @@
...
+// Write records an operation and its metadata to a Journal accepting variadic key-value
+// pairs.
+func (mw *MemoryWriter) Write(event string, kvs ...interface{}) {
+	var ent entry
+	ent.time = mw.clock.Now()
+	ent.event = event
+	extracted, err := extractKeyValues(kvs...)
+	if err != nil {
+		panic(err)
+	}
```
frrist: The zapjournal writes a error message to a configurable location, currently this is set to stderr: https://github.com/filecoin-project/go-filecoin/pull/3485/files#diff-c65947145146f8d0a4bae969afded6abR18

I _really_ don't want the `Write` interface to return an error, we could do something similar to zap here or just ignore the errors, thoughts?

journal/zap.go
```diff
@@ -0,0 +1,51 @@
...
+
+// NewZapJournal returns a Builder backed by a zap logger.
+func NewZapJournal(filepath string) (Journal, error) {
+	zapCfg := zap.NewProductionConfig()
+	zapCfg.Encoding = "json"
+	zapCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	zapCfg.EncoderConfig.LevelKey = ""
+	zapCfg.EncoderConfig.CallerKey = ""
+	zapCfg.EncoderConfig.MessageKey = "operation"
+	zapCfg.EncoderConfig.NameKey = "topic"
```
anorth: To avoid name collisions, I recommend a prefix, e.g. `_topic`. And "event" rather than "operation"?

message/outbox.go
```diff
@@ -78,6 +82,7 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
 		if err != nil {
 			msgSendErrCt.Inc(ctx, 1)
 		}
+		ob.journal.Write("send", "to", to.String(), "from", from.String(), "method", method, "error", err)
```
anorth: Value, gas price, gas limit, params - all of them

message/outbox_test.go
```diff
@@ -29,7 +29,8 @@ func TestOutbox(t *testing.T) {
 		provider := message.NewFakeProvider(t)
 		bcast := true
 
-		ob := message.NewOutbox(w, message.FakeValidator{RejectMessages: true}, queue, publisher, message.NullPolicy{}, provider, provider)
+		ob := message.NewOutbox(w, message.FakeValidator{RejectMessages: true}, queue, publisher,
+			message.NullPolicy{}, provider, provider, newOutboxTestJournal())
```
anorth: Don't depend across test files like this. Move the function here, and just use a single variable in handler_integration_test.

repo/fsrepo.go
anorth: Non-configurable is fine. But drop the redundant `filecoin` in the file name.

frrist:
Follow on issue to instrument elsewhere: https://github.com/filecoin-project/go-filecoin/issues/3556

Issue #3491, opened 2019-09-24
Nightly network has network name of alpha1


travisperson:
**Describe the bug**:

Currently we use the same process to deploy all the networks. When we added the network name for alpha1 and merged it into the master branch this made the nightly also deploy with the network name of alpha.

**Expected behavior**:

- nightly should have it's own network name

Ideally this would be different on each launch (git commit + timestamp). This will require that we have a fallback for the protocol upgrade table such that we don't have to have the network defined there.

travisperson:
Note: The nightly network now has the name of `go-filecoin-test`.

PR #3494, opened 2019-09-24
Change user devnet name to be distinct from staging network. Change setup.json to use go-filecoin-test.

anorth:
This will allow us to deploy the user devnet with a network name distinct from staging.

Future work will make this more flexible.

Context in #3491.


anorth:
On the staging release branch, the setup.json will be changed back to alpha2

PR #3495, opened 2019-09-24
close connection on bad genesis block hello

frrist:
When a peer does not share our genesis block close the libp2p **connection** with them.


anorth:
Previous state: https://github.com/filecoin-project/go-filecoin/commit/2545cd1a2742d7750f05d4c5d47d99bd4e8686a8#diff-a348bd8f91eeee23380b03033d246a96L102

This placed it in a weird place where the node would maintain connections to nodes that had both the wrong genesis and wrong code, but not if they had only the wrong code.

Then the commit SHA check was then removed in #3385.

PR #3504, opened 2019-09-25
Prover restarts PoSt computation if challenge seed changes

anorth:


anorth:
This functionality will move out to the storage miner shared module, so I'm closing this now. FYI @gcristea

PR #3506, opened 2019-09-25
Remove @phritz from maintainer list

anorth:
@phritz departed the project 5 months ago and dropped maintainer status, but this doc wasn't updated and may mislead people.

While we'd welcome a return, I think regaining that status would require re-nomination.

FYI @whyrusleeping


phritz:
makes sense :(

mishmosh:
approving with sadness.

PR #3512, opened 2019-09-26
key protocol versions off network name prefix

### LINKED ISSUE #3496: ProtocolUpgradeTable supports a default protocol version for unknown networks
### Description

We want to support dynamic network names. The primary use case for this is to support changing the nightly network name without requiring the addition of the network to the protocol table.

The ProtocolVersionTableBuilder would support a method which would define the protocol version returned by `VersionAt` of the `ProtocolVersionTableBuilder` when no version information is defined.

This work is in support of #3491.

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/master/version/protocol_versions.go

@acruikshank @anorth

acruikshank:
fixes #3496

### Why this PR?

We need to spin up networks on a nightly basis for validation. We do not want these networks to communicate with each other. We also do not want to require new protocol versions on a nightly basis for each network.

### What's in this PR?

This PR introduces the ability to specified a dot delimited network (e.g. "nightly.2019-10-8") in the genesis block. For the purposes of determining the protocol version, only the part before the dot "nightly" will be considered. So "nightly.2019-10-8" and "nightly.2019-10-9" will get the same protocol table (so long as the nightly protocol table hasn't changed in code).

The full name in genesis ensures nodes on different networks won't be able to sync with each other. It also ensures that nodes on different networks won't be able to communicate on the same protocols.

### DIFF
--- a/cmd/go-filecoin/miner_daemon_test.go
+++ b/cmd/go-filecoin/miner_daemon_test.go
@@ -375,7 +375,7 @@ func minerDaemonTestConfig(t *testing.T) *gengen.GenesisCfg {
 				SectorSize:       constants.DevSectorSize,
 			},
 		},
-		Network: "go-filecoin-test",
+		Network: "gfctest",
 		Time:    123456789,
 	}
 }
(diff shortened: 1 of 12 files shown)


anorth:
Discussion had. @acruikshank can you do anything necessary to get this ready for review?

PR #3513, opened 2019-09-26
WIP: Graphsync sessions

hannahhoward:
# Goals

Break out the mechanics of making graphsync requests from the process of fetching tipsets

# Implementation

Much like a bitswap session, a graphsync session handles delegating requests to peers (eventually in an intelligent manner), dealing with hangups, possibly making retries till the request completes.

# For discussion

One open question is what to do with a partially completed request -- unlike BS, where you either have a block or you don't, you might receive part of a GS query with valuable data without getting the whole thing. So simply delegating the session to retry till it completes may not make sense.


ZenGround0:
This may be useful for future work optimizing peer requests.  Before picking this up note that this work is unfinished and potentially not the optimal path forward for the node wholistically -- for example there is some latency signal from hello protocol that isn't being considered here, and it likely will make sense to decouple session management from the graphsync transport.

PR #3515, opened 2019-09-27
Real state evaluator for the chain builder

anorth:
This is a WIP, pairing with @frrist. FYI @hannahhoward @ZenGround0


ZenGround0:
This may be useful for testing but we should consider go-filecoin testing wholistically and understand the utility of this before trying to land it.

PR #3529, opened 2019-10-04
Feat/spec weight

ZenGround0:
Our first protocol upgrade for using the spec's latest weighting function

- [x] Refactor consensus into chain validation and selection
- [x] Remove TestWeightTipSetDeep
- [x] Add unit tests for the new weight function
- [ ] Add new integration tests that capture the spirit of TestWeightTipSetDeep using better tools and with the context of our new weighting function
- [x] Add the new weighting function
- [ ] Add a protocol upgrade to use this weighting function past a particular height


anorth:
Please separate this work into at least three PRs:
- refactors and cleanup
- implementation and testing of alternate weight function
- protocol upgrade to switch to the new weight function

ZenGround0:
Closing down to break into three above as suggested

PR #3530, opened 2019-10-04
thread builtin actors through processor

acruikshank:
### What is in this PR?

This is movement towards implementing protocol upgrades. One thing blocking that is that the builtin actor lookup has simply been a global variable containing a map. This is bad practice for many reasons. We have been passing it around in state trees for reasons I've never understood (probably expediency). This refactor make many improvements around builtin actors in preparation for upgrades.

1. `builtin.Actors` is now a struct that represents an immutable lookup of actors.
2. `builtin.Actors` is now constructed with a builder.
3. `builtin.Actors` supports the concept of a protocol version although it is currently always set to 0.
4. `CallQueryMethod` and `PreviewQueryMethod` are now methods on `DefaultProcessor`. This lets us dependency inject components (`builtin.Actors` for now) into the processor and use it in those methods rather than passing them in as parameters.
5. Remove `builtin.Actors` from `state.Tree`
6. Many alterations to tests to support these changes.

### DIFF
--- a/actor/builtin/builtin.go
+++ b/actor/builtin/builtin.go
@@ -27,11 +27,11 @@ type Actors struct {
 // GetActorCode returns executable code for an actor by code cid at a specific protocol version
 func (ba Actors) GetActorCode(code cid.Cid, version uint64) (exec.ExecutableActor, error) {
 	if !code.Defined() {
-		return nil, fmt.Errorf("missing code")
+		return nil, fmt.Errorf("undefined code cid")
 	}
 	actor, ok := ba.actors[codeVersion{code: code, protocolVersion: version}]
 	if !ok {
-		return nil, fmt.Errorf("unknown code: %s", code.String())
+		return nil, fmt.Errorf("unknown code: %s, version: %d", code.String(), version)
 	}
 	return actor, nil
 }
@@ -40,7 +40,8 @@ type BuiltinActorsBuilder struct {
 	actors map[codeVersion]exec.ExecutableActor
 }
 
-func NewActorsBuilder() *BuiltinActorsBuilder {
+// NewBuilder creates a builder to generate a builtin.Actor data structure
+func NewBuilder() *BuiltinActorsBuilder {
 	return &BuiltinActorsBuilder{actors: map[codeVersion]exec.ExecutableActor{}}
 }
 
@@ -62,7 +63,7 @@ func (bab *BuiltinActorsBuilder) Build() Actors {
 
 // DefaultActors is list of all actors that ship with Filecoin.
 // They are indexed by their CID.
-var DefaultActors = NewActorsBuilder().
+var DefaultActors = NewBuilder().
 	Add(types.AccountActorCodeCid, 0, &account.Actor{}).
 	Add(types.StorageMarketActorCodeCid, 0, &storagemarket.Actor{}).
 	Add(types.PaymentBrokerActorCodeCid, 0, &paymentbroker.Actor{}).
--- a/actor/builtin/builtin.go
+++ b/actor/builtin/builtin.go
@@ -24,7 +24,8 @@ type Actors struct {
 	actors map[codeVersion]exec.ExecutableActor
 }
 
-func (ba Actors) GetBuiltinActorCode(code cid.Cid, version uint64) (exec.ExecutableActor, error) {
+// GetActorCode returns executable code for an actor by code cid at a specific protocol version
+func (ba Actors) GetActorCode(code cid.Cid, version uint64) (exec.ExecutableActor, error) {
 	if !code.Defined() {
 		return nil, fmt.Errorf("missing code")
 	}
@@ -39,16 +40,15 @@ type BuiltinActorsBuilder struct {
 	actors map[codeVersion]exec.ExecutableActor
 }
 
-func NewBuiltinActorsBuilder() *BuiltinActorsBuilder {
+func NewActorsBuilder() *BuiltinActorsBuilder {
 	return &BuiltinActorsBuilder{actors: map[codeVersion]exec.ExecutableActor{}}
 }
 
-func BuiltinActorsExtender(actors Actors) *BuiltinActorsBuilder {
-	newActors := make(map[codeVersion]exec.ExecutableActor, len(actors.actors))
-	for k, v := range actors.actors {
-		newActors[k] = v
+func (bab *BuiltinActorsBuilder) AddAll(actors Actors) *BuiltinActorsBuilder {
+	for cv, a := range actors.actors {
+		bab.Add(cv.code, cv.protocolVersion, a)
 	}
-	return &BuiltinActorsBuilder{actors: newActors}
+	return bab
 }
 
 func (bab *BuiltinActorsBuilder) Add(c cid.Cid, version uint64, actor exec.ExecutableActor) *BuiltinActorsBuilder {
@@ -62,7 +62,7 @@ func (bab *BuiltinActorsBuilder) Build() Actors {
 
 // DefaultActors is list of all actors that ship with Filecoin.
 // They are indexed by their CID.
-var DefaultActors = NewBuiltinActorsBuilder().
+var DefaultActors = NewActorsBuilder().
 	Add(types.AccountActorCodeCid, 0, &account.Actor{}).
 	Add(types.StorageMarketActorCodeCid, 0, &storagemarket.Actor{}).
 	Add(types.PaymentBrokerActorCodeCid, 0, &paymentbroker.Actor{}).
--- a/actor/builtin/builtin.go
+++ b/actor/builtin/builtin.go
@@ -14,9 +14,10 @@ import (
 	cid "github.com/ipfs/go-cid"
 )
 
+// codeVersion identifies an ExecutableActor by its code and protocol version
 type codeVersion struct {
-	code    cid.Cid
-	version uint64
+	code            cid.Cid
+	protocolVersion uint64
 }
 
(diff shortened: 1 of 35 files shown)


consensus/expected.go
```diff
@@ -79,6 +78,12 @@ type Processor interface {
 
 	// ProcessTipSet processes all messages in a tip set.
 	ProcessTipSet(context.Context, state.Tree, vm.StorageMap, types.TipSet, [][]*types.SignedMessage, []types.TipSet) (*ProcessTipSetResponse, error)
+
+	// CallQueryMethod calls a method on an actor in the given state tree.
+	CallQueryMethod(ctx context.Context, st state.Tree, vms vm.StorageMap, to address.Address, method string, params []byte, from address.Address, optBh *types.BlockHeight) ([][]byte, uint8, error)
+
+	// PreviewQueryMethod estimates the amount of gas that will be used by a method
+	PreviewQueryMethod(ctx context.Context, st state.Tree, vms vm.StorageMap, to address.Address, method string, params []byte, from address.Address, optBh *types.BlockHeight) (types.GasUnits, error)
```
ZenGround0: Do we need these on this interface for the purposes of expected?  If not where is this interface used to motivate this change and can we consider moving that distinct subset to the caller's package?

node/builder.go
```diff
@@ -163,12 +164,20 @@ func (nc *Builder) build(ctx context.Context) (*Node, error) {
 		return nil, err
 	}
 
+	// set up processor
+	var processor consensus.Processor
+	if nc.Rewarder == nil {
+		processor = consensus.NewDefaultProcessor()
+	} else {
+		processor = consensus.NewConfiguredProcessor(consensus.NewDefaultMessageValidator(), consensus.NewDefaultBlockRewarder(), builtin.DefaultActors)
```
ZenGround0: This looks like a copy error where the second argument should be nc.Rewarder.  If not there is something weird going on we should try to fix up.

plumbing/cst/chain_state.go
```diff
@@ -146,12 +149,8 @@ func (chn *ChainStateReadWriter) GetActorSignature(ctx context.Context, actorAdd
 		return nil, ErrNoActorImpl
 	}
 
-	st, err := chn.readWriter.GetTipSetState(ctx, chn.readWriter.GetHead())
-	if err != nil {
-		return nil, err
-	}
-
-	executable, err := st.GetBuiltinActorCode(actor.Code)
+	// TODO: use chain height to determine protocol version
```
ZenGround0: nit: add an issue for this TODO and link

vm/context.go
```diff
@@ -193,7 +202,8 @@ func (ctx *Context) CreateNewActor(addr address.Address, code cid.Cid, initializ
 	newActor.Code = code
 
 	childStorage := ctx.storageMap.NewStorage(addr, newActor)
-	execActor, err := ctx.state.GetBuiltinActorCode(code)
+	// TODO: need to use blockheight derived version
```
ZenGround0: nit: link TODO to issue

vm/vm.go
```diff
@@ -42,7 +42,8 @@ func send(ctx context.Context, deps sendDeps, vmCtx *Context) ([][]byte, uint8,
 		return nil, 0, nil
 	}
 
-	toExecutable, err := vmCtx.state.GetBuiltinActorCode(vmCtx.to.Code)
+	// TODO: use chain height based protocol version here
```
ZenGround0: nit: link issue

actor/builtin/builtin.go
```diff
@@ -12,16 +14,58 @@ import (
 	cid "github.com/ipfs/go-cid"
 )
 
-// Actors is list of all actors that ship with Filecoin.
-// They are indexed by their CID.
-var Actors = map[cid.Cid]exec.ExecutableActor{}
+type codeVersion struct {
+	code    cid.Cid
+	version uint64
```
anorth: Is this the protocol version? Please name and document it as such.

actor/builtin/builtin.go
```diff
@@ -12,16 +14,58 @@ import (
...
+	actors map[codeVersion]exec.ExecutableActor
+}
+
+func (ba Actors) GetBuiltinActorCode(code cid.Cid, version uint64) (exec.ExecutableActor, error) {
+	if !code.Defined() {
+		return nil, fmt.Errorf("missing code")
+	}
+	actor, ok := ba.actors[codeVersion{code: code, version: version}]
+	if !ok {
+		return nil, fmt.Errorf("unknown code: %s", code.String())
```
anorth: Include the version in the error message

actor/builtin/builtin.go
```diff
@@ -12,16 +14,58 @@ import (
...
+type codeVersion struct {
+	code    cid.Cid
+	version uint64
+}
+
+type Actors struct {
+	actors map[codeVersion]exec.ExecutableActor
+}
+
+func (ba Actors) GetBuiltinActorCode(code cid.Cid, version uint64) (exec.ExecutableActor, error) {
```
anorth: Since the package is `builtin`, remove this from the type and function names. e.g. `builtin.GetActorCode`

actor/builtin/builtin.go
```diff
@@ -12,16 +14,58 @@ import (
...
 
-func init() {
-	// Instance Actors
-	Actors[types.AccountActorCodeCid] = &account.Actor{}
-	Actors[types.StorageMarketActorCodeCid] = &storagemarket.Actor{}
-	Actors[types.PaymentBrokerActorCodeCid] = &paymentbroker.Actor{}
-	Actors[types.MinerActorCodeCid] = &miner.Actor{}
-	Actors[types.BootstrapMinerActorCodeCid] = &miner.Actor{Bootstrap: true}
-	Actors[types.InitActorCodeCid] = &initactor.Actor{}
+func NewBuiltinActorsBuilder() *BuiltinActorsBuilder {
```
anorth: `builtin.NewBuilder()`?

actor/builtin/builtin.go
```diff
@@ -12,16 +14,58 @@ import (
...
+	return &BuiltinActorsBuilder{actors: map[codeVersion]exec.ExecutableActor{}}
 }
+
+func BuiltinActorsExtender(actors Actors) *BuiltinActorsBuilder {
```
anorth: This is a bit clunky. How about an `AddAll` method instead so the call is `builtin.NewBuilder().AddAll(actors).Build()`

Or just rename this method to `Extend()`.

consensus/actor_state.go
```diff
@@ -29,11 +29,13 @@ type ActorStateStore struct {
 	cst *hamt.CborIpldStore
 	// For vm storage.
 	bs bstore.Blockstore
+	// executable actors
+	processor Processor
```
anorth: I'm not convinced this is a desirable dependency. I gather it's introduced by removing the actor code map from the state tree? I do see the code was already invoking `CallQueryMethod` from the processor. Maybe we need an abstraction that wraps up the state tree *and* the code table (for want of a better term), i.e. *all* the "memory" of the VM.

No immediate action required.

anorth:
The new spec clearly attaches the actor lookup to the state tree:

```
type StateTree struct {
    SystemActors  {ActorName: addr.Address}
    ActorStates   {addr.Address: actor.ActorState}

    // TODO: API ConvenienceAPI
}
```

consensus/actor_state.go
acruikshank: I'm not either, but here's the motivation: builtin actors and the protocol version table are required for protocol version variant message execution. These objects are singleton for the execution lifetime of the app and `ActorState` doesn't use them directly. It simply passes them on to processor whenever it calls Query. Other code running state transitions already keeps a reference to the processor to ApplyMessages. This change standardizes patterns of usage while reducing `ActorState`s exposure to execution details.

I'll work towards the cleaner abstraction you mention in the next PR.

acruikshank: I'm not either, but here's the motivation: Without giving ActorState a reference to processor, it would now be forced to keep a reference to bulitin.Actors and pass it on to the Query method without actually using it. This will soon be true for the ProtocolVersionTable (or something containing it) as well. Other code that runs state transitions are already injected with a reference to Protocol to call ApplyMessage and its variants. This change standardizes patterns of usage while reducing `ActorStates` exposure to execution details.

Your idea for a new abstraction is a good one, and i'll be working towards that in the next PR.

node/builder.go
acruikshank: Good catch!

consensus/expected.go
acruikshank: good call

acruikshank:
@anorth which version of the spec states that? Mine (pulled tonight) has:
```
type StateTree map[Address]Actor
```
Even if the version you quote is accurate, the `SystemActors` appear to be a mapping from name to address. Presumably this is to tie a well known name to an ID address for singleton actors. This would be problematic for the same reason tying addresses to `ExecutableActor`s is. Code capable of producing a state tree should not be required to store details about message execution. Either way it's not to bad, but if the spec doesn't say otherwise, its better to keep dispatch to executable code closer to execution.

anorth:
@acruikshank looks like the spec says both. 5.2.8 `#state-tree` has your version, 2.3.1.3 `#systems__filecoin_blockchain__vm__state_tree` has mine.

@sternhenri the spec is confusing here.

sternhenri:
Mmh thanks for pointing it out. @jbenet I think you are working on VM completion this week. Can you handle? Happy to otherwise!

PR #3531, opened 2019-10-07
Refactor/weight prep

ZenGround0:
In preparation for the weight protocol upgrade this PR refactors the consensus component by moving the chain selection logic into its own object.  It also removes TestTipSetWeightDeep.  A modern version of this test is introduced in a later PR.

### DIFF
--- a/chain/syncer.go
+++ b/chain/syncer.go
@@ -54,14 +54,16 @@ type syncerChainReaderWriter interface {
 	GetTipSetAndStatesByParentsAndHeight(pTsKey types.TipSetKey, h uint64) ([]*TipSetAndState, error)
 }
 
+type syncChainSelector interface {
+	// IsHeaver returns true if tipset a is heavier than tipset b and false if
+	// tipset b is heavier than tipset a.
+	IsHeavier(ctx context.Context, a, b types.TipSet, aStateID, bStateID cid.Cid) (bool, error)
+}
+
 type syncStateEvaluator interface {
 	// RunStateTransition returns the state root CID resulting from applying the input ts to the
 	// prior `stateRoot`.  It returns an error if the transition is invalid.
 	RunStateTransition(ctx context.Context, ts types.TipSet, tsMessages [][]*types.SignedMessage, tsReceipts [][]*types.MessageReceipt, ancestors []types.TipSet, stateID cid.Cid) (cid.Cid, error)
-
-	// IsHeaver tests whether tipset `a` is heavier than tipset `b`.
-	// The state IDs identify the state to which the tipset applies (i.e. prior to its messages).
-	IsHeavier(ctx context.Context, a, b types.TipSet, aStateID, bStateID cid.Cid) (bool, error)
 }
 
 // Syncer updates its chain.Store according to the methods of its
@@ -95,6 +97,8 @@ type Syncer struct {
 
 	// Evaluates tipset messages and stores the resulting states.
 	stateEvaluator syncStateEvaluator
+	// Selects the heaviest of two chains
+	chainSelector syncChainSelector
 	// Provides and stores validated tipsets and their state roots.
 	chainStore syncerChainReaderWriter
 	// Provides message collections given cids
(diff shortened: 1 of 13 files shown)


chain/syncer_integration_test.go
```diff
@@ -112,267 +108,44 @@ func TestLoadFork(t *testing.T) {
...
-// Current issue: when creating miner mining with addr0, addr0's storage head isn't found in the blockstore
-// and I can't figure out why because we pass in the correct blockstore to createStorageMinerWithpower.
-func TestTipSetWeightDeep(t *testing.T) {
-	// This test takes many seconds, the bottleneck is gengen.
-	tf.IntegrationTest(t)
+// Power table weight comparisons impact syncer's selection.
+// One fork has more blocks but less total power.
+// Verify that the heavier fork is the one with more power.
+func TestSyncerWeighsPower(t *testing.T) {
+	t.Skip("fill this out when making weight upgrade")
```
acruikshank: It would make me more comfortable leaving this in the code if you either added a TODO with an issue to point to the future work or include the issue # in the skip message (this applies to the next test too).

ZenGround0: Woops I meant to delete this.  Anorth requested a rebase and this tagged along unintended.

ZenGround0:
> It's probably worth asking if the reason we want to break apart these larger structs (I've been doing it too) is because our intferfaces are producer instead of consumer interfaces. 

I don't think I understand the question and would love to talk more.  The super concrete reason I did this is that testing 2 halves of a multipurpose thing is a lot more tedious than setting up and testing the two things on their own.

chain/syncer.go
```diff
@@ -54,14 +54,16 @@ type syncerChainReaderWriter interface {
 	GetTipSetAndStatesByParentsAndHeight(pTsKey types.TipSetKey, h uint64) ([]*TipSetAndState, error)
 }
 
+type syncChainSelector interface {
+	// IsHeaver returns 1 if tipset a is heavier than tipset b and -1 if
+	// tipset b is heavier than tipset a.
```
anorth: It returns a boolean

consensus/chain_selector.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Weight returns the expected consensus weight of this TipSet in uint64
+// encoded fixed point representation.
+func (c *ChainSelector) Weight(ctx context.Context, ts types.TipSet, pSt state.Tree) (uint64, error) {
+	ctx = log.Start(ctx, "Expected.Weight")
+	log.LogKV(ctx, "Weight", ts.String())
+	if ts.Len() == 1 && ts.At(0).Cid().Equals(c.genesisCid) {
+		return uint64(0), nil
+	}
```
anorth: You only moved this here, but for follow-up, this should probably be a function of height rather than coupled to specific genesis CID.

consensus/chain_selector.go
```diff
@@ -0,0 +1,152 @@
...
+		// a is heavier if b's ticket is greater than a's ticket.
+		return cmp == 1, nil
+	}
+
+	// Tie break on cid ids.
+	// TODO: I think this is drastically impacted by number of blocks in tipset
+	// i.e. bigger tipset is always heavier.  Not sure if this is ok, need to revist.
+	cmp = strings.Compare(a.String(), b.String())
+	if cmp == 0 {
+		// Caller is mistakenly calling on two identical tipsets.
```
anorth: When you update this code, this constraint should be in the function comment (or dropped)

consensus/protocol.go
```diff
@@ -25,13 +24,6 @@ import (
 // the most theoretically obvious or pleasing and should not be considered
 // finalized.
 type Protocol interface {
-	// Weight returns the weight given to the input ts by this consensus protocol.
-	Weight(ctx context.Context, ts types.TipSet, pSt state.Tree) (uint64, error)
-
-	// IsHeaver returns 1 if tipset a is heavier than tipset b and -1 if
-	// tipset b is heavier than tipset a.
-	IsHeavier(ctx context.Context, a, b types.TipSet, aStateID, bStateID cid.Cid) (bool, error)
-
```
anorth: 🎉

PR #3532, opened 2019-10-07
Feat/new weight

ZenGround0:
- [x] new weight to latest spec
- [x] unit tests

--edit--
Integration test is now grouped in with the upgrade so that it can (1) avoid complexity of testing across upgrade boundary (2) not require protocol breaking change to introduce new weighting function on master.

### DIFF
--- a/consensus/chain_selector.go
+++ b/consensus/chain_selector.go
@@ -16,6 +16,27 @@ import (
 	"github.com/filecoin-project/go-filecoin/types"
 )
 
+// Parameters used by the latest weight function "NewWeight"
+const (
+	// newECV is the constant V defined in the EC spec.
+	newECV uint64 = 2
+
+	// pi is the multiplicand in the null penalty term
+	pi float64 = 0.87
+
+	// nullThresh is the min number of null rounds before the penalty kicks in
+	nullThresh = 3
+)
+
+// Parameters used by the deprecated weight function before the alphanet upgrade
+const (
+	// ecV is the constant V defined in the EC spec.
+	ecV uint64 = 10
+
+	// ecPrM is the power ratio magnitude defined in the EC spec.
+	ecPrM uint64 = 100
+)
+
 // ChainSelector weighs and compares chains according to the Storage Power
 // Consensus Protocol
 type ChainSelector struct {
@@ -33,8 +54,62 @@ func NewChainSelector(cs *hamt.CborIpldStore, actorState SnapshotGenerator, gCid
 	}
 }
 
-// Weight returns the expected consensus weight of this TipSet in uint64
-// encoded fixed point representation.
+// NewWeight returns the EC weight of this TipSet in uint64 encoded fixed point
+// representation.
+//
+// w(i) = w(i-1) + (pi)^(P_n) * [V * num_blks + X ]
+// P_n(n) = if n < 3:0 else: n, n is number of null rounds
+// X = log_2(total_storage(pSt))
+func (c *ChainSelector) NewWeight(ctx context.Context, ts types.TipSet, pSt state.Tree) (uint64, error) {
+	ctx = log.Start(ctx, "Expected.Weight")
+	log.LogKV(ctx, "Weight", ts.String())
+	if ts.Len() > 0 && ts.At(0).Cid().Equals(c.genesisCid) {
+		return uint64(0), nil
+	}
+	// Retrieve parent weight.
+	parentW, err := ts.ParentWeight()
+	if err != nil {
+		return uint64(0), err
+	}
+
+	w, err := types.FixedToBig(parentW)
+	if err != nil {
+		return uint64(0), err
+	}
+
+	// Each block adds ECV to the weight's inner term
+	innerTerm := new(big.Float)
+	floatECV := new(big.Float).SetInt64(int64(newECV))
+	floatNumBlocks := new(big.Float).SetInt64(int64(ts.Len()))
+	innerTerm.Mul(floatECV, floatNumBlocks)
+
+	// Add bitnum(total storage power) to the weight's inner term
+	powerTableView := c.createPowerTableView(pSt)
+	totalBytes, err := powerTableView.Total(ctx)
+	if err != nil {
+		return uint64(0), err
+	}
+	roughLogTotalBytes := new(big.Float).SetInt64(int64(totalBytes.BigInt().BitLen()))
+	innerTerm.Add(innerTerm, roughLogTotalBytes)
+
+	// Attenuate weight by the number of tickets
+	numTickets := len(ts.At(0).Tickets)
+	P := new(big.Float).SetInt64(int64(1))
+	if numTickets >= nullThresh {
+		bigPI := new(big.Float).SetFloat64(pi)
+		// P = pi^numNull
+		for i := 0; i < numTickets; i++ {
+			P.Mul(P, bigPI)
+		}
+	}
+	update := new(big.Float)
+	update.Mul(innerTerm, P)
+	w.Add(w, update)
+	return types.BigToFixed(w)
+}
+
+// Weight returns the EC weight of this TipSet in uint64 encoded fixed point
+// representation.
 func (c *ChainSelector) Weight(ctx context.Context, ts types.TipSet, pSt state.Tree) (uint64, error) {
 	ctx = log.Start(ctx, "Expected.Weight")
 	log.LogKV(ctx, "Weight", ts.String())
@@ -54,14 +129,14 @@ func (c *ChainSelector) Weight(ctx context.Context, ts types.TipSet, pSt state.T
 
 	powerTableView := c.createPowerTableView(pSt)
 
-	// Each block in the tipset adds ECV + ECPrm * miner_power to parent weight.
+	// Each block in the tipset adds ecV + ECPrm * miner_power to parent weight.
 	totalBytes, err := powerTableView.Total(ctx)
 	if err != nil {
 		return uint64(0), err
 	}
 	floatTotalBytes := new(big.Float).SetInt(totalBytes.BigInt())
-	floatECV := new(big.Float).SetInt64(int64(ECV))
-	floatECPrM := new(big.Float).SetInt64(int64(ECPrM))
+	floatECV := new(big.Float).SetInt64(int64(ecV))
+	floatECPrM := new(big.Float).SetInt64(int64(ecPrM))
 	for _, blk := range ts.ToSlice() {
 		minerBytes, err := powerTableView.Miner(ctx, blk.Miner)
 		if err != nil {
--- a/consensus/weight_test.go
+++ b/consensus/weight_test.go
@@ -1,0 +1,156 @@
+package consensus_test
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/ipfs/go-hamt-ipld"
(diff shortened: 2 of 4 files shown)


ZenGround0:
Will merge after #3531

consensus/chain_selector.go
```diff
@@ -17,23 +19,96 @@ import (
...
+	roughLogTotalBytes := new(big.Float).SetInt64(int64(totalBytes.BigInt().BitLen()))
+	innerTerm.Add(innerTerm, roughLogTotalBytes)
+
+	// Attenuate weight by the number of tickets
+	numTickets := len(ts.At(0).Tickets)
+	P := new(big.Float).SetInt64(int64(1))
+	if numTickets >= NullThresh {
+		bigPI := new(big.Float).SetFloat64(PI)
+		// P = PI^numNull
+		for i := 0; i < numTickets; i++ {
```
steven004: It is intended to have at least PI to the power of NullThresh (0.87^3 = 0.6585) as the ratio to attenuate weight? This happens when there are 2 null blocks between tipsets. It looks too much, while there is no attenuation when one null block exists between tipsets. 

I would suggest changing it to: 
```golang
          for i := 0; i < numTickets-1; i++ {
              ... 
         }
```

ZenGround0: Thanks for looking at this closely @steven004.  From the version of the spec I implemented against 
- `Weight = ParentWeight + P(V*NumBlocksMinedInThisRound + X)`
- `P will be equal to some term P_i^P_n`
- `P_n = the number of tickets in this block's Ticket array if that number is  >= 3, 0 otherwise.`

So this implementation is up to that version of the spec.

Unfortunately I can't link this to you because rapid changes to the spec have shifted the weighting function definition yet again :(  

However this might mean that your concerns are addressed by [the new function](https://filecoin-project.github.io/specs/#chain-selection).  Please leave github issues in [the specs repo](https://github.com/filecoin-project/specs) for concerns about that latest speced weighting function.

consensus/chain_selector.go
```diff
@@ -16,6 +16,21 @@ import (
...
+const PI float64 = 0.87
+
+// NullThresh is the min number of null rounds before the penalty kicks in
+const NullThresh = 3
+
+// ECV is the constant V defined in the EC spec.
+const ECV uint64 = 10
+
+// ECPrM is the power ratio magnitude defined in the EC spec.
+const ECPrM uint64 = 100
```
anorth: Don't export these constants. Prefer to collapse into one or a couple of `const (... )` blocks.

anorth: These don't correspond to the present state of https://filecoin-project.github.io/specs/#chain-weighting, though I believe that is still in flux. I understand that you've implemented the weight function from the pre-October spec (consistent with the rest of the implementation).

That's fine, but I have therefore not mentally verified the spec alignment.

consensus/chain_selector.go
```diff
@@ -33,8 +48,62 @@ func NewChainSelector(cs *hamt.CborIpldStore, actorState SnapshotGenerator, gCid
...
+//
+// w(i) = w(i-1) + (PI)^(P_n) * [V * num_blks + X ]
+// P_n(n) = if n < 3:0 else: n, n is number of null rounds
+// X = log_2(total_storage(pSt))
+func (c *ChainSelector) NewWeight(ctx context.Context, ts types.TipSet, pSt state.Tree) (uint64, error) {
+	ctx = log.Start(ctx, "Expected.Weight")
+	log.LogKV(ctx, "Weight", ts.String())
+	if ts.Len() == 1 && ts.At(0).Cid().Equals(c.genesisCid) {
+		return uint64(0), nil
+	}
```
anorth: This would be more robust as a function of height, but ok to leave that until we come back to update yet again.

consensus/chain_selector.go
```diff
@@ -16,6 +16,21 @@ import (
...
+const NewECV uint64 = 2
+
+// PI is the multiplicand in the null penalty term
+const PI float64 = 0.87
+
+// NullThresh is the min number of null rounds before the penalty kicks in
+const NullThresh = 3
+
+// ECV is the constant V defined in the EC spec.
+const ECV uint64 = 10
```
anorth: I initially thought this was unused, but realise it's there for the old function. Please clearly separate the constants for the old and new functions.

consensus/weight_test.go
```diff
@@ -0,0 +1,155 @@
...
+		intWeight := requireFixedToInt(t, fixWeight)
+		assert.Equal(t, 7, intWeight)
+	})
+
+	t.Run("total power adjusts as expected", func(t *testing.T) {
+		asLowerX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(15), make(map[address.Address]address.Address))
+		asSameX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(31), make(map[address.Address]address.Address))
+		asHigherX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(32), make(map[address.Address]address.Address))
+
+		// Weight is 1 lower than with total = 16 with total = 15
```
anorth:
```suggestion
		// Weight is 1 lower than total = 16 with total = 15
```

consensus/weight_test.go
```diff
@@ -0,0 +1,155 @@
...
+		for i := 0; i < 15; i++ {
+			fifteenTickets = append(fifteenTickets, consensus.MakeFakeTicketForTest())
+			expected *= consensus.PI
+		}
+		toWeighFifteenNull := types.RequireNewTipSet(t, &types.Block{
+			ParentWeight: 0,
+			Tickets:      fifteenTickets,
+		})
+
+		expected *= 7
```
anorth: Why?

consensus/weight_test.go
```diff
@@ -0,0 +1,155 @@
...
+		asLowerX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(15), make(map[address.Address]address.Address))
+		asSameX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(31), make(map[address.Address]address.Address))
+		asHigherX := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(32), make(map[address.Address]address.Address))
+
+		// Weight is 1 lower than with total = 16 with total = 15
+		selLower := consensus.NewChainSelector(cst, asLowerX, types.CidFromString(t, "genesisCid"))
+		fixWeight, err := selLower.NewWeight(ctx, toWeigh, fakeTree)
+		assert.NoError(t, err)
+		intWeight := requireFixedToInt(t, fixWeight)
+		assert.Equal(t, 6, intWeight)
```
anorth: Consider factoring out an `assertEqualInt(t, int, fixed)` to compress this noise.

consensus/weight_test.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		toWeighFifteenNull := types.RequireNewTipSet(t, &types.Block{
+			ParentWeight: 0,
+			Tickets:      fifteenTickets,
+		})
+
+		expected *= 7
+		bigExpected := big.NewFloat(expected)
+		fixExpected, err := types.BigToFixed(bigExpected) // do fixed point rounding for cmp
+		require.NoError(t, err)
```
anorth: Consider creating `assertEqualBig(t, bigfloat, fixed)`.

consensus/chain_selector.go
ZenGround0: Alright. Tradeoff is we have more magic numbers in tests because testing packages can't read them.  I already do a lot of this and am fine with doubling down.

consensus/weight_test.go
ZenGround0: will add a comment to make it clearer

consensus/chain_selector.go
```diff
@@ -33,8 +48,62 @@ func NewChainSelector(cs *hamt.CborIpldStore, actorState SnapshotGenerator, gCid
...
+	innerTerm.Add(innerTerm, roughLogTotalBytes)
+
+	// Attenuate weight by the number of tickets
+	numTickets := len(ts.At(0).Tickets)
+	P := new(big.Float).SetInt64(int64(1))
+	if numTickets >= NullThresh {
+		bigPI := new(big.Float).SetFloat64(PI)
+		// P = PI^numNull
+		for i := 0; i < numTickets; i++ {
+			P.Mul(P, bigPI)
```
acruikshank: Your assumption that numTickets is small enough for this for loop to be reasonable also implies that big.NewFloat(math.Pow(PI, float64(numTickets))) should be sufficient for this calculation. Does this spec specify this algorithm?

consensus/weight_test.go
```diff
@@ -0,0 +1,155 @@
...
+		Tickets:      tickets,
+	})
+	sel := consensus.NewChainSelector(cst, as, types.CidFromString(t, "genesisCid"))
+
+	t.Run("basic happy path", func(t *testing.T) {
+		// 0 + 1[2*1 + 5] = 7
+		fixWeight, err := sel.NewWeight(ctx, toWeigh, fakeTree)
+		assert.NoError(t, err)
+		intWeight := requireFixedToInt(t, fixWeight)
+		assert.Equal(t, 7, intWeight)
```
acruikshank: Even with the comment above, this 7 is going to be hard to reason about later. Better to put the equation in the comment as the expected and populate it with all the constants.

consensus/weight_test.go
```diff
@@ -0,0 +1,156 @@
...
+		ParentWeight: 0,
+		Tickets:      tickets,
+	})
+	sel := consensus.NewChainSelector(cst, as, types.CidFromString(t, "genesisCid"))
+
+	t.Run("basic happy path", func(t *testing.T) {
+		// 0 + 1[2*1 + 5] = 7
+		fixWeight, err := sel.NewWeight(ctx, toWeigh, fakeTree)
+		assert.NoError(t, err)
+		assertEqualInt(t, 7, fixWeight)
```
acruikshank: Later it's going to be difficult to decode this 7 and the equation in the comment above. It'd be better to put the equation as the expected, populate it will all the known constants, create variables for all the expected values.

consensus/chain_selector.go
```diff
@@ -33,8 +48,62 @@ func NewChainSelector(cs *hamt.CborIpldStore, actorState SnapshotGenerator, gCid
...
+	floatNumBlocks := new(big.Float).SetInt64(int64(ts.Len()))
+	innerTerm.Mul(floatECV, floatNumBlocks)
+
+	// Add X to the weight's inner term
+	powerTableView := c.createPowerTableView(pSt)
+	totalBytes, err := powerTableView.Total(ctx)
+	if err != nil {
+		return uint64(0), err
+	}
+	roughLogTotalBytes := new(big.Float).SetInt64(int64(totalBytes.BigInt().BitLen()))
```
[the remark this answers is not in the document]
ZenGround0: bit length thing should be captured [here](https://github.com/filecoin-project/specs/pull/538).

consensus/chain_selector.go
ZenGround0: ah didn't even notice the power function.  When we move to the actual stable speced weight function I'll remove this sloppiness.

PR #3536, opened 2019-10-08
Upgrade/alphanet weight

ZenGround0:
WIP should be reviewed after 3532 merges

- [x] weight function upgrade
- [x] refactor weight slightly to accomodate weight validation
- [x] weight validation upgrade
- [x] integration test for upgraded weight

### DIFF
--- a/chain/syncer_integration_test.go
+++ b/chain/syncer_integration_test.go
@@ -6,12 +6,15 @@ import (
 	"time"
 
 	bserv "github.com/ipfs/go-blockservice"
+	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-hamt-ipld"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
 	"github.com/ipfs/go-ipfs-exchange-offline"
 
+	"github.com/filecoin-project/go-filecoin/actor"
 	"github.com/filecoin-project/go-filecoin/address"
 	"github.com/filecoin-project/go-filecoin/chain"
+	"github.com/filecoin-project/go-filecoin/consensus"
 	"github.com/filecoin-project/go-filecoin/repo"
 	"github.com/filecoin-project/go-filecoin/state"
 	th "github.com/filecoin-project/go-filecoin/testhelpers"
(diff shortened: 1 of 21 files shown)


chain/testing.go
```diff
@@ -65,12 +65,16 @@ func NewBuilder(t *testing.T, miner address.Address) *Builder {
...
-	require.NoError(t, err)
+	nullState := types.CidFromString(t, "null")
 	b.tipStateCids[types.NewTipSetKey().String()] = nullState
 	return b
 }
 
+// SetStateBuilder sets the state builder
+func (f *Builder) SetStateBuilder(sb StateBuilder) {
+	f.stateBuilder = sb
+}
```
anorth: Please replace this mutation with a new constructor, `NewBuilderWithState(...)` to which the existing constructor can delegate.

chain/testing.go
```diff
@@ -301,6 +305,8 @@ type StateBuilder interface {
 	Weigh(tip types.TipSet, state cid.Cid) (uint64, error)
 }
 
+//
+
```
anorth: Delete

consensus/block_validation.go
```diff
@@ -46,29 +47,44 @@ type MessageSyntaxValidator interface {
 type DefaultBlockValidator struct {
 	clock.Clock
 	blockTime time.Duration
+	ptv       *version.ProtocolVersionTable
```
anorth: `pvt`?

consensus/block_validation.go
```diff
@@ -46,29 +47,44 @@ type MessageSyntaxValidator interface {
...
 
 // ValidateSemantic validates a block is correctly derived from its parent.
-func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *types.Block, parents *types.TipSet) error {
+func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *types.Block, parents *types.TipSet, parentWeight uint64) error {
 	pmin, err := parents.MinTimestamp()
 	if err != nil {
 		return err
 	}
 
+	// Alphanet upgrade
```
anorth: Move this comment down to the new code, and change to "Protocol Version 1 upgrade"

consensus/block_validation.go
```diff
@@ -46,29 +47,44 @@ type MessageSyntaxValidator interface {
...
 	}
 
+	parentVersion, err := dv.ptv.VersionAt(types.NewBlockHeight(ph))
+	if err != nil {
+		return err
+	}
+	if parentVersion >= version.Protocol1 {
+		if uint64(child.ParentWeight) != parentWeight {
+			fmt.Printf("consensus fault: parentweight (%d) != expected parentweight (%d)\n", uint64(child.ParentWeight), parentWeight)
+			return fmt.Errorf("block %s has invalid parent weight %d", child.Cid().String(), parentWeight)
```
anorth: Don't print and error, just return all the information in the error string.

Remove the debugging print on 85 too.

consensus/block_validation.go
```diff
@@ -46,29 +47,44 @@ type MessageSyntaxValidator interface {
...
 	ph, err := parents.Height()
 	if err != nil {
 		return err
 	}
 
+	parentVersion, err := dv.ptv.VersionAt(types.NewBlockHeight(ph))
+	if err != nil {
+		return err
+	}
+	if parentVersion >= version.Protocol1 {
```
anorth: Add a significant comment here explaining the old and new behaviour, and what has changed.

consensus/block_validation_test.go
```diff
@@ -63,22 +66,42 @@ func TestBlockValidSemantic(t *testing.T) {
...
-		err = validator.ValidateSemantic(ctx, c, &parents)
+		err = validator.ValidateSemantic(ctx, c, &parents, 0)
 		assert.Error(t, err)
 		assert.Contains(t, err.Error(), "too far")
 
 	})
+
+	t.Run("reject block mined with invalid parent weight after alphanet upgrade", func(t *testing.T) {
+		hUpgrade := 300
+		c := &types.Block{Height: types.Uint64(hUpgrade) + 50, ParentWeight: 5000, Timestamp: types.Uint64(ts.Add(blockTime).Unix())}
```
anorth: Factor out a helper method to make this timestamp more concise.

consensus/block_validation_test.go
```diff
@@ -63,22 +66,42 @@ func TestBlockValidSemantic(t *testing.T) {
...
+		p := &types.Block{Height: types.Uint64(hUpgrade) + 49, Timestamp: types.Uint64(ts.Unix())}
+		parents := consensus.RequireNewTipSet(require.New(t), p)
+
+		err := validator.ValidateSemantic(ctx, c, &parents, 30)
```
anorth: The `30` is opaque, can you give it a name?

node/builder.go
```diff
@@ -189,12 +189,18 @@ func (b *Builder) build(ctx context.Context) (*Node, error) {
 		return nil, errors.Wrap(err, "failed to build node.Network")
 	}
 
+	// TODO: inject protocol upgrade table into code that requires it (#3360)
```
anorth: Looks done.

version/protocol_versions.go
```diff
@@ -28,5 +31,9 @@ func ConfigureProtocolVersions(network string) (*ProtocolVersionTable, error) {
 		Add(DEVNET4, Protocol0, types.NewBlockHeight(0)).
 		Add(LOCALNET, Protocol0, types.NewBlockHeight(0)).
 		Add(TEST, Protocol0, types.NewBlockHeight(0)).
+		Add(USER, Protocol1, types.NewBlockHeight(40000)).
+		Add(DEVNET4, Protocol1, types.NewBlockHeight(300)).
+		Add(LOCALNET, Protocol1, types.NewBlockHeight(300)).
+		Add(TEST, Protocol1, types.NewBlockHeight(300)).
```
anorth: For almost all tests I would expect to use version 1 from the beginning. Only the tests specifically exercising both behaviours should have a non-zero upgrade height.

Don't call into this method from those tests if this isn't building the right table - just build a custom table with only the TEST network. Reserve this common table for "shared" networks.

So I expect this table to be:
USER: Protocol0, 0
USER: Protocol1, 40000
LOCALNET: Protocol1, 0
TEST: Protocol1, 0

@acruikshank

node/node.go
```diff
@@ -795,19 +796,38 @@ func (node *Node) getStateTree(ctx context.Context, ts types.TipSet) (state.Tree
 
 // getWeight is the default GetWeight function for the mining worker.
 func (node *Node) getWeight(ctx context.Context, ts types.TipSet) (uint64, error) {
+	h, err := ts.Height()
+	if err != nil {
+		return 0, err
+	}
+	var wFun func(context.Context, types.TipSet, cid.Cid) (uint64, error)
```
acruikshank: This function would be easier to read if the if statement were moved to the bottom and we didn't need to pass functions around.

node/node.go
```diff
@@ -795,19 +796,38 @@ func (node *Node) getStateTree(ctx context.Context, ts types.TipSet) (state.Tree
...
 	if parent.Len() == 0 {
-		return node.Chain.ChainSelector.Weight(ctx, ts, nil)
+		return wFun(ctx, ts, cid.Undef)
 	}
 	pSt, err := node.Chain.ChainReader.GetTipSetState(ctx, parent)
 	if err != nil {
 		return uint64(0), err
 	}
-	return node.Chain.ChainSelector.Weight(ctx, ts, pSt)
+	root, err := pSt.Flush(ctx)
```
acruikshank: ChainReader has a `GitTipSetStateRoot` function that avoids the need to instantiate a tree and then flush it. It would also eliminate the need to special case genesis.

chain/syncer_integration_test.go
```diff
@@ -107,3 +110,197 @@ func TestLoadFork(t *testing.T) {
...
+
+	// sync fork 1
+	assert.NoError(t, syncer.HandleNewTipSet(ctx, types.NewChainInfo("", head1.Key(), heightFromTip(t, head1)), true))
+	assert.Equal(t, head1.Key(), store.GetHead())
+	// sync fork 2
+	assert.NoError(t, syncer.HandleNewTipSet(ctx, types.NewChainInfo("", head2.Key(), heightFromTip(t, head1)), true))
+	assert.Equal(t, head2.Key(), store.GetHead())
+}
+
+type integrationStateBuilder struct {
```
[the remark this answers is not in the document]
ZenGround0: It's working around the inherent coupling of the state machine and chain data structures.  The chain testing's state builder is our existing attempt at decoupling these things for tests and it worked pretty well here.

PR #3545, opened 2019-10-11
Separate BLS and Secp messages in block header.

ZenGround0:
### Why this PR?

This PR is a step towards BLS message signature support.

We need to support both BLS and Secp (elliptic curve) signed messages. The mechanism for this is to maintain separate data structures for each. The BLS messages will have a single signature in the block header, while the Secp messages will each have their own signature. The spec defines a TxMeta struct to contain the keys forr both these structures that replaces a single key for the Secp messages.

### What is in this PR?

This PR introduces the `TxMeta` struct in the block header. It contains a `SecpRoot` that points to the messages, and a `BLSRoot` which, for now, always points to an empty message list. All the rest of the work in this PR exists to permit access to the Secp messages through this new structure.

### DIFF
--- a/chain/message_store.go
+++ b/chain/message_store.go
@@ -49,7 +49,7 @@ func (ms *MessageStore) LoadMessages(ctx context.Context, meta types.TxMeta) ([]
 	err = ms.ipldStore.Get(ctx, meta.BLSRoot, &bls)
 	if err != nil {
 		return nil, nil, errors.Wrapf(err, "failed to load bls messages %s", meta.BLSRoot.String())
-	}	
+	}
 	return []*types.SignedMessage(secp), []*types.SignedMessage(bls), nil
 }
 
@@ -70,7 +70,7 @@ func (ms *MessageStore) StoreMessages(ctx context.Context, secpMessages, blsMess
 	if err != nil {
 		return ret, err
 	}
-	
+
 	ret.SecpRoot = secpRoot
 	ret.BLSRoot = blsRoot
 	return ret, nil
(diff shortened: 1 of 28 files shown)


chain/message_store.go
```diff
@@ -13,14 +13,14 @@ import (
 // MessageProvider is an interface exposing the load methods of the
 // MessageStore.
 type MessageProvider interface {
-	LoadMessages(context.Context, cid.Cid) ([]*types.SignedMessage, error)
+	LoadMessages(context.Context, types.TxMeta) ([]*types.SignedMessage, []*types.SignedMessage, error)
```
anorth: Please name these return values

anorth: It might be too late, but since the TxMeta and the execution order of these messages is defined to be BLS and then SECP, returning them in that order might slightly reduce cognitive friction. Maybe this will be more obvious when the BLS ones are not `SignedMessage`, which I assume is coming in a follow-up.

net/graphsync_fetcher.go
```diff
@@ -224,7 +224,9 @@ func (gsf *GraphSyncFetcher) fetchRemainingTipsets(ctx context.Context, starting
 // non-recursively
 func (gsf *GraphSyncFetcher) fetchBlocks(ctx context.Context, cids []cid.Cid, targetPeer peer.ID) error {
 	selector := gsf.ssb.ExploreFields(func(efsb selectorbuilder.ExploreFieldsSpecBuilder) {
-		efsb.Insert("messages", gsf.ssb.Matcher())
+		efsb.Insert("messages", gsf.ssb.ExploreFields(func(messagesSelector selectorbuilder.ExploreFieldsSpecBuilder) {
```
anorth: I think we need an issue to track that chain sync omits BLS messages, and then to fix and test it!

chain/message_store.go
acruikshank: This is definitely covered in the next PR.

Issue #3558, opened 2019-10-14
Validation for forwarding pubsub messages on message topic 


ZenGround0:
### Description
Splitting off the message forwarding validation from #3552

PR #3559, opened 2019-10-14
Pubsub message validation

ZenGround0:
### Motivation
message topic pubsub validation

### Proposed changes
New message topic validator for constructing pubsub validator messages

Closes issue #3558

### DIFF
--- a/internal/pkg/net/validators.go
+++ b/internal/pkg/net/validators.go
@@ -10,6 +10,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/metrics"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 )
 
 var blockTopicLogger = logging.Logger("net/block_validator")
--- a/internal/pkg/net/validators.go
+++ b/internal/pkg/net/validators.go
@@ -13,8 +13,11 @@ import (
 )
 
 var blockTopicLogger = logging.Logger("net/block_validator")
+var messageTopicLogger = logging.Logger("net/message_validator")
 var mDecodeBlkFail = metrics.NewInt64Counter("net/pubsub_block_decode_failure", "Number of blocks that fail to decode seen on BlockTopic pubsub channel")
 var mInvalidBlk = metrics.NewInt64Counter("net/pubsub_invalid_block", "Number of blocks that fail syntax validation seen on BlockTopic pubsub channel")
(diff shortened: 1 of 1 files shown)


net/validators.go
```diff
@@ -58,3 +61,48 @@ func (btv *BlockTopicValidator) Validator() pubsub.Validator {
 func (btv *BlockTopicValidator) Opts() []pubsub.ValidatorOpt {
 	return btv.opts
 }
+
+// MessageTopicValidator may be registered on go-libp3p-pubsub to validate
```
anorth: `libp2p`

net/validators.go
```diff
@@ -58,3 +61,48 @@ func (btv *BlockTopicValidator) Validator() pubsub.Validator {
...
+				messageTopicLogger.Debugf("message %s from peer: %s failed to validate: %s", mCid.String(), p.String(), err.Error())
+				mInvalidMsg.Inc(ctx, 1)
+				return false
+			}
+			return true
+		},
+	}
+}
+
+// Topic returns the topic string BlockTopic
```
anorth: MessageTopic

node/builder.go
```diff
@@ -376,14 +376,21 @@ func (b *Builder) buildNetwork(ctx context.Context, config *config.BootstrapConf
 }
 
 func (b *Builder) buildMessaging(ctx context.Context, network *NetworkSubmodule, chain *ChainSubmodule, wallet *WalletSubmodule) (MessagingSubmodule, error) {
-	msgPool := message.NewPool(b.Repo.Config().Mpool, consensus.NewIngestionValidator(chain.State, b.Repo.Config().Mpool))
+	poolValidator := consensus.NewIngestionValidator(chain.State, b.Repo.Config().Mpool)
+	msgPool := message.NewPool(b.Repo.Config().Mpool, poolValidator)
```
anorth: Does this mean we're doing validation twice, once in the pubsub and then again in the pool? We shouldn't do that, but have clear expectations about what is validated when. 

Validating syntax at the network makes sense, but I'm much less confident that the more expensive validation of comparing the message with blockchain state does. Maybe the spec people have some insight there?

ZenGround0: > Does this mean we're doing validation twice

Yup

> We shouldn't do that, but have clear expectations about what is validated when

I agree, this is one of those imperfect steps towards untangling the massive knot of what-gets-validated-when.  There is no reasonable single PR that is going to get us where we want to go.  However it is noted that I have under-communicated this objective and I will create an issue documenting thought process and evolving plan.

> Maybe the spec people have some insight there?

We can wait for more details but right now the spec claims " the account in question must have enough funds to cover the actions specified" in order for propagation validation to pass.  This is enough for me to want to plug full validation into this part of the system as a first pass.

anorth:
> This is enough for me to want to plug full validation into this part of the system as a first pass

Ok, that's fine. Let's make that clear in the comments. To the extent that the code doesn't represent our ideal, describe the difference.

Does this represent the totality of message validation, or is there more to come (and if so, is there an issue)?

PR #3562, opened 2019-10-15
BLS signed message processing

### LINKED ISSUE #1610: Miners need to aggregate signatures when mining a block
### Description
This implies a new field will need to be added to blocks, said field should contain the signature representing the aggregate signature of all messages contained by the block.
Details to come when they surface.

### Acceptance criteria
TODO

acruikshank:
### Motivation

BLS signatures have been adopted in the spec to increase bandwidth by removing individual message signatures from mined message. For BLS signatures to work, message senders must have a BLS private key. This must be used to create their address (which is a protocol + BLS public key). 

When messages from these users are mined, the miner can aggregate all the BLS signatures into a single signature that will be attached to the block header, and then all the individual signatures can be discarded. Block validation must validate the BLS signature in the block header against all the BLS signed messages, before processing.

### Proposed changes

Sorry this is a long PR. It's best read as series of commits which address different concerns:

1. The first commit adds a `BLSAggregateSig` field to the block header (which will be set in a later commit). Mostly this commit is concerned with altering the order in which messages are mined. The spec declares that all BLS messages should be processed prior to Secp messages.
2. The second commit deals with the realization that although go-filecoin implements BLS addresses, it's missing most of the functionality needed to sign and validate using BLS. This commit expands the options for KeyInfo's to include BLS and adds conditionals to sign or verify based on what type of cryptography system is required by the key or the address.
3. This commit updates go-bls-signatures. The old version would panic when attempting to verify with new data. This has been fixed in the updated version.
4. This commit continues the work started in the second commit by adding code to support BLS signing and verification and cleaning up the wallet and crypto packages.
5. This commit finally gets back to BLS processing for block mining. This takes the BLS messages split out in the first commit, creates a new BLS aggregate signature from all their individual signatures, sets the signature in the block, and stores the BLS messages sans SignedMessage wrapper.
6. This commit handles the verification side. The strategy is to start by verifying that the block BLS signature validates against all the public keys of the BLS messages. It then wraps all the BLS messages, combines them with the secp messages (BLS first) and then sends them to the processor. It modifies the default processor used by the processor to skip signature processing if the message is BLS to cope with the fact that the BLS messages are wrapped in empty SignedMessage wrappers.

closes issue #1609 
closes #1610 
(more or less, these issues are old)

### Future work

* This PR does not switch the default address generation to BLS. This will be the subject of a future PR.
* This PR leaves the `MeteredMessage` wrapper around `Message` that is not described in the spec. Combining these structures is worthwhjile, but would add a lot of noise to an already long and noisy PR. This is saved for future work.


chain/syncer.go
```diff
@@ -189,7 +192,7 @@ func (syncer *Syncer) syncOne(ctx context.Context, grandParent, parent, next typ
 
 	// Run a state transition to validate the tipset and compute
 	// a new state to add to the store.
-	root, err := syncer.stateEvaluator.RunStateTransition(ctx, next, nextMessages, nextReceipts, ancestors, parentWeight, stateRoot)
+	root, err := syncer.stateEvaluator.RunStateTransition(ctx, next, nextBlsMessages, nextSecpMessages, nextReceipts, ancestors, parentWeight, stateRoot)
```
anorth: Open question about at what point we should validate all the signatures and then strip them and just deal with a single list of messages. I think there are arguments that it could be as high as this point, i.e. immediately after reading them from a block (via the message provider). The BLS message signatures of course must have been already verified - checking all the signatures at the same time, and then assuming them checked, would be simpler in some ways.

We could probably introduce a SigVerifiedMessages wrapper for type checking to help us track that state.

chain/syncer_integration_test.go
```diff
@@ -205,7 +205,7 @@ func (isb *integrationStateBuilder) ComputeState(prev cid.Cid, blocksMessages []
 		return genRoot, nil
 	}
 	// setup fork with state we associate with more power
-	if len(blocksMessages[0]) > 0 {
+	if len(secpMessages[0]) > 0 {
```
anorth: This is only ok because it's test code. I think we should leave a comment stating the assumption of secp messages made here. Or just generalise it now to handle possibly-empty secp messages but also check bls messages.

consensus/validation.go
```diff
@@ -36,14 +36,15 @@ type SignedMessageValidator interface {
...
+	allowHighNonce          bool
+	bypassBLSSignatureCheck bool
 }
 
 // NewDefaultMessageValidator creates a new default validator.
 // A default validator checks for both permanent semantic problems (e.g. invalid signature)
 // as well as temporary conditions which may change (e.g. actor can't cover gas limit).
 func NewDefaultMessageValidator() SignedMessageValidator {
-	return &defaultMessageValidator{}
+	return &defaultMessageValidator{bypassBLSSignatureCheck: true}
```
anorth: Please TODO and link to an issue for changing this.

types/testing.go
```diff
@@ -67,6 +66,37 @@ func NewMockSignersAndKeyInfo(numSigners int) (MockSigner, []KeyInfo) {
 	return signer, ki
 }
 
+// MustGenerateMixedKeyInfo produces m bls keys and n secp keys.
+// BLS and Secp will be interleaved. The keys will be valid, but not deterministic.
```
anorth: Non-determinism is really sucky for tests (e.g. block CIDs can't be relied upon). Please make this deterministic or file and link to an issue for whatever makes that currently impossible.

ZenGround0: Filed #3570 for this.  Easy way is to generate in advance and store in a package constant.  Hard way is to change BLS sigs interface and write some rust.

gengen/util/gengen.go
```diff
@@ -161,14 +165,15 @@ func GenGen(ctx context.Context, cfg *GenesisCfg, cst *hamt.CborIpldStore, bs bl
...
+	if !emptySignedMessagesCid.Equals(types.EmptyMessagesCID) ||
+		!emptyMessagesCid.Equals(types.EmptyMessagesCID) ||
 		!emptyReceiptsCid.Equals(types.EmptyReceiptsCID) {
 		return nil, errors.New("bad CID for empty messages/receipts")
 	}
 
 	geneblk := &types.Block{
 		StateRoot:       stateRoot,
-		Messages:        types.TxMeta{SecpRoot: emptyMessagesCid, BLSRoot: emptyMessagesCid},
+		Messages:        types.TxMeta{SecpRoot: emptySignedMessagesCid, BLSRoot: emptySignedMessagesCid},
```
ZenGround0: nit: bls root should probably be `emptyMessagesCid`

mining/worker_test.go
```diff
@@ -357,6 +358,27 @@ func TestApplyBLSMessagesFist(t *testing.T) {
...
+			digests = append(digests, bls.Hash(msgBytes))
+
+			pubKey := bls.PublicKey{}
+			copy(pubKey[:], msg.From.Payload())
+			keys = append(keys, pubKey)
+		}
+
+		blsSig := bls.Signature{}
+		copy(blsSig[:], block.BLSAggregateSig)
+		bls.Verify(&blsSig, digests, keys)
```
ZenGround0: I think this needs an assert

types/message.go
```diff
@@ -133,6 +133,34 @@ func DecodeMessages(b []byte) ([]*SignedMessage, error) {
...
+		panic(err)
+	}
+
+	return obj
+}
+
+// MessageCollection tracks a group of messages and assigns it a cid.
+type MessageCollection []*MeteredMessage
+
+// DecodeMessages decodes raw bytes into an array of signed messages
```
ZenGround0: into an array of metered messages

chain/syncer.go
ZenGround0: This is a great point.  If the validation order were to remain unchanged I am strongly in favor of the argument to do these checks as soon as possible and then deal only with an ordered list of messages across the consensus boundary.  

However because of changes to and ongoing churn in the spec our existing validation pipeline is expected to change a lot.  It's not worth doing this reorganization until we have a solid understanding of the order all of the validation steps at which point we can rearrange everything more or less at once.

consensus/expected.go
```diff
@@ -257,7 +263,9 @@ func (c *Expected) runMessages(ctx context.Context, st state.Tree, vms vm.Storag
 			return nil, errors.Wrap(err, "error validating block state")
 		}
 
-		receipts, err := c.processor.ProcessBlock(ctx, cpySt, vms, blk, tsMessages[i], ancestors)
+		// wrap bls messages and combine to process bls messages first
+		msgs := append(wrapMessages(blsMessages[i]), secpMessages[i]...)
```
ZenGround0: note for the future: this feels like one step backwards.  We should be verifying secp signatures and unwrapping them to match BLS, not rewrapping BLS in signed message with an empty signature.  Totally fine for now though.

consensus/expected.go
```diff
@@ -297,3 +306,40 @@ func (c *Expected) createPowerTableView(st state.Tree) PowerTableView {
 func (c *Expected) loadStateTree(ctx context.Context, id cid.Cid) (state.Tree, error) {
 	return state.LoadStateTree(ctx, c.cstore, id)
 }
+
+// verifyBLSMessageAggregate errors if the bls signature is not a valid aggregate of message signatures
+func verifyBLSMessageAggregate(sig types.Signature, msgs []*types.MeteredMessage) error {
```
ZenGround0: Does this work when there are no bls messages?  Does `VerifyBLSAggregate(nil, nil, nil)` evaluate to true?

consensus/validation.go
ZenGround0: Where does this come from?  Shouldn't we be validating individual message signatures?

wallet/dsbackend_test.go
```diff
@@ -26,7 +27,7 @@ func TestDSBackendSimple(t *testing.T) {
 	assert.Len(t, fs.Addresses(), 0)
```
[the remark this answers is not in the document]
ZenGround0: I think this is invalidated by a later commit that exercises bls keys / signatures sufficiently.  Disregard.

chain/syncer.go
acruikshank: The question is a good one, but pushing validation higher would change the relationship between syncer and expected (which currently handles all validation). So, not in this PR.

consensus/expected.go
acruikshank: An empty signature does not validate against an empty slice of messages. Blocks will need to contain a valid `BLSAggregateSig` of a an empty list of bls messages (i.e. `bls.Aggregate([]bls.Signature{})`).

crypto/crypto_test.go
```diff
@@ -26,39 +28,52 @@ func TestGenerateKey(t *testing.T) {
...
 	digest3 := make([]byte, 70)
 	copy(digest3, digest)
-	assert.False(t, crypto.Verify(pk, msg, digest3))
+	assert.False(t, crypto.VerifySecp(pk, msg, digest3))
 
 	recovered, err := crypto.EcRecover(msg, digest)
 	assert.NoError(t, err)
 	assert.Equal(t, recovered, crypto.PublicKey(sk))
 }
+
```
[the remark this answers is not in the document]
acruikshank: I added some similar failure cases.

types/testing.go
acruikshank: Thanks @ZenGround0!

wallet/wallet.go
```diff
@@ -107,46 +107,21 @@ func (w *Wallet) SignBytes(data []byte, addr address.Address) (types.Signature,
 	return backend.SignBytes(data, addr)
 }
 
-// GetAddressForPubKey looks up a KeyInfo address associated with a given PublicKey
-func (w *Wallet) GetAddressForPubKey(pk []byte) (address.Address, error) {
```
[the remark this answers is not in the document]
acruikshank: It wasn't used. If we kept it we'd have to modify it to specify which type of address or break it out into two methods or check for both types of addresses. Without any motivating usages, it's hard to say which.

PR #3563, opened 2019-10-16
Pull message signature validation higher up the call tree

anorth:
### Motivation
Validation (and probably other future test code) wants to be able to invoke the processor with a message, assuming the signature has already been checked or is otherwise irrelevant.

### Proposed changes
This pulls the signature validation up the call tree, then passes the MeteredMessage to ApplyMessage (which validation code can then also call directly, bypassing signature validation).

WIP, nearly done. FYI @frrist @acruikshank, but I'm also thinking about the opposite approach of forcing signed messages in validation tools.


anorth:
On further thought, you can probably ignore this. I think going with signed messages might be the best approach anyway.

PR #3575, opened 2019-10-18
move block and tipset related structures to new block package

acruikshank:
### Motivation

We have too many data structures in our `types` package.

### Proposed changes

Move `Block`, `TipSet` and related structs to a new `block` package (and change almost everything to accommodate that change).


block/tipset_test.go
```diff
@@ -1,42 +1,44 @@
-package types
+package block
```
anorth: This should be block_test. I think that will then require moving `RequireNewTipSet` back to a block/testing.go file.

acruikshank:
@anorth My first attempt was moving this to `chain`, but almost everything already in chain would have to come out to avoid circular dependencies. We're in the same situation with the `message` package and the message structs in `types`. FWIW the spec describes these structures as being in a `block` package.

block/tipset_test.go
acruikshank: I moved this into `block_test` since this is the only test that requires it, I'm leaving `RequireNewTipSet` in `tipset_test.go`.

PR #3581, opened 2019-10-22
refactor: move code to serialize using the new encoding pkg

icorderi:


encoding/encoding.go
```diff
@@ -12,38 +17,189 @@ type Decodable interface {
...
+	Bytes() []byte
 }
 
 // Decoder represents types that can decode values.
 type Decoder interface {
-	// SetBytes sets the initializer internal bytes to match the input.
-	SetBytes([]byte)
-	// DecodeObject decodes an object.
-	DecodeObject(obj Decodable) error
+	// DecodeValue encodes a primitive value.
```
ZenGround0: nit: looks like these all say `encode` and should say `decode`

encoding/encoding.go
```diff
@@ -12,38 +17,189 @@ type Decodable interface {
...
+	if encodable, ok := obj.(Encodable); ok {
+		if err = encodable.Encode(encoder); err != nil {
+			return nil, err
+		}
+		return encoder.Bytes(), nil
+	}
+
+	switch v.Kind() {
+	case reflect.Uint:
+		// Review: is the size of `uint` fixed or depends on the architecture?
```
ZenGround0: Both int and uint are platform dependent.  You can read with [unsafe](https://golang.org/pkg/unsafe/#Sizeof)

net/graphsync_fetcher_test.go
```diff
@@ -50,8 +51,8 @@ type notDecodable struct {
 }
 
 func init() {
-	cbor.RegisterCborType(notDecodable{})
-}
+	encoding.RegisterIpldCborType(notDecodable{})
```
ZenGround0: Why do we register some types inline in regular source files but register others using the code generating tool?

icorderi: the ones left behind are because they are doing non-standard things I slowly need to add support for

encoding/testing.go
```diff
@@ -34,6 +34,13 @@ func newTestDecoder() testDecoder {
 	}
 }
 
+// EncodeUint8 encodes a uint8.
```
ZenGround0: nit: encodes a uint

encoding/testing.go
```diff
@@ -62,6 +69,13 @@ func (encoder *testEncoder) EncodeUint64(obj uint64) error {
 	return nil
 }
 
+// EncodeInt8 encodes a int8.
+func (encoder *testEncoder) EncodeInt(obj int) error {
```
ZenGround0: nit: encodes an int

Issue #3580, opened 2019-10-22
Headers-only support in graphsync fetcher


ZenGround0:
### Description
A [recent spec proposal](https://github.com/filecoin-project/specs/pull/568) reiterates the need to separate header fetching from full block fetching.  A first step to achieve this is to add headers-only fetching support to the graphsync fetcher.

### Acceptance criteria
The graphsync fetcher is extended to include functions for fetching headers only.

### Risks + pitfalls
* Selector foo is hard
* tests are already big and unwieldy, let's try to make tests *cleaner* with this PR

### Where to begin
Probably the vast majority of fetcher logic can be abstracted into a common method figure out a nice shared abstraction.

PR #3584, opened 2019-10-23
Graphsync Fetcher can fetch only headers of chain

ZenGround0:
### Motivation
Recent guidelines from the spec call for fetching headers and messages separately during syncing.  We need support from this in our underlying fetcher.

### Proposed changes
Add `FetchTipsetHeaders` to the graphsync fetcher.  This method will fetch only headers, no messages or receipts

It also breaks some graphsync test helpers into a separate file for my own sanity.

Closes issue #3580


net/graphsync_helpers_test.go
```diff
@@ -0,0 +1,441 @@
+package net_test
```
acruikshank: This file is a good idea.

PR #3586, opened 2019-10-24
Use AMT (sharrays) for messages and receipts in blocks.

acruikshank:
### Motivation

The spec says messages (secp and bls) and receipts should included in a block in an AMT data structure.

### Proposed changes

This PR replaces `[]*types.SignedMessage`, `[]*types.UnsignedMessage` and `[]*types.MessageReceipt` data structures previously pointed to by block headers with [AMT](https://github.com/filecoin-project/go-amt-ipld) data structures.

This PR is structured as 3 commits:

1. Replace the message and receipt collections with AMTs in the block header.
2. Introduce Graphsync queries that can retrieve AMT data recursively.
3. Fix the chain export/import to support the new data structures.


anorth:
Not for this PR, but @ZenGround0 as part of chain sync updates we should *not be fetching receipts*. Receipts should always be created by local validation (and then the AMT root compared with block header).

chain/car.go
```diff
@@ -106,6 +108,43 @@ func Export(ctx context.Context, headTS block.TipSet, cr carChainReader, mr carM
...
+func exportAMTReceipts(ctx context.Context, out io.Writer, receipts []*types.MessageReceipt) error {
+	ms := carWritingMessageStore(out)
+
+	_, err := ms.StoreReceipts(ctx, receipts)
+	return err
+}
+
+func carWritingMessageStore(out io.Writer) *MessageStore {
+	d := datastore.NewMapDatastore()
+	bs := blockstore.NewBlockstore(d)
```
anorth: This is pretty confusing in this context. This will accumulate the entire thing in memory, unnecessarily. I think the `carExportBlockstore` should be write-only; the read methods should panic (or return error). Then it won't need this in-memory store at all.

If that doesn't work, using a MessageStore to intermediate this export is probably too indirect a path.

chain/car_test.go
```diff
@@ -257,20 +263,57 @@ func validateBlockstoreImport(t *testing.T, start, stop block.TipSetKey, bstore
...
+			err = secpAMT.ForEach(func(_ uint64, d *typegen.Deferred) error {
+				if err := cbornode.DecodeInto(d.Raw, &c); err != nil {
+					return err
+				}
+
+				b, err := bstore.Get(c)
+				if err != nil {
+					return err
+				}
+				return cbornode.DecodeInto(b.RawData(), &smsg)
```
anorth: Factor out the duplicated code?

chain/testing.go
```diff
@@ -60,18 +65,22 @@ func NewBuilderWithState(t *testing.T, miner address.Address, sb StateBuilder) *
...
+		cstore:       hamt.CSTFromBstore(bs),
+		messages:     NewMessageStore(bs),
 		tipStateCids: make(map[string]cid.Cid),
-		messages:     make(map[types.TxMeta][]*types.SignedMessage),
-		receipts:     make(map[cid.Cid][]*types.MessageReceipt),
 	}
 
-	b.messages[types.TxMeta{SecpRoot: types.EmptyMessagesCID, BLSRoot: types.EmptyMessagesCID}] = []*types.SignedMessage{}
-	b.receipts[types.EmptyMessagesCID] = []*types.MessageReceipt{}
+	ctx := context.Background()
```
anorth: Please use `context.TODO` so it's more clear that this probably isn't ideal

net/graphsync_fetcher.go
```diff
@@ -276,6 +280,14 @@ func (gsf *GraphSyncFetcher) fetchBlocks(ctx context.Context, selGen func() ipld
 	return anyError
 }
 
+func (gsf *GraphSyncFetcher) fetchThroughAMTSelector(recursionDepth uint32) selectorbuilder.SelectorSpec {
+	return gsf.ssb.ExploreIndex(2,
+		gsf.ssb.ExploreRecursive(10,
```
anorth: What are `2` and `10`? Please factor out named constants.

net/graphsync_fetcher.go
```diff
@@ -236,9 +239,10 @@ func (gsf *GraphSyncFetcher) fetchRemainingTipsets(ctx context.Context, starting
 func (gsf *GraphSyncFetcher) fullBlockSel() ipld.Node {
 	selector := gsf.ssb.ExploreFields(func(efsb selectorbuilder.ExploreFieldsSpecBuilder) {
 		efsb.Insert("messages", gsf.ssb.ExploreFields(func(messagesSelector selectorbuilder.ExploreFieldsSpecBuilder) {
-			messagesSelector.Insert("secpRoot", gsf.ssb.Matcher())
+			messagesSelector.Insert("secpRoot", gsf.fetchThroughAMTSelector(10))
+			messagesSelector.Insert("bLSRoot", gsf.fetchThroughAMTSelector(10))
 		}))
-		efsb.Insert("messageReceipts", gsf.ssb.Matcher())
+		efsb.Insert("messageReceipts", gsf.fetchThroughAMTSelector(10))
```
anorth: Please factor out this `10`. I think the intention is that this fetches "everything" so either leave an extensive comment justifying the value 10, or set this to something like 1000 so this doesn't cause a very hard to diagnose failure later.

chain/testing.go
```diff
@@ -195,17 +204,20 @@ func (f *Builder) Build(parent block.TipSet, width int, build func(b *BlockBuild
...
-		require.True(f.t, ok)
-		b.StateRoot, err = f.stateBuilder.ComputeState(prevState, [][]*types.UnsignedMessage{}, [][]*types.SignedMessage{msgs})
+		smsgs, umsgs, err := f.messages.LoadMessages(ctx, b.Messages)
+		require.NoError(f.t, err)
+		b.StateRoot, err = f.stateBuilder.ComputeState(prevState, [][]*types.UnsignedMessage{umsgs}, [][]*types.SignedMessage{smsgs})
 		require.NoError(f.t, err)
 
-		f.blocks[b.Cid()] = b
+		// add block to cstore
+		_, err = f.cstore.Put(ctx, b)
```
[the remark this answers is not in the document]
ZenGround0: Looking more I think it lets us do `GetBlockStoreValue` so we can get a `blocks.Block` hassle free. This change introduces too much "global store" code into the builder in my opinion.  One of the nice things about the builder conceptually is that it has so far avoided these pervasive patterns.  

I would prefer if we kept the blocks map for satisfying `GetBlock(s)` and `GetTipSets`.  I think a better way to structure things is to keep the `cstore` relatively hidden, either scoping it to a blockstorevaluegetter subtype or instantiating it on `GetBlockStoreValue` . Then within `GetBlockStoreValue` we could grab the block from the map, Put it into the narrowly scoped cstore and then immediately Get from the blockstore.  If this blockstore and the message store blockstore are different that would also make me happy.

Is there something I'm missing that the above doesn't handle?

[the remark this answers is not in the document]
ZenGround0: Argh we need `GetBlockStoreValue` for fetching messages and AMT nodes for graphsync don't we?  It feels quite bad that we're slowly letting the complexity of the whole system leak into the default case of our test-faker.  We're like slowly boiling frogs enjoying getting warmer.

That said this step isn't THAT bad.  I won't block this PR on this.  We should consider intervening soon though.

net/graphsync_fetcher_test.go
```diff
@@ -118,17 +129,24 @@ func TestGraphsyncFetcher(t *testing.T) {
...
-			expectedMessages, _, err := builder.LoadMessages(ctx, blk.Messages)
+
+			// use fetcher blockstore to retrieve messages
+			msgStore := chain.NewMessageStore(bs)
+			secpMsgs, blsMsgs, err := msgStore.LoadMessages(ctx, blk.Messages)
 			require.NoError(t, err)
-			require.True(t, reflect.DeepEqual(messages, expectedMessages))
-			rawBlock, err = bs.Get(blk.MessageReceipts)
+
+			// get messages from builders block store
```
ZenGround0: nit get expectedMessages from

net/graphsync_fetcher_test.go
```diff
@@ -118,17 +129,24 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+			expectedSecpMessages, expectedBLSMsgs, err := builder.LoadMessages(ctx, blk.Messages)
 			require.NoError(t, err)
-			receipts, err := types.DecodeReceipts(rawBlock.RawData())
+
+			require.True(t, reflect.DeepEqual(secpMsgs, expectedSecpMessages))
+			require.True(t, reflect.DeepEqual(blsMsgs, expectedBLSMsgs))
+
+			// use fetcher blockstore to retrieve messages
```
ZenGround0: nit receipts

net/graphsync_fetcher_test.go
```diff
@@ -118,17 +129,24 @@ func TestGraphsyncFetcher(t *testing.T) {
...
+			receipts, err := msgStore.LoadReceipts(ctx, blk.MessageReceipts)
 			require.NoError(t, err)
+
+			// get messages from builders block store
```
ZenGround0: nit expected receipts

net/graphsync_helpers_test.go
```diff
@@ -360,35 +361,33 @@ func (mv mockSyntaxValidator) ValidateMessagesSyntax(ctx context.Context, messag
...
 type blockAndMessageProvider interface {
-	chain.MessageProvider
-	chain.BlockProvider
+	GetBlockstoreValue(ctx context.Context, c cid.Cid) (blocks.Block, error)
 }
 
-func tryBlockNode(ctx context.Context, f chain.BlockProvider, c cid.Cid) (format.Node, error) {
-	if block, err := f.GetBlock(ctx, c); err == nil {
-		return block.ToNode(), nil
+func tryBlockMessageReceiptNode(ctx context.Context, f blockAndMessageProvider, c cid.Cid) (format.Node, error) {
```
ZenGround0: nit: change name to `tryBlockStoreValue`

types/message.go
```diff
@@ -21,22 +24,29 @@ type GasUnits = Uint64
 // BlockGasLimit is the maximum amount of gas that can be used to execute messages in a single block
 var BlockGasLimit = NewGasUnits(10000000)
 
+// EmptyMessagesCID is the cid of an empty collection of messages.
+var EmptyMessagesCID cid.Cid
+
+// EmptyReceiptsCID is the cid of an empty collection of receipts.
+var EmptyReceiptsCID cid.Cid
```
ZenGround0: There are a few places you might be able to use this and don't during genesis construction assuming those functions can take types as dependency.

chain/message_store.go
```diff
@@ -26,72 +30,219 @@ type MessageWriter interface {
...
+	for i, msg := range receipts {
+		cids[i], err = ms.storeBlock(msg)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return cids, nil
+}
+
+func (ms *MessageStore) storeBlock(data interface{}) (cid.Cid, error) {
```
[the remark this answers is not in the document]
acruikshank: I started by trying to do this. Go doesn't allow you to cast, say, a []*types.UnsignedMessage to []interface{}, so we'd have to copy the slice into a new []interface{} using a for loop before passing it in. At this point the refactor doesn't feel worth it, which is why I have what I have here.

chain/testing.go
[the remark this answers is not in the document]
acruikshank: I agree the builder has gotten more complicated in some ways. It's gotten simpler in others (it no longer has to maintain the broken out structure of the block in multiple maps). The added complexity is entirely driven by the additional complexity of block topology.

I agree that we should work hard to keep functional logic from leaking into the builder, but, in this case, we are only adding our data structure logic. This logic has gotten significantly more complicated, so I'd rather not replicate it in the builder.

We should definitely keep this concern in mind as we make future changes to the builder.

mining/worker_test.go
```diff
@@ -56,7 +56,7 @@ func Test_Mine(t *testing.T) {
 	baseBlock := &block.Block{Height: 2, StateRoot: stateRoot, Tickets: []block.Ticket{{VRFProof: []byte{0}}}}
 	tipSet := th.RequireNewTipSet(t, baseBlock)
 
-	st, pool, addrs, cst, bs := sharedSetup(t, mockSignerVal)
+	st, pool, addrs, _, bs := sharedSetup(t, mockSignerVal)
```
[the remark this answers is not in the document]
acruikshank: nope

net/graphsync_fetcher.go
acruikshank: Recursion depth of 10 allows us to retrieve a amt with a billion elements. Recursion depth of 1000 makes my head hurt. Arguably it should be much lower. I extracted it to a constant and documented the choice.

Issue #3583, opened 2019-10-23
Bound Syncer Dispatcher's work queue


ZenGround0:
### Description
We expect many syncing targets to propagate to the syncer dispatcher.  We also expect the work done by the syncer to potentially take a very long time to complete.  To prevent the work queue from building up a large number of targets, many of which will be useless by the time the syncer is ready for another target, we should bound the work queue. 

Proposal: In the case the work queue hits the bound we can insert all incoming targets onto the work queue and then use the heap `Remove` method to remove the last `totalTargets - workQueueBound` targets on the queue.  This way we keep the bound and only drop the lowest priority targets.

### Acceptance criteria
Bounded dispatcher worker queue that keeps highest priority targets

### Risks + pitfalls
There is a remote possibility this introduces time dependence into the tests in the case something is going really wrong there.  Probably we are never hitting any >10 bound in tests so its not going to be an issue.

### Where to begin
It makes sense to add this checking and removing to the `TargetQueue` type.  Either we can do it each time within `Push` or we can add a `MaybeTrim` (or something) method.

anorth:
I'm not sure this is safe. If "high priority" means high reported height, then malicious messages with unverified heights could crowd out the real chain.

Safer and simpler would be to just drop inbound on the floor if the queue is full, wait for space to appear. (Dropping older, rather than newer, could be slightly more efficient, but probably not worth the bookkeeping). More sophisticated rate limiting would probably take the source into account too.

PR #3587, opened 2019-10-25
Bounded dispatcher queue

ZenGround0:
### Motivation

We can't afford to let memory grow arbitrarily, this PR safely bounds the number of incoming sync requests held in memory.

### Proposed changes
Add a parameter, do as @anorth's suggested and drop any incoming targets without adding to queue.

Closes issue #3583


syncer/dispatcher.go
```diff
@@ -11,9 +11,11 @@ import (
...
 var log = logging.Logger("sync.dispatch")
 
-// This is the size of the channel buffer used for receiving targets from
-// producers.
-const incomingBufferSize = 5
+// IncomingBufferSize is the size of the channel used for receiving targets from producers.
+const IncomingBufferSize = 5
+
+// WorkQueueBound is the size of the work queue
+const WorkQueueBound = 20
```
ZenGround0: I exposed these for tests, if its a problem I can add as constructor inputs or introduce assumptions into the tests that need a small amount of manual intervention.

syncer/dispatcher.go
```diff
@@ -25,8 +27,8 @@ func NewDispatcher(catchupSyncer syncer) *Dispatcher {
 	return &Dispatcher{
 		workQueue:     NewTargetQueue(),
 		catchupSyncer: catchupSyncer,
-		incoming:      make(chan Target, incomingBufferSize),
-		control:       make(chan interface{}),
+		incoming:      make(chan Target, IncomingBufferSize),
+		control:       make(chan interface{}, 1),
```
ZenGround0: The test was quite tricky and this solved my problem.  I'm not sure I could have tested this with a 0 buffer as before.  I think it is ok because it does not change much from a user uninterested in callback registration, but it opens up way simpler patterns for consumers (potentially non-test in the future) trying to synchronize with a callback.

PR #3593, opened 2019-10-25
add syncer back to chain submodule

frrist:
### Motivation
Without this change cli commands that use the syncer will panic as the syncer is nil.


frrist:
Closing as this has been resolved

Issue #3555, opened 2019-10-14
go-filecoin can export CAR file with all "fringe" (childless) tips from a chain store


frrist:
### Description
As a follow on to #3473 go-filecoin should be able to export a CAR file with all "fringe" (childless) tips from a chain store. One motivating goal of this feature is to allow a developer to force a go-filecoin daemon to replay chains from different forks.

### Acceptance criteria
- go-filecoin chain export command is modified to optionally allow a node to export its chain with all childless tipsets included in the exported file.

### Where to begin
TODO fill in details when #3473 merges.

PR #3594, opened 2019-10-25
standalone binary for chain export and inspection

frrist:
### Motivation
Create a standalone binary for exporting a blockchain from a go-filecoin repo to a car file.


anorth:
This appears to duplicate a lot of code that already exists for traversing and exporting. The intention of "standalone" is not that it doesn't depend on existing code, just that it can run without requiring a daemon.

For #3555 the head/s must be read from the chain store. The CLI argument should be optional, defaulting to whatever the chain store has stored as the head (and later when 3555 is addressed, all of the heads).

frrist:
Not sure what happened with CI here, going to ignore it until I moved this from draft to "ready".
```
# pkg-config --cflags  -- /go/src/github.com/filecoin-project/go-filecoin/go-bls-sigs/libbls_signatures.pc
Failed to open '/go/src/github.com/filecoin-project/go-filecoin/go-bls-sigs/libbls_signatures.pc': No such file or directory
No package '/go/src/github.com/filecoin-project/go-filecoin/go-bls-sigs/libbls_signatures.pc' found
pkg-config: exit status 1
# pkg-config --cflags  -- /go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc /go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc
Failed to open '/go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc': No such file or directory
No package '/go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc' found
Failed to open '/go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc': No such file or directory
No package '/go/src/github.com/filecoin-project/go-filecoin/go-sectorbuilder/sector_builder_ffi.pc' found
pkg-config: exit status 1
FAIL	github.com/filecoin-project/go-filecoin/actor [build failed]
FAIL	github.com/filecoin-project/go-filecoin/actor/builtin/account [build failed]
FAIL	github.com/filecoin-project/go-filecoin/actor/builtin/miner [build failed]
FAIL	github.com/filecoin-project/go-filecoin/actor/builtin/storagemarket [build failed]
FAIL	github.com/filecoin-project/go-filecoin/block [build failed]
```

tools/chain-util/pkg/export/export.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, errors.Wrapf(err, "failed to load link from dagservice %s", c)
+	}
+	csc.addState(nd)
+	return nd.Links(), nil
+
+}
+
+func (csc *chainStateCollector) addState(nd format.Node) {
+	csc.state = append(csc.state, nd)
+}
```
frrist: Copy-Pasta from here
https://github.com/filecoin-project/go-filecoin/blob/4f71dca6ec0a18e3b19445c39ef06033e51a859e/internal/app/go-filecoin/plumbing/cst/chain_state.go#L233-L277

I'm not happy about this code duplication, but it seems better than exporting `newChainStateCollector` from chain_state.go, thoughts?

go.mod
```diff
@@ -62,7 +60,6 @@ require (
 	github.com/libp2p/go-libp2p-circuit v0.1.3
 	github.com/libp2p/go-libp2p-core v0.2.4-0.20191022172111-a5bf2487c11d
 	github.com/libp2p/go-libp2p-kad-dht v0.1.1
-	github.com/libp2p/go-libp2p-peer v0.2.0
 	github.com/libp2p/go-libp2p-peerstore v0.1.3
```
frrist: red in this file is from running `go mod tidy`

tools/chain-util/pkg/export/export.go
```diff
@@ -0,0 +1,173 @@
...
+// NewChainExporter returns a ChainExporter.
+func NewChainExporter(repoPath string, out io.Writer) (*ChainExporter, error) {
+	// open the datastore in readonly mode
+	badgerOpt := &badgerds.DefaultOptions
+	badgerOpt.ReadOnly = true
+
+	log.Infof("opening filecoin datastore: %s", repoPath)
+	badgerPath := filepath.Join(repoPath, "badger/")
+	log.Infof("opening badger datastore: %s", badgerPath)
+	badgerDS, err := badgerds.NewDatastore(badgerPath, badgerOpt)
```
ZenGround0: Since we are not using the go-filecoin/repo abstraction there doesn't seem to be any protection against multiple processes obtaining a handle to a datastore wrapping the same path.

I'm a bit worried that this could lead to accidental corruption if someone runs this tool while a daemon is running.

Lowest effort solution is to include an all CAPs user facing warning somewhere.

I think "the right thing" here is to open up a go-filecoin repo abstraction and then grab the datastores with `Datastore` and `ChainDatastore`.  Even if we're lucky and concurrent read-read and read-write works for badger right now it adds a hidden requirement to our datastore choice that could erode with time.

tools/chain-util/pkg/export/export.go
ZenGround0: Could you elaborate on why importing from chain_state.go is worse?  Without your context importing seems like the better choice.  You could put this type in the chain package or in a new `ipld`/`dag` or the util package if you don't want to export free functions from plumbing.

tools/chain-util/pkg/export/export.go
```diff
@@ -0,0 +1,173 @@
...
+	out io.Writer
+}
+
+// Export will export a chain (all blocks and their messages) to the writer `out`.
+func (ce *ChainExporter) Export(ctx context.Context) error {
+	msgStore := chain.NewMessageStore(ce.bstore)
+	return chain.Export(ctx, ce.Head, ce, msgStore, ce, ce.out)
+}
+
+// GetTipSet gets all the TipSet for a given TipSetKey from the ChainExporter blockstore.
```
ZenGround0: nit: gets the TipSet

tools/chain-util/pkg/export/export.go
frrist: > Since we are not using the go-filecoin/repo abstraction there doesn't seem to be any protection against multiple processes obtaining a handle to a datastore wrapping the same path.

Since we are using badger to open the repo it will fail to open if a [repo lock exists](https://github.com/dgraph-io/badger#opening-a-database), from the linked README.md:
> Please note that Badger obtains a lock on the directories so multiple processes cannot open the same database at the same time.

Here is what a user will see if they attempt to run this binary on a repo that currently has a lock held by another (go-filecoin) process.
```
./chain-util export --repo=/home/frrist/.filecoin/ --out=/tmp/test.car
2019-11-11T11:25:10.586-0800    INFO    chain-util/export       export/export.go:44     opening repo: /home/frrist/.filecoin/
2019-11-11T11:25:10.586-0800    WARN    chain-util      chain-util/main.go:75   Cannot acquire directory lock on "/home/frrist/.filecoin/badger".  Another process is using this Badger database.: resource temporarily unavailable
```

Also note that we are opening the repo in read-only mode:
https://github.com/filecoin-project/go-filecoin/pull/3594/files#diff-2bbea98c52d3e94072ef6b7ba1a9250dR27-R29

ZenGround0: That's good!  It moves this from a correctness to a bitrot thing. It still doesn't address this part of the comment:

>  Even if we're lucky and concurrent read-read and read-write works for badger right now it adds a hidden requirement to our datastore choice that could erode with time.

Can we assume that all ipfs datastore implementations MUST repo lock?  If we can, great.  If not the moment we pivot to another datastore or we open up user configured datastores we/users are going to need to satisfy a hidden requirement for interop with this tool which means there's some odds somebody *won't* satisfy this requirement and they'll be left wondering why the tool is corrupting their datastore.

ZenGround0: Grabbing the filecoin repo lock enforces safety across all future datastore implementations.

tools/chain-util/pkg/export/export.go
frrist: Mainly because I didn't like the idea of exporting a free function from plumbing. I will move this into dag plumbing exposed as a `RecursiveGet` method.

tools/chain-util/pkg/export/export.go
frrist: I think I might have misunderstood the initial comment, let me try again.

> Since we are not using the go-filecoin/repo abstraction

Meaning since we are not calling the go-filecoin method [OpenFSRepo](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/repo/fsrepo.go#L131), correct?

> doesn't seem to be any protection against multiple processes obtaining a handle to a datastore wrapping the same path

There _is_ protection against multiple processes getting a handle to a _datastore_  (as I have shown this is provided by the badger datastore package), however since go-filecoin [creates is own repo lock](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/repo/fsrepo.go#L164) and since we are not checking for it to exists here, we are left open to possible corruption in the event that we change our datastore/repo implementation? 

Is the ask that a check for filecoin repo lock is added? (as is done [here](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/repo/fsrepo.go#L164))

ZenGround0: Yeah this is all exactly what I'm thinking.  If we use openfsrepo we get the guarantee of go-filecoin repo locking preventing contention with other software that opens the repo (go-filecoin, the migration tool...) regardless of what datastore we use.

Originally I wasn't sure if badger had its own protection which is why I blocked approval and why I approved when you told me that it does.

PR #3595, opened 2019-10-25
init accepts a car file with full chain

frrist:
### Motivation
The work done in https://github.com/filecoin-project/go-filecoin/pull/3594 creates a car file contaning a chain with genesis block state s.t. a go-filecoin node may be initialized with it e.g. `go-filecoin init --genesisfile=chain.car`.

### DIFF
--- a/cmd/go-filecoin/init.go
+++ b/cmd/go-filecoin/init.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ipfs/go-ipfs-blockstore"
 	"github.com/ipfs/go-ipfs-cmdkit"
 	"github.com/ipfs/go-ipfs-cmds"
+	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-core/crypto"
 
 	"github.com/filecoin-project/go-filecoin/fixtures"
@@ -26,6 +27,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
+
+var logInit = logging.Logger("commands/init")
 
 var initCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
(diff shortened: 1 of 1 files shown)


commands/init.go
```diff
@@ -188,18 +188,58 @@ func loadGenesis(ctx context.Context, rep repo.Repo, sourceName string) (consens
...
+	for !cur.Parents.Equals(block.UndefTipSet.Key()) {
+		bsBlk, err := bs.Get(cur.Parents.ToSlice()[0])
+		if err != nil {
+			return nil, err
+		}
+		cur, err = block.DecodeBlock(bsBlk.RawData())
+		if err != nil {
 			return nil, err
 		}
+	}
```
frrist: I'm not fond of this walking the whole file to find the genesis block business. A couple alternatives that come to mind:
- Store the genesis block as the first entry after the CAR file header.
- (Ab)use the CAR file header and put the genesis block cid in `ch.Roots[0]` and the head TipSets in `ch.Roots[1:]`
- Wrap the CAR file in something specifically made for this use case.

Other ideas?

commands/init.go
```diff
@@ -188,18 +188,58 @@ func loadGenesis(ctx context.Context, rep repo.Repo, sourceName string) (consens
 		return nil, err
 	}
 
-	if len(ch.Roots) != 1 {
-		return nil, fmt.Errorf("expected car with only a single root")
+	// need to check if we are being handed a car file with a single genesis block or an entire chain.
```
frrist: Do we want to support initializing with a CAR file that has all state tree data as well?

node/testing.go
```diff
@@ -64,29 +64,30 @@ func MakeChainSeed(t *testing.T, cfg *gengen.GenesisCfg) *ChainSeed {
...
 
 	var blk block.Block
 	if err := cst.Get(context.TODO(), cs.info.GenesisCid, &blk); err != nil {
-		return nil, err
+		return nil, block.UndefTipSet, err
 	}
 
-	return &blk, nil
+	head, err := block.NewTipSet(&blk)
+	return &blk, head, err
```
anorth: It seems strictly redundant to return both the tipset and block. I don't understand the intent of this change.

commands/init.go
anorth: I don't think any special handling should be required (at least not here). As I understand it, `car.LoadCar` will load all blocks from the CAR into the store: blockchain, messages, receipts, state trees, whatever it has.

node/testing.go
frrist: As the chain init code reads now, the genesis block is required when initializing the store and the head tipset could be passed to the SetHead call:
https://github.com/filecoin-project/go-filecoin/pull/3595/files#diff-d5beb8fa8ed2190fa7862dfc6120a4deR22-R42

commands/init.go
frrist: Sorry I should have been more clear here, there are 4 cases I can think of for how a node could be initialized:
1. No genesis block is passed, just a vanilla `go-filecoin init`
2. A genesis block/CAR-file is passed:
  a. It contains only genesis block and state.
  b. It contains a chain, genesis block, and its state.
  c. It contains a chain with all state for every transition. 

go-filecoin has cases 1 and 2.a covered, and the code added here mostly addresses 2.b- all that's left is to sync up to the head returned by the modified genesis function here. 
I believe 2.c may require some special handing here to communicate to the chain init method that it should persist the state roots contained in the CAR file to its chain store.

I am realizing now that adding support for 2.c (if we choose to do so) is outside the scope of this PR and have filed https://github.com/filecoin-project/go-filecoin/issues/3606.

cmd/go-filecoin/init.go
```diff
@@ -180,44 +219,63 @@ func loadGenesis(ctx context.Context, rep repo.Repo, sourceName string) (consens
...
+		gensisBlk = cur
 
-func getNodeInitOpts(peerKeyFile string) ([]node.InitOpt, error) {
-	var initOpts []node.InitOpt
-	if peerKeyFile != "" {
-		data, err := ioutil.ReadFile(peerKeyFile)
-		if err != nil {
-			return nil, err
+		// the root cids of the car file are a tipset, they form the head of a chain
+		var heads []*block.Block
```
ZenGround0: Because this looks redundant at first glance (why not just return the cids?) please leave a comment explaining that the block data and tipset construction is needed to sort the cids in the correct order.

frrist:
>This PR doesn't seem to include the actual init flag for receiving a chain from a local file but only allows remote car files to contain whole chains. Is that right?

This PR extends the method `loadGenesis` to accept a car file that contains a chain while leave the functionality in place that allows it to accept a car file containing just a genesis block. The method `loadGenesis` operates on the value of `--genesisfile` meaning no new flags were needed here.

cmd/go-filecoin/init.go
frrist: You are right, this is redentant -- will remove and just log the CID's.

Issue #3590, opened 2019-10-25
Revert ticket array


ZenGround0:
### Description
The spec reverted most of #2222 #2223 #2225 . Move back to incrementing nonces to generate election proofs during rounds with no winners instead of moving up the ticket chain and generating a new ticket

### Acceptance criteria
Single ticket, no array.  Election proofs use nonce increment.

### Where to begin
Look at old code before we made this change.

PR #3598, opened 2019-10-26
Feat/ticket array revert

ZenGround0:
### Motivation
Align with spec on ticket not including VDF outputs and no longer being an array

### Proposed changes
- remove vdf fields
- make field a ticket not an array
- revert back to nonce generation for election 

Closes issue #3590


internal/pkg/consensus/expected.go
```diff
@@ -221,27 +221,18 @@ func (c *Expected) validateMining(ctx context.Context, st state.Tree, ts block.T
 		}
 
 		// Validate ElectionProof
-		numTickets := len(blk.Tickets)
-		result, err := c.IsElectionWinner(ctx, pwrTableView, blk.Tickets[numTickets-1], blk.ElectionProof, workerAddr, blk.Miner)
+		nullBlkCount := uint64(blk.Height) - prevHeight - 1
```
icorderi: I would go for `blk.Height.Uint64()` which will give us more control over who is doing arithmetic on it in case we ever change the size. Doing `uint64(x)` could drop bytes without us knowing.

ZenGround0: `blk.Height` is a `types.Uint64` which is [defined as a `uint64`](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/types/uint64.go#L32) under the hood.  I think casting to the underlying type is just a formality and won't ever lose us bytes.

Side note I think we're also safe from byte loss when the block's height field is decoded because we force decode into [an 8 byte array](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/types/uint64.go#L57)

PR #3603, opened 2019-10-28
Dayu

dayu26:
### Motivation
I add a sub command sendsigned <message> in "go-filecoin message" which can used to send signed message to message pool directly.

You can use this command to send message like:
{"meteredMessage":{"message":{"to":"t17iitsud4axsddakaguhpwnoqv4koijyo4hjeyfa","from":"t1s47wbhtafgvei7pskkvng2ogqj5ikv5m53coami","nonce":"1","value":"23","method":"","params":null},"gasPrice":"0.001","gasLimit":"100"},"signature":"5wlIwOCJ2EdVgtT6Z7mOvCTr0%2BX24FxieNGG9EiVfdYx84gtaH5PyH4xo9%2BQIhzZZ73G%2BW1NY%2FSp55PUJ8h7%2BQE="}


internal/pkg/message/outbox.go
```diff
@@ -139,6 +140,44 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
...
+	m := types.SignedMessage{}
+
+	b := []byte(message)
+	err = json.Unmarshal(b, &m)
+	if err != nil {
+		return
+	}
+	signed := &m
+
+	height, err := tipsetHeight(ob.chains, head)
```
anorth: Factor the duplicated code from here to the end of the method into a helper function shared by `Send`.

cmd/go-filecoin/message.go
```diff
@@ -136,6 +137,46 @@ var msgSendCmd = &cmds.Command{
 	},
 }
 
+var signedMsgSendCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Send a signed message", // This feels too generic...
+	},
+	Arguments: []cmdkit.Argument{
+		//cmdkit.StringArg("target", true, false, "Address of the actor to send the message to"),
```
anorth: Delete

cmd/go-filecoin/message.go
```diff
@@ -136,6 +137,54 @@ var msgSendCmd = &cmds.Command{
 	},
 }
 
+var signedMsgSendCmd = &cmds.Command{
+	Helptext: cmdkit.HelpText{
+		Tagline: "Send a signed message", // This feels too generic...
```
anorth: Remove the copied commentary.

internal/pkg/message/outbox.go
```diff
@@ -121,16 +121,30 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
...
+	return ob.SignedSend(ctx, signed,bcast)
+}
+
+// Send a signed message, retaining it in the outbound message queue.
+// If bcast is true, the publisher broadcasts the message to the network at the current block height.
+func (ob *Outbox) SignedSend(ctx context.Context, signed *types.SignedMessage, bcast bool) (out cid.Cid, err error) {
+	defer func() {
+		if err != nil {
+			msgSendErrCt.Inc(ctx, 1)
+		}
```
anorth: When an error occurs and this is called from `Send`, this will cause the metric to be incremented twice. I think you need to factor out the body of this method (136 onwards) to a private function/method, called by both Send and SignedSend.

PR #3610, opened 2019-10-29
Remove protocol upgrade and deprecated Weight()

ZenGround0:
### Motivation
We want to reset the alphanet soon.  We don't want to carry along the baggage of the old broken weighting function.

### Proposed changes
- Remove old weight function
- Remove protocol version table from code no longer checking it
- Remove tests testing for upgrade behavior

Closes issue #

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/syncer_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/syncer_submodule.go
@@ -18,7 +18,6 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net/pubsub"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/syncer"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/version"
 )
 
 // SyncerSubmodule enhances the node with chain syncing capabilities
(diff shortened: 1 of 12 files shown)


internal/app/go-filecoin/internal/submodule/syncer_submodule.go
```diff
@@ -57,16 +56,15 @@ type nodeSyncDispatcher interface {
...
 	IsHeavier(ctx context.Context, a, b block.TipSet, aStateID, bStateID cid.Cid) (bool, error)
 }
 
 // NewSyncerSubmodule creates a new chain submodule.
-func NewSyncerSubmodule(ctx context.Context, config syncerConfig, repo chainRepo, blockstore *BlockstoreSubmodule, network *NetworkSubmodule, discovery *DiscoverySubmodule, chn *ChainSubmodule, pvt *version.ProtocolVersionTable) (SyncerSubmodule, error) {
+func NewSyncerSubmodule(ctx context.Context, config syncerConfig, repo chainRepo, blockstore *BlockstoreSubmodule, network *NetworkSubmodule, discovery *DiscoverySubmodule, chn *ChainSubmodule) (SyncerSubmodule, error) {
 	// setup block validation
 	// TODO when #2961 is resolved do the needful here.
-	blkValid := consensus.NewDefaultBlockValidator(config.BlockTime(), config.Clock(), pvt)
+	blkValid := consensus.NewDefaultBlockValidator(config.BlockTime(), config.Clock())
```
acruikshank: I probably would have left the pvt in to aid future upgrades. Removing it also make sense, though, so this is fine.

Issue #3554, opened 2019-10-14
Initialize go-filecoin with an exported CAR file


frrist:
### Description
As a follow on to #3473 a filecoin node should be able to initialize from a chain CAR file produced by running `go-filecoin export chain.car`. This will require a modification to the chain export method such that it writes the genesis state to the CAR file being exported.

### Acceptance criteria
- go-filecoin can accept export chain CAR file as initialization arg, i.e. `./go-filecoin init --genesisfile=chain.car`

frrist:
closed via https://github.com/filecoin-project/go-filecoin/pull/3595

PR #3612, opened 2019-10-29
Frrist/export statetree

frrist:
### Motivation
A filecoin node should be initalizable from a car file export by the `go-filecoin chain export` command.

### Proposed changes
Add a method [`ChainStateTree`](https://github.com/filecoin-project/go-filecoin/compare/frrist/export-statetree?expand=1#diff-ab19a401e17ca256e756c4c7637bb5d2R228-R231) to the `ChainStateReader` that exports the chains state tree at a given CID. The method returns a slice of IPLD nodes representing the state tree. 

Closes issue #3554


internal/app/go-filecoin/internal/submodule/chain_submodule.go
```diff
@@ -64,7 +64,7 @@ func NewChainSubmodule(ctx context.Context, config chainConfig, repo chainRepo,
 	}
 	actorState := consensus.NewActorStateStore(chainStore, blockstore.CborStore, blockstore.Blockstore, processor)
 	messageStore := chain.NewMessageStore(blockstore.Blockstore)
-	chainState := cst.NewChainStateReadWriter(chainStore, messageStore, blockstore.CborStore, builtin.DefaultActors)
+	chainState := cst.NewChainStateReadWriter(chainStore, messageStore, blockstore.Blockstore, builtin.DefaultActors)
```
frrist: We accepts a blockstore here since it allows the state tree to be easily exported using a dag service, which requires a blockstore.

internal/app/go-filecoin/plumbing/cst/chain_state.go
```diff
@@ -211,10 +217,59 @@ func (chn *ChainStateReadWriter) ChainExport(ctx context.Context, head block.Tip
...
+	csc.addState(dagNd)
+	seen := cid.NewSet()
+	for _, l := range dagNd.Links() {
+		if err := dag.Walk(ctx, csc.getLinks, l.Cid, seen.Visit); err != nil {
+			return nil, err
+		}
+	}
+	return csc.state, nil
+
+}
```
frrist: recursively collect all StateTree links to a slice.

internal/pkg/chain/car.go
```diff
@@ -101,6 +106,19 @@ func Export(ctx context.Context, headTS block.TipSet, cr carChainReader, mr carM
...
+				logCar.Debugf("writigin state tree: %s", hdr.StateRoot)
+				stateRoots, err := sr.ChainStateTree(ctx, hdr.StateRoot)
+				if err != nil {
+					return err
+				}
+				for _, r := range stateRoots {
+					if err := carutil.LdWrite(out, r.Cid().Bytes(), r.RawData()); err != nil {
+						return err
+					}
+				}
```
frrist: Write the state tree to the car file.

internal/pkg/consensus/genesis.go
```diff
@@ -190,6 +190,24 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
 				return nil, err
 			}
 		}
+		if err := cst.Blocks.AddBlock(types.StorageMarketActorCodeObj); err != nil {
```
acruikshank: This is ok. The only thing I would recommend is moving this to a function pkg/actor/actor (where these values are called) and having both genesis.go and gengen call it.

I would like to be able to add a method on builtin.Actors that does this for all its actors, but there's no way to go from the code Cid to the object that created it.

internal/app/go-filecoin/plumbing/cst/chain_state.go
```diff
@@ -211,10 +217,59 @@ func (chn *ChainStateReadWriter) ChainExport(ctx context.Context, head block.Tip
...
+
+func (csc *chainStateCollector) collectState(ctx context.Context, stateRoot cid.Cid) ([]format.Node, error) {
+	dagNd, err := csc.dagserv.Get(ctx, stateRoot)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to load stateroot from dagservice %s", stateRoot)
+	}
+	csc.addState(dagNd)
+	seen := cid.NewSet()
+	for _, l := range dagNd.Links() {
+		if err := dag.Walk(ctx, csc.getLinks, l.Cid, seen.Visit); err != nil {
```
acruikshank: Does this have a side effect? It looks like the purpose is to validate. If that's the case, this should be `collectAndValidateState` or just `validateState`. If not, the side effect should be documented in a comment.

frrist: Its purpose is to recursively walk the state tree starting at `stateroot` and call `getLinks` for each IPLD node it visits, `getLinks` adds each node to the slice of IPLD nodes in the chainStateCollector called `state`. Visiting each node should not have any side effects, I will leave a comment at the top of this method stating its intent.

PR #3616, opened 2019-10-29
Stop fetching message receipts

acruikshank:
### Motivation

We want to move from blocks that contain the state root and receipt CID AFTER applying their messages, to storing the state root and receipts cid of the tipset it is built upon.

### Proposed changes

This PR transitions message receipts from a data structure transmitted with individual blocks to a collection only generated when a Tipset is processed and stored locally. This transition has a few consequences:

1. We need to save the receipt root as well as the state root after running a state transition on a TipSet. This happens in `syncer.syncOne`. In a future PR this will be used to populate and validate the parent receipt root in block headers.
2. We stop retrieving receipts in graphsync fetcher.
3. The waiter must now locate a receipt by first locating the message by Cid and then find the receipt that corresponds to the index of the message when run in the tipset. This necessitates a refactor in processor that extracts the message deduplication in TipSet processing so that it may be reused in waiter.

### DIFF
--- a/internal/app/go-filecoin/plumbing/msg/waiter.go
+++ b/internal/app/go-filecoin/plumbing/msg/waiter.go
@@ -133,7 +133,14 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
 				log.Errorf("Waiter.Wait: %s", e)
 				return nil, false, e
 			case block.TipSet:
-				return w.receiptForTipset(ctx, raw, msgCid)
+				msg, found, err := w.receiptForTipset(ctx, raw, msgCid)
+				if err != nil {
+					return nil, false, err
+				}
+				if found {
+					return msg, found, nil
+				}
+				// otherwise continue waiting
 			default:
 				return nil, false, fmt.Errorf("unexpected type in channel: %T", raw)
 			}
--- a/internal/app/go-filecoin/plumbing/msg/waiter.go
+++ b/internal/app/go-filecoin/plumbing/msg/waiter.go
@@ -102,7 +102,7 @@ func (w *Waiter) Wait(ctx context.Context, msgCid cid.Cid, cb func(*block.Block,
 // if now block with the given CID exists in the chain.
 func (w *Waiter) findMessage(ctx context.Context, ts block.TipSet, msgCid cid.Cid) (*ChainMessage, bool, error) {
 	var err error
-	for iterator := chain.IterAncestors(ctx, w.chainReader, ts); !iterator.Complete(); err = iterator.Next() {
+	for iterator := chain.IterAncestors(ctx, w.chainReader, ts); err == nil && !iterator.Complete(); err = iterator.Next() {
 		msg, found, err := w.receiptForTipset(ctx, iterator.Value(), msgCid)
 		if err != nil {
 			log.Errorf("Waiter.Wait: %s", err)
@@ -112,7 +112,7 @@ func (w *Waiter) findMessage(ctx context.Context, ts block.TipSet, msgCid cid.Ci
 			return msg, true, nil
 		}
 	}
-	return nil, false, nil
+	return nil, false, err
 }
 
 // waitForMessage looks for a message CID in a channel of tipsets and returns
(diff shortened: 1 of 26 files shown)


internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -149,149 +134,94 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
...
-		return nil, err
-	}
-	ancestors, err := chain.GetRecentAncestors(ctx, parentTs, w.chainReader, ancestorHeight)
-	if err != nil {
-		return nil, err
-	}
+		for i, msg := range blkMsgs {
+			if cids[i].Equals(msgCid) {
+				// wrapped cid might be different from given cid
+				targetCid, err := msg.Cid()
```
ZenGround0: potentially clearer to call this wrappedCid here.

ZenGround0: Probably a good idea to file an issue sorting the UX out for BLS messages.  Doing it this way requires that all users get the cid of an unsigned message as output from the CLI for BLS and a signed message as output for SECP so that they can correctly wait.

internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -149,149 +134,94 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
...
-				continue
+				return nil, err
 			}
-			duplicates[c] = struct{}{}
-			if c.Equals(msgCid) {
-				return msgCnt, nil
+
+			if msgCid.Equals(targetCid) {
+				if receiptIndex >= len(receipts) {
+					return nil, errors.Errorf("could not find message receipt at index %d", receiptIndex)
```
ZenGround0: I think this only gets hit if a programmer is violating system invariants and that this error should shout louder.  The only way this can happen is if there is an inconsistency in the stored receipts between initial dedupping and the waiter's deduping.

internal/pkg/chain/car_test.go
```diff
@@ -164,9 +164,8 @@ func TestChainImportExportMessages(t *testing.T) {
 		mm.NewSignedMessage(alice, 4),
 		mm.NewSignedMessage(alice, 5),
 	}
-	rcts := types.EmptyReceipts(5)
 	ts2 := cb.BuildOneOn(ts1, func(b *chain.BlockBuilder) {
-		b.AddMessages(msgs, []*types.UnsignedMessage{}, rcts)
```
ZenGround0: 🎉 This is going to be one of *those* commits

internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -149,149 +134,94 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
...
-						if c.Equals(msgCid) {
-							recpt, err := w.receiptFromTipSet(ctx, msgCid, raw)
-							if err != nil {
-								return nil, false, errors.Wrap(err, "error retrieving receipt from tipset")
-							}
-							return &ChainMessage{msg, blk, recpt}, true, nil
-						}
-					}
-				}
+				return w.receiptForTipset(ctx, raw, msgCid)
```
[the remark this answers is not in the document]
acruikshank: good catch!

internal/app/go-filecoin/plumbing/msg/waiter.go
acruikshank: This is true, unless there is some other sort of corruption. I've been violating these system invariants often enough while programming this that I still feel the check is worth it.

internal/pkg/types/message.go
```diff
@@ -211,34 +211,6 @@ func (mC MessageCollection) ToNode() ipld.Node {
 	return obj
 }
 
-// ReceiptCollection tracks a group of receipts and assigns it a cid.
```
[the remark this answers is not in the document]
acruikshank: Good call. SignedMessageCollection and UnsignedMessageCollection are now completely unused, so I removed them.

Issue #3312, opened 2019-08-28
Message Syntax/Semantics Validation in fetcher


hannahhoward:
### Description

This is a place holder issue for defining what we want to validate about message / message receipt syntax and semantics. As part of #3232 we're defining an interface for doing syntax validation. This issue is to discuss what should be validated about the syntax, and if a semantic validator should also be build

### Acceptance criteria

TBD -- this is for discussion

### Where to begin

This issue needs further definition before it's actionable so the first place is to define what should be validated

anorth:
This question should be answered in the spec before we can write much useful code. I've filed https://github.com/filecoin-project/specs/issues/464

anorth:
There is no longer any message semantic validation (until execution), but we still need syntax validation.

anorth:
Just a thought, maybe graphsync shouldn't do message sytax validation, b/c if it does then the messages won't be available to block syntax validation.

PR #3617, opened 2019-10-30
Change ApplyMessagesAndPayRewards to accept unsigned messages

anorth:
# Motivation
The processor doesn't use the signatures. Signatures in a message or block must be validated before passing to the processor, as part of block syntax validation (BLS message signatures can only be validated in aggregate, not individually) or before the message pool.

This removes signatures from the processor.

Note that blocks and messages fetched by graphsync aren't validated at all! #3312 (cc @ZenGround0)

### Proposed changes

Change DefaultProcessor methods to accept `UnsignedMessage`s.

### DIFF
--- a/internal/app/go-filecoin/plumbing/msg/waiter.go
+++ b/internal/app/go-filecoin/plumbing/msg/waiter.go
@@ -90,6 +90,9 @@ func (w *Waiter) Wait(ctx context.Context, msgCid cid.Cid, cb func(*block.Block,
 	}
 
 	chainMsg, found, err = w.waitForMessage(ctx, ch, msgCid)
+	if err != nil {
+		return err
+	}
 	if found {
 		return cb(chainMsg.Block, chainMsg.Message, chainMsg.Receipt)
 	}
@@ -148,8 +151,14 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
 	}
 }
 
-func (w *Waiter) receiptForTipset(ctx context.Context, ts block.TipSet, msgCid cid.Cid) (*ChainMessage, bool, error) {
-	tsMessages := make([][]*types.SignedMessage, ts.Len())
+func (w *Waiter) receiptForTipset(ctx context.Context, ts block.TipSet, targetMsg cid.Cid) (*ChainMessage, bool, error) {
+	// The targetMsg might be the CID of either a signed SECP message or an unsigned
+	// BLS message.
+	// This accumulates the CIDs of the messages as they appear on chain (signed or unsigned)
+	// but then unwraps them all, obtaining the CID of the unwrapped SECP message body if
+	// applicable. This unwrapped message CID is then used to find the target message in the
+	// unwrapped de-duplicated tipset messages, and thence the corresponding receipt by index.
+	tsMessages := make([][]*types.UnsignedMessage, ts.Len())
 	for i := 0; i < ts.Len(); i++ {
 		blk := ts.At(i)
 		secpMsgs, blsMsgs, err := w.messageProvider.LoadMessages(ctx, blk.Messages)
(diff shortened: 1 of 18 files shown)


internal/pkg/consensus/expected.go
```diff
@@ -324,21 +322,11 @@ func verifyBLSMessageAggregate(sig types.Signature, msgs []*types.UnsignedMessag
...
-	}
-	return messages
-}
-
-func wrapMessages(blsMessages []*types.UnsignedMessage) []*types.SignedMessage {
-	signed := []*types.SignedMessage{}
-	for _, msg := range blsMessages {
-		signed = append(signed, &types.SignedMessage{Message: *msg})
+// Unwraps nested slices of signed messages.
+func unwrap2(smsgs [][]*types.SignedMessage) [][]*types.UnsignedMessage {
```
acruikshank: Why the 2 in `unwrap2`?

internal/pkg/consensus/expected.go
```diff
@@ -289,7 +287,7 @@ func (c *Expected) runMessages(ctx context.Context, st state.Tree, vms vm.Storag
 	// NOTE: It is possible to optimize further by applying block validation
 	// in sorted order to reuse first block transitions as the starting state
 	// for the tipSetProcessor.
-	allMessages := combineMessages(blsMessages, secpMessages)
+	allMessages := append(blsMessages, secpMessages...)
```
acruikshank: This is wrong. I realized its wrong after I approved. It's wrong because it replicates incorrect behavior I introduced with BLS messages. ProcessTipSet expects a slice of slice of messages where each inner slice corresponds to a block. This is creating a slice of slices that is twice as long as the number of blocks. It's troubling that this is not being causing tests to fail.

I've fixed this in my pending PR, so it's ok to leave this broken here.

I need to examine to spec to determine whether tipset message processing should be ordered by block, type, gas price, or type, block, gas price. If the latter, we're going to need to pass in a different data structure that lets us tie a block miner to an arbitrary message.

internal/pkg/consensus/expected.go
anorth: For the nested slices (originally types.Unwrap was here too). Removed.

internal/pkg/consensus/expected.go
anorth: Thanks, I will leave this broken here.

It would seem worthwhile to be for us to create an explicit type representing the messages in a tipset, encapsulating the ordering and de-duping logic. As a bonus, consider arranging it so that it unavoidably checks all the message signature while extracting them from the message, thereby giving some type-safety to signature validation.

internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -148,55 +151,64 @@ func (w *Waiter) waitForMessage(ctx context.Context, ch <-chan interface{}, msgC
...
 		secpMsgs, blsMsgs, err := w.messageProvider.LoadMessages(ctx, blk.Messages)
 		if err != nil {
 			return nil, false, err
 		}
 
-		cids := make([]cid.Cid, len(blsMsgs)+len(secpMsgs))
-		blkMsgs := make([]*types.SignedMessage, len(blsMsgs)+len(secpMsgs))
+		originalCids := make([]cid.Cid, len(blsMsgs)+len(secpMsgs))
+		unwrappedMsgs := make([]*types.UnsignedMessage, len(blsMsgs)+len(secpMsgs))
+		wrappedMsgs := make([]*types.SignedMessage, len(blsMsgs)+len(secpMsgs))
```
anorth: FYI @acruikshank rebasing against your earlier change caused some gymnastics in the waiter, as de-duping now occurs over unsigned messages.

Changing the waiter callback result value from SignedMessage to UnsignedMessage could reduce the pain a bit, but is also slightly wrong.

We might need a union-style type that can represent either signed or unsigned, presenting a canonical CID in either case.

Issue #3525, opened 2019-10-01
Segfault in miner actor SubmitPoSt while generating block


coder-lb:
**Describe the bug**:

The filecoin daemon crashes because of nil pointer deference.

**Diagnostic information:**

Here is the log:

18:19:36.881  INFO     mining: [TIMER] DefaultWorker.Generate baseTipset: { bafy2bzacedupc6xucqlpqn7fcbudpajlzi5nzxsxwd5grenjwtknusggjddiw } - elapsed time: 2.764s block_generate.go:26
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x98559f]

goroutine 38955419 [running]:
github.com/filecoin-project/go-filecoin/types.(*BlockHeight).Add(0x0, 0xc030fa7528, 0x15)
        /go/src/github.com/filecoin-project/go-filecoin/types/block_height.go:108 +0x4f
github.com/filecoin-project/go-filecoin/actor/builtin/miner.(*Actor).SubmitPoSt.func1(0x1ae2be0, 0xc051e98300, 0x1638860, 0xc02433c480)
        /go/src/github.com/filecoin-project/go-filecoin/actor/builtin/miner/miner.go:874 +0x106
github.com/filecoin-project/go-filecoin/actor.WithState(0x1ae2be0, 0xc051e98300, 0x1638860, 0xc02433c480, 0xc04381c680, 0xc04381c6f8, 0xa, 0xc013f77b60, 0x1787501)
        /go/src/github.com/filecoin-project/go-filecoin/actor/storage.go:50 +0x87
github.com/filecoin-project/go-filecoin/actor/builtin/miner.(*Actor).SubmitPoSt(0xc0001ac778, 0x1ae2be0, 0xc051e98300, 0xc011745500, 0xc0, 0xc0, 0x0, 0x1ae4560, 0xc02cce8990, 0x1ae4560, ...)
        /go/src/github.com/filecoin-project/go-filecoin/actor/builtin/miner/miner.go:867 +0x2c1
reflect.Value.call(0xc024086900, 0xc030fa74c8, 0x13, 0x182c96d, 0x4, 0xc0008df080, 0x5, 0x8, 0x8, 0xc0008df080, ...)
        /usr/local/go/src/reflect/value.go:447 +0x461
[...]

PR #3618, opened 2019-10-30
Fix initialization of storage miner actor state

anorth:
### Motivation
Actor state must be initialized, especially things that are pointers. The `ProvingPeriodEnd` was uninitialized and caused a panic if a SubmitPos was received before CommitSector. 

There are other actors that need more attention too, but I'm deferring until we re-implement them with the new spec.

### Proposed changes
Initialize all the state in StorageMinerActor.

Closes issue #3525


anorth:
I was a bit sloppy not running all tests before pushing this. There were a bunch of places in actor and test code that checked for `nil`s explicitly that had to be updated. We need to make sure the new spec is explicit about this (but am guessing it won't have nil-able semantics).

PR #3619, opened 2019-10-31
refactor: change actor method dispatch to be num based 

icorderi:

### DIFF
--- a/internal/pkg/consensus/expected_test.go
+++ b/internal/pkg/consensus/expected_test.go
@@ -161,7 +161,7 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
 		require.NoError(t, err)
 
 		blsMessages := make([][]*types.UnsignedMessage, tipSet.Len())
-		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), "", []byte{})
+		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), types.InvalidMethodID, []byte{})
 		blsMessages[0] = append(blsMessages[0], msg)
 
 		_, _, err = exp.RunStateTransition(ctx, tipSet, blsMessages, emptyMessages, []block.TipSet{pTipSet}, uint64(blocks[0].ParentWeight), blocks[0].StateRoot)
@@ -191,7 +191,7 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
 		require.NoError(t, err)
 
 		secpMessages := make([][]*types.SignedMessage, tipSet.Len())
-		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), "", []byte{})
+		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), types.InvalidMethodID, []byte{})
 		smsg := &types.SignedMessage{
 			Message:   *msg,
 			Signature: []byte("not a signature"),
--- a/internal/pkg/consensus/testing.go
+++ b/internal/pkg/consensus/testing.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor/builtin"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor/builtin/miner"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor/builtin/storagemarket"
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
@@ -57,18 +59,20 @@ type FakePowerTableViewSnapshot struct {
 }
 
 // Query produces test logic in response to PowerTableView queries.
-func (tq *FakePowerTableViewSnapshot) Query(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
-	if method == "getTotalStorage" {
+func (tq *FakePowerTableViewSnapshot) Query(ctx context.Context, optFrom, to address.Address, method types.MethodID, params ...interface{}) ([][]byte, error) {
+	// Note: this currently happens to work as is, but it's wrong
+	// Note: a better mock is recommended to make sure the correct methods get dispatched
+	if method == storagemarket.GetTotalStorage {
 		if tq.TotalPower != nil {
 			return [][]byte{tq.TotalPower.Bytes()}, nil
 		}
 		return [][]byte{}, errors.New("something went wrong with the total power")
-	} else if method == "getPower" {
+	} else if method == miner.GetPower {
 		if tq.MinerPower != nil {
 			return [][]byte{tq.MinerPower.Bytes()}, nil
 		}
 		return [][]byte{}, errors.New("something went wrong with the miner power")
-	} else if method == "getWorker" {
+	} else if method == miner.GetWorker {
 		if tq.MinerToWorker != nil {
 			return [][]byte{tq.MinerToWorker[to].Bytes()}, nil
 		}
--- a/internal/pkg/types/message.go
+++ b/internal/pkg/types/message.go
@@ -10,15 +10,21 @@ import (
 	"github.com/filecoin-project/go-amt-ipld"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-ipfs-blockstore"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	ipld "github.com/ipfs/go-ipld-format"
 	errPkg "github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
-	"github.com/whyrusleeping/cbor-gen"
+	typegen "github.com/whyrusleeping/cbor-gen"
 )
+
+// MethodID is an identifier of a method (in an actor).
+type MethodID UVarint
+
+// InvalidMethodID is the value of an invalid method id.
+var InvalidMethodID = MethodID(99999999999)
 
 // GasUnits represents number of units of gas consumed
 type GasUnits = Uint64
@@ -59,8 +65,8 @@ type UnsignedMessage struct {
 
 	Value AttoFIL `json:"value"`
 
-	Method string `json:"method"`
-	Params []byte `json:"params"`
+	Method MethodID `json:"method"`
+	Params []byte   `json:"params"`
 
 	GasPrice AttoFIL  `json:"gasPrice"`
 	GasLimit GasUnits `json:"gasLimit"`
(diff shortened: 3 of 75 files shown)


internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -20,29 +22,41 @@ type State struct {
...
-}
-
-// NewActor returns a init actor.
-func NewActor() *actor.Actor {
-	return actor.NewActor(types.InitActorCodeCid, types.ZeroAttoFIL)
+// Method returns method definition for a given method id.
+func (a *Actor) Method(id types.MethodID) (exec.Method, *exec.FunctionSignature, bool) {
+	switch id {
+	case GetNetwork:
+		return reflect.ValueOf((*Impl)(a).GetNetwork), signatures[GetNetwork], true
```
anorth: Reflection is heavy overhead for this dispatch. It wasn't necessary in the existing code, and this is a good opportunity to fix it. It would be more direct (both understandable and in execution) to look up an array indexed by method id. In fact, something very similar to the existing Exports list (but including the method pointer) would probably work great.

If you're planning to replace this mechanism in a follow-up, 👌 .

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -192,6 +114,38 @@ type State struct {
...
+	GetSectorSize
+	GetAsks
+	GetAsk
+	GetLastUsedSectorID
+	GetProvingSetCommitments
+	IsBootstrapMiner
+	GetPoStState
+	GetProvingWindow
+	CalculateLateFee
+	GetActiveCollateral
```
anorth: When we come to align method ids with the spec, we'll probably want to specify the value of each of these constants explicitly, rather than rely on iota magic.

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -299,55 +237,201 @@ var minerExports = exec.Exports{
...
 //
 
+// Impl is the VM implementation of the actor.
+type Impl Actor
+
+var log = logging.Logger("actor.miner")
+
+// XXX: import cycle, need cleanup
+var XXX_storagemarket_UpdateStorage = types.MethodID(1)
+var XXX_storagemarket_GetProofsMode = types.MethodID(3)
```
anorth: Please fix anything that can be immediately done so and remove XXX comments before merging. 
- If there's work to be deferred, please write a `TODO` and reference a GitHub issue with the details (for easy grepping).
- If it's just ugly but no immediate action, just leave a comment explaining the situation but without the XXX.

This one is probably worth a discussion offline to choose a path forward.

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -21,82 +22,10 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/exec"
 )
 
-var log = logging.Logger("miner-actor")
-
```
anorth: The re-arrangement makes it difficult for reviewers to understand what has and hasn't changed. Please resist the urge to make cosmetic changes when performing an extensive functional change. Land the re-arrangement in a "prefactor" PR, or just leave it for another day.

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -299,55 +237,201 @@ var minerExports = exec.Exports{
...
+
+const (
+	PoStStateNoStorage = iota
+	PoStStateWithinProvingPeriod
+	PoStStateAfterProvingPeriod
+	PoStStateUnrecoverable
+)
+
 // AddAsk adds an ask to this miners ask list
-func (ma *Actor) AddAsk(ctx exec.VMContext, price types.AttoFIL, expiry *big.Int) (*big.Int, uint8,
```
anorth: Similarly, I don't object to this rename on its own, but it would be far better to have happened in a different PR. This also means that the functional change can't be reverted independently to these renames etc etc.

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -1170,16 +1254,18 @@ func getPoStChallengeSeed(ctx exec.VMContext, state State, sampleAt *types.Block
 	return seed, nil
 }
 
+// XXX: consider moving this out of this file
```
anorth: Don't commit XXX or "consider" comments, they just add noise.

internal/pkg/consensus/expected_test.go
```diff
@@ -161,7 +161,7 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
 		require.NoError(t, err)
 
 		blsMessages := make([][]*types.UnsignedMessage, tipSet.Len())
-		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), "", []byte{})
+		msg := types.NewUnsignedMessage(blsAddr, address.TestAddress2, 0, types.NewAttoFILFromFIL(0), types.InvalidMethodID, []byte{})
```
anorth: Blocking: The spec specifies the method id `0` for value transfers ("send"). `InvalidMethodID` is good for distinguishing an invalid id from one with meaning, but using it for sending is wrong. 

It looks like most uses of InvalidMethodID here are intended as sends, so renaming it is probably the easiest path.

- https://filecoin-project.github.io/specs/#systems__filecoin_vm__interpreter
- https://filecoin-project.github.io/specs/#systems__filecoin_vm__actor

internal/pkg/vm/actor/builtin/initactor/init.go
icorderi: Yes, I want all of this to go away during the `VMContext to spec` and `method execution up to spec` tasks. Out-of-scope to cleanup on this PR.

internal/pkg/vm/actor/builtin/miner/miner.go
icorderi: When they get specified, agreed. 
https://github.com/filecoin-project/specs/issues/616

internal/pkg/types/message.go
```diff
@@ -10,16 +10,22 @@ import (
...
 	errPkg "github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
-	"github.com/whyrusleeping/cbor-gen"
+	typegen "github.com/whyrusleeping/cbor-gen"
 )
 
+// MethodID is an identifier of a method (in an actor).
+type MethodID UVarint
```
anorth: I like that we have an explicit type name for method ids, distinguishing them from other integers. Help me understand the encoding. A MethodID in the message is supposed to be encoded as a variable-length integer. This new `UVarint` is represented as a Go `uint` and is not encoded with variable-length encoding as far as I can see.

If you want to defer the varint encoding for follow-up, that's ok. Just use 'uint' here and leave a comment referencing a GitHub issue for fixing the encoding. Indirecting through this new `UVarint` isn't adding anything, and just obscures the fact that it's not in fact varint encoded. Since this is the only reference to `UVarint` outside its definition, inserting this indirection again later will be trivial (and, I would suggest, be named after its runtime representation if we chose a fixed-size).

If you want to use the correct encoding now, use the existing Uint64 type which implements that. Yes, 64-bits is excessive for runtime representation, so we could also file an issue to replace this alias with a Uint32 or even Uint16 later, purely as an optimization.

internal/pkg/vm/actor/export.go
```diff
@@ -20,20 +20,13 @@ import (
 // TODO: the work of creating the wrapper should be ideally done at compile time, otherwise at least only once + cached
 // TODO: find a better name, naming is hard..
 // TODO: Ensure the method is not empty. We need to be paranoid we're not calling methods on transfer messages.
-func MakeTypedExport(actor exec.ExecutableActor, method string) exec.ExportedFunc {
-	f, ok := reflect.TypeOf(actor).MethodByName(strings.Title(method))
+func MakeTypedExport(actor exec.ExecutableActor, method types.MethodID) (exec.ExportedFunc, bool) {
+	fn, signature, ok := actor.Method(method)
```
acruikshank: BLOCKING: There's a problem retrieving the executable function as a value rather than as a method.

The miner actor has a `Bootstrap` variable set in the executable actor's state. This permits two different types of actors to share most of the same code. Since the function is now a free function, that state does not get passed into the miner and no miner will ever act like a bootstrap miner.

This behavior appears to be under-tested. It will only appear when a bootstrap miner is asked to Seal or Post. 

There are two quick fixes:

1. Have actor.Method return a method instead of a function and pass `actor` in as the first parameter.
2. Create a new `bootstrapMiner` struct that duplicates all the `miner` code not excluded by the bootstrap flag and set it under the BootstrapMiner code cid in `actor.Builtin`.

There are other solutions including coming up with a better story around bootstrapping in general, but the suggestions above are probably the best way to move past this PR.

internal/pkg/vm/vm.go
```diff
@@ -36,7 +36,7 @@ func send(ctx context.Context, deps sendDeps, vmCtx *Context) ([][]byte, uint8,
 		}
 	}
 
-	if vmCtx.message.Method == "" {
+	if vmCtx.message.Method == types.InvalidMethodID {
```
acruikshank: No method is a valid method ID. This should be something like `types.NoMethodID`.

internal/pkg/consensus/testing.go
```diff
@@ -57,18 +59,20 @@ type FakePowerTableViewSnapshot struct {
 }
 
 // Query produces test logic in response to PowerTableView queries.
-func (tq *FakePowerTableViewSnapshot) Query(ctx context.Context, optFrom, to address.Address, method string, params ...interface{}) ([][]byte, error) {
-	if method == "getTotalStorage" {
+func (tq *FakePowerTableViewSnapshot) Query(ctx context.Context, optFrom, to address.Address, method types.MethodID, params ...interface{}) ([][]byte, error) {
+	// Note: this currently happens to work as is, but it's wrong
```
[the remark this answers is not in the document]
icorderi: The code was working before because the method names happen to not be repeated between actors, but it was not correct nor is it correct. Name collision chances were a bit lower than ID collisions which all start at the same number between actors.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -61,14 +75,21 @@ func (ia *Actor) InitializeState(storage exec.Storage, networkInterface interfac
...
+// vm methods for actor
+//
+
+// Impl is the VM implementation of the actor.
+type Impl Actor
+
 // GetNetwork returns the network name for this network
-func (sma *Actor) GetNetwork(vmctx exec.VMContext) (string, uint8, error) {
-	if err := vmctx.Charge(actor.DefaultGasCost); err != nil {
+func (*Impl) GetNetwork(ctx exec.VMContext) (string, uint8, error) {
```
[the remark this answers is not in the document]
icorderi: keeping it, it will be moved to `vm.Context` in the near future

internal/pkg/types/message.go
```diff
@@ -170,3 +176,7 @@ type TxMeta struct {
 func (m TxMeta) String() string {
 	return fmt.Sprintf("secp: %s, bls: %s", m.SecpRoot.String(), m.BLSRoot.String())
 }
+
+func (id MethodID) String() string {
```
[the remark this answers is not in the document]
icorderi: it didnt, might be `String` is well understood/built-in ? or the linter took the day off.

internal/pkg/types/message.go
icorderi: I disagree on principle, we can discuss it offline. 
I removed it so we can get this merged and move on.

PR #3621, opened 2019-10-31
Election works off of a lookback hardcoded to 5

ZenGround0:
### Motivation
We need our consensus election to look back into chain history so that our elections can prevent grinding per the spec.

### Proposed changes
Add lookback parameter to consensus election to get us up to spec.  Hardcoding to 5.  This PR involves 0 cleanup and sticks to old ugly patterns.  On the plus side it is very simple and doesn't make anything worse.

Closes issue #2223

### DIFF
--- a/internal/pkg/consensus/expected_test.go
+++ b/internal/pkg/consensus/expected_test.go
@@ -31,56 +31,9 @@ func TestNewExpected(t *testing.T) {
 	t.Run("a new Expected can be created", func(t *testing.T) {
 		cst, bstore := setupCborBlockstore()
 		as := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(5), make(map[address.Address]address.Address))
-		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), th.NewFakeBlockValidator(), as, types.CidFromString(t, "somecid"), th.BlockTimeTest, &consensus.FakeElectionMachine{}, &consensus.FakeTicketMachine{})
+		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), as, th.BlockTimeTest, &consensus.FakeElectionMachine{}, &consensus.FakeTicketMachine{})
 		assert.NotNil(t, exp)
 	})
-}
-
-func requireNewValidTestBlock(t *testing.T, baseTipSet block.TipSet, stateRootCid cid.Cid, height uint64, minerAddr address.Address, minerWorker address.Address, signer types.Signer) *block.Block {
-	b, err := th.NewValidTestBlockFromTipSet(baseTipSet, stateRootCid, height, minerAddr, minerWorker, signer)
-	require.NoError(t, err)
-	return b
-}
-
-// requireMakeBlocks sets up 3 blocks with 3 owner actors and 3 miner actors and puts them in the state tree.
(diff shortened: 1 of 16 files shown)


ZenGround0:
It is very important to read commit by commit as I ended up taking a tangential but sorely needed detour to clean up the expected tests before I could write new validation tests.

internal/pkg/consensus/expected_test.go
```diff
@@ -31,58 +31,11 @@ func TestNewExpected(t *testing.T) {
 	t.Run("a new Expected can be created", func(t *testing.T) {
 		cst, bstore := setupCborBlockstore()
 		as := consensus.NewFakeActorStateStore(types.NewBytesAmount(1), types.NewBytesAmount(5), make(map[address.Address]address.Address))
-		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), th.NewFakeBlockValidator(), as, types.CidFromString(t, "somecid"), th.BlockTimeTest, &consensus.FakeElectionMachine{}, &consensus.FakeTicketMachine{})
+		exp := consensus.NewExpected(cst, bstore, consensus.NewDefaultProcessor(), as, th.BlockTimeTest, &consensus.FakeElectionMachine{}, &consensus.FakeTicketMachine{})
 		assert.NotNil(t, exp)
 	})
 }
 
-func requireNewValidTestBlock(t *testing.T, baseTipSet block.TipSet, stateRootCid cid.Cid, height uint64, minerAddr address.Address, minerWorker address.Address, signer types.Signer) *block.Block {
```
acruikshank: 🎉

internal/pkg/mining/worker_test.go
```diff
@@ -31,14 +31,130 @@ import (
...
+	return &mockTicketGen{fn: f}
 }
 
 func (mtg *mockTicketGen) NextTicket(ticket block.Ticket, genAddr address.Address, signer types.Signer) (block.Ticket, error) {
-	mtg.ticketGen = true
+	mtg.fn(ticket)
 	return consensus.MakeFakeTicketForTest(), nil
 }
 
+type mockElectionMachine struct {
```
acruikshank: This test infrastructure should be moved below the tests themselves.

internal/pkg/mining/worker_test.go
```diff
@@ -31,14 +31,130 @@ import (
...
+	})
+
+	t.Run("Ticket gensees ticket 1 ancestor back", func(t *testing.T) {
+		genTicket, err := ancestors[0].MinTicket()
+		require.NoError(t, err)
+		mtm := newMockTicketGen(func(ticket block.Ticket) {
+			assert.Equal(t, genTicket, ticket)
+		})
+
+		ctx, cancel := context.WithCancel(context.Background())
```
acruikshank: This context (and the one in the previous test) should also have a timeout to keep this test from hanging.

anorth: Only if it's very long, else the test will flake on heavily-loaded systems. Test code can't make good judgements about the performance of the system it's running on. A timeout can also be provided from the outside.

ZenGround0: I opted not to add the timeout for @anorth's reason above.  We can revisit in an issue if you'd like to file one @acruikshank. We should probably clean up the entire channel-based interface and return structure of the worker in one go and this could be a subset of that technical debt work.

PR #3622, opened 2019-10-31
Block state root and receipt cid now reference result of parent tip set message processing

acruikshank:
### Motivation

The spec dictates (or will dictate) that state root and receipt root in the block header reflect the result of messages processed in the parent tipset, not messages processed in the currrent  block as currently implemented.

### Proposed changes

1. Set state root and receipt root to the values saved with the parent tipset when generating a block. The block generator must still run a state transition in order to determine which messages are valid to include.
2. When validating a block, only validate that the block miner has included the correct state root and message receipts for the parent tipset. Here we stop running state transitions on individual blocks.
3. Modify the message waiter to pull receipts for an entire tipset. This involves some slightly involved logic around determining the index of a receipt given multiple message sets associated with its blocks.
4. Since receipts are now associated with a tipset rather than a block, we remove them from our block analysis UI, including `show block`.

Closes issue #

### DIFF
--- a/internal/pkg/consensus/expected.go
+++ b/internal/pkg/consensus/expected.go
@@ -42,8 +42,12 @@ func init() {
 var (
 	// ErrStateRootMismatch is returned when the computed state root doesn't match the expected result.
 	ErrStateRootMismatch = errors.New("blocks state root does not match computed result")
+	// ErrInvalidBase is returned when the chain doesn't connect back to a known good block.
+	ErrInvalidBase = errors.New("block does not connect to a known good chain")
 	// ErrUnorderedTipSets is returned when weight and minticket are the same between two tipsets.
 	ErrUnorderedTipSets = errors.New("trying to order two identical tipsets")
+	// ErrReceiptRootMismatch is returned when the block's receipt root doesn't match the receipt root computed for the parent tipset.
+	ErrReceiptRootMismatch = errors.New("blocks receipt root does not match parent tip set")
 )
 
 // DefaultBlockTime is the estimated proving period time.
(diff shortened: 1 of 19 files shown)


ZenGround0:
Talking with @anorth this shouldn't merge without a conversation on how to handle errors from `ProcessTipSet` it looks like this function drops errors that `ProcessBlock` used to propagate and we can get into trouble if we don't add in that propagation.

internal/pkg/consensus/processor_test.go
```diff
@@ -272,9 +217,9 @@ func TestProcessBlockReward(t *testing.T) {
 		Height:    20,
 		StateRoot: stCid,
 	}
-	ret, err := NewDefaultProcessor().ProcessBlock(ctx, st, vms, blk, []*types.SignedMessage{}, nil)
+	ret, err := NewDefaultProcessor().ProcessTipSet(ctx, st, vms, RequireNewTipSet(require.New(t), blk), [][]*types.SignedMessage{{}}, nil)
```
anorth: Rename the `TestProcessBlockxxx` tests.

internal/pkg/consensus/expected.go
```diff
@@ -248,60 +250,39 @@ func (c *Expected) validateMining(ctx context.Context, st state.Tree, ts block.T
 // An error is returned if individual blocks contain messages that do not
 // lead to successful state transitions.  An error is also returned if the node
```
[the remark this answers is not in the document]
acruikshank: good catch

internal/pkg/consensus/processor_test.go
acruikshank: thanks.

anorth:
Discussion in https://github.com/filecoin-project/specs/issues/619 suggests we should just accept failing messages in a block, which is also a simple policy.

internal/pkg/consensus/expected.go
```diff
@@ -241,67 +242,43 @@ func (c *Expected) validateMining(ctx context.Context, st state.Tree, ts block.T
...
-func (c *Expected) runMessages(ctx context.Context, st state.Tree, vms vm.StorageMap, ts block.TipSet, blsMessages [][]*types.UnsignedMessage, secpMessages [][]*types.UnsignedMessage, ancestors []block.TipSet) (state.Tree, []*types.MessageReceipt, error) {
-	var cpySt state.Tree
-	var results []*ApplicationResult
-
-	// TODO: don't process messages twice
+// An error is returned if the state root or receipt root of any block
+// does not match that of the parent tipset.  An error is also returned if
+// the node faults while running aggregate state computation.
+func (c *Expected) runMessages(ctx context.Context, st state.Tree, vms vm.StorageMap, ts block.TipSet, blsMessages [][]*types.UnsignedMessage, secpMessages [][]*types.UnsignedMessage, parentReceipts cid.Cid, ancestors []block.TipSet) (state.Tree, []*types.MessageReceipt, error) {
+	parentStateRoot, err := st.Flush(ctx)
```
anorth: This flush just to get hold of the parentStateRoot is a bit icky. The caller has this parentStateRoot - that's now the tree was loaded in the first place. The name `runMessages` is also a bit misleading if it also does this other validation.

I suggest pulling the state and receipt root match checks up to `RunStateTransition`.

internal/pkg/consensus/processor.go
```diff
@@ -173,6 +116,10 @@ func (p *DefaultProcessor) ProcessTipSet(ctx context.Context, st state.Tree, vms
 	}
 	bh := types.NewBlockHeight(h)
 
+	var res ProcessTipSetResponse
+	res.Failures = make(map[cid.Cid]struct{})
+	res.Successes = make(map[cid.Cid]struct{})
```
anorth: This method has a named return value `tsResult` - use that one.

acruikshank: good catch.

internal/pkg/consensus/expected.go
acruikshank: Agreed, but it really makes more sense to put it in `validateMining` since that function is already iterating through blocks and checking for mistakes. I'm now passing the state root and receipt cids into that function so we don't need to flush the state tree.

PR #3623, opened 2019-11-01
refactor: new vm runtime and context

icorderi:


internal/pkg/vm2/vminternal/errors/vminternal.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// This contents can be slowly placed on the vm2 internal.
+package vminternal
+
+import (
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/errors"
+)
+
+// TODO fritz require actors to define their exit codes and associate
+// an error string with them.
```
anorth: We can delete this comment.

PR #3625, opened 2019-11-01
Remove misleading ProcessTipsetResponse which implied that errors are not ignored.

anorth:
### Motivation
The `ProcessTipsetResponse` contained fields which indicated which messages failed, but these were ignored by callers.

### Proposed changes
Removed the custom response type in favour of returning a slice of the message results (pass thru from ApplyMessagesEtc) and document the dropping of errors.

I suggest this land before #3622, and that that PR must now handle these errors somehow.


anorth:
This appears to have broken all the daemon tests that use messagewait. It might take me a while to figure out why - please be on the lookout for dumdums.

anorth:
I've rebased against #3622 and it's much simpler now.

PR #3626, opened 2019-11-01
refactor: move context to new vm folder 

icorderi:
This PR finishes the cleanup of vm (that was using the vm2 pkg).

They key changes are in the following files: 
- some changes around `vmcontext.go`, methods now take an interface in instead of the concrete type. 
- some movements around `state/tree.go`, mostly visibility and file breakdown. 
- moved `export.go` from actor to vmcontext (the MakeExport function will likely be removed in a future PR in favor of something that gives us direct dispatch instead of all that reflection).
- vm2 no longer exists since I finished moving everything back to vm2 then renamed it back to vm.


internal/pkg/vm/vm_test.go
```diff
@@ -1,140 +0,0 @@
...
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/errors"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/state"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm2/vminternal/dispatch"
-
-	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
-	"github.com/stretchr/testify/assert"
-)
-
-func TestTransfer(t *testing.T) {
```
acruikshank: BLOCKING: It appears that `TestSendErrorHandling` made it `vmcontext_test`, but this test seems to have been lost. Please add it back or justify its exclusion.

icorderi: test is there, just different file. 
Moved it to vmcontext tests.

internal/pkg/vm/state/testing.go
```diff
@@ -124,10 +115,19 @@ func (m *MockStateTree) GetActorCode(c cid.Cid, protocol uint64) (dispatch.Execu
...
+	tree := NewTree(cst)
 	strAddr, err := address.NewActorAddress([]byte(s))
 	require.NoError(t, err)
 	err = tree.SetActor(context.Background(), strAddr, &actor.Actor{})
 	require.NoError(t, err)
 	return tree
 }
+
+// MustGetActor gets the actor or panics if it can't.
+func MustGetActor(st Tree, a address.Address) *actor.Actor {
```
acruikshank: If `MustGetActor` needs to move, `MustSetActor` should too.

internal/pkg/vm/state/tree.go
```diff
@@ -41,45 +23,34 @@ type Tree interface {
...
-		return nil, errors.Wrapf(err, "failed to load node for %s", c)
-	}
-	stateTree := newEmptyStateTree(store.(*hamt.CborIpldStore))
-	stateTree.root = root
-
-	return stateTree, nil
-}
+const (
+	// TreeBitWidth is the bit width of the HAMT used to store a state tree
+	TreeBitWidth = 5
```
acruikshank: This constant should be moved back to the top of the file. This is inconsistent with how constants are defined throughout the codebase. 

Most of the code reordering in this file is unnecessary. There is no code style for how elements get ordered in this project. If we decide to implement a style for order of elements, we should make it explicit and implement and fix the codebase in a single PR. Piecemeal changes like just increase entropy and destroy git history.

internal/pkg/vm/state/tree.go
```diff
@@ -41,45 +23,34 @@ type Tree interface {
...
+)
 
-// NewEmptyStateTree instantiates a new state tree with no data in it.
-func NewEmptyStateTree(store *hamt.CborIpldStore) Tree {
+// NewTree instantiates a new state tree.
+func NewTree(store *hamt.CborIpldStore) Tree {
```
acruikshank: nice.

internal/pkg/vm/vm.go
```diff
@@ -1,88 +1,60 @@
...
-	fromActor.Balance = fromActor.Balance.Sub(value)
-	toActor.Balance = toActor.Balance.Add(value)
-
-	return nil
+	return vmcontext.Transfer(fromActor, toActor, value)
 }
+
+// FunctionSignature describes the signature of a single function.
+//
+// Dragons: this signature should not be neede on the outside
```
acruikshank: I'm not sure I agree with this dragon. It seems reasonable that actors inside the VM should be able to communicate their method signatures to the outside. That will requires something like this type.

Please elaborate or remove.

internal/pkg/vm/internal/storagemap/storagemap.go
```diff
@@ -16,69 +23,70 @@ import (
...
 // persisted.
-func (s *storageMap) NewStorage(addr address.Address, actor *actor.Actor) Storage {
-	storage, ok := s.storageMap[addr]
+func (s *storageMap) NewStorage(addr address.Address, actor *actor.Actor) runtime.Storage {
+	st, ok := s.storageMap[addr]
 	if ok {
 		// Return a hybrid storage with the pre-existing chunks, but the given instance of the actor.
 		// This ensures changes made to the actor appear in the state tree cache.
-		storage = Storage{
+		st = storage{
```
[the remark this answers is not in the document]
icorderi: leaving it `st`. `s` is already taken, `as = storage{}` doesn't read well.

internal/pkg/vm/state/testing.go
icorderi: got scrambled on an intermediate refactor, MustGetActor and MustSetActor are both here.

internal/pkg/vm/vm.go
icorderi: The only things needing to know the signature are whomever is doing encoding the input and decoding the output. This does not necessarily need to exist in data, it could just be code.
The VM runtime doesn't necessarily have to support reflection for someone to be able to call methods on actors and interpret the results correctly. 

We currently send `[]interface{}` down and the `export.go` file is turning it into ABI (this occurs inside the VM). 
The return says `[]byte` which is very unclear on what bytes. We are currently returning ABI bytes which is wrong, there is an issue opened on that already.

PR #3630, opened 2019-11-05
Feat/pubsub topics

### LINKED ISSUE #3561: Gossipsub -- make the node wait for the mesh before publishing anything
### Description
Because of the way gossipsub works there is a long delay between node startup and a node's ability to publish over pubsub channels (see [this issue for more detail](https://github.com/libp2p/go-libp2p-pubsub/issues/211)).

Using the default timing parameters with gossipsub exacerbated an existing race in the majority of our integration tests and created an unworkable amount of flakiness.  To get around this in the short term we modified the heartbeat interval to be short enough to roughly force mesh construction before publish.

This is an imperfect solution that makes races manageable but does not eliminate them.  We could remove them once and for all by blocking on node init until after receiving an event from a gossipsub eventbus.  We haven't done this yet because it requires a change to the gossipsub implementation / api.

### Acceptance criteria
No more pubsub related races by consuming a gossipsub event bus and waiting for a "ready to publish" event.

### Risks + pitfalls
This work is mostly outside of our control and belongs to libp2p.

### Where to begin
See if we can get this scheduled in gossipsub roadmap.  After that most of our work is done.

ZenGround0:
### Motivation
We want to move to gossipsub and keep race conditions out of tests.  This required waiting for recent upgrades to libp2p.  This PR takes the first step of 1) updating libp2p 2) refactoring our internal wrappers to make use of the new topic abstraction

### Proposed changes
- update pubsub dep to 0.2.1
- go-filecoin uses topic abstraction.  syncer and messaging submodules own block and message topics.  internal go-filecoin wrapper changed to a single `Topic` type.  Construction modified.  Publish and subscribe calls modified to not take in topic string.

Helps close #3561


internal/app/go-filecoin/internal/submodule/messaging_submodule.go
```diff
@@ -38,14 +39,22 @@ func NewMessagingSubmodule(ctx context.Context, config messagingConfig, repo mes
 	msgPool := message.NewPool(repo.Config().Mpool, consensus.NewIngestionValidator(chain.State, repo.Config().Mpool))
 	inbox := message.NewInbox(msgPool, message.InboxMaxAgeTipsets, chain.ChainReader, chain.MessageStore)
 
+	// setup messaging topic
+	topic, err := network.Fsub.Join(net.MessageTopic(network.NetworkName))
+	if err != nil {
+		return MessagingSubmodule{}, err
+	}
```
anorth: I'm a bit confused about the difference between this `Join` and `Topic.Subscribe()` calls below. I've been code diving to try to reconcile this with the code I see in libp2p (on master) but haven't 100% figured it out.

Subscribing (having any network side-effect) at module construction is undesirable. It looks like in libp2p, this `Join()` is a no-op (but also with no return value?). If that's an assumption we're making in order to not have side-effects here, we should document as such. But with GossipSub, `Join` does do a bunch, so we're not getting any abstraction here.

If constructing a Topic object involves interacting with libp2p in a way that has side effects, I think we're better representing a topic as a string until we actually want to subscribe to it.

ZenGround0: I want to move away from the deprecated endpoints so I'm keeping Topics.  However your point on side effects during construction is well taken.  I'll construct both the topic and subscription during start.

PR #3632, opened 2019-11-06
first integration with chain-validation library

frrist:
### Motivation
Integrate go-filecoin with the first version of the chain-validation library.

### Proposed changes
- Implement the interface required by the chain validation library.
- Add calls to the chain-validation library value transfer tests.

### Status
Ready for review

### DIFF
--- a/internal/pkg/validation/gfc_test.go
+++ b/internal/pkg/validation/gfc_test.go
@@ -4,9 +4,12 @@ import (
 	"testing"
 
 	"github.com/filecoin-project/chain-validation/pkg/suites"
+	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
 )
 
 func TestValueTransfer(t *testing.T) {
+	tf.UnitTest(t)
+
 	factory := NewFactories()
 	suites.AccountValueTransferSuccess(t, factory, 0)
 	suites.AccountValueTransferZeroFunds(t, factory, 0)
--- a/internal/pkg/validation/gfc_test.go
+++ b/internal/pkg/validation/gfc_test.go
@@ -1,0 +1,19 @@
+package validation
+
+import (
+	"testing"
+
+	"github.com/filecoin-project/chain-validation/pkg/suites"
+)
+
+func TestValueTransfer(t *testing.T) {
+	factory := NewFactories()
+	suites.AccountValueTransferSuccess(t, factory, 0)
+	suites.AccountValueTransferZeroFunds(t, factory, 0)
+	suites.AccountValueTransferOverBalanceNonZero(t, factory, 0)
+	suites.AccountValueTransferOverBalanceZero(t, factory, 0)
+	suites.AccountValueTransferToSelf(t, factory, 0)
+	suites.AccountValueTransferFromKnownToUnknownAccount(t, factory, 0)
+	suites.AccountValueTransferFromUnknownToKnownAccount(t, factory, 0)
+	suites.AccountValueTransferFromUnknownToUnknownAccount(t, factory, 0)
+}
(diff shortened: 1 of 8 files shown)


internal/pkg/validation/message.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, errors.Wrapf(err, "failed to encode abi param values: %v", params)
+	}
+	if int(method) >= len(methods) {
+		return nil, errors.Errorf("No method name for method %v", method)
+	}
+	methodName := methods[method]
+	msg := types.NewMeteredMessage(fromDec, toDec, nonce, valueDec, methodName, paramsDec, types.NewAttoFIL(gasPrice), types.NewGasUnits(uint64(gasUnit)))
+
+	return msg, nil
+	//return types.NewSignedMessage(*msg, mf.signer)
```
anorth: Delete

internal/pkg/validation/message.go
```diff
@@ -0,0 +1,60 @@
...
+
+// FromSingletonAddress returns the address for a Singleton Actor.
+func (mf *MessageFactory) FromSingletonAddress(addr state.SingletonActorID) state.Address {
+	return fromSingletonAddress(addr)
+}
+
+// Maps method enumeration values to method names.
+// This will change to a mapping to method ids when method dispatch is updated to use integers.
+var methods = []types.MethodID{
+	chain.NoMethod: 0,
```
anorth: Go-filecoin now has method IDs. Please fill in at least a subset to demonstrate the pattern.

internal/pkg/validation/state.go
```diff
@@ -0,0 +1,295 @@
...
+	bs := blockstore.NewBlockstore(datastore.NewMapDatastore())
+	cst := hamt.CSTFromBstore(bs)
+	treeImpl := state.NewTree(cst)
+	storageImpl := vm.NewStorageMap(bs)
+	return &StateWrapper{treeImpl, storageImpl, newKeyStore()}
+}
+
+// Cid returns the cid of the state wrappers current state
+func (s *StateWrapper) Cid() cid.Cid {
+	panic("implement me")
```
anorth: Blocking: I think we should replace the state.Tree with a pattern similar to that used by the Lotus adapter, storing the state root CID and hamt.CborIpldStore in the state wrapper, and then loading the state on demand when used.

internal/pkg/validation/state.go
```diff
@@ -0,0 +1,295 @@
...
+		if err != nil {
+			return nil, nil, err
+		}
+		if err := s.Tree.SetActor(ctx, fcAddr, intAct); err != nil {
+			return nil, nil, err
+		}
+		_, err = s.Tree.Flush(ctx)
+		if err != nil {
+			return nil, nil, err
+		}
```
anorth: Factor out the repeated lines.

internal/pkg/validation/gfc_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+	factory := NewFactories()
+	suites.AccountValueTransferSuccess(t, factory, 0)
+	suites.AccountValueTransferZeroFunds(t, factory, 0)
+	suites.AccountValueTransferOverBalanceNonZero(t, factory, 0)
+	suites.AccountValueTransferOverBalanceZero(t, factory, 0)
+	suites.AccountValueTransferToSelf(t, factory, 0)
+	suites.AccountValueTransferFromKnownToUnknownAccount(t, factory, 0)
+	suites.AccountValueTransferFromUnknownToKnownAccount(t, factory, 0)
+	suites.AccountValueTransferFromUnknownToUnknownAccount(t, factory, 0)
```
anorth: This pattern is very fragile, and will break or miss new tests every time the suites change.
Can we instead export a list of test funcs from the validation package, and then iterate that here?

What's the `0`? Please put it in a named constant.

frrist:
A lot has changed in the validation library wrt serialization and types. Details can be seen in https://github.com/filecoin-project/chain-validation/tree/v0.0.1. I am going to close this out and revisit later with a more concrete approach.

PR #3636, opened 2019-11-06
[WIP] review: `Runtime` diffs with spec

icorderi:
>  **NOTICE:** this will **NOT** be merged, this is a conversation placeholder will we narrow down the contracts.  

Please take a look at the `Review:` comments, there are few questions, proposed changes, some things we could ignore from the spec as well as some things that no longer show up on the spec.

I don't believe the actors should have access to **more** than what is specified in the spec, so some discussion on the things that disappeared from from `Runtime` is welcome. 

Feel free to ignore the newtypes around the file for now, they are just so signatures read as close to the spec as possible.

### Motivation

Discuss the differences with the spec and define the new Runtime/VMContext types. 

### Proposed changes

Check commends in the code.

### DIFF
--- a/internal/pkg/vm/internal/runtime/runtime.go
+++ b/internal/pkg/vm/internal/runtime/runtime.go
@@ -8,27 +8,158 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
 
+
+
+// Runtime2 is the VM.
+type Runtime2 struct {
+	CurerntEpoch() Epoch
+	// Review: why is this taking an offset? verify size of offset
+	Randomness(epoch Epoch, offset uint64) Randomness
+	// Review: the spec still has a TODO on this
+	// Review: not sure why the original has a messagereceipt instead of an InvocOutput
+	Send(input InvocInput) InvocOutput
+	// Review: how does this work? What qualifies as an "error"?
+	// Review: how is this any different than the caller just ignoring a returned error?
+	// SendAllowingErrors(input msg.InvocInput) msg.MessageReceipt
+
+	// Review: we might want a global Storage() Storage here
+}
+
+// InvocationContext is passed to the actors on each method call.
+type InvocationContext interface {
+	// Review: grouped the VM level things out of the invocation itself
+	Runtime() Runtime2
+	Caller() address.Address
+	// Review: (is in spec, dont think we need it) idem to Caller().Equal(addr) :/
+	// ValidateCallerIs(caller addr.Address)
+	// Review: not sure what all this pattern matching adds up to, the "singleton" pattern does an equality check
+	// ValidateCallerMatches(CallerPattern)
+	// Review: the spec handles state and state changes a bit differently
+	// Review: our Storage is a bit of a mix of AcquireState and the Ipld{Get-Put}
+	// AcquireState() ActorStateHandle
+
+	// Review: The spec suggest handling return construction by the runtime, interesting but not strictly necessary
+	// SuccessReturn()     msg.InvocOutput
+	// ValueReturn(Bytes)  msg.InvocOutput
+	// ErrorReturn(exitCode exitcode.ExitCode) msg.InvocOutput
+
+	// Review: how does this work? is the spec expecting us to Panic? or just terminate the message execution?
+	// Review: this should be `Abort(string) -> msg.InvocOutput` if we follow the pattern
+	// Abort(string)
+
+	// Review: this is a bit trickier than abort in Go. This is a conditional early exit on the function which we cannot do.
+	// Review: `if <cond> { return AssertFailure("blah") }` is a bit closer to go if we are to try to follow this pattern.
+	// Assert(bool)
+
+	// Review: we didnt have this before, why does the actor need to know the value supplied?
+	ValueSupplied() types.AttoFIL
+
+	// Review: if this is just for the `init` actor then it should not be here.
+	// Review: the init actor is a special system actor, its ok for it to have a different interface with a bit more power.
+    // CreateActor(
+    //     cid                actor.StateCID
+    //     a                  addr.Address
+    //     constructorParams  actor.MethodParams
+    // )
+
+	// Review: we provide this through the Storage()
+    // IpldGet(c ipld.CID) union {Bytes, error}
+	// IpldPut(x ipld.Object) ipld.CID
+	// Review: is the expectation fo storage to be global or actor specific? this has security implications down the line
+	Storage() Storage
+}
+
+type innitInvocationContext interface {
+	InvocationContext
+	// Review: the spec has no error on this? what if the addr is already taken?
+	CreateActor(cid cid.Cid, addr address.Address, constructorParams MethodParams)
+}
+
+// Storage2 defines the storage module exposed to actors.
+// Review: check the old `Storage` for some more comments on why this looks like it does
+type Storage2 interface {
+	Put(interface{}) (cid.Cid, error)
+	Get(cid.Cid, interface{}) error
+}
+
+// Epoch is the current chain epoch.
+// Dragons: this is just to match names on the VM, Height got renamed to Epoch on the spec
+// TODO: rename types.BlockHeight to block.Epoch (issue: ???)
+type Epoch types.BlockHeight
+
+// Randomness is a string of random bytes
+type Randomness []byte
+
+// InvocInput are the params to invoke a method in an Actor.
+// Review: why is this a struct instead of arguments?
+type InvocInput struct {
+    To      address.Address
+    Method  types.MethodID
+    Params  MethodParams
+    Value   types.AttoFIL
+}
+
+// ExitCode is the exit code of a method executing inside the VM.
+// Review: do we have reserved codes?
+// TODO: this has type UVarint in the spec
+type ExitCode types.Int64
+
+// InvocOutput is waht os returned by an actor executing a method.
+type InvocOutput struct {
+	ExitCode     ExitCode
+    ReturnValue  []byte
+}
+
+// MethodParam is the parameter to an Actor method already encoded for execution in the VM.
+type MethodParam []byte
+// MethodParams is a list of `MethodParam` to be passed into an Actor method.
+type MethodParams []MethodParam
+
 // Runtime defines the ABI interface exposed to actors.
 type Runtime interface {
+	// Review: this is not in the spec anymore
 	Message() *types.UnsignedMessage
+
+	// Review: the spec splits this up and brings the get/put to be first class
 	Storage() Storage
+
+	// Review: changed from arguments to a struct
 	Send(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error)
+
+	// Review: this is not in the spec anymore
 	AddressForNewActor() (address.Address, error)
+
+ 	// Review: renamed
 	BlockHeight() *types.BlockHeight
+
+	// Review: this is not in the spec anymore
 	MyBalance() types.AttoFIL
+
+	// Review: this changed a bit, not sure why it cant be determined off the result of `Caller()`
 	IsFromAccountActor() bool
+
+	// Review: this is not in the spec anymore
 	Charge(cost types.GasUnits) error
+
+	// Review: this got renamed and gained an offset arg
 	SampleChainRandomness(sampleHeight *types.BlockHeight) ([]byte, error)
 
+	// unchanged
 	CreateNewActor(addr address.Address, code cid.Cid, initalizationParams interface{}) error
 
+	// Review: this is not in the spec anymore
 	Verifier() verification.Verifier
 }
 
 // Storage defines the storage module exposed to actors.
 type Storage interface {
-	Put(interface{}) (cid.Cid, error)
-	Get(cid.Cid) ([]byte, error)
-	Commit(cid.Cid, cid.Cid) error
 	Head() cid.Cid
+	Put(interface{}) (cid.Cid, error)
+	// Dragons: this interface is wrong, the caller does not know how the object got serialized (Put/1 takes an interface{} not bytes)
+	// TODO: change to `Get(cid, interface{}) error`
+	Get(cid.Cid) ([]byte, error)
+	// Review: why is this needed on the actor API? Actor commit is implicit upon a succesfull termination.
+	// Review: an explicit Commit() on actor code leads to bugs
+	// Review: what situation requires state to change, succesfully execute the method, but not to be commited?
+	Commit(cid.Cid, cid.Cid) error
 }
--- a/internal/pkg/vm/internal/runtime/runtime.go
+++ b/internal/pkg/vm/internal/runtime/runtime.go
@@ -11,6 +11,7 @@ import (
 
 
 // Runtime2 is the VM.
+// Review: the object that implements this outlives actor calls.
 type Runtime2 struct {
 	CurerntEpoch() Epoch
 	// Review: why is this taking an offset? verify size of offset
@@ -26,6 +27,7 @@ type Runtime2 struct {
 }
 
 // InvocationContext is passed to the actors on each method call.
+// Review: this type + Runtime2 is mostly what the spec defines as Runtime.
 type InvocationContext interface {
 	// Review: grouped the VM level things out of the invocation itself
 	Runtime() Runtime2
@@ -69,6 +71,7 @@ type InvocationContext interface {
 	Storage() Storage
 }
 
+// Review: more powerfull contract for the initactor.
 type innitInvocationContext interface {
 	InvocationContext
 	// Review: the spec has no error on this? what if the addr is already taken?
@@ -88,6 +91,7 @@ type Storage2 interface {
 type Epoch types.BlockHeight
 
 // Randomness is a string of random bytes
+// Review: placeholder for matching newtypes with the spec.
 type Randomness []byte
 
 // InvocInput are the params to invoke a method in an Actor.
@@ -105,6 +109,7 @@ type InvocInput struct {
 type ExitCode types.Int64
 
 // InvocOutput is waht os returned by an actor executing a method.
+// Review: multiple returns vs a return struct. There are some minor implications with swapping vm implementations written in another language.
 type InvocOutput struct {
 	ExitCode     ExitCode
     ReturnValue  []byte
--- a/internal/pkg/vm/internal/runtime/runtime.go
+++ b/internal/pkg/vm/internal/runtime/runtime.go
@@ -8,74 +8,85 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
 
-
-
 // Runtime2 is the VM.
-// Review: the object that implements this outlives actor calls.
 type Runtime2 struct {
 	CurerntEpoch() Epoch
-	// Review: should we turn the offset into a RandomnessEvent?
+	// Review: for @go-filecoin, should we turn the offset into a RandomnessEvent?
 	Randomness(epoch Epoch, offset uint64) Randomness
-	// Review: not sure why the original has a messagereceipt instead of an InvocOutput
-	Send(input InvocInput) InvocOutput
-	// Review: how does this work? What qualifies as an "error"?
-	// Review: how is this any different than the caller just ignoring a returned error?
-	// SendAllowingErrors(input msg.InvocInput) msg.MessageReceipt
-
-	// Review: we might want a global Storage() Storage here
+	Send(input InvocInput) result.InvocOutput
+	// SharedStorage is an IPLD store shared by all actors.
+	// Review: for @go-filecoin, we can skip this until we have actor code that actually needs this
+	// Review: for @spec, Add expiration (Epoch's) on PUT? (the question is mostly for global, but the local store might want to advise on this too).
+	// Review: for @spec, how will gas accounting work on global? PUT fee, GET fee, OWNER Fee?
+	SharedStorage() IpldStorage2
 }
 
-// InvocationContext is passed to the actors on each method call.
-// Review: this type + Runtime2 is mostly what the spec defines as Runtime.
+// DispatchContext is the context the actor dispatcher receives.
+// Review: for @go-filecoin, do we want to have the `ctx.ValidateCaller(blahblah)` at the top of each method, or hidden away on the dispatch table?
+type DispatchContext interface {
+	// Review: for @go-filecoin, simplified the API to be just one method.
+	ValidateCaller(CallerPattern)
+	// InvocationContext is the context to be passed down the actor method.
+	InvocationContext() InvocationContext
+}
+
+// InvocationContext is passed to the actors on each method call..
 type InvocationContext interface {
-	// Review: grouped the VM level things out of the invocation itself
+	// Review: for @go-filecoin, composition or inheritance?
 	Runtime() Runtime2
 	Caller() address.Address
-	// Review: (is in spec, dont think we need it) idem to Caller().Equal(addr) :/
-	// ValidateCallerIs(caller addr.Address)
-	// Review: not sure what all this pattern matching adds up to, the "singleton" pattern does an equality check
-	// ValidateCallerMatches(CallerPattern)
-	// Review: the spec handles state and state changes a bit differently
-	// Review: our Storage is a bit of a mix of AcquireState and the Ipld{Get-Put}
-	// AcquireState() ActorStateHandle
+	StateHandler() ActorStateHandler
 
-	// Review: The spec suggest handling return construction by the runtime, interesting but not strictly necessary
-	// SuccessReturn()     msg.InvocOutput
-	// ValueReturn(Bytes)  msg.InvocOutput
-	// ErrorReturn(exitCode exitcode.ExitCode) msg.InvocOutput
-
-	// Review: how does this work? is the spec expecting us to Panic? or just terminate the message execution?
-	// Review: this should be `Abort(string) -> msg.InvocOutput` if we follow the pattern
-	// Abort(string)
-
-	// Review: this is a bit trickier than abort in Go. This is a conditional early exit on the function which we cannot do.
-	// Review: `if <cond> { return AssertFailure("blah") }` is a bit closer to go if we are to try to follow this pattern.
-	// Assert(bool)
-
-	// Review: we didnt have this before, why does the actor need to know the value supplied?
-	ValueSupplied() types.AttoFIL
-
-	// Review: is the expectation fo storage to be global or actor specific? this has security implications down the line
-	Storage() Storage
-
-	// Dragons: this was not in the spce, but it was likely a bug
-	// Note: @jzimmerman I think that's an error in the spec -- balance should be accessible via the runtime.
+	// Review: for @go-filecoin+@spec, renamed `ValueSupplied` to `ValueReceived`. The "supplied" is not clear by whom.
+	ValueReceived() types.AttoFIL
 	Balance() types.AttoFIL
+
+	LocalStorage() IpldStorage2
 }
 
 // initInvocationContext is a more powerfull context for the "init" actor.
 type initInvocationContext interface {
 	InvocationContext
-	// Review: added an error that is not yet on the spec but should be
+	// Review: for @spec, added an error that is not yet on the spec but should be
 	// Note: @jzimmerman: "Agreed, this should support an error return."
 	CreateActor(cid cid.Cid, addr address.Address, constructorParams MethodParams) error
 }
 
-// Storage2 defines the storage module exposed to actors.
-// Review: simpler API (check type Storage for comments on the things that got removed)
-type Storage2 interface {
+// minerInvocationContext has some special sauce for the miner.
+// Review: for @go-filecoin, do we want custom interfaces for the miners instead of grouping a bunch fo methods on the InvocationContext?
+type minerInvocationContext interface {
+	Verifier() verification.Verifier
+}
+
+// ActorStateHandler handles the actor state, allowing actors to lock on the state.
+type ActorStateHandler interface {
+	// Take returns the `Cid` for the state and locks the state.
+	//
+	// Any futre calls to `Take` by other instances will block until `Release` or `UpdateRelease` is called.
+	// Review: for @spec, the spec impl of `Take` is not doing any blocking, just panic'ing on the second call on the SAME object instance (this objects are not kept around, so it is always a new instance..).
+	Take() cid.Cid
+	// UpdateRelease updates the actor state and releases the lock on it.
+	//
+	// No future calls to `Take` are allowed on this object.
+	// Review: for @spec, this method can currently be called without having called `Take` first.
+	UpdateRelease(newStateCID ActorSubstateCID)
+	// Release asserts that the state has not changed and releases the lock on it.
+	//
+	// No future calls to `Take` are allowed on this object.
+	// Review: for @spec, this method can currently be called without having called `Take` first.
+	// Review: for @spec, why is this method required?
+	Release(checkStateCID ActorSubstateCID)
+}
+
+// IpldStorage2 defines the storage module exposed to actors.
+type IpldStorage2 interface {
 	Put(interface{}) (cid.Cid, error)
 	Get(cid.Cid, interface{}) error
+}
+
+// CallerPattern checks if the caller matches the pattern.
+type CallerPattern interface {
+	IsMatch() bool
 }
 
 // Epoch is the current chain epoch.
@@ -88,8 +99,6 @@ type Epoch types.BlockHeight
 type Randomness []byte
 
 // InvocInput are the params to invoke a method in an Actor.
-// Review: unless anyone in @go-filecoin has a strong preference for arguments, we are chaging to the struct the spec suggests.
-// Note: @jzimmerman mentioned we might want to reffer to the inputs by Cid in the future
 type InvocInput struct {
     To      address.Address
     Method  types.MethodID
@@ -97,63 +106,33 @@ type InvocInput struct {
     Value   types.AttoFIL
 }
 
-// ExitCode is the exit code of a method executing inside the VM.
-type ExitCode types.Int64
-
-// InvocOutput is waht os returned by an actor executing a method.
-// Review: multiple returns vs a return struct. There are some minor implications with swapping vm implementations written in another language.
-type InvocOutput struct {
-	ExitCode     ExitCode
-    ReturnValue  []byte
-}
-
 // MethodParam is the parameter to an Actor method already encoded for execution in the VM.
 type MethodParam []byte
 // MethodParams is a list of `MethodParam` to be passed into an Actor method.
 type MethodParams []MethodParam
+
+// Abort will stop the VM execution and return an abort error to the caller.
+// Review: @go-filecoin, do we want to handle this with panics?
+func Abort(msg string) {
+	panic("byte me")
+}
+
+// Assert will abort if the condition is `False` and return an abort error to the caller.
+func Assert(cond bool) {
+	if !cond {
+		Abort("assertion failure in actor code.")
+	}
+}
+
+//
+// Dragons: this will get deleted, still here while we finish some discussions on the PR.
+//
 
 // Runtime defines the ABI interface exposed to actors.
 type Runtime interface {
 	// Review: this is not in the spec ~anymore~ yet..
 	// Note: @jzimmerman: still a WIP, things like the miner actor might need it
 	Message() *types.UnsignedMessage
-
-	// The spec splits this up and brings the get/put to be first class,
-	// Review: go-filecoin should keep it as a composition
-	Storage() Storage
-
-	// Changed from arguments to a struct
-	// Note: we will likely go with the struct, check comments on `type InvocInput`
-	Send(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error)
-
-	// This is not in the spec anymore
-	// Note: @jzimmerman: this functionality is delicate enough security-wise (especially with respect to different addresses on different forks) that I think it makes sense not to expose it directly.
-	AddressForNewActor() (address.Address, error)
-
- 	// Review: renamed and has a new type
-	BlockHeight() *types.BlockHeight
-
-	// This is not in the spec anymore, but it should be
-	// Note: @jzimmerman I think that's an error in the spec -- balance should be accessible via the runtime.
-	MyBalance() types.AttoFIL
-
-	// got replaceed by ValidateCallerIs and ValidateCallerMAtches
-	IsFromAccountActor() bool
-
-	// This is not in the spec anymore.
-	// Note: @jzimmerman: "Gas accounting is still WIP"
-	// TODO: come back to this when Gas accounting lands on spec
-	Charge(cost types.GasUnits) error
-
-	// Note: this got renamed and gained an offset arg
-	SampleChainRandomness(sampleHeight *types.BlockHeight) ([]byte, error)
-
-	// unchanged
-	CreateNewActor(addr address.Address, code cid.Cid, initalizationParams interface{}) error
-
-	// Review: this is not in the spec anymore
-	// Note: the spec seems to do a direct call, that has some implications (see github comments).
-	Verifier() verification.Verifier
 }
 
 // Storage defines the storage module exposed to actors.
--- a/internal/pkg/vm/internal/runtime/runtime.go
+++ b/internal/pkg/vm/internal/runtime/runtime.go
@@ -14,9 +14,8 @@ import (
 // Review: the object that implements this outlives actor calls.
 type Runtime2 struct {
 	CurerntEpoch() Epoch
-	// Review: why is this taking an offset? verify size of offset
+	// Review: should we turn the offset into a RandomnessEvent?
 	Randomness(epoch Epoch, offset uint64) Randomness
-	// Review: the spec still has a TODO on this
 	// Review: not sure why the original has a messagereceipt instead of an InvocOutput
 	Send(input InvocInput) InvocOutput
 	// Review: how does this work? What qualifies as an "error"?
@@ -56,26 +55,20 @@ type InvocationContext interface {
 	// Review: we didnt have this before, why does the actor need to know the value supplied?
 	ValueSupplied() types.AttoFIL
 
-	// Review: if this is just for the `init` actor then it should not be here.
-	// Review: the init actor is a special system actor, its ok for it to have a different interface with a bit more power.
-    // CreateActor(
-    //     cid                actor.StateCID
-    //     a                  addr.Address
-    //     constructorParams  actor.MethodParams
-    // )
-
-	// Review: we provide this through the Storage()
-    // IpldGet(c ipld.CID) union {Bytes, error}
-	// IpldPut(x ipld.Object) ipld.CID
 	// Review: is the expectation fo storage to be global or actor specific? this has security implications down the line
 	Storage() Storage
+
+	// Dragons: this was not in the spce, but it was likely a bug
+	// Note: @jzimmerman I think that's an error in the spec -- balance should be accessible via the runtime.
+	Balance() types.AttoFIL
 }
 
-// Review: more powerfull contract for the initactor.
-type innitInvocationContext interface {
+// initInvocationContext is a more powerfull context for the "init" actor.
+type initInvocationContext interface {
 	InvocationContext
-	// Review: the spec has no error on this? what if the addr is already taken?
-	CreateActor(cid cid.Cid, addr address.Address, constructorParams MethodParams)
+	// Review: added an error that is not yet on the spec but should be
+	// Note: @jzimmerman: "Agreed, this should support an error return."
+	CreateActor(cid cid.Cid, addr address.Address, constructorParams MethodParams) error
 }
 
 // Storage2 defines the storage module exposed to actors.
@@ -95,7 +88,8 @@ type Epoch types.BlockHeight
 type Randomness []byte
 
 // InvocInput are the params to invoke a method in an Actor.
-// Review: why is this a struct instead of arguments?
+// Review: unless anyone in @go-filecoin has a strong preference for arguments, we are chaging to the struct the spec suggests.
+// Note: @jzimmerman mentioned we might want to reffer to the inputs by Cid in the future
 type InvocInput struct {
     To      address.Address
     Method  types.MethodID
@@ -104,8 +98,6 @@ type InvocInput struct {
 }
 
 // ExitCode is the exit code of a method executing inside the VM.
-// Review: do we have reserved codes?
-// TODO: this has type UVarint in the spec
 type ExitCode types.Int64
 
 // InvocOutput is waht os returned by an actor executing a method.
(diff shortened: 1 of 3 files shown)


internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 )
 
+
+
+// Runtime2 is the VM.
+// Review: the object that implements this outlives actor calls.
+type Runtime2 struct {
+	CurerntEpoch() Epoch
+	// Review: why is this taking an offset? verify size of offset
+	Randomness(epoch Epoch, offset uint64) Randomness
```
ZenGround0: I think the offset will be different for different operations.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+    // CreateActor(
+    //     cid                actor.StateCID
+    //     a                  addr.Address
+    //     constructorParams  actor.MethodParams
+    // )
+
+	// Review: we provide this through the Storage()
+    // IpldGet(c ipld.CID) union {Bytes, error}
+	// IpldPut(x ipld.Object) ipld.CID
+	// Review: is the expectation fo storage to be global or actor specific? this has security implications down the line
```
ZenGround0: Could you elaborate on the security implications?  At first glance I think it should not matter with hash links everywhere.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	Storage() Storage
+}
+
+// Review: more powerfull contract for the initactor.
+type innitInvocationContext interface {
```
ZenGround0: +1

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: this is not in the spec anymore
 	Charge(cost types.GasUnits) error
+
+	// Review: this got renamed and gained an offset arg
 	SampleChainRandomness(sampleHeight *types.BlockHeight) ([]byte, error)
 
+	// unchanged
 	CreateNewActor(addr address.Address, code cid.Cid, initalizationParams interface{}) error
 
+	// Review: this is not in the spec anymore
```
ZenGround0: The spec's direct import in the miner code (`new(proving.StorageProvingSubsystem_I).VerifySeal(&sector.SealVerifyInfo_I{`) seems worse for testing and like something we can do our own way.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 // Storage defines the storage module exposed to actors.
 type Storage interface {
+	Head() cid.Cid
 	Put(interface{}) (cid.Cid, error)
+	// Dragons: this interface is wrong, the caller does not know how the object got serialized (Put/1 takes an interface{} not bytes)
+	// TODO: change to `Get(cid, interface{}) error`
 	Get(cid.Cid) ([]byte, error)
+	// Review: why is this needed on the actor API? Actor commit is implicit upon a succesfull termination.
+	// Review: an explicit Commit() on actor code leads to bugs
+	// Review: what situation requires state to change, succesfully execute the method, but not to be commited?
```
ZenGround0: These are good points / questions.  cc @acruikshank

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Splitting global/actor-local storage means that the hashspace the actor has access to through the actor storage is restricted to objects he put in there in the past. Other actors cannot access those objects even if they have the CID (unless the CID is also present in their storage). 

It also allows us in the future to establish object provenance and storage consumption per actor.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: Yes, that's the intent -- we don't want accidental collisions when different computations (or different phases during a single computation) need to draw randomness from the same epoch. Ideally, the offsets would be aggregated in a central place to make it easier to enforce uniqueness, similar to `MethodNum`.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 
+
+
+// Runtime2 is the VM.
+// Review: the object that implements this outlives actor calls.
+type Runtime2 struct {
+	CurerntEpoch() Epoch
+	// Review: why is this taking an offset? verify size of offset
+	Randomness(epoch Epoch, offset uint64) Randomness
+	// Review: the spec still has a TODO on this
```
jzimmerman: I believe that TODO is now outdated; `InvocOutput` and `MessageReceipt` are fairly stable at this point.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: not sure why the original has a messagereceipt instead of an InvocOutput
```
jzimmerman: I don't think it makes much of a difference, but `MessageReceipt` allows actor code to check the gas used by its subcalls, which I suppose could be a useful capability in the future. (On the other hand, we should also analyze the security implications.) @zixuanzh @jbenet Thoughts on this?

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	Send(input InvocInput) InvocOutput
+	// Review: how does this work? What qualifies as an "error"?
+	// Review: how is this any different than the caller just ignoring a returned error?
```
jzimmerman: Any error encountered by the subcall (either `rt.Fatal`, `panic`, or an ordinary error return code) cascades and immediately causes the caller to also fail with `MethodSubcallError`, unless the caller explicitly uses `SendAllowingErrors`. This is primarily to prioritize "fail safe" behavior -- i.e., if actor A calls actor B and doesn't realize the subcall might fail, we would rather the entire transaction fail to apply, than for actor A to continue erroneously and complete the transaction.

(Related question, though -- if naming the error-cascading version `Send` is too error-prone, we could rename it to `SendCascadingErrors` or something similar, and not have an unspecified general `Send` method.)

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// SendAllowingErrors(input msg.InvocInput) msg.MessageReceipt
+
+	// Review: we might want a global Storage() Storage here
```
jzimmerman: What would be the use case -- i.e. what data would be stored here (outside of the actor's own state in the state tree)?

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+}
+
+// InvocationContext is passed to the actors on each method call.
+// Review: this type + Runtime2 is mostly what the spec defines as Runtime.
+type InvocationContext interface {
+	// Review: grouped the VM level things out of the invocation itself
+	Runtime() Runtime2
+	Caller() address.Address
+	// Review: (is in spec, dont think we need it) idem to Caller().Equal(addr) :/
+	// ValidateCallerIs(caller addr.Address)
```
jzimmerman: The idea here is that the runtime can _enforce_ that the actor code calls `ValidateCallerIs` (or `ValidateCallerMatches`), and automatically raise an error if it's never called. (The rationale being that by default, methods are callable by any party introducing a message to the chain, and there is a large class of security-critical bugs that could arise if actors, particularly system actors, don't check the identity of the caller.)

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: not sure what all this pattern matching adds up to, the "singleton" pattern does an equality check
+	// ValidateCallerMatches(CallerPattern)
```
jzimmerman: `ValidateCallerMatches` is intended for methods that can be invoked more than one way, e.g., by some non-singleton actor A and also by Cron. The runtime will check that there's exactly one call to `ValidateCaller{Is|Matches}`, so we can't do this the natural way by just checking `Caller().Equals(A) || Caller().Equals(Cron)`.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: the spec handles state and state changes a bit differently
```
jzimmerman: Yes -- we introduced this change as what we thought would be the simplest way to mitigate reentrancy attacks. The basic invariants are:
- `AcquireState` and `Release|UpdateRelease` calls must be balanced and only one state lock may be held at a time
- `Send*` calls cannot be made while the state lock is held
The runtime should enforce these properties dynamically (see, e.g., `_checkStateLock`). We would like to enforce as much as possible statically, but couldn't see a good way to do this without introducing a lot of complexity at the API level and making the actor code highly non-Go-idiomatic.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: our Storage is a bit of a mix of AcquireState and the Ipld{Get-Put}
+	// AcquireState() ActorStateHandle
+
+	// Review: The spec suggest handling return construction by the runtime, interesting but not strictly necessary
+	// SuccessReturn()     msg.InvocOutput
+	// ValueReturn(Bytes)  msg.InvocOutput
+	// ErrorReturn(exitCode exitcode.ExitCode) msg.InvocOutput
+
+	// Review: how does this work? is the spec expecting us to Panic? or just terminate the message execution?
```
jzimmerman: The runtime will panic, causing the actor code to immediately abort (and thus the return value isn't needed), but the runtime will then catch the error and convert it into an ordinary failure exit code. (See `Abort`, `_throwErrorFull`, and `_catchRuntimeErrors` in `systems/filecoin_vm/runtime/runtime.go`.)

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	// Review: this should be `Abort(string) -> msg.InvocOutput` if we follow the pattern
+	// Abort(string)
+
+	// Review: this is a bit trickier than abort in Go. This is a conditional early exit on the function which we cannot do.
+	// Review: `if <cond> { return AssertFailure("blah") }` is a bit closer to go if we are to try to follow this pattern.
+	// Assert(bool)
```
jzimmerman: As currently specified, `Assert` just calls `rt.Abort` if the condition is false (`runtime.go:149`).

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+
+	// Review: we didnt have this before, why does the actor need to know the value supplied?
```
jzimmerman: This may be necessary for actors that require some funds as payment in order to perform an operation. Otherwise, the receiving actor will have no way of confirming during the method execution that the funds were actually transferred.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+	ValueSupplied() types.AttoFIL
+
+	// Review: if this is just for the `init` actor then it should not be here.
+	// Review: the init actor is a special system actor, its ok for it to have a different interface with a bit more power.
```
jzimmerman: Yes, the intent is that `CreateActor` is only callable by the init actor (enforced by the runtime). I believe it simplifies the implementation if we can treat the init actor like any other actor in this regard, but if it turns out to be easier to make it a special case, I don't see any particular problems this would cause (@zixuanzh @jbenet anything I've missed here?)

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: This is a tough question -- actor-specific storage is desirable for the reasons you mention, but on the other hand, we would like it to be the case that actor A can send a large object to actor B just by CID reference. If this also requires the runtime to inspect the contents of the message and copy all constituent CIDs to actor B's address space, that could get pretty messy (not to mention causing a lot of duplication for common data objects). @zixuanzh @jbenet Thoughts on this question?

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+    // IpldGet(c ipld.CID) union {Bytes, error}
+	// IpldPut(x ipld.Object) ipld.CID
+	// Review: is the expectation fo storage to be global or actor specific? this has security implications down the line
+	Storage() Storage
+}
+
+// Review: more powerfull contract for the initactor.
+type innitInvocationContext interface {
+	InvocationContext
+	// Review: the spec has no error on this? what if the addr is already taken?
```
jzimmerman: Agreed, this should support an error return.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+// Dragons: this is just to match names on the VM, Height got renamed to Epoch on the spec
+// TODO: rename types.BlockHeight to block.Epoch (issue: ???)
+type Epoch types.BlockHeight
+
+// Randomness is a string of random bytes
+// Review: placeholder for matching newtypes with the spec.
+type Randomness []byte
+
+// InvocInput are the params to invoke a method in an Actor.
+// Review: why is this a struct instead of arguments?
```
jzimmerman: It's used in several places (and conceivably in the future we might want to refer to it by CID).

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+type InvocInput struct {
+    To      address.Address
+    Method  types.MethodID
+    Params  MethodParams
+    Value   types.AttoFIL
+}
+
+// ExitCode is the exit code of a method executing inside the VM.
+// Review: do we have reserved codes?
```
jzimmerman: See `systems/filecoin_vm/runtime/exitcode/vm_exitcodes.go`

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+// TODO: this has type UVarint in the spec
```
jzimmerman: Re. integer types, see https://github.com/filecoin-project/specs/issues/615

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+}
+
+// MethodParam is the parameter to an Actor method already encoded for execution in the VM.
+type MethodParam []byte
+// MethodParams is a list of `MethodParam` to be passed into an Actor method.
+type MethodParams []MethodParam
+
 // Runtime defines the ABI interface exposed to actors.
 type Runtime interface {
+	// Review: this is not in the spec anymore
```
jzimmerman: We need some way to clearly distinguish between information from the original on-chain `Message`, vs. the invocation input "message" to this particular method. Currently the runtime only provides the latter information, but in some cases it's likely we will need the former (e.g., the identity of the miner who won the block in which this transaction appears). This is still WIP at the spec level, so if there are particular additional pieces of information that are needed by actors, I would feel free to add them and make a note to discuss later (we do want to limit these to what's strictly necessary, though).

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 	Message() *types.UnsignedMessage
+
+	// Review: the spec splits this up and brings the get/put to be first class
```
jzimmerman: I don't think this is fundamental, if it simplifies things to have a proxy object that should be fine too

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 	Storage() Storage
+
+	// Review: changed from arguments to a struct
 	Send(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error)
+
+	// Review: this is not in the spec anymore
```
jzimmerman: This got moved to `_generateActorAddress` in `runtime.go`, but I think the public handle to this API is still pending. This functionality is delicate enough security-wise (especially with respect to different addresses on different forks) that I think it makes sense not to expose it directly, but rather have it be an operation executed via the init actor (who can then derive a fresh address via looking up the address of its caller). @zixuanzh @jbenet may also want to weigh in here

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 	AddressForNewActor() (address.Address, error)
+
+ 	// Review: renamed
 	BlockHeight() *types.BlockHeight
+
+	// Review: this is not in the spec anymore
```
jzimmerman: I think that's an error in the spec -- balance should be accessible via the runtime.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
 	MyBalance() types.AttoFIL
+
+	// Review: this changed a bit, not sure why it cant be determined off the result of `Caller()`
 	IsFromAccountActor() bool
+
+	// Review: this is not in the spec anymore
```
jzimmerman: Gas accounting is still WIP in the spec, so I would hold off on changing this for now

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: I don't think this is fundamental; `verifySeal` is intended to be a pure function, but beyond that, it shouldn't matter how the source modules are split up

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Getting both the isolation and the cheap "send" (with just the CID) is totally doable, the question is whether or not the spec really wants/needs either/both of those characteristics to achieve its goals.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: I think the importance of that is that if it's not a runtime hook, it then is expected to be either: VM code, or the VM needs an ABI for calling external functions defined. 

Right now "VM code" happens to be happy Go, and calling some other happy random Go code seems ok, and we are not thinking on when we are crossing VM boundaries that much. But it has consequences when swapping VM implementations (specially ones written in another language) or when getting an actual VM defined with its own language.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Got it.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Ok, that is interesting. So the actor might want to do something about the gas used by the functions it called, yet it has not control or visibility over its own gas consumption?
(the `Runtime` proposed has no mechanism for the actor code to consume gas).

If there is no present need for it on any of the current actors, I would suggest it is left out of the spec. The cost of changing it in the future is minimal.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: This is an implementation detail and the mechanism to make that happen is very language specific.

If you are peusocoding the actors in "Go-ish" then ignore the error in go style, or give your pseudocode language a `try/catch` and just do:

```
try
   runtime.Send(blah)
catch
    // ignore 
end
```

It cleans up the interface and lets the implementors decide how to build the VM primitives the way they want and in a way that is closer to the language it is in. 

Btw, what actor needs to do calls on other actors and ignore errors? And why would we want actors to be able to call other actors that fail and carry on like nothing happen? Do we expect to reach the same state as other instances that did not see the error? How does that even work? An actor providing a pure function that fails?

internal/pkg/vm/internal/runtime/runtime.go
icorderi: It is more of a question for the spec team.. not sure what might be needed by the actors. If the answer is nothing then we don't need it now, it easy to add later. 

From my viewpoint, I see the chain as something immutable present in the global store for actor to potentially read pieces of.

Btw this only makes sense if the local-store the actor gets is JUST for things the actor wrote himself. If there is not actor-local store, then the only thing we have IS a global store.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Not sure this gets you want you want.. you are just enforcing that a call existed, not that it is the correct call. It sill requires the same security review. 
What are you doing about the methods that do NOT have a requirement? How in the runtime telling them apart? 
How will the runtime know you are validating the correct address?

From a security review standpoint, reviewing..
```
runtime.Caller().IsActor(actor.FooBar)
// super sensitive code
```

..or reviewing..
```
runtime.ValidateCallerIs(actor.FooBar)
// super sensitive code
```

..is the same.

If the spec requires certain methods to be only callable by a specific `Actor`, it might be best for it to just say that, and indicate which methods fall under this restriction. 

Implementations can decide how they want to deal with that (interfaces, method annotations, in-line checks, etc.).

internal/pkg/vm/internal/runtime/runtime.go
icorderi: `ValidateCallerIs()` seems just like a simpler version of `ValidateCallerMatches()`, I understand the problem you are trying to solve. 

We can discuss the approach behind this methods in the previous comment.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: I have too many questions on this one..
Do we have any docs describing this "reentrancy attack"? I want to see what it is you are trying to protect against.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Unless there is a need for it to really be a method dispatched on the VMContext I would take it out of the interface. Most languages will not want to resolve this as a method you call on an interface, since you want to take over control of the callers execution flow, and some languages might not want that to be done with exceptions, or recovering from panics.  

somewhere in the vm package/namespace/module:
```
// Abort this execution of the vm call stack. 
// The caller should receive error 777 `PanicError`. 
func abort(message string) 
```

actor code..
```go
func somemethod(r Runtime) Result {
// ...
abort("blahblah")
// ...
}
```

most likely it will be implemented as `return abort("blahblah")` in most places and have it build the correct `InvOutput`, with it being actual panic/exceptions on languages that enjoy catching them all.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: This is very unimplementable without using exceptions or macros. With the `abort()` you can at least return the object, this is a conditional. 

I would also remove it from the interface and go ahead and write `assert(<cond>)` on the pseudocode and just explain the `assert` method magic somewhere else. 

```go
// Assert with abort the execution of the vm call stack if the condition is `False`.
// The caller should receive error 66 `AssertError`. 
func assert(|| cond ||)
```

Implementors can add it back to the interface if they are fine with having it there, but same as with abort, unless the `assert/1` actually needs the `VMContext` object, then it should not be on that interface. Those two seem like free functions/macros.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Thanks.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: It's minor, it was more of an internal comment than anything. 
Personally I prefer reflecting the composition on the context, rather than flattening it out and turning it into a facade.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Got it, we can come back at this once Gas Accounting lands.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: If that case is the case I consider having a `type RandomnessEvent uint64`, and change the signature to `Randomness(epoch Epoch, event RandomnessEvent) Randomness`.

That being said, those offsets are not all hardcoded. The `InitActor` seems to call `Randomess/2` with the `ActorID` as an argument. 

Bonus question: Are the actors allowed to call `Randomness/2` with a first argument other than `Current Epoch`?

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Sounds good, we will probably keep it separate on our implementation so other actors don't see it.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Thanks @jzimmerman. 

Then unless @anorth, @acruikshank or @ZenGround0 have a strong preference for positional arguments and want to discuss this some more I'll move forward with the struct and change all the signature we have.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: yes thanks, we will have back to clean that up once that issue is resolved.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Thanks, makes sense.
 I'll wait for @zixuanzh and @jbenet to pitch in before resolving it.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Cool, I'll add it for now. Who can confirm?

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Good, I'm adding it on my end for now. Who can confirm the error is going into the spec?

internal/pkg/vm/internal/runtime/runtime.go
anorth: I sent you some references out of band.

internal/pkg/vm/internal/runtime/runtime.go
anorth: We can still implement it the way you suggest, @icorderi.

I concur we probably want to reserve `panic/recover` for "faults", i.e. nondeterministic failures like OOM, disk full.

internal/pkg/vm/internal/runtime/runtime.go
anorth: I think we did have this before. E.g. for payment of late fees in the storage miner actor.

My current understanding is that value will still be transferred "behind the scenes" before message execution (i.e. so the receiving actor has access to the value just transferred to it).

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+// Dragons: this is just to match names on the VM, Height got renamed to Epoch on the spec
+// TODO: rename types.BlockHeight to block.Epoch (issue: ???)
+type Epoch types.BlockHeight
+
+// Randomness is a string of random bytes
+// Note: placeholder for matching newtypes with the spec.
+type Randomness []byte
+
+// InvocInput are the params to invoke a method in an Actor.
+// Review: unless anyone in @go-filecoin has a strong preference for arguments, we are chaging to the struct the spec suggests.
```
anorth: 👌

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,163 @@ import (
...
+    Method  types.MethodID
+    Params  MethodParams
+    Value   types.AttoFIL
+}
+
+// ExitCode is the exit code of a method executing inside the VM.
+type ExitCode types.Int64
+
+// InvocOutput is waht os returned by an actor executing a method.
+// Review: multiple returns vs a return struct. There are some minor implications with swapping vm implementations written in another language.
```
anorth: I think this forms the receipt, which means we need the struct defined in order to build the receipt list and put its hash in the block header.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: > Bonus question: Are the actors allowed to call Randomness/2 with a first argument other than Current Epoch?

jzimmerman: > Why not just let the caller do the arithmetic of epoch - offset and reduce to a single parameter?

The offset here isn't measured in epochs; rather, each epoch `E` determines a pseudorandom stream, and then the user may provide an offset into that stream.

jzimmerman: > consider having a type RandomnessEvent uint64, and change the signature to Randomness(epoch Epoch, event RandomnessEvent) Randomness

I think this would be a good approach, if the event is the _only_ seed we need. However, it sounds like that's not the case (e.g., if `ActorID` is used as an argument). Given that requirement, I'd propose something like the following:

```
type RandomSeed []byte
type Serialization []byte
Randomness(epoch Epoch, seed RandomSeed)
DeriveRandomSeedExtended(event RandomnessEvent, s Serialization)
DeriveRandomSeed(event RandomnessEvent)
Serialize_T(T) Serialization  // for each type T, may be automatically generated in the future
```

What do you think?

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: My sense is that the isolation property isn't critical at the moment, but others should feel free to weigh in.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: I see, that makes sense. It would be nice if these pure functions (and, in fact, all functions involving nontrivial computation) were proxied through the runtime, since this would make gas accounting much easier. On the other hand, we don't want to expand the runtime API surface without limit. Would it make sense to have a single runtime API method (e.g., `Call`) that takes an ID of a pure function, along with arguments (either serialized or with some standard ABI)?

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: A few notes here:

- Using return values instead of exceptions/panics may be error-prone in languages that make error-propagation difficult. E.g., actor method A calls helper function B (not itself an actor method), and B returns some combination of values indicating failure, but A fails to check for this condition. If B instead panics, A is forced to terminate also. Exceptions/panics are less idiomatic Go, but I believe are much less error-prone in cases like this.

- If we had language support for robust return-value-based error handling (e.g. Rust's `try` macros and option types), then I would suggest to use that instead of exceptions. In general, I think the conclusion is that the desired functionality is clearly specified, but in each language, we should use the most secure (not necessarily the most idiomatic) way of realizing that functionality. In Go, this seems to be panic/recover.

- If we allow panic/recover for nondeterministic faults such as disk full, then we have to decide what to do in this case. Presumably this means the runtime shouldn't handle the panic (since it's not part of the VM/actor logic), and just crash the process as a normal panic would. I think we can achieve this functionality, while still using panic/recover for ordinary VM errors also, by using different exception types (see `_catchRuntimeErrors` in `runtime.go`).

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: - `rt.Assert` is a function that's callable from VM code, and thus, I think has to be on the `VMContext` (or what we previously called the `Runtime`, i.e., whatever the API object is that is accessible to each actor method). I think it's fine, in languages that make it natural, to use `return rt.AbortReturn(...)` instead of `rt.Abort(...)`, but I'm not convinced that Go is one of those languages (see above).

- We can't write `assert` as a macro in pseudocode, because the intent is for the spec (or at least these parts of it) to be actual Go code -- correct, compilable, and testable (albeit possibly too inefficient to serve as a reference implementation).

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: > somewhat related question: the current go-filecoin code was moving funds behind the scenes at the vmruntime level. Will we continue to handle it that way or will the actor code handle money movement now that it can see it?

The runtime should handle this. See `_transferFunds` and `_sendInternal` in `runtime.go`

jzimmerman: > My current understanding is that value will still be transferred "behind the scenes" before message execution (i.e. so the receiving actor has access to the value just transferred to it).

Yes -- the change here is that not only will the receiving actor have access to the value (in the sense that its balance will have been incremented), it will also have _knowledge_ of the value in that the runtime will confirm precisely how much value was transferred accompanying this particular message send.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: Agreed -- I think it's fine to have it return an `InvocOutput` for now

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: > Btw, what actor needs to do calls on other actors and ignore errors? And why would we want actors to be able to call other actors that fail and carry on like nothing happen? Do we expect to reach the same state as other instances that did not see the error? How does that even work? An actor providing a pure function that fails?

Errors that are handled in this fashion must be actual VM/actor errors, not nondeterministic failures (see below). For the former, one example is the cron actor -- suppose actors A, B, C, ... register with cron; then it should not be possible for A to DoS the cron calls to B and C by raising an error within A itself.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: See below -- I don't think the isolation property is critical.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: - It's an additional safety check. It has exactly the same review cost, but just in case security review misses some control flow path that completely fails to check the caller identity, we would still prefer to "fail safe" than allow the method to complete execution.

- For methods that don't need to check the caller identity, they should call `ValidateCallerMatches(...accept-everything pattern...)` to make that clear in a way that's (to the extent possible) checkable by the runtime.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: I'll submit a PR to the spec to add this

jzimmerman: https://github.com/filecoin-project/specs/pull/630

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Not sure why we would need serialization. 
I thought you wanted an offset into the random stream, why are you redefining it as a `[]byte`?

```go
type seed interface {
    seed() uint64
}

func Randomness(epoch Epoch, seeder seed) {}

func (x RandomnessEvent) seed() uint64 {
    // impl...
}

func (x ActorID) seed() uint64 {
    // impl...
}
```

internal/pkg/vm/internal/runtime/runtime.go
icorderi: As currently written in the spec, that `SendAllowingErrors` would allow the caller to trap an "abort()" and also an unhandled `panic()`, is that intended?

internal/pkg/vm/internal/runtime/runtime.go
icorderi: How would a reviewer fail to read that line? shouldn't be a check at the top of the method? 
Or is the spec intending to have dynamic checks on the code depending on input arguments?

internal/pkg/vm/internal/runtime/runtime.go
icorderi: The spec is not actually locking, and is not clear how this is `StateHandle` state is shared between reentrant calls to actor. 

I created the `ActorStateHandler` type and added the methods, left some `Review:` comments on it too.
We can take the discussion over there since there are quite a few distinct concerns.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: All good, I would propose a rename to `ValueReceived`.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+type Runtime2 struct {
+	CurerntEpoch() Epoch
+	// Review: for @go-filecoin, should we turn the offset into a RandomnessEvent?
+	Randomness(epoch Epoch, offset uint64) Randomness
+	Send(input InvocInput) result.InvocOutput
+	// SharedStorage is an IPLD store shared by all actors.
+	// Review: for @go-filecoin, we can skip this until we have actor code that actually needs this
+	// Review: for @spec, Add expiration (Epoch's) on PUT? (the question is mostly for global, but the local store might want to advise on this too).
+	// Review: for @spec, how will gas accounting work on global? PUT fee, GET fee, OWNER Fee?
+	SharedStorage() IpldStorage2
```
anorth: Yes, ok to skip until actor code needs it, but I don't think that will be very long. Maps in actor state are represented as [H]AMTs which in turn get/put their nodes here.

I suggest we skip questions of expiration and garbage collection for now here, but start a discussion in specs repo. I have a slightly better understanding of it now since we last talked.

I've filed an [issue for gas](https://github.com/filecoin-project/specs/issues/631) and suggest we leave it out of scope here.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+}
+
+// DispatchContext is the context the actor dispatcher receives.
+// Review: for @go-filecoin, do we want to have the `ctx.ValidateCaller(blahblah)` at the top of each method, or hidden away on the dispatch table?
```
anorth: Slight preference for within the actor methods, for transparent explicitness.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+	// Review: for @go-filecoin, simplified the API to be just one method.
+	ValidateCaller(CallerPattern)
+	// InvocationContext is the context to be passed down the actor method.
+	InvocationContext() InvocationContext
+}
+
+// InvocationContext is passed to the actors on each method call..
+type InvocationContext interface {
+	// Review: for @go-filecoin, composition or inheritance?
+	Runtime() Runtime2
```
anorth: You mean Go type extension? Yes, that could be good here, but your discretion.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+	Caller() address.Address
```
anorth: Comments need to be very explicit whether, in the case of nested calls, this is the immediate caller or the account actor at the top of the call tree. I'm assuming the former (and the same for value transferred and any other per-message values). cc @jzimmerman

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+	StateHandler() ActorStateHandler
+
+	// Review: for @go-filecoin+@spec, renamed `ValueSupplied` to `ValueReceived`. The "supplied" is not clear by whom.
+	ValueReceived() types.AttoFIL
+	Balance() types.AttoFIL
+
+	LocalStorage() IpldStorage2
```
anorth: Pending information to the contrary, let's drop this in favour of the `SharedStorage`. Even if that changes, I expect it to be resolved pretty quickly such that we don't have any migration friction to worry about.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+// initInvocationContext is a more powerfull context for the "init" actor.
+type initInvocationContext interface {
+	InvocationContext
+	// Review: for @spec, added an error that is not yet on the spec but should be
+	// Note: @jzimmerman: "Agreed, this should support an error return."
+	CreateActor(cid cid.Cid, addr address.Address, constructorParams MethodParams) error
+}
 
+// minerInvocationContext has some special sauce for the miner.
+// Review: for @go-filecoin, do we want custom interfaces for the miners instead of grouping a bunch fo methods on the InvocationContext?
```
anorth: Yeah nice, this will help keep testing deps small.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+type minerInvocationContext interface {
 	Verifier() verification.Verifier
 }
 
+// ActorStateHandler handles the actor state, allowing actors to lock on the state.
+type ActorStateHandler interface {
+	// Take returns the `Cid` for the state and locks the state.
+	//
+	// Any futre calls to `Take` by other instances will block until `Release` or `UpdateRelease` is called.
+	// Review: for @spec, the spec impl of `Take` is not doing any blocking, just panic'ing on the second call on the SAME object instance (this objects are not kept around, so it is always a new instance..).
```
anorth: There's no concurrency, so blocking doesn't make sense. IMO this should cause the call to abort. Comparable to an assertion failure, indicating programmer error.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+
+// CallerPattern checks if the caller matches the pattern.
+type CallerPattern interface {
+	IsMatch() bool
+}
+
+// Epoch is the current chain epoch.
+// Dragons: this is just to match names on the VM, Height got renamed to Epoch on the spec
+// TODO: rename types.BlockHeight to block.Epoch (issue: ???)
+type Epoch types.BlockHeight
```
anorth: I *think* that we should be replacing go-filecoin's `types.BlockHeight` (which wraps bigint) with a newtype for `types.Uint64`, and then yes replacing with `Epoch` everywhere.

Related to https://github.com/filecoin-project/specs/issues/615, we need our own issue for that busywork.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+    Value   types.AttoFIL
+}
+
+// MethodParam is the parameter to an Actor method already encoded for execution in the VM.
+type MethodParam []byte
+// MethodParams is a list of `MethodParam` to be passed into an Actor method.
+type MethodParams []MethodParam
+
+// Abort will stop the VM execution and return an abort error to the caller.
+// Review: @go-filecoin, do we want to handle this with panics?
```
anorth: I am pretty torn on this. Our code will line up better with the spec if we do use a panic mechanism. If we do, we should wrap the argument to `panic` in a struct so that a `recover` at the top of the call tree can distinguish an explicit call abort (which all nodes should reproduce) from an unreproducible environmental failure (e.g. out of memory), the latter of which should crash the node.

It's my understanding that an aborted call will not result in a receipt. It does consume gas and increment the caller's sequence number, though. I.e. equivalent to running out of gas. I get confused every time I think about this tho, so could be wrong.

We need also to distinguish between this and an "application" failure that "successfully" returns a non-zero exit code.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -8,27 +8,142 @@ import (
...
+}
+
+//
+// Dragons: this will get deleted, still here while we finish some discussions on the PR.
+//
+
+// Runtime defines the ABI interface exposed to actors.
+type Runtime interface {
+	// Review: this is not in the spec ~anymore~ yet..
+	// Note: @jzimmerman: still a WIP, things like the miner actor might need it
```
anorth: Please prefer to expose the individual values that are needed, if at all possible @jzimmerman

internal/pkg/vm/internal/runtime/runtime.go
anorth: I too think this is internal, not exposed to actors.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: ok, skipping for now.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Agreed.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: SGTM

internal/pkg/vm/internal/runtime/runtime.go
icorderi: I don't particularly like using panics for control flow.

> It's my understanding that an aborted call will not result in a receipt. It does consume gas and increment the caller's sequence number, though. I.e. equivalent to running out of gas. I get confused every time I think about this tho, so could be wrong.

The abort consuming gas has some implications on how we manage gas and handle the panic. 
The gas consumed by a function 3 levels deep which panics, needs to be accounted when the top level call traps the panic.

> We need also to distinguish between this and an "application" failure that "successfully" returns a non-zero exit code.

The abort already has a unique error status code (code=8).

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Except for a single call I think we are safe to remove.

All calls to `.Message()` access the `.From()` and `.Value()`, except for `paymentbroker.go` which once, access `.CallSeqNum()` to build a `ChannelID`.

The `From`, and `Value` calls can be replaced with the `Caller()` and `ValueReceived()` in the new context.

I'm not certain about the `CallSeqNum()` or how the new actor code will determine the ChannelID without that info being available (randomness perhaps?) @jzimmerman?.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: > Not sure why we would need serialization.

This is to avoid errors that might arise from duplication of code otherwise. E.g., if we have:

```go
type Foo struct {
  A int
  B int
}
```

and we would like to use objects of type `Foo` directly as random seeds, then in the underlying implementation, we want to derive some integer seed from any `Foo` instance such that distinct instances produce distinct seeds. `Serialize` (followed by a collision-resistant hash) and `CID` would both satisfy this criterion, so we shouldn't add a third independent implementation for it as well. (In fact, ideally, CID should also be implemented in terms of serialization.)

jzimmerman: > I thought you wanted an offset into the random stream, why are you redefining it as a []byte?

I think this is just a question of notation -- if we model it as an _infinite_ stream, any bit string can be used as an integer to index into it. But for practical purposes, we can model it as a finite stream of length `2^64`, and hash the `[]byte` index into that range instead.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: No -- the intent is for `SendAllowingErrors` (N.B.: renamed to `SendCatchingErrors`) to only intercept actual VM/actor errors, not nondeterministic failures or unhandled panics. Will draft a PR to change this in the spec.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: Yes, ideally it should always be a simple check at the top of the method (and its absence should easily be caught during security review), but we'd like to allow for the possibility of more complex checks in the future. (And even if we don't ever need more complex checks, I think it still doesn't hurt to have this extra check in the runtime, just in case.)

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: > The spec is not actually locking

In the context of a single method invocation, the effect of the runtime checks is to lock the state -- we don't allow concurrent calls, so we don't use inter-thread synchronization. But to the actor code, the state appears locked: it can't be re-acquired, it must be released, and no `Send` call can be made while it's held.

jzimmerman: > not clear how this is StateHandle state is shared between reentrant calls to actor

Reentrant calls should never need to share a state handle -- in fact, this can't happen since method calls should be the only mechanism for reentrancy, but this is prohibited if a state handle is held.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: @anorth @icorderi Yes, I would also assume the former (i.e. immediate caller). Also agree comments should be very explicit on this.

internal/pkg/vm/internal/runtime/runtime.go
jzimmerman: In case it's helpful, there's already a reference implementation of this error struct mechanism (`RuntimeError`) in the spec: `systems/filecoin_vm/runtime/runtime.go`, see in particular the `_catchRuntimeErrors` function.

(Also note I have a pending PR to remove the `MethodPanic` case in that function, per our earlier discussion, since we shouldn't be intercepting nondeterministic/nonreproducible panics here.)

jzimmerman: (Side note: @icorderi, for what it's worth, I also don't like using panics for control flow in general, but in Go we don't have a robust error-handling mechanism such as option types or Rust's `try!` macros, so I didn't see a good alternative.)

anorth: I may have been wrong about aborted calls not resulting in a receipt. I was talking about old behaviour.

Re gas: I have heard somewhere that Eth consumes the entire gas limit for a message when it aborts, rather than attempt consensus about the cost (and also disincentivize malformed messages). So the accounting may be simple. However we need this to be clearly specified and to check it doesn't mess up incentives with block validation rules etc (e.g. miners including messages out of sequence order in order to collect max gas).

internal/pkg/vm/internal/runtime/runtime.go
anorth: There's no payment broker in the spec, go-filecoin has an implementation from a prior spec. In the new world, each channel is a distinct actor.

jzimmerman: I don't see an issue with exposing `CallSeqNum` to actor code, but we should make clear that this is a sequence number across _top-level_ calls, and isn't incremented by actor method subcalls. (If we need a sequence number with the latter definition, which may be useful in some cases, we should add that separately and make the terminology very clear.)

internal/pkg/vm/internal/runtime/runtime.go
icorderi: lets talk offline

internal/pkg/vm/internal/runtime/runtime.go
icorderi: went for composition.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Let's wait for the actor pseudocode to land and see what we need.

Issue #804, opened 2018-08-28
Switch to gossipsub


dignifiedquire:

anorth:
@dignifiedquire this old issue predates many of the current contributors' tenure. Could you add some more detail so we can understand the motivation and what's needed?

dignifiedquire:
There are two different implementations of pubsub in libp2p land, one the older one is called floodsub and a newer one called gossipsub. Gossipsub is an improvement over floodsub, but was just being finished when the work on go-filecoin started. This is about switching to it, to get the new benefits. 

More details on gossipsub can be found here: 
- https://github.com/libp2p/specs/tree/master/pubsub/gossipsub
- https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go

anorth:
This is blocked on GossipSub exposing some needed functionality (@zenground0 has the details)

PR #3637, opened 2019-11-06
Feat/gossip sub

ZenGround0:
### Motivation
We want to use gossipsub.  We also want to avoid races in the tests.  In the past attempts to integrate to gossipsub led to a race of the form
1) node A comes online
2) node A publishes a message node B needs to hear about
3) node B isn't around to hear it and the message silently drops
4) node B comes online and waits forever
5) node A waits forever on node B

This race has always been around, gossipsub just makes it worse by adding in extra steps to setup a topic mesh during subscription that caused tons of tests to hit the race.

### Proposed changes
We now have gossipsub without races by
1) updating to a new version of gossipsub
2) adding a noop discovery to our pubsub that allows publish to block until a peer exists for it
3) calling publish with an option that forces it to block until there is one other peer subscribing to the topic

This required a slight rethinking of the code that hit Publish: message send and add new block, in order to prevent daemon deadlock

4) add new block just introduces an extra goroutine wrapping publish and logs publish errors instead of propagating them
5) message send also wraps publish in a goroutine.  To help tests and for potentially useful handling down the road I've introduced a third argument from message send, an error channel that will output the result of publish.  All production code ignores the new error channel, maintaining the assumption that you can run `message send` without any peers and add the message to your pool.  The error channel seems like a step in the right direction if we keep the current architecture, though we might want to separate out the 3 different things happening with a send (queue push, pool put, pubsub publish) at a level closer to commands.

Closes issue #3630 #804

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/network_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/network_submodule.go
@@ -2,6 +2,7 @@ package submodule
 
 import (
 	"context"
+	"time"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/config"
@@ -31,7 +32,8 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/pkg/errors"
 
-	"github.com/filecoin-project/go-filecoin/internal/pkg/net"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/discovery"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/net"	
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/state"
 )
@@ -111,9 +113,13 @@ func NewNetworkSubmodule(ctx context.Context, config networkConfig, repo network
 	}
 
 	// Set up libp2p network
+	// The gossipsub heartbeat timeout needs to be set sufficiently low
+	// to enable publishing on first connection.  The default of one
+	// second is not acceptable for tests.
+	libp2pps.GossipSubHeartbeatInterval = 100 * time.Millisecond
 	// TODO: PubSub requires strict message signing, disabled for now
 	// reference issue: #3124
-	gsub, err := libp2pps.NewGossipSub(ctx, peerHost, libp2pps.WithMessageSigning(false))
+	gsub, err := libp2pps.NewGossipSub(ctx, peerHost, libp2pps.WithMessageSigning(false), libp2pps.WithDiscovery(&discovery.NoopDiscovery{}))
 	if err != nil {
 		return NetworkSubmodule{}, errors.Wrap(err, "failed to set up network")
 	}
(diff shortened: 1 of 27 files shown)


internal/app/go-filecoin/internal/submodule/network_submodule.go
```diff
@@ -111,9 +113,13 @@ func NewNetworkSubmodule(ctx context.Context, config networkConfig, repo network
 	}
 
 	// Set up libp2p network
+	// The gossipsub heartbeat timeout needs to be set sufficiently low
+	// to enable publishing on first connection.  The default of one
+	// second is not acceptable for tests.
+	libp2pps.GossipSubHeartbeatInterval = 100 * time.Millisecond
```
anorth: This might be too low for real-world use. Could you move it into the config?

internal/pkg/discovery/noop_discovery.go
```diff
@@ -0,0 +1,29 @@
...
+	libp2pdisc "github.com/libp2p/go-libp2p-core/discovery"
+	pstore "github.com/libp2p/go-libp2p-peerstore" // nolint: staticcheck
+)
+
+// NoopDiscovery satisfies the discovery interface without doing anything
+type NoopDiscovery struct{}
+
+// FindPeers returns a dead channel that is always closed
+func (sd *NoopDiscovery) FindPeers(ctx context.Context, ns string, opts ...libp2pdisc.Option) (<-chan pstore.PeerInfo, error) { // nolint: staticcheck
+	stupidCh := make(chan pstore.PeerInfo) // nolint: staticcheck
```
anorth: I don't think it's stupid. `emptyCh`? `closedCh`?

internal/app/go-filecoin/plumbing/api.go
```diff
@@ -270,12 +270,12 @@ func (api *API) Snapshot(ctx context.Context, baseKey block.TipSetKey) (consensu
 // message using the wallet. This call "sends" in the sense that it enqueues the
 // message in the msg pool and broadcasts it to the network; it does not wait for the
 // message to go on chain. Note that no default from address is provided.
-func (api *API) MessageSend(ctx context.Context, from, to address.Address, value types.AttoFIL, gasPrice types.AttoFIL, gasLimit types.GasUnits, method types.MethodID, params ...interface{}) (cid.Cid, error) {
+func (api *API) MessageSend(ctx context.Context, from, to address.Address, value types.AttoFIL, gasPrice types.AttoFIL, gasLimit types.GasUnits, method types.MethodID, params ...interface{}) (cid.Cid, chan error, error) {
```
anorth: Document the error channel's semantics, which I understand to be "receives either nil or an error and is then closed after the message is propagated to pubsub", making it clear that a caller can always rely on receiving from the channel as a signal that the send is done.

internal/app/go-filecoin/node/message_propagate_test.go
```diff
@@ -56,7 +56,7 @@ func TestMessagePropagation(t *testing.T) {
 	connect(t, nodes[0], nodes[1])
 	connect(t, nodes[1], nodes[2])
 	// Wait for network connection notifications to propagate
-	time.Sleep(time.Millisecond * 50)
+	time.Sleep(time.Millisecond * 200)
```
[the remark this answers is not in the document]
ZenGround0: No one likes this less than me but the tools we have from gossipsub don't help us with the race here unless we go into a laborious and kafkaesque setup.  If we can get a "block on topic / subscription until there are peers on the mesh" call from pubsub we could use it here.

Issue #3568, opened 2019-10-17
Fix MessageReceipt exit code and gas value


anorth:
### Description
The receipt structure should look like this.

```
type MessageReceipt struct {
    ExitCode       uint8
    ReturnValue  []byte
    GasUsed      GasUnits
}
```

We need to drop the unnecessary nested array, and change the gas to be the number of gas units, rather than the product of units and price.

FYI @acruikshank

anorth:
Done

PR #3646, opened 2019-11-11
Change StorageMinerActor.GetProvingWindow to return UIntArray instead of multiple returns.

anorth:
### Motivation
This is the only method utilizing multiple returns. Changing this will let us change the
`dispatch.FunctionSignature` to return a single value instead of array, and then propagate
this to the MessageReceipt. See #3568.

I avoided creating a custom return type for the tuple just to save on noise, but can do so if requested. This also uses uint64 instead of `types.BlockHeight` because I expect the spec to define epoch numbers to be uint64 anyway (and we can ditch the type wrapper). Retaining the `BlockHeight` would require a custom ABI type.

### Proposed changes
Return a `UIntArray` (always of 2 elements) instead of multiple values.

### DIFF
--- a/internal/app/go-filecoin/porcelain/client.go
+++ b/internal/app/go-filecoin/porcelain/client.go
@@ -79,13 +79,13 @@ func listAsksFromActorResult(ctx context.Context, plumbing claPlubming, actorRes
 		return err
 	}
 
-	var asksIds []uint64
+	var asksIds []types.Uint64
 	if err := encoding.Decode(ret[0], &asksIds); err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


anorth:
Right, thanks. I've changed the `UintArray` to use `types.Uint64` which defines leb128 encoding. If we later change blockheight to be similar, it will be easy to switch this to use that.

My reading of the latest comments in https://github.com/filecoin-project/specs/issues/615#issuecomment-549498308 suggests that's we'll probably use CBOR's default encoding for all integers that are not explicitly BigInt.

PR #3648, opened 2019-11-12
Init actor Exec, GetAdressByActorID, GetActorIDByAddress

acruikshank:
### Motivation

We need to move actor creation into the init actor so we get ID addresses.

### Proposed changes

This PR sets the following goals:

1. `StorageMinerActor`s are instantiated in the InitActor's `Exec` method.
2. `StorageMinerActor`s get ID addresses instead of actor addresses.

The [InitActor spec](https://filecoin-project.github.io/specs/#InitActor) and the VM methods on which it relies are very rough. This PR attempts to implement the Spec faithfully where doing so will lead to a working implementation. It attempts to make reasonable guesses where the spec has gaps. And it sticks to the existing implementation where those options aren't available. Decisions are documented in the code.

### DIFF
--- a/cmd/go-filecoin/util_test.go
+++ b/cmd/go-filecoin/util_test.go
@@ -18,7 +18,7 @@ func TestOptionalAddr(t *testing.T) {
 
 		opts := make(cmdkit.OptMap)
 
-		specifiedAddr, err := address.NewActorAddress([]byte("a new test address"))
+		specifiedAddr, err := address.NewSecp256k1Address([]byte("a new test address"))
 		require.NoError(t, err)
 		opts["from"] = specifiedAddr.String()
 
--- a/internal/app/go-filecoin/node/node.go
+++ b/internal/app/go-filecoin/node/node.go
@@ -370,7 +370,7 @@ func (node *Node) SetupMining(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get mining address")
 	}
-	_, err = node.PorcelainAPI.ActorGet(ctx, minerAddr)
+	_, err = node.PorcelainAPI.ActorGetStable(ctx, minerAddr)
 	if err != nil {
 		return errors.Wrap(err, "failed to get miner actor")
 	}
@@ -410,7 +410,7 @@ func (node *Node) StartMining(ctx context.Context) error {
 
 	err := node.SetupMining(ctx)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed to setup mining")
 	}
 
 	minerAddr, err := node.MiningAddress()
--- a/internal/app/go-filecoin/porcelain/protocol_test.go
+++ b/internal/app/go-filecoin/porcelain/protocol_test.go
@@ -35,9 +35,9 @@ func (tppp *testProtocolParamsPlumbing) ChainHeadKey() block.TipSetKey {
 }
 
 func (tppp *testProtocolParamsPlumbing) MessageQuery(ctx context.Context, optFrom, to address.Address, method types.MethodID, _ block.TipSetKey, params ...interface{}) ([][]byte, error) {
-	if method == storagemarket.GetProofsMode {
+	if to == address.StorageMarketAddress && method == storagemarket.GetProofsMode {
 		return [][]byte{{byte(types.TestProofsMode)}}, nil
-	} else if method == initactor.GetNetwork {
+	} else if to == address.InitAddress && method == initactor.GetNetwork {
 		return [][]byte{[]byte("protocolTest")}, nil
 	}
 	return [][]byte{}, errors.Errorf("call to unknown method %s", method)
--- a/internal/pkg/chain/traversal_test.go
+++ b/internal/pkg/chain/traversal_test.go
@@ -16,7 +16,7 @@ import (
 
 func TestIterAncestors(t *testing.T) {
 	tf.UnitTest(t)
-	miner, err := address.NewActorAddress([]byte(fmt.Sprintf("address")))
+	miner, err := address.NewSecp256k1Address([]byte(fmt.Sprintf("address")))
 	require.NoError(t, err)
 
 	t.Run("iterates", func(t *testing.T) {
--- a/internal/pkg/sectorbuilder/utils.go
+++ b/internal/pkg/sectorbuilder/utils.go
@@ -8,21 +8,19 @@ import (
 
 // AddressToProverID creates a prover id by padding an address hash to 31 bytes
 func AddressToProverID(addr address.Address) [31]byte {
-	// this code will no longer WAE when ID's or BLS pub keys are added
-	// as they will break the assumption of addresses all being the same length
-	if addr.Protocol() != address.Actor && addr.Protocol() != address.SECP256K1 {
-		panic("cannot create prover hash from new address protocol")
+	payload := addr.Payload()
+	if len(payload) > 31 {
+		panic("cannot create prover id from address because address is too long")
 	}
-	hash := addr.Payload()
 
-	dlen := 31          // desired length
-	hlen := len(hash)   // hash length
-	padl := dlen - hlen // padding length
+	dlen := 31           // desired length
+	hlen := len(payload) // hash length
+	padl := dlen - hlen  // padding length
 
 	var prid [31]byte
 
 	// will copy dlen bytes from hash
-	copy(prid[:], hash)
+	copy(prid[:], payload)
 
 	if padl > 0 {
 		copy(prid[hlen:], bytes.Repeat([]byte{0}, padl))
--- a/internal/pkg/types/message.go
+++ b/internal/pkg/types/message.go
@@ -29,6 +29,8 @@ const (
 	InvalidMethodID = MethodID(0xFFFFFFFFFFFFFFFF)
 	// SendMethodID is the method ID for sending money to an actor.
 	SendMethodID = MethodID(0)
+	// ConstructorMethodID is the method ID used to initialize an actor's state.
+	ConstructorMethodID = MethodID(1)
 )
 
 // GasUnits represents number of units of gas consumed
--- a/internal/pkg/version/protocol_version_table_test.go
+++ b/internal/pkg/version/protocol_version_table_test.go
@@ -3,7 +3,7 @@ package version
 import (
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
-	"github.com/magiconair/properties/assert"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"testing"
@@ -54,7 +54,7 @@ func TestUpgradeTable(t *testing.T) {
 	t.Run("constructing a table with no versions is an error", func(t *testing.T) {
 		_, err := NewProtocolVersionTableBuilder(network).Build()
 		require.Error(t, err)
-		assert.Matches(t, err.Error(), "no protocol versions specified for network testnetwork")
+		assert.Contains(t, err.Error(), "no protocol versions specified for network testnetwork")
 	})
 
 	t.Run("constructing a table with no version at genesis is an error", func(t *testing.T) {
@@ -62,7 +62,7 @@ func TestUpgradeTable(t *testing.T) {
 			Add(network, 2, types.NewBlockHeight(20)).
 			Build()
 		require.Error(t, err)
-		assert.Matches(t, err.Error(), "no protocol version at genesis for network testnetwork")
+		assert.Contains(t, err.Error(), "no protocol version at genesis for network testnetwork")
 	})
 
 	t.Run("ignores versions from wrong network", func(t *testing.T) {
@@ -98,7 +98,7 @@ func TestUpgradeTable(t *testing.T) {
 			Add(network, 4, types.NewBlockHeight(40)).
 			Build()
 		require.Error(t, err)
-		assert.Matches(t, err.Error(), "protocol version 2 effective at 30 is not greater than previous version, 2")
+		assert.Contains(t, err.Error(), "protocol version 2 effective at 30 is not greater than previous version, 2")
 	})
 
 	t.Run("does not permit version numbers to decline", func(t *testing.T) {
@@ -110,6 +110,6 @@ func TestUpgradeTable(t *testing.T) {
 			Add(network, 0, types.NewBlockHeight(40)).
 			Build()
 		require.Error(t, err)
-		assert.Matches(t, err.Error(), "protocol version 3 effective at 10 is not greater than previous version, 4")
+		assert.Contains(t, err.Error(), "protocol version 3 effective at 10 is not greater than previous version, 4")
 	})
 }
--- a/internal/pkg/vm/abi/abi.go
+++ b/internal/pkg/vm/abi/abi.go
@@ -5,6 +5,7 @@ import (
 	"math/big"
 	"reflect"
 
+	"github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p-core/peer"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
@@ -30,6 +31,8 @@ const (
 	BytesAmount
 	// ChannelID is a *types.ChannelID
 	ChannelID
+	// Cid is a cid.Cid
+	Cid
 	// BlockHeight is a *types.BlockHeight
 	BlockHeight
 	// Integer is a *big.Int
@@ -83,6 +86,8 @@ func (t Type) String() string {
 		return "*types.BytesAmount"
 	case ChannelID:
 		return "*types.ChannelID"
+	case Cid:
+		return "cid.Cid"
 	case BlockHeight:
 		return "*types.BlockHeight"
 	case Integer:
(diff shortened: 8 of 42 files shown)


go.mod
```diff
@@ -20,10 +20,9 @@ require (
 	github.com/filecoin-project/go-leb128 v0.0.0-20190212224330-8d79a5489543
 	github.com/filecoin-project/go-sectorbuilder v0.0.0-20190801004428-e75bc9b0aaea
 	github.com/go-check/check v0.0.0-20190902080502-41f04d3bba15 // indirect
-	github.com/go-critic/go-critic v0.0.0-20181204210945-ee9bf5809ead // indirect
 	github.com/golang/groupcache v0.0.0-20191027212112-611e8accdfc9 // indirect
 	github.com/golang/mock v1.3.1 // indirect
-	github.com/golangci/golangci-lint v1.17.1
+	github.com/golangci/golangci-lint v1.18.0
```
acruikshank: This attempts to adress golang 13 build issue.

internal/app/go-filecoin/porcelain/protocol_test.go
```diff
@@ -35,9 +35,9 @@ func (tppp *testProtocolParamsPlumbing) ChainHeadKey() block.TipSetKey {
 }
 
 func (tppp *testProtocolParamsPlumbing) MessageQuery(ctx context.Context, optFrom, to address.Address, method types.MethodID, _ block.TipSetKey, params ...interface{}) ([][]byte, error) {
-	if method == storagemarket.GetProofsMode {
+	if to == address.StorageMarketAddress && method == storagemarket.GetProofsMode {
```
acruikshank: Method id's aren't unique, so this needs both.

internal/pkg/sectorbuilder/utils.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // AddressToProverID creates a prover id by padding an address hash to 31 bytes
 func AddressToProverID(addr address.Address) [31]byte {
-	// this code will no longer WAE when ID's or BLS pub keys are added
+	// this code will no longer WAE when BLS pub keys are added
 	// as they will break the assumption of addresses all being the same length
-	if addr.Protocol() != address.Actor && addr.Protocol() != address.SECP256K1 {
+	if addr.Protocol() != address.Actor && addr.Protocol() != address.SECP256K1 && addr.Protocol() != address.ID {
```
acruikshank: ID address payloads are much less than 32 bytes, so they still work as ProverIDs

internal/pkg/types/message.go
```diff
@@ -29,6 +29,8 @@ const (
 	InvalidMethodID = MethodID(0xFFFFFFFFFFFFFFFF)
 	// SendMethodID is the method ID for sending money to an actor.
 	SendMethodID = MethodID(0)
+	// ConstructorMethodID is the method ID used to initialize an actor's state.
+	ConstructorMethodID = MethodID(1)
```
acruikshank: An actor's constructor is always at method id 1. See the [VM actor spec](https://filecoin-project.github.io/specs/#systems__filecoin_vm__actor).

internal/pkg/vm/abi/abi.go
```diff
@@ -30,6 +31,8 @@ const (
 	BytesAmount
 	// ChannelID is a *types.ChannelID
 	ChannelID
+	// Cid is a cid.Cid
+	Cid
```
acruikshank: Cid is an argument for the first time in `InitActor.Exec`

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -21,12 +24,26 @@ type Actor struct{}
...
+	id := s.NextID
+	s.NextID++
+	return id
 }
 
 // Actor methods
 const (
-	GetNetwork types.MethodID = iota + 32
+	Exec                 types.MethodID = 3
+	GetActorIDForAddress types.MethodID = 4
```
acruikshank: The [spec](https://filecoin-project.github.io/specs/#InitActor) only specifies ids for these two methods.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -21,12 +24,26 @@ type Actor struct{}
...
+	// currently unspecified
+	GetAddressForActorID types.MethodID = iota + 32
```
acruikshank: This is added to access the mapping defined in the spec.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -99,3 +138,190 @@ func (*Impl) GetNetwork(ctx runtime.Runtime) (string, uint8, error) {
...
+
+	err = lookup.Set(ctx, addr.String(), actorID)
+	if err != nil {
+		return cid.Undef, errors.FaultErrorWrapf(err, "could not set id")
+	}
+
+	return lookup.Commit(ctx)
+}
+
+func (a *Impl) keyForActorID(actorID types.Uint64) (string, error) {
```
acruikshank: This is unspecified. The spec suggests ipld maps can have integer keys, but our HAMT only supports string keys. Using our default encoding (leb128) and casting that to a string makes a pretty space efficient key.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -99,3 +138,190 @@ func (*Impl) GetNetwork(ctx runtime.Runtime) (string, uint8, error) {
...
+	key, err := encoding.Encode(actorID)
+	if err != nil {
+		return "", errors.FaultErrorWrapf(err, "could not encode actor id: %d", actorID)
+	}
+
+	return string(key), nil
+}
+
+func (a *Impl) isBuiltinActor(code cid.Cid) bool {
+	return code.Equals(types.StorageMarketActorCodeCid) ||
```
acruikshank: This is awful but in the spec. This should really be available from the context and derived from BuiltinActor data. Also these checks are unnecessary if only non-singleton, builtin actors have constructor methods.

internal/pkg/vm/actor/builtin/miner/miner_test.go
```diff
@@ -113,9 +113,8 @@ func TestChangeWorker(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	st, vms := th.RequireCreateStorages(ctx, t)
```
acruikshank: Remove a test cross dependency.

internal/pkg/vm/actor/storage.go
```diff
@@ -85,6 +85,23 @@ func ReadState(ctx runtime.Runtime, st interface{}) error {
 	return nil
 }
 
+// WriteState stores state and commits it as the actor's head
+func WriteState(ctx runtime.Runtime, state interface{}) error {
```
acruikshank: Finally breaking from the nested callback pattern for actor state manipulation.

internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -20,7 +20,7 @@ type Runtime interface {
 	Charge(cost types.GasUnits) error
 	SampleChainRandomness(sampleHeight *types.BlockHeight) ([]byte, error)
 
-	CreateNewActor(addr address.Address, code cid.Cid, initalizationParams interface{}) error
+	CreateNewActor(addr address.Address, code cid.Cid) error
```
acruikshank: Initialization now happens in a subsequent call to the constructor.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -313,6 +298,10 @@ func send(ctx context.Context, transfer TransferFn, vmCtx ExtendedRuntime) ([][]
 		panic("trying to execute fake method on the actual VM, fix test")
 	}
 
+	if msg.Method == types.ConstructorMethodID && !vmCtx.From().Code.Equals(types.InitActorCodeCid) {
```
acruikshank: Without this, a network user could call an actor's constructor directly and reinitialize it's state. Conceivably the constructor method itself could make this check, but constructors should be pretty consistent so this is safer.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -99,3 +138,190 @@ func (*Impl) GetNetwork(ctx runtime.Runtime) (string, uint8, error) {
...
+		return address.Undef, errors.CodeError(err), errors.RevertErrorWrapf(err, "could not load lookup for cid: %s", state.IdMap)
+	}
+
+	key, err := a.keyForActorID(actorID)
+	if err != nil {
+		return address.Undef, errors.CodeError(err), errors.FaultErrorWrapf(err, "could not encode actor id: %d", actorID)
+	}
+
+	var addr address.Address
+	err = lookup.Find(ctx, key, &addr)
```
ZenGround0: aside: The fact that the `LoadLookup` `Find` pattern is so common it might make sense to create a helper in actor/storage.go in the same style as `SetKeyValue`

internal/pkg/vm/actor/builtin/initactor/init.go
ZenGround0: worth writing this down in a spec PR / issue / go-filecoin issue accounting for differences

internal/pkg/vm/actor/builtin/initactor/init.go
ZenGround0: Power actor could go in here too

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -99,3 +138,190 @@ func (*Impl) GetNetwork(ctx runtime.Runtime) (string, uint8, error) {
...
+		code.Equals(types.InitActorCodeCid) ||
+		code.Equals(types.MinerActorCodeCid) ||
+		code.Equals(types.BootstrapMinerActorCodeCid) ||
+		code.Equals(types.PaymentBrokerActorCodeCid)
+}
+
+func (a *Impl) isSingletonActor(code cid.Cid) bool {
+	return code.Equals(types.StorageMarketActorCodeCid) ||
+		code.Equals(types.InitActorCodeCid) ||
+		code.Equals(types.PaymentBrokerActorCodeCid)
```
ZenGround0: Power actor should go here if it goes above.

internal/pkg/vm/actor/storage.go
```diff
@@ -85,6 +85,23 @@ func ReadState(ctx runtime.Runtime, st interface{}) error {
...
+	stage := ctx.Storage()
+
+	cid, err := stage.Put(state)
+	if err != nil {
+		return vmerrors.RevertErrorWrap(err, "Could not stage memory chunk")
```
ZenGround0: I know this is just taking from `WithState` but I think these should be fault errors.  They all trace back to ipld cbor and blockstore errors.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -21,12 +24,26 @@ type Actor struct{}
 
 // State is the init actor's storage.
 type State struct {
-	Network string
+	Network    string
+	AddressMap cid.Cid `refmt:",omitempty"`
+	IdMap      cid.Cid `refmt:",omitempty"`
```
anorth: As we discussed in person, the map from ID to actor address is unnecessary if the state tree is keyed by ID address (for non-account actors). I know it's in the spec but I think it will be removed.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -99,3 +138,190 @@ func (*Impl) GetNetwork(ctx runtime.Runtime) (string, uint8, error) {
...
+		return 0, internal.ErrInsufficientGas, errors.RevertErrorWrap(err, "Insufficient gas")
+	}
+
+	var state State
+	err := actor.ReadState(rt, &state)
+	if err != nil {
+		return 0, errors.CodeError(err), err
+	}
+
+	ctx := context.TODO()
```
anorth: A thought @icorderi : would be handy for the Runtime to provide a context, since numerous Go calls will need it (as a Go-specific thing). This also gives some hook to terminating infinite loops and the like.

cmd/go-filecoin/util_test.go
```diff
@@ -18,7 +18,7 @@ func TestOptionalAddr(t *testing.T) {
 
 		opts := make(cmdkit.OptMap)
 
-		specifiedAddr, err := address.NewActorAddress([]byte("a new test address"))
+		specifiedAddr, err := address.NewSecp256k1Address([]byte("a new test address"))
```
acruikshank: Actor addresses are now always converted to Id addresses before processing, so Secp256k1 addresses are safer for testing.

internal/app/go-filecoin/node/node.go
```diff
@@ -370,7 +370,7 @@ func (node *Node) SetupMining(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get mining address")
 	}
-	_, err = node.PorcelainAPI.ActorGet(ctx, minerAddr)
+	_, err = node.PorcelainAPI.ActorGetStable(ctx, minerAddr)
```
acruikshank: This new function does the Actor address -> Id address conversion.

internal/pkg/version/protocol_version_table_test.go
```diff
@@ -3,7 +3,7 @@ package version
 import (
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
-	"github.com/magiconair/properties/assert"
+	"github.com/stretchr/testify/assert"
```
acruikshank: we should standardize on testify assert

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -135,7 +143,13 @@ func (ctx *VMContext) Send(to address.Address, method types.MethodID, value type
 		return nil, 1, errors.NewFaultErrorf("unhandled: sending to self (%s)", msg.From)
 	}
 
-	toActor, err := deps.GetOrCreateActor(context.TODO(), msg.To, func() (*actor.Actor, error) {
+	// fetch id address from init actor if necessary
+	toAddr, err := ctx.resolveActorAddress(msg.To)
```
ZenGround0: Why do we go through all of the trouble of doing this outside the state machine if we are just going to do it inside?  Why are both of these resolutions necessary?

internal/pkg/chain/traversal_test.go
```diff
@@ -16,7 +16,7 @@ import (
 
 func TestIterAncestors(t *testing.T) {
 	tf.UnitTest(t)
-	miner, err := address.NewActorAddress([]byte(fmt.Sprintf("address")))
+	miner, err := address.NewSecp256k1Address([]byte(fmt.Sprintf("address")))
```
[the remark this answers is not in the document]
ZenGround0: More succinctly, what problem does this solve?

internal/pkg/vm/actor/builtin/storagemarket/storagemarket.go
```diff
@@ -196,6 +202,21 @@ func (*impl) createStorageMiner(vmctx runtime.Runtime, sectorSize *types.BytesAm
 	return ret.(address.Address), 0, nil
 }
 
+// retriveActorId uses init actor to map an actorAddress to an id address
+func retreiveActorID(vmctx runtime.Runtime, actorAddr address.Address) (address.Address, error) {
```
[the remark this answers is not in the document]
ZenGround0: Figured I'd drop some more info: https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

(2 page Go design doc, pretty helpful)

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -48,7 +52,9 @@ type VMContext struct {
 type NewContextParams struct {
 	From        *actor.Actor
 	To          *actor.Actor
+	ToAddr      address.Address
```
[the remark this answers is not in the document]
acruikshank: @ZenGround0 This ToAddr is the address that may have been translated from an actor address to an ID address. The ID address is not represented in the message anywhere. I could have modified the To address of the message, but I'm not totally sure what's going to be looking at that, so I'd rather not have the mutation. Hence the added field.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -48,7 +52,9 @@ type VMContext struct {
...
 	Message     *types.UnsignedMessage
+	OriginMsg   *types.UnsignedMessage
```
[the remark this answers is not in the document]
acruikshank: @ZenGround0 `Message` and `OriginMsg` are the same for messages that originate from outside the VM. When one of these messages trigger an internal message, The OriginMessage will still be the message from outside the VM. We need this to create a stable actor address. If we use the the message 'From` and `Nonce` from an internal message, storage market requests to the init actor will always generate the same address.

internal/pkg/vm/internal/vmcontext/vmcontext.go
acruikshank: @ZenGround0 Because we are generating the VM context in two different places. I can't answer why we're doing that (history is gone). I rather we not do it in two places, but that's not a refactor I want to include in this PR.

internal/pkg/vm/internal/vmcontext/vmcontext.go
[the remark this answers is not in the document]
ZenGround0: Sorry for any confusion.  I agree this was the right thing to do, I was just ranting that it hadn't been done before and to all the data we need from the actual message.  It seems bad to actually bring the message along.

Issue #3653, opened 2019-11-12
Genesis block contains a timestamp


frrist:
### Description
Inorder for go-filecoin to further align with the spec the genesis block should contain a timestamp so that a nodes system clock may be initialized with a GenesisTime (see [spec definition](https://filecoin-project.github.io/specs/#systems__filecoin_nodes__clock))

### Acceptance criteria
- Add a timestamp to the genesis block produced by go-filecoins default genesis function as well as the gengen tool 

### Risks + pitfalls
- Some tests/code in go-filecoin expect that genesis block creation is deterministic, time breaks these assumptions
- current block validation code has special handling for the genesis block, it may need to be adjusted to accommodate the added timestamp.

frrist:
closed via https://github.com/filecoin-project/go-filecoin/pull/3655

PR #3655, opened 2019-11-13
Add timestamp to genesis block

frrist:
### Motivation
Genesis block requires a timestamp field so that a GenesisTime value may be derived as defined in the [clock section of the spec](https://filecoin-project.github.io/specs/#systems__filecoin_nodes__clock). Further motivation may also be found here: https://github.com/filecoin-project/go-filecoin/issues/3653

### Proposed changes
Modify `GenGenesisCar`, `GenGen`, `MustGenerateGenesis`, and genesis Config to accept a timestamp for the genesis block.

When a genesis block is required for a tests its timestamp is set to `time.Unix(123456789, 0)` for consistency

When calling the gengen binary, time is accepted as a command line argument (`date +%s`) defaulting to 123456789 if no value is given for the flag.

### Alternatives
Instead of modifying the aforementioned methods to accept a timestamp as an argument the `GenesisCfg` structure could be modified to contain the value. I decided against this since the `GenesisCfg` values have file based configuration and requiring a configuration file change in order to change the timestamp in the genesis block looked ugly.

Closes issue #3653

### DIFF
--- a/internal/pkg/testhelpers/iptbtester/genesis.go
+++ b/internal/pkg/testhelpers/iptbtester/genesis.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/filecoin-project/go-filecoin/cmd/go-filecoin"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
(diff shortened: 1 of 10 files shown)


internal/pkg/testhelpers/iptbtester/genesis.go
```diff
@@ -27,7 +28,7 @@ type idResult struct {
 
 // MustGenerateGenesis constructs the requires information and files to build a single
```
ZenGround0: old nit: required information

internal/pkg/testhelpers/iptbtester/genesis.go
```diff
@@ -27,7 +28,7 @@ type idResult struct {
...
 // filecoin node with the provided funds. The GenesisInfo can be used with MustImportGenesisMiner
-func MustGenerateGenesis(t *testing.T, funds int64, dir string) *GenesisInfo {
+func MustGenerateGenesis(t *testing.T, funds int64, dir string, genesisTime time.Time) *GenesisInfo {
```
ZenGround0: another old nit: since you are touching all of these calls consider renaming the function to `RequireGenerateGenesis` to follow our convention: require means exit with t.Fatal in case of error, must means panic in case of error.

frrist: done

PR #3658, opened 2019-11-14
feat: new vm runtime

icorderi:
### Motivation

Spec changed. We need to make several changes to get closer to the spec.
We have an opportunity to make some refactors.

(see https://github.com/filecoin-project/go-filecoin/pull/3636 for conv with the spec team)

### Proposed changes

- split the existing Runtime into two abstractions a `Runtime` which represents the VM and an `InvocationContext` that is tied to the current executing method.
- New `ActorStateHandle`, still in conv with spec on some details. 

### Not changed
- actors are not using new `ActorStateHandle` to retrieve state, followup PR 
- VMContext implementation is still one monolith, followup PR 
- `Storage` visibility is not clear until new actors pseudocode lands.  
- Left gas, message, and address calls in place until we delete the actors using them.

### DIFF
--- a/internal/pkg/vm/actor/builtin/power/power.go
+++ b/internal/pkg/vm/actor/builtin/power/power.go
@@ -183,7 +183,7 @@ func (*impl) createStorageMiner(vmctx invocationContext, ownerAddr, workerAddr a
 
 		// Update power table.
 		ctx := context.Background()
-		newPowerTable, err := actor.WithLookup(ctx,  vmctx.Storage(), state.PowerTable, func(lookup storage.Lookup) error {
+		newPowerTable, err := actor.WithLookup(ctx, vmctx.Storage(), state.PowerTable, func(lookup storage.Lookup) error {
 			// Do not overwrite table entry if it already exists
 			err := lookup.Find(ctx, addr.String(), nil)
 			if err != hamt.ErrNotFound { // we expect to not find the power table entry
@@ -229,7 +229,7 @@ func (*impl) removeStorageMiner(vmctx invocationContext, delAddr address.Address
 	var state State
 	_, err := actor.WithState(vmctx, &state, func() (interface{}, error) {
 		ctx := context.Background()
-		newPowerTable, err := actor.WithLookup(ctx,  vmctx.Storage(), state.PowerTable, func(lookup storage.Lookup) error {
+		newPowerTable, err := actor.WithLookup(ctx, vmctx.Storage(), state.PowerTable, func(lookup storage.Lookup) error {
 			// Find entry to delete.
 			var delEntry TableEntry
 			err := lookup.Find(ctx, delAddr.String(), &delEntry)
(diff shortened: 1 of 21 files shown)


internal/pkg/vm/internal/vmcontext/actor_state_handle_test.go
```diff
@@ -0,0 +1,134 @@
...
+	})
+
+	t.Run("abort when calling update release without taking", func(t *testing.T) {
+		defer mustPanic(t)
+
+		ts := setup()
+
+		ts.h.UpdateRelease("fake")
+	})
+}
```
anorth: Thanks for the tests!

internal/pkg/vm/actor/builtin/power/power.go
```diff
@@ -150,8 +150,16 @@ var Errors = map[uint8]error{
 	ErrDuplicateEntry:       errors.NewCodedRevertError(ErrDuplicateEntry, "duplicate create power table entry attempt"),
 }
 
+// Dragons: shouldnt this be calling the initacor to create an actor?
```
ZenGround0: I'm not sure it even should be in charge of creating a miner actor.  It seems like the miner actor's constructor should call a `createEntry` method on the power actor instead: https://github.com/filecoin-project/specs/issues/635

If miner creation does stay here then yes, it should be calling init (the PR introducing power actor landed before init actually did actor creation)

internal/pkg/vm/internal/exitcode/exitcode.go
```diff
@@ -0,0 +1,65 @@
...
+	MethodPanic = ExitCode(9)
+
+	// MethodSubcallError is returned when an actor method's Send call has
+	// returned with a failure error code (and the Send call did not specify
+	// to ignore errors).
+	MethodSubcallError = ExitCode(10)
+
+	// Runtime failed to encode the object.
+	// Review: for @spec, @go-filecoin, added this error
+	EncodingError = ExitCode(11)
```
ZenGround0: This looks like it is as worthy of its own code as `IOFailure`.  Did you consider adding that one in too?

internal/pkg/vm/internal/result/result.go
```diff
@@ -0,0 +1,38 @@
...
+func Success() InvocOutput {
+	return InvocOutput{ExitCode: 0}
+}
+
+// Value returns a successful code with the value encoded.
+//
+// Callers do NOT need to encode the value before calling this method.
+// Review: for @spec, @go-filecoin, changed this to an interface so the caller doesnt have to do the serialization himself
+func Value(obj interface{}) InvocOutput {
+	aux, err := encoding.Encode(obj)
```
ZenGround0: Since we're not going through the specialty abi encodings is this implicitly getting rid of the abi?  Or do all of those encodings simply wrap `encoding` anyway?

PR #3659, opened 2019-11-14
feat: new actor state handle

icorderi:
### Motivation

We need to use the new contracts from #3658

### Proposed changes

We have options.. lets talk.

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/actor_state_handle.go
+++ b/internal/pkg/vm/internal/vmcontext/actor_state_handle.go
@@ -74,7 +74,7 @@ type transactionFn = func() (interface{}, error)
 // `Transaction(F) -> (T, Error) where F: Fn(S) -> (T, error), S: ActorState`.
 //
 // Note: the actual Go signature is a bit different due to the lack of type system magic,
-//       and also wanting to avoid some unnecesary reflection.
+//       and also wanting to avoid some unnecessary reflection.
 //
 // Review: we might want to spend an hour or four making the signature look like it's supposed to..
 // Hack: In order to know `S` and save some code, the actual signature looks like:
--- a/internal/pkg/vm/internal/vmcontext/actor_state_handle.go
+++ b/internal/pkg/vm/internal/vmcontext/actor_state_handle.go
@@ -16,7 +16,7 @@ type actorStateHandle struct {
 	// used_obj holds the pointer to the obj that has been used with this handle.
 	//
 	// any subsequent calls needs to be using the same variable.
-	used_obj interface{}
+	usedObj interface{}
 }
 
 // validateFn returns True if it's valid.
@@ -48,9 +48,9 @@ var _ runtime.ActorStateHandle = (*actorStateHandle)(nil)
 // Readonly is the implementation of the ActorStateHandle interface.
 func (h *actorStateHandle) Readonly(obj interface{}) {
 	// track the variable used by the caller
-	if h.used_obj == nil {
-		h.used_obj = obj
-	} else if h.used_obj != obj {
+	if h.usedObj == nil {
+		h.usedObj = obj
+	} else if h.usedObj != obj {
 		runtime.Abort("Must use the same state variable on repeated calls")
 	}
 
@@ -74,9 +74,9 @@ func (h *actorStateHandle) Transaction(obj interface{}, f transactionFn) (interf
 	}
 
 	// track the variable used by the caller
-	if h.used_obj == nil {
-		h.used_obj = obj
-	} else if h.used_obj != obj {
+	if h.usedObj == nil {
+		h.usedObj = obj
+	} else if h.usedObj != obj {
 		runtime.Abort("Must use the same state variable on repeated calls")
 	}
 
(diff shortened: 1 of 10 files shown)


anorth:
I like your proposal. It looks like it should be functionally equivalent to the externally-observable behaviour the spec is aiming for. This PR is obviously missing implementation details of how those mechanics are achieved, but if achievable 👍 .

The spec also aims to prevent message send during what would be the protected regions. I think this could be added to your proposed semantics.

internal/pkg/vm/testing.go
```diff
@@ -213,6 +229,9 @@ func (ts testStorage) Commit(cid.Cid, cid.Cid) error {
 
 // Head returns an empty Cid to satisfy the Storage interface
 func (ts testStorage) Head() cid.Cid {
+	if ts.state == nil {
+		return cid.Undef
+	}
```
icorderi: @acruikshank any idea how some tests were passing before? 
`encoding.Encode({{ nil }}.state)` was blowing up during initactor tests.

acruikshank: Which tests weren't passing with what changes? I don't know how to answer this question.

internal/pkg/vm/internal/vmcontext/actor_state_handle.go
```diff
@@ -9,62 +9,110 @@ import (
...
+	// call user code allowing mutation but not side-effects
+	h.ctx.AllowSideEffects(false)
+	out, err := f()
+	h.ctx.AllowSideEffects(true)
+
+	// forward user error
+	if err != nil {
+		// re-load state from storage
+		// Note: this will throw away any changes done to the object
+		h.get(oldcid, obj)
```
anorth: I don't think this is correct. Prior to this function call, obj was some arbitrary object pointer, probably nil. This does not restore it to its old state, but to a different state (equiv to Readonly). A correct rollback would be something like a deepcopy, but I don't think it's worth it. I propose instead that the obj will be left as the transaction left it, but won't be written to storage or have the head updated.

internal/pkg/vm/internal/vmcontext/actor_state_handle.go
```diff
@@ -77,35 +125,57 @@ func (h *actorStateHandle) UpdateRelease(obj interface{}) {
...
 	h.head = &newcid
+
+	// add validation
+	h.addCidValidation(obj, newcid)
+
+	return out, nil
+}
+
+func (h *actorStateHandle) addCidValidation(obj interface{}, expected cid.Cid) {
+	// checks that the cid of the obj has not changed
```
anorth:
```suggestion
	// checks that the cid of the obj matches an expected value
```

internal/pkg/vm/internal/vmcontext/actor_state_handle.go
```diff
@@ -77,35 +125,57 @@ func (h *actorStateHandle) UpdateRelease(obj interface{}) {
 		runtime.Abort("Storage commit error")
 	}
 
-	// make new head available to take
+	// update head
 	h.head = &newcid
+
+	// add validation
+	h.addCidValidation(obj, newcid)
```
anorth: A list of validation fns is not the right semantics here. If a method contains two well-formed transactions in sequence then the validation fn from the first one will not pass after the second. I think we want semantics of a single expected CID which is updated with each transaction.

internal/pkg/vm/internal/vmcontext/actor_state_handle.go
```diff
@@ -77,35 +125,57 @@ func (h *actorStateHandle) UpdateRelease(obj interface{}) {
...
+		if !vfn() {
+			runtime.Abort("State mutated outside of Transaction() scope")
+		}
 	}
+}
 
+// Dragons: move this to the storage API
+func (h *actorStateHandle) cidOf(obj interface{}) cid.Cid {
+	// get Cid for object
+	// Dragons: this should not be storing
```
anorth: This is easy enough to fix now. Factor the `if` block out of `storage.Put()` into an exported `storage.ToNode(interface{})` and call that.

internal/pkg/vm/internal/vmcontext/actor_state_handle_test.go
```diff
@@ -11,120 +13,241 @@ import (
...
+		var out testActorStateHandleState
+
+		ts.h.Transaction(&out, func() (interface{}, error) {
+			out.FieldA = "changed!"
+			return nil, nil
+		})
+
+		out.FieldA = "changed again!"
 	})
+}
```
anorth: Add a test for sequential transactions, and for readonly followed by transaction.

internal/pkg/vm/testing.go
anorth: Update the comment on line 230

icorderi: `builtin/miner/miner_test.go:184` the `TestConstructor` added a ~week ago was not passing now if didn't add that `nil` check.

internal/pkg/vm/internal/vmcontext/actor_state_handle.go
```diff
@@ -9,62 +9,110 @@ import (
...
 // NewActorStateHandle returns a new `actorStateHandle`
 func NewActorStateHandle(ctx actorStateHandleContext, head cid.Cid) runtime.ActorStateHandle {
-	return &actorStateHandle{
-		ctx:  ctx,
-		head: &head,
+	aux := newActorStateHandle(ctx, head)
+	return &aux
+}
+
+func newActorStateHandle(ctx actorStateHandleContext, head cid.Cid) actorStateHandle {
```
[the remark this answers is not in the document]
icorderi: resolved offline

Issue #3664, opened 2019-11-20
Finality prefactor: staging head


ZenGround0:
### Description
In order to handle finality properly the syncer needs to track intermediate state of heads that have been synced and deemed heaviest but cannot be exposed to the user.

To achieve this we can add a staging field to the syncer and set this during our update of the heaviest seen tipset.

### Acceptance criteria
syncer field tracks heaviest tipset independent of the chain store

### Risks + pitfalls
Initialization of the staging field requires jumping through some hoops because it can only happen after the chain store has been properly loaded.

PR #3665, opened 2019-11-20
Finality refactor

ZenGround0:
### Motivation
Prefactor for finality that allows the chain store update to be deferred until a safe chain syncing state.

### Proposed changes
- Extract out heaviest chain check / head setting
- Add staging field to syncer

Closes issue #3664

### DIFF
--- a/internal/pkg/chainsync/internal/syncer/syncer_test.go
+++ b/internal/pkg/chainsync/internal/syncer/syncer_test.go
@@ -2,6 +2,7 @@ package syncer_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


internal/pkg/chainsync/internal/syncer/syncer_test.go
```diff
@@ -457,7 +460,7 @@ func TestSyncerStatus(t *testing.T) {
 
 	// advance the chain head, ensure status changes
 	t2 := builder.AppendOn(t1, 1)
-	require.NoError(t, syncer.HandleNewTipSet(ctx, block.NewChainInfo(peer.ID(""), "", t2.Key(), heightFromTip(t, t2)), false))
+	require.NoError(t, syncer.HandleNewTipSet(ctx, block.NewChainInfo(peer.ID(""), "", t2.Key(), heightFromTip(t, t2))))
```
acruikshank: I find it a little disturbing that altering the `stageIfHeaviest` logic hasn't broken any tests. Does the fact that these tests still pass despite the fact we are not passing in the `trusted=true` flag indicate that the new code is exercised and working properly?

ZenGround0: I removed the status report on trustedness in order to get this to pass.  Trustedness of peers is a deprecated concept so I decided to throw it out here while doing other cleanup.

The stageIfHeaviest logic is adding in a hidden side effect to the syncer that is not yet consumed so it is expected that tests don't fail.

Issue #3652, opened 2019-11-12
update system clock to match spec


frrist:
### Description
The Filecoin spec contains a [definition for a clock](https://filecoin-project.github.io/specs/#systems__filecoin_nodes__clock), it is used for synchrony amongst participants in the filecoin network. The [go-filecoin implementation of the clock](https://github.com/filecoin-project/go-filecoin/tree/master/internal/pkg/clock) should be updated to better align with the spec.

### Acceptance criteria
- Clock receives a `GenesisTime` value from the genesis block when a filecoin node is initialized. 
- Add `EpochAtTime(t Time)` method to go-filecoin clock. `EpochAtTime` should return the ChainEpoch corresponding to time t, where `epoch ~= (current_time - genesis_time) / epoch_time`

anorth:
Please distinguish generic clock functionality (that we already have) from the epoch clock, so that code requiring only access to wall-clock time is not polluted with unwanted information about epochs. You can likely achieve this pretty easily by wrapping the existing clock in a new thingy that also includes the genesis time.

PR #3669, opened 2019-11-21
Implement ChainEpochClock similar to spec definition 

frrist:
### Motivation
[Spec Clock](https://filecoin-project.github.io/specs/#systems__filecoin_nodes__clock)

### Proposed changes
- Implement the ChainEpoch clock on the Node as described in the spec exposing it as an interface wrapping the existing System Clock.
- Pass the ChainEpoch clock to the syncer module as it will be used during block validation later.

### Whats Left?
- [x] Unit Tests ChainEpochClock
- [x]  Get clarification on  filecoin-project/specs#653

Closes issue #3652

### DIFF
--- a/cmd/go-filecoin/daemon.go
+++ b/cmd/go-filecoin/daemon.go
@@ -20,8 +20,8 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node"
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/paths"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/config"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/journal"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/repo"
 )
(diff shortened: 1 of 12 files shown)


cmd/go-filecoin/daemon.go
```diff
@@ -95,7 +94,6 @@ func daemonRun(req *cmds.Request, re cmds.ResponseEmitter) error {
 		return errors.Wrap(err, "Bad block time passed")
 	}
 	opts = append(opts, node.BlockTime(blockTime))
-	opts = append(opts, node.ClockConfigOption(clock.NewSystemClock()))
```
frrist: This is now done in the node builder since we need the genesis timestamp to init the chain clock.

internal/pkg/chain/store.go
```diff
@@ -280,6 +280,11 @@ func (store *Store) GetGenesisState(ctx context.Context) (state.Tree, error) {
 	return store.stateTreeLoader.LoadStateTree(ctx, store.stateAndBlockSource.cborStore, genesis.StateRoot)
 }
 
+// GetGenesisBlock returns the genesis block held by the chain store.
+func (store *Store) GetGenesisBlock(ctx context.Context) (*block.Block, error) {
```
frrist: I think this is much cleaner than using the `GetTipSet` method. Any objections?

internal/pkg/clock/chainclock.go
```diff
@@ -0,0 +1,42 @@
+package clock
```
frrist: Spec for reference: https://filecoin-project.github.io/specs/#systems__filecoin_nodes__clock

internal/pkg/clock/chainclock.go
```diff
@@ -0,0 +1,42 @@
...
+type ChainEpochClock interface {
+	EpochAtTime(t time.Time) ChainEpoch
+	Clock
+}
+
+// chainClock is a clock that represents epochs of the protocol.
+type chainClock struct {
+	// GenesisTime is the time of the first block. EpochClock counts
+	// up from there.
+	GenesisTime time.Time
```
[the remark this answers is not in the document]
frrist: Filed issue here: https://github.com/filecoin-project/specs/issues/653. Agreed this seems unnecessary.

Issue #3666, opened 2019-11-20
Syncer catchup and follow


ZenGround0:
### Description
The syncer should be able to detect catchup to follow and follow to catchup state transitions and modify the state externalized by the chain store accordingly.

The head should not be set during catchup.

Catchup should occur once all gaps are closed

Follow should occur whenever gaps exist

anorth:
@ZenGround0 done?

PR #3670, opened 2019-11-26
Catchup Follow transitions

ZenGround0:
### Motivation
The syncer needs to be able to transition between catchup and follow states based on gaps in tipset epochs and only set chain heads during the follow state.

### Proposed changes
* Introduce a transitioner type and new dispatcher data and logic calling it to manage chain sync transitions 
* Reintroduce the boolean into HandleNewTipSet so that the caller can differentiate modes to control head setting
* Expose a method on the syncer for directly setting the staged head to call during transitions 
* Cleanup syncer by removing unused lock (all calls to syncer are serialized through the dispatcher's event loop)

Closes issue #3666

### Testing
Testing this is going to be quite challenging.  I'm working on something I'd like to include in this PR but wanted to surface this work first.  In the interest of moving faster I might defer the tests until a later PR.

Advice on how to test is appreciated.  Right now I'm planning to build out calls on the node to allow for in process integration tests.  Likely this is going to require evaluating and doing useful but non-trivial testing-specific development such as
* improve node mining interface / tools so that it is simple to setup the chains on a set of nodes exercising the catchup to follow transition
* remove race in mine once by providing a blocking call back to the dispatcher
* add blocking sync call directly to the node interface 
* add state dump to dispatcher's event loop

There is also some basic unit testing I should do in the syncer and new transitioner.

### DIFF
--- a/internal/pkg/chainsync/internal/syncer/syncer.go
+++ b/internal/pkg/chainsync/internal/syncer/syncer.go
@@ -2,7 +2,7 @@ package syncer
 
 import (
 	"context"
-	
+
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/chain"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/chainsync/status"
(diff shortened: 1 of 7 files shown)


internal/pkg/chainsync/internal/syncer/syncer.go
```diff
@@ -425,22 +419,28 @@ func (syncer *Syncer) widen(ctx context.Context, ts block.TipSet) (block.TipSet,
...
+	if err != nil {
+		return err
+	}
+	if catchup {
+		return nil
+	}
+	return syncer.SetStagedHead(ctx)
+}
+
+func (syncer *Syncer) handleNewTipSet(ctx context.Context, ci *block.ChainInfo) (err error) {
```
acruikshank: Like that you're not making this method even bigger, but refactoring into nested function calls isn't the best solution for breaking code out. It would be nice if this could get an overall refactor, but it's not necessary for this PR.

PR #3671, opened 2019-11-27
add Git tool version require

marco-storswift:
### Motivation
git tools does not support git rev-parse @:"${__submodule_path} if git version less v2.16.5 , you can get error like this
failed to find or obtain precompiled assets for bls-signatures, falling back to local build
fatal: Invalid object name '@'.
make: *** [.install-bls] ERROR 128
Command './scripts/install-go-bls-sigs.sh' failed: exit status 1
fatal: Invalid object name '@'.
### Proposed changes

Closes issue #

git tools does not support git rev-parse @:"${__submodule_path} if git version less v2.16.5 , you can get error like this:
failed to find or obtain precompiled assets for bls-signatures, falling back to local build
fatal: Invalid object name '@'.
make: *** [.install-bls] ERROR 128
Command './scripts/install-go-bls-sigs.sh' failed: exit status 1
fatal: Invalid object name '@'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,15 @@ Now install the tools and dependencies listed below. If you have **any problems 
 #### Install Go and Rust
 
 The build process for go-filecoin requires:
-
+- [Git](https://git-scm.com) >= v2.16.5
+  - git tools does not support git rev-parse @:"${__submodule_path} if git version less v2.16.5 , you can get error like this 
+```sh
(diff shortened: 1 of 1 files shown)


ZenGround0:
This is helpful to know, but closing since 2.16 is 6 minor versions into not being maintained so it is acceptable to leave figuring this out up to the user.

PR #3673, opened 2019-11-28
feat: match latest vm spec PR

icorderi:
### Motivation

Spec changed, we change. 

### Proposed changes

- using implementer discretion on:
    * Send signature
    * Grouping of exposed methods

### DIFF
--- a/internal/pkg/vm/actor/builtin/initactor/init.go
+++ b/internal/pkg/vm/actor/builtin/initactor/init.go
@@ -12,7 +12,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/errors"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/dispatch"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/errors"
+	internal "github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/errors"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/runtime"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-hamt-ipld"
@@ -126,7 +126,7 @@ type invocationContext interface {
 	runtime.InvocationContext
 	AddressForNewActor() (address.Address, error)
 	CreateNewActor(addr address.Address, code cid.Cid) error
-	Message() *types.UnsignedMessage
+	OriginalMessage() *types.UnsignedMessage
 }
 
 // Impl is the VM implementation of the actor.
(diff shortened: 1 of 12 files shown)


internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -270,7 +270,7 @@ func (a *Impl) Exec(rt invocationContext, codeCID cid.Cid, params []interface{})
 	}
 
 	// send message containing actor's initial balance to construct it with the given params
-	_, _, err = rt.Runtime().Send(idAddr, types.ConstructorMethodID, rt.Message().Value, params)
+	_, _, err = rt.Runtime().Send(idAddr, types.ConstructorMethodID, rt.OriginalMessage().Value, params)
```
anorth: You can change this to rt.ValueTransferred() and remove the message.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -126,7 +126,7 @@ type invocationContext interface {
 	runtime.InvocationContext
 	AddressForNewActor() (address.Address, error)
 	CreateNewActor(addr address.Address, code cid.Cid) error
-	Message() *types.UnsignedMessage
+	OriginalMessage() *types.UnsignedMessage
```
anorth: As discussed, let's say `LegacyMessage`. This is the immediate message, not the top-level one.

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -470,7 +470,7 @@ func (*Impl) AddAsk(ctx invocationContext, price types.AttoFIL, expiry *big.Int)
 
 	var state State
 	out, err := actor.WithState(ctx, &state, func() (interface{}, error) {
-		if ctx.Message().From != state.Worker {
+		if ctx.OriginalMessage().From != state.Worker {
```
anorth: You can change this to ctx.Caller and remove the message. Same goes for all the other uses of From in this file (and everywhere else).

internal/pkg/vm/actor/builtin/miner/miner.go
```diff
@@ -730,7 +730,7 @@ func (a *Impl) CommitSector(ctx invocationContext, sectorID uint64, commD, commR
 			copy(req.CommR[:], commR)
 			copy(req.CommRStar[:], commRStar)
 			req.Proof = proof
-			req.ProverID = sectorbuilder.AddressToProverID(ctx.Message().To)
+			req.ProverID = sectorbuilder.AddressToProverID(ctx.OriginalMessage().To)
```
anorth: FYI I'm adding `Receiver` to the spec, which can replace this use of message

icorderi: 👍 thanks!

PR #3674, opened 2019-11-29
refactor: remove LegacyCreateNewActor

icorderi:


internal/pkg/vm/internal/runtime/runtime.go
```diff
@@ -68,8 +68,12 @@ type InvocationContext interface {
 // and ensure the context implementation exposes them.
 type ExtendedInvocationContext interface {
 	InvocationContext
-	// Dragons: add new CreateActor (this is just for the init actor)
-	// CreateActor(code cid.Cid, params []interface{}) address.Address
+	// Create an actor in the state tree.
+	//
+	// This will determine an address for the actor and call its `Constructor()` method.
+	//
+	// WARNING: May only be called by InitActor.
```
anorth: Document which address is returned, the id-address, or the reorg-stable address. Maybe we should return both.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -252,38 +252,80 @@ func (ctx *VMContext) Charge(cost types.GasUnits) error {
...
-// LegacyCreateNewActor creates an actor at the given address.
-// If the address is occupied by a non-empty actor, this method will fail.
-func (ctx *VMContext) LegacyCreateNewActor(addr address.Address, code cid.Cid) error {
+// CreateActor implemenets the ExtendedInvocationContext interface.
+func (ctx *VMContext) CreateActor(actorID types.Uint64, code cid.Cid, params []interface{}) address.Address {
+	if !isBuiltinActor(code) {
+		runtime.Abort("Can only create built-in actors.")
+	}
+
+	// Review: why are we checking this here? it assumes singleton actors always exist
```
anorth: Yes, they are assumed to exist in the genesis state, not by the application of a message that could somehow invoke this method.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -252,38 +252,80 @@ func (ctx *VMContext) Charge(cost types.GasUnits) error {
...
-		return errors.FaultErrorWrap(err, "Error retrieving or creating actor")
+		runtime.Abort("Could not get or create actor")
 	}
 
 	if !newActor.Empty() {
-		return errors.NewRevertErrorf("attempt to create actor at address %s but a non-empty actor is already installed", addr.String())
+		runtime.Abort("Actor address already exists")
 	}
 
+	// Review: how does this work? we are modifying the value without commiting chages to the store
```
anorth: I think this may be an ugly reach-around because we have a pointer to the newActor state struct that is rightly owned by the state tree.

internal/pkg/vm/internal/runtime/runtime.go
icorderi: Good point.

Issue #3681, opened 2019-12-02
Update VM message execution to match spec


icorderi:
- [x] state rollback when a send returns non-zero exit code
- [ ] check interpreter behaviour against spec (miner penalty, gas tracking etc)

PR #3699, opened 2019-12-04
feat: new Send api to match spec

icorderi:
### Motivation

The new spec defines methods to result either in a success with a value returned or they Abort. 

### Proposed changes

- This PR renames the old Send to `LegacySend` while
- Adds a new `Send` with a single value being returned. 

See #3681.

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/vmcontext.go
+++ b/internal/pkg/vm/internal/vmcontext/vmcontext.go
@@ -116,8 +116,8 @@ func (ctx *VMContext) Randomness(epoch types.BlockHeight, offset uint64) runtime
 	return rnd
 }
 
-// Send allows actors to invoke methods on other actors
-func (ctx *VMContext) Send(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error) {
+// LegacySend allows actors to invoke methods on other actors
+func (ctx *VMContext) LegacySend(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) ([][]byte, uint8, error) {
 	// check if side-effects are allowed
 	if !ctx.allowSideEffects {
 		runtime.Abort("Calling Send() is not allowed during side-effet lock")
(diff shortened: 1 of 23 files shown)


internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -306,14 +423,16 @@ func (ctx *VMContext) CreateActor(actorID types.Uint64, code cid.Cid, params []i
...
-	if err != nil {
-		runtime.Abort("Constructor failed on actor")
-	}
+	ctx.Runtime().Send(idAddr, types.ConstructorMethodID, ctx.Message().ValueReceived(), params)
 
 	return actorAddr
 }
 
+// VerifySignature implemenets the ExtendedInvocationContext interface.
+func (*VMContext) VerifySignature(signer address.Address, signature types.Signature, msg []byte) bool {
```
acruikshank: This method appears to be unused. It doesn't seem to belong in this PR. It should either be tested or removed.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -184,6 +185,122 @@ func (ctx *VMContext) Send(to address.Address, method types.MethodID, value type
 	return out, ret, nil
 }
 
+// Send allows actors to invoke methods on other actors
+func (ctx *VMContext) Send(to address.Address, method types.MethodID, value types.AttoFIL, params []interface{}) interface{} {
```
acruikshank: This method is mostly the same as `LegacySend`. You should extract that code into a new function rather than copying it. This will preserve git history when `LegacySend` goes away.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -184,6 +185,122 @@ func (ctx *VMContext) Send(to address.Address, method types.MethodID, value type
...
+		BlockHeight: ctx.blockHeight,
+		Ancestors:   ctx.ancestors,
+		Actors:      ctx.actors,
+	}
+	innerCtx := NewVMContext(innerParams)
+
+	return deps.Apply(innerCtx)
+}
+
+func apply(ctx *VMContext) interface{} {
```
acruikshank: This is mostly a duplicate of `send`. It should either replace it in this PR, one of `apply` or `send` should call the other, or the common code should be extracted into a function.

PR #3700, opened 2019-12-06
ID addresses for all actors

acruikshank:
### Motivation

In order to reduce bandwidth, all actors should be keyed in the state tree by id address, and messages should use id addresses whenever possible. We already store miner actors by id. Since signed messages are always sent from account actors, we need account actors to store a public key based address in state, so that the signatures of messages sent from id addresses can be verified. All our code that creates account actors (either by sending a message to an unoccupied address or during genesis) needs to do so by interacting with the Init actor.

### Proposed changes

1. Modify all code that produces account actors to do so by calling `Init.Exec`.
2. Modify vmcontext.CreateActor so that the actorAddress will be the public key address passed in as a parameter rather than a generated address. The actorAddress is the address mapped to the id address in `InitActor` state.


internal/pkg/vm/actor/builtin/account/account.go
```diff
@@ -22,15 +30,21 @@ func NewActor(balance types.AttoFIL) (*actor.Actor, error) {
 	return actor.NewActor(types.AccountActorCodeCid, balance), nil
 }
 
-// UpgradeActor converts the given actor to an account actor, leaving its balance and nonce in place.
```
ZenGround0: So this is going away?  I never quite understood the security implications but did we change our approach to handle the malicious-pre-creation behavior motivating this?

internal/pkg/vm/actor/builtin/account/account.go
```diff
@@ -39,14 +53,58 @@ func UpgradeActor(a *actor.Actor) error {
...
 // InitializeState for account actors does nothing.
-func (*Actor) InitializeState(_ runtime.Storage, _ interface{}) error {
-	return nil
+func (*Actor) InitializeState(storage runtime.Storage,  initializerData interface{}) error {
+	state, ok := initializerData.(*State)
+	if !ok {
+		return errors.NewFaultError("Initial state to account actor is not a account.State struct")
+	}
+
+	if state.Address.Protocol() != address.SECP256K1 && state.Address.Protocol() != address.BLS {
```
ZenGround0: It seems off that we create new account actors with `NewActor` that don't satisfy this condition in consensus/genesis.  Their states will always need to be uninitialized.  It seems like these things shouldn't even be considered account actors.

internal/pkg/vm/actor/builtin/account/account.go
```diff
@@ -39,14 +53,58 @@ func UpgradeActor(a *actor.Actor) error {
...
+
+//
+// vm methods for actor
+//
+
+// Impl is the VM implementation of the actor.
+type Impl Actor
+
+// Constructor initializes the actor's state
+func (impl *Impl) Constructor(ctx runtime.InvocationContext, addr address.Address) (uint8, error) {
```
ZenGround0: Having trouble finding where this gets called

internal/pkg/vm/actor/builtin/storagemarket/storagemarket_test.go
```diff
@@ -68,32 +68,16 @@ func TestStorageMarketCreateStorageMinerDoesNotOverwriteActorBalance(t *testing.
...
-	createdAddress, err := address.NewFromBytes(result.Receipt.Return[0])
-	require.NoError(t, err)
-	assert.Equal(t, derivedAddr, createdAddress)
-	miner, err := st.GetActor(ctx, minerAddr)
-	require.NoError(t, err)
+	_, err = th.ApplyTestMessage(st, vms, msg, types.NewBlockHeight(0))
 
-	// miner balance should be sum of messages
-	assert.Equal(t, types.NewAttoFILFromFIL(300).String(), miner.Balance.String())
+	// message fails because we do not create actors when message to is an ID address.
```
ZenGround0: Is this true with the given code?  I can't find the account actor init state function (which I think is supposed to raise the error) getting called from the `getOrCreateActor` call.

internal/pkg/vm/internal/vmcontext/vmcontext_test.go
```diff
@@ -482,3 +494,20 @@ func TestTransfer(t *testing.T) {
 		assert.EqualError(t, Transfer(actor2, actor3, negval), "cannot transfer negative values")
 	})
 }
+
+func mustCreateInitActor(t *testing.T, bs blockstore.Blockstore) *actor.Actor {
```
ZenGround0: nit `require` instead of `must` because we are not panicing

tools/gengen/util/gengen.go
```diff
@@ -123,7 +123,7 @@ func GenGen(ctx context.Context, cfg *GenesisCfg, cst *hamt.CborIpldStore, bs bl
 		return nil, err
 	}
 
-	if err := setupPrealloc(st, keys, cfg.PreAlloc); err != nil {
```
ZenGround0: We could have avoided lots of the trouble in the consensus/genesis function if we instead used one shared genesis state / block implementation.

The function options that modify the consensus.Config could all be modified to operate over a gengen config.  The interfaces are slightly different but IMO we'd be much much better off adopting usage of the other interface to work with gengen configs OR modifying gengen configs to be slightly more expressive.

Wrote this up here: https://github.com/filecoin-project/go-filecoin/issues/3701

internal/pkg/consensus/genesis.go
```diff
@@ -158,17 +159,36 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
...
+
+		// retrieve init actor so we can register addresses
+		cachedState := state.NewCachedTree(st)
+		init, err := cachedState.GetActor(ctx, address.InitAddress)
+		if err != nil {
+			return nil, err
+		}
+
 		// Initialize account actors
 		for addr, val := range genCfg.accounts {
```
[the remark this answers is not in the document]
acruikshank: Good catch. Done.

internal/pkg/vm/actor/builtin/account/account.go
acruikshank: The difference is that we no longer have "Empty" actors. You either send a message to an unknown address and it becomes an Account actor, or you call Exec (usually you call another actor's method that calls Exec) and create an actor of another type. Actors will always have a CID now, though.

internal/pkg/vm/actor/builtin/account/account.go
acruikshank: The network actor is the only Account actor that isn't derived from a public key address. It is added to the state tree directly. This condition should hold for all other actors.

internal/pkg/vm/actor/builtin/account/account.go
acruikshank: This gets called by `vmcontext.CreateActor`. It might not have been wired that way when you made this comment.

internal/pkg/consensus/processor.go
```diff
@@ -567,18 +592,26 @@ func NewDefaultBlockRewarder() *DefaultBlockRewarder {
 var _ BlockRewarder = (*DefaultBlockRewarder)(nil)
 
 // BlockReward transfers the block reward from the network actor to the miner's owner.
-func (br *DefaultBlockRewarder) BlockReward(ctx context.Context, st state.Tree, minerOwnerAddr address.Address) error {
+func (br *DefaultBlockRewarder) BlockReward(ctx context.Context, st state.Tree, vms vm.StorageMap, minerOwnerAddr address.Address) error {
 	cachedTree := state.NewCachedTree(st)
```
ZenGround0: Not related to this PR, but tracing through how rewards are calculated (some in ApplyMessage, some in ApplyMessageAndPayRewards) strengthens my impression that these functions have grown to become significantly more complicated than they inherently need to be.

internal/pkg/consensus/processor.go
```diff
@@ -592,15 +625,13 @@ func (br *DefaultBlockRewarder) BlockRewardAmount() types.AttoFIL {
...
+func rewardTransfer(ctx context.Context, fromAddr, toAddr address.Address, value types.AttoFIL, st *state.CachedTree, vms vm.StorageMap, gt *vm.GasTracker) error {
 	fromActor, err := st.GetActor(ctx, fromAddr)
 	if err != nil {
 		return errors.FaultErrorWrap(err, "could not retrieve from actor for reward transfer.")
 	}
 
-	toActor, err := st.GetOrCreateActor(ctx, toAddr, func() (*actor.Actor, error) {
-		return &actor.Actor{}, nil
-	})
+	toActor, _, err := getOrCreateActor(ctx, st, vms, toAddr, gt)
```
ZenGround0: Unrelated to this change, but we really should audit this behavior with a complete spec (I don't think spec has an opinion about this yet).  Right now I don't think it's possible to setup a miner with an owner that has not been created yet.  Maybe the protocol wants to be more lenient about this.  If not then this is a complicated piece of dead code among protocol compliant nodes.

internal/pkg/consensus/processor.go
```diff
@@ -645,3 +676,53 @@ func (p *DefaultProcessor) minerOwnerAddress(ctx context.Context, st state.Tree,
...
+
+	if addr.Protocol() != address.ID {
+		initAct, err := st.GetActor(ctx, address.InitAddress)
+		if err != nil {
+			return nil, address.Undef, err
+		}
+
+		// this should never fail due to lack of gas since gas doesn't have meaning here
+		noopGT := vm.NewGasTracker()
+		noopGT.MsgGasLimit = 10000
```
ZenGround0: nit: ok for now but this magic number carries lots of questions with it for someone like me who hasn't fully internalized gas processing rules.  An explanation pre-emptively answering some of these (is this always enough, is this some special protocol defined number, what about the gas price?) would be helpful.

internal/pkg/consensus/processor.go
```diff
@@ -645,3 +676,53 @@ func (p *DefaultProcessor) minerOwnerAddress(ctx context.Context, st state.Tree,
...
+	if err != nil {
+		return nil, address.Undef, err
+	}
+
+	if found {
+		act, err := st.GetActor(ctx, idAddr)
+		return act, idAddr, err
+	}
+
+	if addr.Protocol() != address.ID {
```
ZenGround0: This condition is always true given the current definition of `ResolveAddress`.  I think the whole if guard can go away

acruikshank: Good catch.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -152,8 +151,9 @@ func (ctx *VMContext) LegacySend(to address.Address, method types.MethodID, valu
 		return nil, 1, errors.FaultErrorWrapf(err, "failed to get id address for actor address")
 	}
 
-	toActor, err := deps.GetOrCreateActor(context.TODO(), toAddr, func() (*actor.Actor, error) {
-		return &actor.Actor{}, nil
+	// TODO: this needs to initialize in init actor state
```
ZenGround0: Looks like this TODO is out of date.  The caller must have been init exec which does this bookkeeping already.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -117,11 +117,47 @@ func (*Actor) InitializeState(storage runtime.Storage, params interface{}) error
 	return storage.LegacyCommit(id, cid.Undef)
 }
 
+//
+// public methods for actor
+//
+
+// InitializeAccountActor returns a function to be called when lookup in the state tree fails
+func XInitializeAccountActor(vmctx invocationContext, addr address.Address, balance types.AttoFIL) (*actor.Actor, address.Address, error) {
```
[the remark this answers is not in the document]
ZenGround0: Reminder from out of band conversation: this function and createMappedIDAddress are dead code and should be removed.

internal/pkg/testhelpers/core.go
```diff
@@ -42,29 +44,74 @@ func RequireMakeStateTree(t *testing.T, cst *hamt.CborIpldStore, acts map[addres
...
-func RequireNewAccountActor(t *testing.T, value types.AttoFIL) *actor.Actor {
-	act, err := account.NewActor(value)
+// RequireLookupActor converts the given address to an id address before looking up the actor in the state tree
+func RequireLookupActor(ctx context.Context, t *testing.T, st state.Tree, vms vm.StorageMap, actorAddr address.Address) (*actor.Actor, address.Address) {
+	if actorAddr.Protocol() == address.ID {
+		return state.MustGetActor(st, actorAddr), actorAddr
+	}
+
+	cachedTree := state.NewCachedTree(st)
+	initActor, _, err := cachedTree.GetOrCreateActor(ctx, address.InitAddress, func() (*actor.Actor, address.Address, error) {
```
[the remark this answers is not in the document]
acruikshank: This pattern allows us to sequence the test helpers without worrying about the order. An init actor is required for this to work (and in some tests it isn't present). We call these a lot so it saves some test setup to leave them in here.

internal/pkg/vm/actor/builtin/account/account.go
acruikshank: You probably could, but the miner is actually really going to be added at an ID address. The init actor will then map the new actor address, which will override the mapping of the message previously sent. So it works out in the end. We may want to limit account creation to Secp and BLS addresses.

internal/pkg/vm/actor/builtin/storagemarket/storagemarket_test.go
acruikshank: It's thrown from https://github.com/filecoin-project/go-filecoin/pull/3700/files#diff-b56a91c5cc23ee17bbf77b69af05eae5R689 when we resolve the address as an ID address and then attempt to get it from the state tree directly.

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -232,14 +247,14 @@ func (a *Impl) Exec(rt invocationContext, codeCID cid.Cid, params []interface{})
 
 		// map id to address and vice versa
 		ctx := context.TODO()
-		state.AddressMap, err = a.setId(ctx, rt.Runtime().Storage(), state.AddressMap, actorAddr, actorID)
+		state.AddressMap, err = setId(ctx, rt.Runtime().Storage(), state.AddressMap, actorAddr, actorID)
```
anorth: In the case of account actors, I don't think this is right. For account actors, we need to map the actor's pubkey address (i.e. params[0]) to the ID address. The `actorAddr` should not be used.

See https://github.com/filecoin-project/specs/issues/762 though before diving in to fix it now.

acruikshank: @anorth I made this work by having `rt.CreateActor` return the public key address in the case where the actor is an Account actor. There wasn't much guidance in the spec at the time. You're right though that nothing would work if this mapping didn't map to the public key address.

Issue #3677, opened 2019-12-02
Block propagation clock time synchrony, "ticket nonce" validation


icorderi:

anorth:
@ZenGround0 done?

PR #3704, opened 2019-12-11
Feat/mine on epoch boundaries

ZenGround0:
### Motivation
We need to produce blocks and schedule the work to do so with respect to a set of time rules.  This PR overhauls the mining scheduler to do just that.  It is a prerequisite for any timing rule validation because it creates correct timestamps.

### Proposed changes
- production of blocks on epoch aligned boundaries using the `ChainEpochClock` (thanks @frrist!)
- automatic shutdown of late mining jobs
- cleanup teardown logic
- introduce a `skipCh` for signaling that the scheduler should not mine for a period.  We'll wire this up to the syncer dispatcher's state transition logic shortly
- New endpoint in addition to default scheduling and mine once that attempts to mine for only one  epoch and might not win.  Not directly used (yet).  The need for something like this will only grow and we'll wire it up to mining api and CLI as the need arises either for operating a testnet or integration testing.

Closes issue #3677's "time synchrony" seems related? I'll file a more fine grained one


internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
-//   - waitgroup: signals that the scheduler and any goroutines it launched
-//             have stopped. (Context cancelation happens async, so you
-//             need some way to know when it has actually stopped.)
-//
-// Once started, a scheduler can be stopped by canceling the context provided to `Start()`,
-// which will signal on the waitgroup when it's actually done.
+// Scheduler is the mining interface consumers use.
 type Scheduler interface {
-	Start(miningCtx context.Context) (<-chan Output, *sync.WaitGroup)
+	Start(miningCtx context.Context, skipCh chan bool) (<-chan Output, *sync.WaitGroup)
```
anorth: Please document behaviour of skipCh

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+		workContext, workCancel = context.WithCancel(miningCtx) // nolint: govet
+		base, err := s.pollHeadFunc()
+		if err != nil {
+			log.Errorf("error polling head from mining scheduler %s", err)
+		}
+		h, err := base.Height()
+		if err != nil {
+			log.Errorf("error getting height from base", err)
+		}
+		nullBlkCount := s.calcNullBlks(h)
```
anorth: This implicitly consults the clock again, after it was already consuted at waitForEpochStart. Slightly cleaner would be to return the epoch number from waitForEpochStart and then pass as a parameter to a pure calcNullBlks function.

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+		go func(ctx context.Context) {
+			s.worker.Mine(ctx, base, nullBlkCount, outCh)
+			doneWg.Done()
+		}(workContext)
+	}
+}
+
+func (s *timingScheduler) waitForEpochStart(miningCtx context.Context) {
+	// waitForEpochStart blocks until the next epoch begins
+	currEpoch := uint64(s.chainClock.EpochAtTime(s.chainClock.Now()))
```
anorth: Push this down into a `chainClock.CurrentEpoch()`  function.

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+	nextEpoch := currEpoch + 1
+	nextEpochStart := s.chainClock.StartTimeOfEpoch(nextEpoch)
+	nowB4 := s.chainClock.Now()
+	waitDur := nextEpochStart.Sub(nowB4)
+	newEpochCh := s.chainClock.After(waitDur)
```
anorth: I think you could push this down to the chain clock too, `newEpochCh := s.chainClock.WaitForEpoch(nextEpoch)`

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+	fmt.Printf("wait for time: %s at %s/%s\n", nextEpochStart, nowB4, s.chainClock.Now())
```
anorth: Remove Printf before landing.

internal/pkg/mining/scheduler.go
```diff
@@ -117,39 +136,61 @@ func (s *timingScheduler) IsStarted() bool {
...
 	}
-	s := NewScheduler(w, md, pollHeadFunc)
-	subCtx, subCtxCancel := context.WithCancel(ctx)
-	defer subCtxCancel()
+	epochStartTime := chainClock.StartTimeOfEpoch(nullCount + h + 1)
+	nextEpochStartTime := chainClock.StartTimeOfEpoch(nullCount + h + 2)
+	epochTime := nextEpochStartTime.Sub(epochStartTime)
+	halfEpochTime := epochTime / time.Duration(2) // mine blocks midway through the epoch
+
+	w.clock = th.NewFakeClock(epochStartTime.Add(halfEpochTime))
```
anorth: This clock futzing is pretty nasty, but I think also avoidable. Now that you've removed the sleep from the worker, it only uses the clock to produce a timestamp. So just pass the timestamp to w.Mine().

Worker should probably be renamed to BlockProducer or similar, and just merge the Generate function into the same file (follow-up).

internal/pkg/mining/scheduler_test.go
```diff
@@ -73,6 +56,8 @@ func TestMineOnce10Null(t *testing.T) {
 	messages := chain.NewMessageStore(bs)
 
 	api := th.NewFakeWorkerPorcelainAPI(addr, 10, minerToWorker)
+	genTime := time.Now()
```
anorth: Is it possible to inject a fake clock here instead? Then there'd be no race.

internal/pkg/mining/scheduler_test.go
```diff
@@ -92,238 +77,302 @@ func TestMineOnce10Null(t *testing.T) {
...
+		assert.Equal(t, genTime.Add(blockTime), chainClock.Now())
+		wg.Done()
+		return true
+	})
+
+	scheduler := NewScheduler(w, headFunc(ts), chainClock)
+	scheduler.Start(ctx, nil)
+	fakeClock.BlockUntil(1)
+	fakeClock.Advance(blockTime / time.Duration(2)) // advance half a blocktime
+	time.Sleep(300 * time.Millisecond)              // Need to race.  Ok to delete this test if we prefer not to.
```
anorth: assert the wg is not done here

internal/pkg/mining/scheduler_test.go
```diff
@@ -92,238 +77,302 @@ func TestMineOnce10Null(t *testing.T) {
...
+			mu.Lock()
+			jobs[null] = true
+			mu.Unlock()
+			return true
+		}
+	})
+
+	scheduler := NewScheduler(w, headFunc(ts), chainClock)
+	_, wg := scheduler.Start(ctx, nil)
+	time.Sleep(600 * time.Millisecond) // run through some epochs
```
anorth: Mark this and other sleepy test as integration tests.

internal/pkg/testhelpers/chainclock.go
```diff
@@ -30,6 +31,11 @@ func (fcc *fakeChainClock) EpochAtTime(t time.Time) int64 {
 	return fcc.epoch
 }
 
+// StartTimeOfEpoch returns the current time
+func (fcc *fakeChainClock) StartTimeOfEpoch(e uint64) time.Time {
+	return fcc.Now()
+}
```
anorth: This fake chain clock doesn't seem right at all. It should use a fake clock but a real genesis timestamp and calculations. In fact I would think a real chain clock (wrapping a fake clock) would be adequate for testing.

internal/pkg/testhelpers/testflags/flags.go
```diff
@@ -55,7 +55,7 @@ func IntegrationTest(t *testing.T) {
 	if !*integrationTest {
 		t.SkipNow()
 	}
-	t.Parallel()
+	//t.Parallel()
```
anorth: Restore

internal/pkg/clock/chainclock.go
```diff
@@ -4,19 +4,13 @@ import (
...
-// of a blockchain epoch.
-const EpochDuration = EpochUnits * time.Duration(EpochCount)
-
-// EpochCount is the number of instances in an Epoch.
-const EpochCount = int64(15)
-
-// EpochUnits is the units an Epoch is measured in.
-const EpochUnits = time.Second
+// DefaultBlockTime is the default duration of epochs
+const DefaultBlockTime = 15 * time.Second
```
anorth: EpochDuration is a far better name for this.

internal/pkg/mining/scheduler.go
icorderi: Why is this a channel on the start and not a `Pause()`/`Resume()` on the interface itself?

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
-//             have stopped. (Context cancelation happens async, so you
-//             need some way to know when it has actually stopped.)
-//
-// Once started, a scheduler can be stopped by canceling the context provided to `Start()`,
-// which will signal on the waitgroup when it's actually done.
+// Scheduler is the mining interface consumers use.
 type Scheduler interface {
-	Start(miningCtx context.Context) (<-chan Output, *sync.WaitGroup)
+	Start(miningCtx context.Context, skipCh chan bool) (<-chan Output, *sync.WaitGroup)
 	IsStarted() bool
```
icorderi: given that mining can be paused/skipped now, we might want to add a `IsMining() bool`

internal/pkg/mining/scheduler.go
icorderi: I would push the entire method to `chainClock.WaitNextEpoch()`

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+	// mineLoop is the main event loop for the timing scheduler.  It waits for
+	// a new epoch to start, polls the heaviest head, includes the correct number
+	// of null blocks and starts a mining job async.
+	//
+	// If the previous epoch's mining job is not finished it is canceled via the context
+	//
+	// The scheduler will skip mining jobs if the skipping flag is set
+	workContext, workCancel := context.WithCancel(miningCtx) // nolint:staticcheck
+	for {
+		s.waitForEpochStart(miningCtx)
```
icorderi: I wouldn't just flatout wait for an epoch, I would:

1.  track the epoch we started mining at
2. mine 
3. check the epoch at the time we finish 
   3.a.1 if we are still in the same epoch, all good, do the post-mining work

   3.b.1 if we are NOT in the same epoch, we missed the window, and the mining wont work 
   3.b.2 log a warning that we failed to produce a block during the epoch mining window. 

4. wait for next epoch
> // Note: we wait even in the case were we are already on the next window to have a chance at a full window to mine a block`

internal/pkg/mining/scheduler.go
```diff
@@ -7,108 +7,127 @@ package mining
...
+	nextEpoch := currEpoch + 1
+	nextEpochStart := s.chainClock.StartTimeOfEpoch(nextEpoch)
+	nowB4 := s.chainClock.Now()
+	waitDur := nextEpochStart.Sub(nowB4)
+	newEpochCh := s.chainClock.After(waitDur)
+	fmt.Printf("wait for time: %s at %s/%s\n", nextEpochStart, nowB4, s.chainClock.Now())
+	select {
+	case currTime := <-newEpochCh:
+		//	fmt.Printf("starting epoch: %d\n", nextEpoch)
+		//		fmt.Printf("exp time: h%d-m%d-s%d-m%d, curr time: h%d-m%d-s%d-m%d\n", nextEpochStart.Hour(), nextEpochStart.Minute(), nextEpochStart.Second(), nextEpochStart.Nanosecond()/1000000, currTime.Hour(), currTime.Minute(), currTime.Second(), currTime.Nanosecond()/1000000)
```
icorderi: delete

internal/pkg/mining/scheduler.go
```diff
@@ -117,39 +136,61 @@ func (s *timingScheduler) IsStarted() bool {
...
+		pollHeadFunc: f,
+		chainClock:   c,
 	}
+}
 
-	return prevNullBlkCount + 1
+// MineOnce mines on a given base until it finds a winner.
+func MineOnce(ctx context.Context, w DefaultWorker, ts block.TipSet, c clock.ChainEpochClock) (Output, error) {
+	var winner *block.Block
+	var err error
```
icorderi: error does not need to be define on this outer scope

internal/pkg/testhelpers/testflags/flags.go
```diff
@@ -65,7 +65,7 @@ func UnitTest(t *testing.T) {
 	if !*unitTest && !testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
+	//t.Parallel()
```
icorderi: restore this one too

internal/pkg/mining/scheduler.go
ZenGround0: Documentation in the Start function comment and Start code should be sufficient.  Are you asking to document in the interface too?

internal/pkg/mining/scheduler.go
ZenGround0: We could do that but I don't think we should.  Right now the timestamp indicates when the block production finished by default.  We would lose that potentially useful signal by default if we timestamped all blocks in the beginning of the epoch.

The clock futzying is mildly ugly but really not so bad and not hit on the default production code path anyway.

I'm fine with moving generate into the same file as a low priority thing.

internal/app/go-filecoin/internal/submodule/block_mining_submodule.go
```diff
@@ -17,7 +17,7 @@ type BlockMiningSubmodule struct {
 	AddNewlyMinedBlock newBlockFunc
 	// cancelMining cancels the context for block production and sector commitments.
 	CancelMining    context.CancelFunc
-	MiningWorker    mining.Worker
+	MiningWorker    *mining.DefaultWorker
```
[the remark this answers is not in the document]
ZenGround0: Nope the interface is great for testing and maybe useful down the road for miner flexibility but the name could use work.

internal/pkg/mining/scheduler.go
ZenGround0: That sounds like a better idea.  I might switch over.

internal/pkg/mining/scheduler.go
ZenGround0: Also not a bad idea, though I might wait until we have a reason to surface that info.

internal/pkg/mining/scheduler.go
ZenGround0: We *are* doing a slightly better version of what you suggest.  We use the goroutine and cancel function to communicate that the time has passed and stop the mining work from the previous epoch which should log the warning.  Might need to fiddle with log statements to get exactly what we want but that stuff can wait after this PR.

The only difference between whats here and whats proposed is that we don't block new mining work on late mining work from a previous epoch which is a better way to handle this by default.  As you say if we end up being late "mining won't work" so it's better to not block tasks that might work on these tasks that we agree won't work.

internal/pkg/clock/chainclock.go
```diff
@@ -25,21 +19,41 @@ type chainClock struct {
...
-	return &chainClock{GenesisTime: gt, Clock: NewSystemClock()}
+// NewChainClock returns a ChainEpochClock wrapping a default clock.Clock
+func NewChainClock(genesisTime uint64, blockTime time.Duration) ChainEpochClock {
+	return NewChainClockFromClock(genesisTime, blockTime, NewSystemClock())
+}
+
+// NewChainClockFromClock returns a ChainEpochClock wrapping the provided
+// clock.Clock
+func NewChainClockFromClock(genesisTime uint64, blockTime time.Duration, c Clock) ChainEpochClock {
+	gt := time.Unix(int64(genesisTime), int64(genesisTime%1000000000))
```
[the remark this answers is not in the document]
ZenGround0: I've been thinking about this.  While sub-second precision was helpful for testing I don't think its necessary using mock clocks.  I became aware working on the next PR that the timestamp only records up to the second.  With that in mind subsecond epochs really don't make sense and a non-second-aligned genesis also doesn't make sense.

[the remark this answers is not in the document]
ZenGround0: In retrospect this is a massive headache for integration testing.  We need to be able to run networks with subsecond blocktimes but the timestamps can't differentiate epochs.  

As a result we need `EpochAtTime` to return a range of epochs for validation instead of a single epoch

internal/pkg/testhelpers/chainclock.go
ZenGround0: I agree with this

internal/pkg/mining/scheduler.go
ZenGround0: Currently deferring this work to the PR that will hook up syncer state changes to mining scheduler pausing

internal/pkg/mining/scheduler.go
ZenGround0: Decided to defer for now

internal/pkg/clock/chainclock.go
```diff
@@ -4,19 +4,13 @@ import (
...
-
-// EpochUnits is the units an Epoch is measured in.
-const EpochUnits = time.Second
+// DefaultBlockTime is the default duration of epochs
+const DefaultBlockTime = 15 * time.Second
 
 // ChainEpochClock is an interface for a clock that represents epochs of the protocol.
 type ChainEpochClock interface {
 	EpochAtTime(t time.Time) int64
+	StartTimeOfEpoch(e uint64) time.Time
```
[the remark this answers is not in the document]
ZenGround0: I'm moving to `types.BlockHeight` but want to call out that this type is mega-overkill.  We're looking at much more than a million years of blocks whose epochs can fit in an int64.  We should just use go's type aliasing-like thing on uint64s.

internal/pkg/clock/chainclock.go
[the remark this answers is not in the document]
anorth: Many integration tests can use fake clocks too. Only forking real processes will require a whole-second clock.

internal/pkg/clock/chainclock.go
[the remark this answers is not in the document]
anorth: Two things:
- we should use the existing type consistently for all epoch/blockheight
- it has the wrong underlying representation, a known issue

When we change BlockHeight to be an int64, all will be well.

PR #3708, opened 2019-12-18
Feat/timing rules

ZenGround0:
### Motivation
The timestamp validation rules should be correctly checked.

### Proposed changes
Spec compliant timestamps validation checks
1) check that the timestamp matches the epoch number
2) check that the timestamp is not from a future epoch

I decided to implement a slightly generalized version of 1) that can handle epoch durations that are not a multiple of a second.   I did this because it was very simple to write compared to moving our integration tests to properly mocked time and added only a small bit of complexity to our clock/validation code.  This can be reduced to the simpler case once we have integration tests that can readily mock out the clock.

Closes issue #3677

### DIFF
--- a/internal/pkg/consensus/block_validation.go
+++ b/internal/pkg/consensus/block_validation.go
@@ -3,7 +3,6 @@ package consensus
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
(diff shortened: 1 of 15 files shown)


internal/pkg/consensus/block_validation.go
```diff
@@ -77,13 +97,6 @@ func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *bl
 		return fmt.Errorf("block %s has invalid height %d", child.Cid().String(), child.Height)
 	}
 
-	// check that child is appropriately delayed from its parents including
-	// null blocks.
-	// TODO replace check on height when #2222 lands
-	limit := uint64(pmin) + uint64(dv.BlockTime().Seconds())*(uint64(child.Height)-ph)
```
ZenGround0: Note: I've moved the epoch alignment check to "syntax" because the clock synchrony assumption means we no longer need the parent block for this.

Issue #3706, opened 2019-12-16
Pause / unpause mining during transition to / from catchup mode


ZenGround0:
### Description
Transitioning to catchup mode should pause the mining scheduler.  Transition out of it should unpause.

### Acceptance criteria
These side effects of transitioning are written and tested.

### Risks + pitfalls
It is in scope to ensure that mining start can't begin mining when the node is not in the follow state yet.

### Where to begin
1. Add a channel to the dispatcher that sends out signals on a channel
2. Receive on this channel in a go routine within node.go
3. When mining start is called wire up these channel results to the mining scheduler Pause buttons

anorth:
@ZenGround0 done?

PR #3709, opened 2019-12-18
Feat/mining pause wiring

ZenGround0:
### Motivation
The mining scheduler should pause block production when the syncing subsystem detects that it is too far behind syncing the chain to have a good estimate for the head.  The mining scheduler should start producing blocks again once the syncer is confident the node is caught up to the rest of the network.

### Proposed changes
This PR adds a channel to the syncing subsystem's state transitioner which signals the node when the state is transitioned.  It adds some glue in the node to (1) prevent these signals from blocking up the syncing dispatcher (2) pass theses signals onto the mining scheduler.  The work from PR #3704 built the simple `Pause`, `Continue` functionality into the scheduler interface, so this PR simply calls these methods from the node upon hearing the right signal

Closes issue #3706


internal/pkg/vm/actor/builtin/account/account_encoding_gen.go
```diff
@@ -1,20 +1,19 @@
 package account
-	
-/* 
+
+/*
```
acruikshank: Looks like this got linted. It's unfortunate that the generated code isn't linter compatible. This could cause churn, but it's probably more likely to get linted multiple times before it gets regenerated, so probably best to leave this the way it is.

Issue #3484, opened 2019-09-23
Use node builder to replace GenNode and associated calls


ZenGround0:
### Description
When building nodes for test we should use the new node builder abstraction.

### Acceptance criteria
- All calls to `GenNode` are replaced with using the builder
- Potentially remove the other test builder `MakeNodeWithChainSeed` though that might be reserved for a follow on issue.

### Risks + pitfalls
May have to augment the node builder to handle these cases.

anorth:
Done

PR #3710, opened 2019-12-19
Node test builder simplification

ZenGround0:
### Motivation
As part of ongoing work to move integration tests to an in-process framework it's in our best interest to simplify that framework.  This PR does so by removing a legacy function `GenNode` with the similar new `node/test.Builder` abstraction

### Proposed changes
All calls to `GenNode` call the node test builder
Node test builder takes in node builder options (need to check in with @anorth about this one)
Move some things around to remove cyclic dependencies.  Mostly achieved by moving our node tests to the `node_test` package from the `node` package(!!)

Closes issue #3484

### DIFF
--- a/internal/app/go-filecoin/node/test/builder.go
+++ b/internal/app/go-filecoin/node/test/builder.go
@@ -11,6 +11,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/config"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/repo"
+	th "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers"
+
 )
 
 // NodeBuilder creates and configures Filecoin nodes for in-process testing.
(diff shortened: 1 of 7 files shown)


internal/app/go-filecoin/node/test/builder.go
```diff
@@ -115,3 +120,37 @@ func (b *NodeBuilder) requireNoError(err error) {
...
+	builder := NewNodeBuilder(t)
+	builder.WithGenesisInit(gif)
+	builderopts = append(builderopts, node.DefaultTestingConfig()...)
+	builder.WithBuilderOpt(builderopts...)
+
+	var out []*node.Node
+	for i := 0; i < numNodes; i++ {
+		nd := builder.Build(context.Background())
+		out = append(out, nd)
+	}
```
anorth: Move this loop into a `builder.BuildMany(numNodes)` and then my same comment as above applies about removing this function and using the builder directly.

"Unstarted" is redundant in the name - we don't have anything that makes them already started.

internal/app/go-filecoin/node/testing.go
```diff
@@ -32,15 +28,6 @@ type ChainSeed struct {
...
 
-// TestNodeOptions is a generalized struct for passing Node options around for testing
-type TestNodeOptions struct {
-	OfflineMode bool
-	BuilderOpts []BuilderOpt
-	GenesisFunc consensus.GenesisInitFunc
-	InitOpts    []InitOpt
-	Seed        *ChainSeed
-}
-
```
anorth: ❤️

Issue #3690, opened 2019-12-02
Consensus fault reporting and slashing


icorderi:

anorth:
All that's missing here is hooking the reporter up to block production in order to mine the slashing message. Don't broadcast the message - it's irrational for any other miner to include it in a block. Attempt to mine it directly.

anorth:
See also #3358

PR #3711, opened 2019-12-19
Spike/consensus slashing

ZenGround0:
### Motivation
We need to slash consensus faults for a secure protocol.  This PR adds consensus fault detection into the syncing pipeline.  It only detects and emits the faults which are currently dropped.  Handling slashing will come in another PR and is currently blocked on a state machine function that actually executes the penalty on chain.

### Proposed changes
1) New consensus fault detector object + unit tests
2) Integrate construction of fault detector into the syncer submodule and add glue for connecting detected faults to a to-be-written slasher within the start method.
3) Integrate fault detection calls into the syncer.

(3) could happen more decoupled from the syncer by adding in a scanning routine on blocks that have been recently added to the store.  Calling inline in `syncOne` was the simplest way to satisfy needed preconditions of valid parent and verified block signature.  Decoupling detection and actual slashing makes this feel very reasonable.

Doesn't close but addresses part 1 of #3690

### DIFF
--- a/cmd/go-filecoin/message_daemon_test.go
+++ b/cmd/go-filecoin/message_daemon_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"strconv"
 	"testing"
+	"fmt"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
--- a/cmd/go-filecoin/message_daemon_test.go
+++ b/cmd/go-filecoin/message_daemon_test.go
@@ -3,7 +3,6 @@ package commands_test
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"strconv"
 	"testing"
 
--- a/cmd/go-filecoin/message_daemon_test.go
+++ b/cmd/go-filecoin/message_daemon_test.go
@@ -3,9 +3,9 @@ package commands_test
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"testing"
-	"fmt"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
--- a/functional-tests/faucet_test.go
+++ b/functional-tests/faucet_test.go
@@ -23,8 +23,7 @@ import (
 var faucetBinary = "../tools/faucet/faucet"
 
 func TestFaucetSendFunds(t *testing.T) {
-	//	tf.FunctionalTest(t)
-	tf.IntegrationTest(t)
+	tf.FunctionalTest(t)
 
 	if _, err := os.Stat(faucetBinary); os.IsNotExist(err) {
 		panic("faucet not found, run `go run build/*.go build` to fix")
--- a/functional-tests/faucet_test.go
+++ b/functional-tests/faucet_test.go
@@ -23,7 +23,8 @@ import (
 var faucetBinary = "../tools/faucet/faucet"
 
 func TestFaucetSendFunds(t *testing.T) {
-	tf.FunctionalTest(t)
+	//	tf.FunctionalTest(t)
+	tf.IntegrationTest(t)
 
 	if _, err := os.Stat(faucetBinary); os.IsNotExist(err) {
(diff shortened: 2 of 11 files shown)


cmd/go-filecoin/message_daemon_test.go
```diff
@@ -126,6 +127,7 @@ func TestMessageSendBlockGasLimit(t *testing.T) {
 	result := struct{ Messages types.TxMeta }{}
 
 	t.Run("when the gas limit is above the block limit, the message fails", func(t *testing.T) {
+		fmt.Printf("WE ARE HERE WE ARE HERE WE ARE HERE\n")
```
icorderi: remove

internal/pkg/slashing/consensus_slashing.go
```diff
@@ -0,0 +1,63 @@
+package slashing
+
+import (
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
+)
+
+// CFaultDetector detects consensus faults -- misbehavior conditions where a single
+// party produces multiple blocks at the same time.
+type CFaultDetector struct {
```
icorderi: Spell out `Consensus` on the name.

internal/pkg/slashing/consensus_slashing.go
```diff
@@ -0,0 +1,63 @@
...
+	// minerIndex tracks witnessed blocks by miner address and epoch
+	minerIndex map[address.Address]map[uint64]*block.Block
+	// sender sends messages on behalf of the slasher
+	faultCh chan ConsensusFault
+}
+
+// ConsensusFault is the information needed to submit a consensus fault
+type ConsensusFault struct {
+	Block1, Block2 *block.Block
```
icorderi: what are this blocks? split them and a oneliner with the details.

cmd/go-filecoin/message_daemon_test.go
acruikshank: Please remove before merging.

functional-tests/faucet_test.go
```diff
@@ -41,7 +42,7 @@ func TestFaucetSendFunds(t *testing.T) {
 	blockTime := time.Second * 5
 
 	// Setup first node, note: Testbed.Name() is the directory
-	genesisTime := time.Unix(123456789, 0)
+	genesisTime := time.Now()
```
acruikshank: We would like this to be deterministic. I'm assuming this is to debug CI. Unless I'm wrong, please revert before merging.

internal/pkg/chainsync/internal/syncer/syncer.go
```diff
@@ -289,6 +299,15 @@ func (syncer *Syncer) syncOne(ctx context.Context, grandParent, parent, next blo
...
 
+	// Now that the tipset is validated preconditions are satisfied to check
+	// consensus faults
+	for i := 0; i < next.Len(); i++ {
+		err := syncer.faultDetector.CheckBlock(next.At(i), parent)
+		if err != nil {
+			return err
+		}
+	}
+
```
anorth: Since this method proceeds whether or not a fault is detected, I would move this to the end, after storing all state. It's kind of an aside, could even be async.

internal/pkg/testhelpers/output.go
```diff
@@ -112,6 +113,7 @@ func (o *CmdOutput) AssertSuccess() *CmdOutput {
 // AssertFail asserts that the output represents a failed execution, with the error
 // matching the passed in error.
 func (o *CmdOutput) AssertFail(err string) *CmdOutput {
+	fmt.Printf("Assert FAIL\n")
```
anorth: Remove

functional-tests/faucet_test.go
ZenGround0: Unfortunately the specced out timing rules mean that we either need to do this, mock out the clock, or modify the test to allow for a time-of-test dependent number of null blocks between genesis and the first non-null.  This seems like the best tradeoff for a functional test (if it were an integration test mocking out clock would probably be preferred).

PR #3712, opened 2020-01-03
Chain daemon tests in process

ZenGround0:
### Motivation
Move integration tests in process for tech debt wins
- tests debuggable by default
- only spawn new processes when we want to test that behavior
- rework lots of dying test daemon code
- avoid situations where production code changes to accommodate tests (most recently clock based validation rules)

### Proposed changes
1) Build out in-process-node integration helpers
2) Make gengen config consistent and self-contained by putting time and seed values within config
3) Add the testing gengen config to fixtures 
4) Remove race from mining once by registering a callback on the dispatcher and a using a signal function
5) Migrate chain daemon tests to in-process-node tests

blocking on #3710


cmd/go-filecoin/chain_daemon_test.go
```diff
@@ -2,34 +2,39 @@ package commands_test
```
anorth: Rename the files to x_integration_test when they no longer use NewDaemon, or maybe x_command_test when explicitly testing the CLI. (though only in this PR if you can do so without clobbering diffs)

internal/pkg/chainsync/chainsync.go
```diff
@@ -16,6 +16,7 @@ type BlockProposer interface {
 	SendHello(ci *block.ChainInfo) error
 	SendOwnBlock(ci *block.ChainInfo) error
 	SendGossipBlock(ci *block.ChainInfo) error
+	WaitForTarget(ci *block.ChainInfo) func()
```
anorth: I think `WaiterForTarget` would be slightly better, since this call doesn't block.

internal/pkg/chainsync/internal/dispatcher/dispatcher.go
```diff
@@ -205,6 +206,19 @@ func (d *Dispatcher) RegisterCallback(cb func(Target)) {
 	d.control <- cbMessage{cb: cb}
 }
 
+// WaitForTarget returns a function that will block until the dispatcher
+// processes the given target
+func (d *Dispatcher) WaitForTarget(ci *block.ChainInfo) func() {
```
anorth: The parameter should be just the tipset key `ci.Head` since other fields are ignored.

tools/gengen/util/gengen.go
```diff
@@ -359,3 +364,9 @@ func ApplyProofsModeDefaults(cfg *GenesisCfg, useLiveProofsMode bool, force bool
 		}
 	}
 }
+
+// DefaultSeed is the default seed used by the gengen tool
+const DefaultSeed = 4
+
+// DefaultTime is the default time used by the gengen tool
+const DefaultTime = 123456789
```
anorth: These appear to be used only in gengen_test, so move them there and unexport.

cmd/go-filecoin/chain_daemon_test.go
ZenGround0: Good idea.  I'll follow on with one file mutation PR at the end of all of the outstanding test PRs to avoid rebase difficulties and bad diffs.

PR #3713, opened 2020-01-04
Move daemon integration tests to in process node tests

ZenGround0:
### Motivation
Follow on to #3712, see that PR for motivation

### Proposed changes
More changes to tests removing dependence on test daemon
Minor improvements to testing infrastructure to enable tests, particularly fixing Run and RunFail to correctly test failed runs.

Closes issue #

### DIFF
--- a/cmd/go-filecoin/actor_daemon_test.go
+++ b/cmd/go-filecoin/actor_daemon_test.go
@@ -2,6 +2,7 @@ package commands_test
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"testing"
 
@@ -9,18 +10,21 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/cmd/go-filecoin"
-	th "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers"
+	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node/test"
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
 )
 
 func TestActorDaemon(t *testing.T) {
 	tf.IntegrationTest(t)
-
+	ctx := context.Background()
 	t.Run("actor ls --enc json returns NDJSON containing all actors in the state tree", func(t *testing.T) {
-		d := th.NewDaemon(t).Start()
-		defer d.ShutdownSuccess()
+		builder := test.NewNodeBuilder(t)
+		n := builder.BuildAndStart(ctx)
+		defer n.Stop(ctx)
+		cmdClient, done := test.RunNodeAPI(ctx, n, t)
+		defer done()
 
-		op1 := d.RunSuccess("actor", "ls", "--enc", "json")
+		op1 := cmdClient.RunSuccess(ctx, "actor", "ls", "--enc", "json")
 		result1 := op1.ReadStdoutTrimNewlines()
 
 		var avs []commands.ActorView
--- a/cmd/go-filecoin/address_daemon_test.go
+++ b/cmd/go-filecoin/address_daemon_test.go
@@ -1,86 +1,114 @@
 package commands_test
 
 import (
+	"context"
 	"os"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-filecoin/fixtures"
+	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node"
+	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node/test"
+
 	th "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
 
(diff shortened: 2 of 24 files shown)


cmd/go-filecoin/actor_daemon_test.go
```diff
@@ -2,25 +2,29 @@ package commands_test
...
-
+	ctx := context.Background()
 	t.Run("actor ls --enc json returns NDJSON containing all actors in the state tree", func(t *testing.T) {
-		d := th.NewDaemon(t).Start()
-		defer d.ShutdownSuccess()
+		builder := test.NewNodeBuilder(t)
+		n := builder.BuildAndStart(ctx)
+		defer n.Stop(ctx)
+		cmdClient, done := test.RunNodeAPI(ctx, n, t)
+		defer done()
```
anorth: This pattern is repeated enough that it's probably worth a helper so you can

```
nd, client, done := builder.BuildAndStartAPI(ctx)
defer done()
```
where `done` shuts down both.

cmd/go-filecoin/address_daemon_test.go
```diff
@@ -1,86 +1,114 @@
 package commands_test
```
anorth: Rename files to x_integration_test or x_cmd_test

cmd/go-filecoin/message_daemon_test.go
```diff
@@ -1,67 +1,83 @@
...
+	require.NoError(t, err)
 
 	t.Log("[failure] invalid target")
-	d.RunFail(
+	cmdClient.RunFail(
+		ctx,
 		address.ErrUnknownNetwork.Error(),
 		"message", "send",
-		"--from", from,
+		"--from", from.String(),
```
anorth: Should we change `RunSuccess/Fail` to take vararg `Stringer`s rather than `string`s? There were more like this above.

internal/app/go-filecoin/node/message.go
```diff
@@ -8,6 +8,13 @@ import (
 )
 
 func (node *Node) processMessage(ctx context.Context, pubSubMsg pubsub.Message) (err error) {
+	sender := pubSubMsg.GetSender()
+
+	// ignore messages from self
+	if sender == node.Host().ID() {
+		return nil
+	}
```
anorth: What necessitated this change?

internal/app/go-filecoin/node/test/api.go
```diff
@@ -68,7 +72,7 @@ func (c *Client) Address() string {
 }
 
 // Run runs a CLI command and returns its output.
-func (c *Client) Run(ctx context.Context, command ...string) *th.CmdOutput {
```
anorth: Blocking: I think this is breaking the explicit intention here, we should discuss and perhaps clarify documentation further before landing this.

Please review the comments on `CmdOutput` and its methods. Failure and error are two different things. The output error should be set only if the command was unable to execute as a unix process (e.g. file not found, permission denied etc) - basically things that should never happen. There is no `RunError()` method that expects this to happen.

Otherwise, no error should be set but the processes exit status and stdout/err captured. RunFail is intended to mean the command executed but returned a non-zero exit status. If instead the process couldn't be executed at all, then we haven't tested what we wanted and the test should fail. Failure is never expected in the shouldFail sense.

The naming of error, failure etc makes this all a bit hard to work out from the code - please clarify names and/or comments if you can.

cmd/go-filecoin/address_daemon_test.go
ZenGround0: will do one big follow up

internal/app/go-filecoin/node/message.go
ZenGround0: The lack of this check caused a very trick race in the mpool remove test.  This guard is what the messaging system appears to assume.  Otherwise we add to the mpool twice with difficult to capture timing semantics.

cmd/go-filecoin/actor_daemon_test.go
ZenGround0: Great idea.  Because this applies across PRs and is a simple change I'm planning to do this along with the file renames in a PR after all of these land.

cmd/go-filecoin/message_daemon_test.go
ZenGround0: Up to you.  Wouldn't the string literals need to be wrapped making that approach more verbose?

internal/app/go-filecoin/node/test/api.go
ZenGround0: Thanks for the clear explanation.

I understand the distinction between failure and error and would prefer to undo these changes if it is not too much effort.  However we will need to solve some problems.

The first problem that shows up when you try to use this code on actual commands is that `commands.Run` consistently propagates the same error that shows up on stderr during error (not fail).  For example `TestSwarmConnectPeersInvalid` propagates an error (`failed to parse multiaddr "/ip4/hello"`) to `Run` that blows up the "client execution error" require.  If you check stderr the same error string shows up there too.  

I'm now realizing this probably means we are implementing commands incorrectly and perhaps we can address this issue at the source. 

If we solve this problem the second problem we will face with the unmodified code is that we will assert that the exitCode is 0 for all `Run` invocations even those made by `RunFail`.  I believe this will make `RunFail` always fail.  This has a simple fix, we can move the exit code checks to `RunFail` and `RunSuccess` by adding a simple helper on `CmdOutput`.

ZenGround0: I'm planning to dig into cli.Run's error propagation.  If I can preserve stderr propagation while modifying commands.Run's propagation to be a cli.ExitError without modifying dependencies or doing anything too ugly I will fix in this PR.  If not I propose we keep what is in the PR.

anorth: Moving exit code checks sounds ok.

PR #3714, opened 2020-01-05
Test/in process integration 2

ZenGround0:
### Motivation
Another follow on to #3712, this PR migrates the extensive fast-based payment channel daemon tests to in process integration tests.

### Proposed changes
This change is very straightforward and does not carry along any improvements to the framework, just migrated tests.

Closes issue #

### DIFF
--- a/cmd/go-filecoin/payment_channel_daemon_test.go
+++ b/cmd/go-filecoin/payment_channel_daemon_test.go
@@ -6,59 +6,51 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ipfs/go-cid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/filecoin-project/go-filecoin/fixtures"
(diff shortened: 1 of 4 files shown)


cmd/go-filecoin/payment_channel_daemon_test.go
```diff
@@ -269,49 +265,84 @@ func TestPaymentChannelReclaimSuccess(t *testing.T) {
...
+	mcid, _, err := rsrc.target.PorcelainAPI.MessageSend(
+		ctx,
+		rsrc.targetAddr,
+		address.PaymentBrokerAddress,
+		types.NewAttoFILFromFIL(0),
+		types.NewAttoFIL(big.NewInt(1)),
+		300,
+		paymentbroker.Redeem,
+		params...,
+	)
```
anorth: I think it's probably worth factoring out a wrapper or sender object to reduce the verbosity of message sends. Either a free function with parameters similar to the FAST method this used to have, or something that curries in parameters like the sender and gas price/limit.

PR #3716, opened 2020-01-07
feat: new vm execution model 

icorderi:
### Motivation

The new spec changes the message processing, gas accounting, error handling, and other misc aspects of actor code execution. 

### Proposed changes

- fresh implementation matching spec

### Missing

- New randomness #3717
- Storage checkpoint and commit #3718
- Move/adjust new create actor method to match latest changes in spec.


internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,308 @@
...
+	// actor does not exist, create an account actor
+	// - precond: address must be a pub-key
+	// - sent init actor a msg to create the new account
+
+	if !target.IsPubKey() {
+		// Don't implicitly create an account actor for an address without an associated key.
+		panic(exitcode.ActorNotFound)
+	}
+
+	// send init actor msg to create the account actor
```
anorth: This differs from the spec at present. Leave a comment pointing to https://github.com/filecoin-project/specs/issues/762

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// 1. fromActor = executing toActor
+	from := ctx.msg.to
+	fromActor := ctx.toActor
+
+	// 2. build internal message
+	encodedParams, err := abi.ToEncodedValues(params...)
+	if err != nil {
+		// Review: should this be a recoverable or unrecoverable panic?
+		panic("failed to encode params")
```
anorth: I think we may have to do encoding in the actor, before Send() - I'll discuss elsewhere. But if we do it this way, this should abort rather than panic - it will be quite possible to pass an unencodable value.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	firstMiner := blocks[0].Miner
+	cronMessage := makeCronTickMessage(firstMiner)
+	vm.applyImplicitMessage(cronMessage)
+
+	// commit state
+	// flush all objects out
+	if err := vm.store.Flush(); err != nil {
+		// Review: what do we want to do with this errors?
+		//         we can change the return to be ([]Receipt, error) although its unclear what to do about it as a caller
+		//         we can just crash, or let the caller retry a few times and then crash
```
anorth: Here and below I think we should return an error value and propagate it up.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	// 2. increment seq number
+	fromActor.IncrementSeqNum()
+
+	// 3. build context
+	ctx := newInvocationContext(vm, imsg, fromActor, &gasTank)
+
+	// 4. invoke message
+	// Note: any errors raised by the invocation will not be trapped
+	// 		 errors are not expected by the spec on implicit messages, this messages must work
+	return ctx.invoke()
```
anorth: I think this should propagate an explicit error

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	// 4. check if _sender_ has enough funds
+	// 5. increment message seq number
+	// 6. withheld maximum gas from _sender_
+	// 7. checkpoint state
+
+	// 1. charge for bytes used in chain
+	ok := gasTank.TryCharge(gascost.OnChainMessage(onChainMsgSize))
+	if !ok {
+		// Invalid message; insufficient gas limit to pay for the on-chain message size.
+		return message.Failure(exitcode.OutOfGas), gasTank.GasConsumed().Cost(msg.GasPrice)
```
anorth: As I understand it, TryCharge will cap out at the GasLimit (even if it's zero), so that GasConsumed() is less than the true cost. But in this case, the miner needs to pay the actual cost of the on-chain message.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	// Dragons: from address need to be normalized here
+
+	// 2. load actor from global state
+	fromActor, err := vm.state.GetActor(context.Background(), msg.From)
+	// Review: when are errors surfaced? this can lead to invalid state transitions
```
anorth: What do you mean by invalid? Differing from the spec / other impls?

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	// 6. Deduct gas limit funds from sender first
+	// Note: this should always succeed, due to the sender balance check above
+	// Note: after this point, we nede to return this funds back before exiting
+	vm.transfer(msg.From, address.BurntFundsAddress, gasLimitCost)
+
+	// 7. checkpoint state
+	// Even if the message fails, the following accumulated changes will be applied:
+	// - CallSeqNumber increment
+	// - sender balance withheld
+	// - (if did not previously exist, the new auto-created target actor)
```
anorth: This last point is no longer true in the spec

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+	//       we need to trap them on this top level call
+	safeProcess := func() (out message.Receipt) {
+		// trap aborts and exitcodes
+		defer func() {
+			// TODO: discard "non-checkpointed" state changes (issue: #3718)
+
+			if r := recover(); r == nil {
+				switch r.(type) {
+				case runtime.AbortPanicError:
+					// Review: abort has a msg inside.. should we log it or what?
```
anorth: Yeah it's just for logging. I would like us to panic with a structure containing both an exit code and an optional message, which is logged here.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+					out = message.Failure(exitcode.MethodAbort).WithGas(gasTank.GasConsumed())
+					return
+				case exitcode.ExitCode:
+					out = message.Failure(r.(exitcode.ExitCode)).WithGas(gasTank.GasConsumed())
```
anorth: I think these message.Failure call sites would be clearer if it just took two arguments, with gas value explicitly zero for the ones above where the miner pays.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+					return
+				default:
+					// Review: how do we rethrow while maintaining call stack?
+					panic(r)
```
anorth: Apparently we can't. We should log it here with debug.PrintStack()

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+		// Note: we are charging the caller not the miner, there is ZERO miner penalty
+		return message.Failure(exitcode.OutOfGas).WithGas(gasTank.GasConsumed()), types.ZeroAttoFIL
+	}
+
+	// 2. Success!
+	return receipt, types.ZeroAttoFIL
+}
+
+func (vm *VM) getMinerOwner(minerAddr address.Address) address.Address {
+	// resolve the target address via the InitActor, and attempt to load state.
```
anorth: You can assume it's already resolved

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+// WARNING: this method will panic if the the amount is negative, accounts dont exist, or have inssuficient funds.
+//
+// Note: this is not idiomatic, it follows the Spec expectations for this method.
+func (vm *VM) transfer(debitFrom address.Address, creditTo address.Address, amount types.AttoFIL) {
+	// allow only for positive amounts
+	if amount.IsNegative() {
+		panic("unreachable: negative funds transfer not allowed")
+	}
+
+	ctx := context.Background()
```
anorth: Please thread the necessary context here.

If the VM were constructed as a use-once object, it would probably be ok to put the context as a field on vm.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+
+	// retrieve debit account
+	fromActor, err := vm.state.GetActor(ctx, debitFrom)
+	if err != nil {
+		panic("unreachable: debit account not found")
```
anorth: In all of these panics, please include the error string in the message

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,543 @@
...
+//
+// implement runtime.Storage for actorStorage
+//
+
+var _ runtime.Storage = (*actorStorage)(nil)
+
+func (s actorStorage) Put(obj interface{}) cid.Cid {
+	cid, err := s.inner.Put(obj)
+	if err != nil {
+		panic("could not put object in store")
```
anorth: Include the error string

internal/pkg/vm/internal/gas/gas.go
```diff
@@ -0,0 +1,62 @@
...
+		gasConsumed: types.ZeroGas,
+	}
+}
+
+// Charge will add the gas charge to the current method gas context.
+//
+// WARNING: this method will panic if there is no sufficient gas left.
+func (t *Tracker) Charge(amount Unit) {
+	if ok := t.TryCharge(amount); !ok {
+		panic(exitcode.OutOfGas)
```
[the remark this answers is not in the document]
anorth: After reading more, I now see that exitcode is trapped, but that was definitely a surprise. It would be much clearer to wrap this.

internal/pkg/vm/address/address.go
```diff
@@ -71,6 +71,12 @@ func (a Address) Protocol() Protocol {
 	return a.str[0]
 }
 
+// IsPubKey returns true if the address is a pub-key style address.
+func (a Address) IsPubKey() bool {
+	p := a.Protocol()
+	return p == SECP256K1 || p == BLS
+}
```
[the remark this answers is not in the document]
icorderi: Ok good, we can change the external address package in a separate PR.

internal/pkg/vm/internal/gas/gas.go
```diff
@@ -0,0 +1,62 @@
+package gas
+
+import (
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/exitcode"
+)
+
+// Unit is the unit of gas.
+// Dragons: move the gasunits to be defined here.
+type Unit = types.GasUnits
```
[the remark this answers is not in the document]
icorderi: Was planning on re-exporting all the externally needed types in `vm/vm.go`.
It might not be idiomatic to go, I'm open to other suggestions.

internal/pkg/vm/internal/gas/gas.go
```diff
@@ -0,0 +1,62 @@
...
+
+// SystemGasLimit is the maximum gas for implicit system messages.
+var SystemGasLimit = (Unit)(types.NewGasUnits(10 ^ 18))
```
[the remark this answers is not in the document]
icorderi: Sure, I'll add it, but shouldn't compilation had already blown up if that was the case. 
That line is defined at module level. Is go doing lazy evaluation on that?

internal/pkg/vm/internal/gas/gas.go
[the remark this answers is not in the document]
icorderi: I changed it to `exitcode.AbortWithCode(exitcode.OutOfGas)`

internal/pkg/vm/internal/storage/storage2.go
```diff
@@ -0,0 +1,176 @@
...
+	return VMStorage{
+		blockstore:       bs,
+		writeBuffer:      map[cid.Cid]ipld.Node{},
+		readCache:        map[cid.Cid]blocks.Block{},
+		readCacheEnabled: false,
+	}
+}
+
+// WithReadCache enable/disables the read chache.
+func (s VMStorage) WithReadCache(enabled bool) VMStorage {
```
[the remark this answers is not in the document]
icorderi: I prefer `with` when returning the object for chaining. 
I changed it a setter if it feels more idiomatic.

internal/pkg/vm/internal/message/result.go
```diff
@@ -0,0 +1,54 @@
...
+func Failure(exitCode exitcode.ExitCode) Receipt {
+	return Receipt{
+		ExitCode:    exitCode,
+		ReturnValue: nil,
+		GasUsed:     types.ZeroGas,
+	}
+}
+
+// WithGas sets the gas used.
+func (r Receipt) WithGas(amount types.GasUnits) Receipt {
```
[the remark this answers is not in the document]
icorderi: It does leave the original unchanged, that receiver is by value.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,308 @@
...
+	// 3. short-circuit _Send_ method
+	// 4. load target actor
+	// 5. load target actor code
+	// 6. create target state handle
+
+	// 1. charge gas for msg
+	ctx.gasTank.Charge(gascost.OnMethodInvocation(&ctx.msg))
+
+	// 2. transfer funds carried by the msg
+	ctx.rt.transfer(ctx.msg.from, ctx.msg.to, ctx.msg.value)
```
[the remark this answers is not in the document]
icorderi: Good catch. Fixed. 

Re: have the to resolved outside. I had it like that originally, but then the "Send" needs it too, so I moved it inside the invoke which is common code to top level messages and sub messages. Since it is now abstracted to a `resolveTarget` method, we could have that call duplicated in both execution paths. 

Re: asserting on address. Sure thing.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,308 @@
...
+	ctx.stateHandle = &stateHandle
+
+	// dispatch
+	// 2. check method exists
+	// 3. invoke method on actor
+
+	// 2. check method exists
+	exportedFn, ok := makeTypedExport(actorImpl, ctx.msg.method)
+	if !ok {
+		panic(exitcode.InvalidMethod)
```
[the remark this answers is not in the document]
icorderi: changed to `exitcode.AbortWithCode(exitcode.InvalidMethod)`

internal/pkg/vm/internal/vmcontext/invocation_context.go
icorderi: Quite a few things changed since mid December, I'll do another pass on a separate PR to bring it up to date again.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: fixed

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: Yea, invalid as in differ from other impls. 

That `GetActor` could be possible hitting disk, in which case an IO error in that path by other implementations might not be surfaced as a `ActorNotFound` exitcode which will result in a receipt being generated.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: I'll have to to another pass on a separate PR to make sure everything matches the spec

internal/pkg/vm/internal/gas/gas.go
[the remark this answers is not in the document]
anorth: Ok sounds good

anorth:
A follow-up PR for spec sync sounds fine, and not urgent.

Ping me when you've pushed change.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: We can merge the two panics objects and have `runtime.Abort` just be the same as doing `runtime.AbortWithCode(exitcodes.Abort, "blah"). 

But I would rather do that on a separate PR at this point.

internal/pkg/vm/internal/interpreter/interpreter.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/message"
+)
+
+// VMInterpreter orchestrates the execution of messages from a tipset on that tipset’s parent state.
+type VMInterpreter interface {
+	// ApplyTipSetMessages applies all the messages in a tipset.
+	//
+	// Note: this method does not error, any processing error will be present as an `ExitCode` in the `MessageReceipt`.
```
anorth:
```suggestion
	// Note: any message processing error will be present as an `ExitCode` in the `MessageReceipt`.
	// An error indicates a non-deterministic failure (e.g. disk read failure).
```

internal/pkg/vm/internal/vmcontext/invocation_context.go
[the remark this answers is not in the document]
anorth: Ok, assertions are fine.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,311 @@
...
+	ctx.rt.transfer(ctx.msg.from, ctx.msg.to, ctx.msg.value)
+
+	// 4. if we are just sending funds, there is nothing else to do.
+	if ctx.msg.method == types.SendMethodID {
+		return message.Ok().WithGas(ctx.gasTank.GasConsumed())
+	}
+
+	// 5. load target actor code
+	// TODO: use chain height based protocol version here (#3360)
+	actorImpl := ctx.rt.getActorImpl(ctx.toActor.Code)
```
anorth: Perhaps pass the height as a parameter here so we can hook it up later.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,311 @@
...
+	// 2. build internal message
+
+	// 1. fromActor = executing toActor
+	from := ctx.msg.to
+	fromActor := ctx.toActor
+
+	// 2. build internal message
+	encodedParams, err := abi.ToEncodedValues(params...)
+	if err != nil {
+		// Review: should this be a recoverable or unrecoverable panic?
```
anorth: So I think for now it should be an on-chain error. The caller could pass something unencodable. This will remain true if we move the encoding up to actor code.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -0,0 +1,566 @@
...
+}
+
+// applyImplicitMessage applies messages automatically generated by the vm itself.
+//
+// This messages do not consume client gas are do not fail.
+func (vm *VM) applyImplicitMessage(imsg internalMessage) (out interface{}, err error) {
+	defer func() {
+		if r := recover(); r == nil {
+			switch r.(type) {
+			case runtime.AbortPanicError:
```
anorth: Can we remove this type?

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (ctx *invocationContext) invoke() interface{} {
+	// pre-dispatch
+	// 1. charge gas for message invocation
+	// 2. transfer optional funds
+	// 3. short-circuit _Send_ method
+	// 4. load target actor
+	// 5. load target actor code
+	// 6. create target state handle
```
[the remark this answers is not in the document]
icorderi: I find that it helps to have a high level breakdown of the contents on this complex business logic methods.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
anorth: Right, I think we need to distinguish those error here. How about returning `(actor, ok, err)`, so errors are only for unrecoverables?

internal/pkg/vm/internal/vmcontext/invocation_context.go
icorderi: it already is, that method is a helper in the context, it internally passes the height.
```go
func (vm *VM) getActorImpl(code cid.Cid) dispatch.ExecutableActor {
	actorImpl, err := vm.actorImpls.GetActorImpl(code, vm.currentEpoch)
	if err != nil {
		exitcode.AbortWithCode(exitcode.ActorCodeNotFound)
	}
	return actorImpl
}
```

icorderi: I think the comment is a copy/pasta from the legacy code. I'll remove it.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: Done, merged the two into:
```go
// ExecutionPanic is used to abort vm execution with an exit code.
type ExecutionPanic struct {
	msg  string
	code exitcode.ExitCode
}
```

And left this two methods: 
```go
func Abort(code exitcode.ExitCode){}

func Abortf(code exitcode.ExitCode, msg string, args ...interface{}){}
```

internal/pkg/vm/internal/vmcontext/invocation_context.go
icorderi: changed to:
```go
runtime.Abortf(exitcode.EncodingError, "failed to encode params. %s", err)
```

PR #3722, opened 2020-01-09
Feat/epost prep

ZenGround0:
### Motivation
If we plan to integrate election post right now we need to fake out the underlying proofs system (go-sectorbuilder repo) and the soon to be deprecated vm (miner and proving set will soon be gone).  This PR does this faking out.

### Proposed changes
I also added a simple `Hasher` util because I was tired of manipulating bytes buffers every time I want to hash something.

### DIFF
--- a/internal/pkg/consensus/election.go
+++ b/internal/pkg/consensus/election.go
@@ -12,9 +12,39 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
 
-// CompareElectionPower return true if the input electionProof is below the
+// ElectionMachine generates and validates PoSt partial tickets and PoSt
+// proofs.
+type ElectionMachine struct{}
+
+// func (em ElectionMachine) GeneratePoStRandomness(ticket block.Ticket, candidateAddr address.Address, signer types.Signer, nullBlockCount uint64) (block.VRFPi, error) {
+// 	seedBuf := make([]byte, binary.MaxVarintLen64)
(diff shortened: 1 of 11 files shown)


internal/pkg/consensus/election.go
```diff
@@ -12,9 +12,39 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
 )
 
-// CompareElectionPower return true if the input electionProof is below the
+// ElectionMachine generates and validates PoSt partial tickets and PoSt
+// proofs.
+type ElectionMachine struct{}
+
+// func (em ElectionMachine) GeneratePoStRandomness(ticket block.Ticket, candidateAddr address.Address, signer types.Signer, nullBlockCount uint64) (block.VRFPi, error) {
```
ZenGround0: This is "Dragons".  I can add that explicitly or remove this commented out stuff right for this PR if preferred.  However this pseudo code is helpful to me and will be replaced with the real deal in a day or two.

internal/pkg/util/hasher/hasher.go
```diff
@@ -0,0 +1,44 @@
...
+	h.toHash = append(h.toHash, bs[:])
+}
+
+// Hash returns the hash value of the provided data
+func (h *Hasher) Hash() []byte {
+	var accum []byte
+	for _, s := range h.toHash {
+		accum = append(accum, s...)
+	}
+	hash := blake2b.Sum256(accum)
```
anorth: I believe that generally most of the Filecoin hashes, are switching to sha2, including the epost-related ones.

PR #3723, opened 2020-01-14
Election machine functionality for epost

ZenGround0:
### Motivation
To implement Election post we need a new set of election machine endpoints to create and verify filecoin block elections.  This PR introduces these new endpoints.

### Proposed changes
Add epost endpoints.  Correctly preprocess data from election machine function calls to provide to the epost interface.  The ticket winning condition is now up to spec.

I've been assessing how close to spec our randomness drawing and VRF generation is and it is pretty lacking.  I'm saving that change for a follow on.

I'm saving election machine unit testing until the behavior is up to spec, likely in the same followup with correct crypto operations.

### DIFF
--- a/internal/pkg/consensus/election.go
+++ b/internal/pkg/consensus/election.go
@@ -5,43 +5,96 @@ import (
 	"encoding/binary"
 	"math/big"
 
-	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/pkg/errors"
 
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/proofs"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/util/hasher"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
+	sector "github.com/filecoin-project/go-sectorbuilder"
 )
 
 // ElectionMachine generates and validates PoSt partial tickets and PoSt
 // proofs.
 type ElectionMachine struct{}
 
-// Dragons -- turn this pseudo code real
-// func (em ElectionMachine) GeneratePoStRandomness(ticket block.Ticket, candidateAddr address.Address, signer types.Signer, nullBlockCount uint64) (block.VRFPi, error) {
-// 	seedBuf := make([]byte, binary.MaxVarintLen64)
-// 	n := binary.PutUvarint(seedBuf, nullBlockCount)
-// 	buf := append(ticket.VRFProof, seedBuf[:n]...)
+// GeneratePoStRandomness returns the PoStRandomness for the given epoch.
+func (em ElectionMachine) GeneratePoStRandomness(ticket block.Ticket, candidateAddr address.Address, signer types.Signer, nullBlockCount uint64) ([]byte, error) {
+	seedBuf := make([]byte, binary.MaxVarintLen64)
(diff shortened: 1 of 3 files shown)


internal/pkg/consensus/election.go
```diff
@@ -5,43 +5,96 @@ import (
...
+// VerifyPoSt verifies a PoSt proof.
+func (em ElectionMachine) VerifyPoSt(ctx context.Context, ep *proofs.ElectionPoster, allSectorInfos sector.SortedSectorInfo, sectorSize uint64, challengeSeed []byte, proof []byte, candidates []*proofs.EPoStCandidate, proverID address.Address) (bool, error) {
+	// filter down sector infos to only those referenced by candidates
+	candidateSectorID := make(map[uint64]struct{})
+	for _, candidate := range candidates {
+		candidateSectorID[candidate.SectorID] = struct{}{}
+	}
+	var candidateSectorInfos []sector.SectorInfo
+	for _, si := range allSectorInfos.Values() {
+		candidateSectorInfos = append(candidateSectorInfos, si)
```
acruikshank: Shouldn't you check the `candidateSectorID` map, before appending `si` here?

internal/pkg/proofs/election_poster.go
```diff
@@ -8,6 +8,9 @@ import (
 	sector "github.com/filecoin-project/go-sectorbuilder"
 )
 
+// SectorChallengeRatioDiv is
```
icorderi: finish comment?

internal/pkg/proofs/election_poster.go
```diff
@@ -51,3 +54,13 @@ func (ep *ElectionPoster) GenerateEPostCandidates(sectorInfo sector.SortedSector
...
 }
+
+// ElectionPostChallengeCount is the total number of partial tickets allowed by
+// the system
+func (ep *ElectionPoster) ElectionPostChallengeCount(sectors, faults uint64) uint64 {
+	if sectors-faults == 0 {
+		return 0
+	}
+	// ceil(sectors / SectorChallengeRatioDiv)
+	return (sectors-faults-1)/SectorChallengeRatioDiv + 1
```
icorderi: Does this pass gofmt? Can we place some spaces between the `-` and also the `/`.

internal/pkg/consensus/election.go
```diff
@@ -5,43 +5,96 @@ import (
...
+	numSectorsSampled := ep.ElectionPostChallengeCount(sectorNum, faultNum)
+
+	lhs := new(big.Int).SetBytes(challengeTicket[:])
+	lhs = lhs.Mul(lhs, big.NewInt(int64(networkPower)))
+	lhs = lhs.Mul(lhs, big.NewInt(int64(numSectorsSampled)))
+
+	// sectorPower * 2^len(H)
+	rhs := new(big.Int).Lsh(big.NewInt(int64(sectorSize)), challengeBits)
+	rhs = rhs.Mul(rhs, big.NewInt(int64(sectorNum)))
+	rhs = rhs.Mul(rhs, big.NewInt(expectedLeadersPerEpoch))
```
acruikshank: It would be great to have the full inequality in a comment. I interpret it as ticket/ticket size < E*sectors/(total power*sectors sampled). Is that close?

internal/pkg/consensus/election.go
ZenGround0: thanks!  this is what I get for not unit testing with the changes

internal/pkg/consensus/election.go
ZenGround0: I will add it.  That is pretty much it.

Issue #3703, opened 2019-12-10
Mining worker calls into election post proof endpoint


ZenGround0:
### Description
Currently the mining worker sleeps a blocktime.  I'm changing this in timing work to put this wait inside the mining scheduler where it belongs.  In the near future we are going to need to put time bound post proving work inside the mining worker.

### Acceptance criteria
Election post requires integrating calls into post proof generation within the mining worker.  This issue tracks that

anorth:
@ZenGround0 done?

PR #3724, opened 2020-01-15
Feat/gen epost on mine

ZenGround0:
### Motivation
To move to election post we need our mining worker to call the new election machine generate / candidate win check endpoints.  We also need to pass this data into the block header when creating a block after winning.

### Proposed changes
This PR forces the miner worker to make the new epost calls and create epost data.  When blocks are created they include the correct epost data.

To pull this off and maintain test coverage I had to tweak the power actor very slightly and expand the power table's fake behavior.

This PR expands the ugly code in the miner worker and block generator which are in need of a good cleanup.  I decided against cleanup here but welcome suggestions on how to trim down tech debt introduced here within this PR.

I've also held off on testing.  I believe that testing at a higher level once we totally remove the deprecated election mechanism and add in consensus checks is going to be more useful than unit testing this intermediate stage.

Closes issue #3703

### DIFF
--- a/internal/pkg/mining/worker.go
+++ b/internal/pkg/mining/worker.go
@@ -17,11 +17,14 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/chain"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/proofs"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/sampling"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/util/hasher"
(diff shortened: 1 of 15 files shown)


internal/pkg/mining/worker.go
```diff
@@ -247,16 +236,108 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
 		outCh <- Output{Err: err}
 		return
 	}
-	weHaveAWinner, err := w.election.DeprecatedIsElectionWinner(ctx, powerTable, electionTicket, nullBlkCount, electionProof, workerAddr, w.minerAddr)
+	sortedSectorInfos, err := powerTable.SortedSectorInfos(ctx, w.minerAddr)
```
icorderi: this entire method could use a bit more grouping of sections of code and comments

ZenGround0: 10,000% yes
I'm hoping to clean this up in the near future.

ZenGround0: The error handling is absolute trash too

PR #3725, opened 2020-01-16
Feat/epost for election

ZenGround0:
### Motivation
We want to run our elections running epost so this PR runs our elections using epost.

### Proposed changes
- Remove deprecated election functions and their use as well as the data field in the block header and the tests (need to remember to go back and use similar test structure on new election machine functions)
- validation now checks for a winning election post ticket in order to accept a block
- all testing dodads are good

With this PR Epost is now more or less a part of the node.  I still need to add a few subtle validation rules and small bug fixes and tweak the VRF/randomness stuff a bit.  I also want to add in proper testing and potentially do some structural cleanups of validation and maybe mining to make testing and configurability less bad there.

Closes issue #


internal/pkg/proofs/election_poster.go
```diff
@@ -19,6 +20,20 @@ type EPoStCandidate struct {
 	SectorChallengeIndex uint64
 }
 
+// ZipEPoStCandidates creates a slice of EPoSt candidates give slices for the fields
```
acruikshank: give -> given

PR #3735, opened 2020-01-30
feat/remove actors

icorderi:
### Motivation

Actors are moving out of repo.

### Proposed changes

- remove actors one by one


functional-tests/network-deployment/miner_workflow_test.go
```diff
@@ -1,186 +0,0 @@
...
-	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
-	minerActor "github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor/builtin/miner"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
-	"github.com/filecoin-project/go-filecoin/tools/fast"
-	"github.com/filecoin-project/go-filecoin/tools/fast/fastesting"
-	"github.com/filecoin-project/go-filecoin/tools/fast/series"
-)
-
-func TestMinerWorkflow(t *testing.T) {
-	network := tf.DeploymentTest(t)
```
anorth: This test was designed to run against live networks to verify automated deploy processes. Just noting that we'll probably want to bring something back later (but it'll probably look pretty different). cc @ZenGround0 @travisperson

internal/app/go-filecoin/plumbing/msg/previewer_test.go
```diff
@@ -1,60 +0,0 @@
...
-	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor/builtin"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
-
-func TestPreview(t *testing.T) {
```
anorth: It's not clear why this test has to go. It uses fake actors and everything. We probably will want to retain this functionality (the user's ability to preview a message execution). Could you attempt to restore it?

internal/pkg/types/constants.go
```diff
@@ -62,8 +56,6 @@ func init() {
 	StorageMarketActorCodeCid = StorageMarketActorCodeObj.Cid()
```
anorth: Aside: these should be replaced later with values imported from specs-actors

PR #3736, opened 2020-02-01
feat: vm integration

icorderi:
### Motivation

We need to integrate the new VM runtime with the rest of the system.

I plan on the next PR to delete the legacy `VMContext` and remove the "legacy methods" from the signatures, that should force us to fix some of the methods outside the processor using the legacy code and bring _some_ of the skipped tests back. 

### Proposed changes

Closes issue #


internal/app/go-filecoin/plumbing/msg/waiter_test.go
```diff
@@ -51,7 +51,7 @@ func setupTest(t *testing.T) (*hamt.CborIpldStore, *chain.Store, *chain.MessageS
 
 func TestWait(t *testing.T) {
 	tf.UnitTest(t)
-
+	t.Skip("this guy is waiting forever")
```
ZenGround0: I'm worried that tests like this that exercise parts of the system adjacent to but decoupled from the vm runtime should actually be working after this PR and that we are unnecessarily breaking functionality.  I'd be a lot less worried if there was a description of why tests are broken and after what work they can be expected to be unskipped.

internal/pkg/consensus/expected.go
```diff
@@ -302,23 +302,42 @@ func (c *Expected) validateMining(
...
-	results, err := c.processor.ProcessTipSet(ctx, st, vms, ts, allMessages, ancestors)
+	// build message information per block
+	for i := 0; i < ts.Len(); i++ {
+		blk := ts.At(i)
+
+		msgInfo := vm.BlockMessagesInfo{
+			BLSMessages:  blsMessages[i],
+			SECPMessages: secpMessages[i],
+			Miner:        blk.Miner,
+		}
```
ZenGround0: 👁 🌅 👁

internal/pkg/consensus/genesis.go
```diff
@@ -144,12 +145,18 @@ func NewEmptyConfig() *Config {
 	}
 }
 
+// GenesisVM is the view into the VM used during genesis block creation.
+type GenesisVM interface {
+	ApplyGenesisMessage(from address.Address, to address.Address, method types.MethodID, value types.AttoFIL, params ...interface{}) (interface{}, error)
```
ZenGround0: 👍

internal/pkg/consensus/genesis.go
```diff
@@ -188,15 +195,7 @@ func MakeGenesisFunc(opts ...GenOption) GenesisInitFunc {
...
-			s := storageMap.NewStorage(addr, a)
-			scid, err := s.Put(val.state)
-			if err != nil {
-				return nil, err
-			}
-			if err = s.LegacyCommit(scid, a.Head); err != nil {
-				return nil, err
-			}
-			if err := st.SetActor(ctx, addr, a); err != nil {
+			if err := st.SetActor(context.Background(), addr, a); err != nil {
```
ZenGround0: I think that this change is you regressing genesis behavior of actually setting up miner actors as specified in the genCfg in order to move to the new state machine.  This is fine, but you should leave a Dragons saying: "actually write miner actor storage" or something so that we don't have to track this down later.

ZenGround0: Reading your `createActor` in `SetupDefaultActors` function below I'm a little confused why you don't abstract that to a top level helper and then run it here on the miner actor state from the genesis config.

internal/pkg/mining/block_generate.go
```diff
@@ -61,53 +55,26 @@ func (w *DefaultWorker) Generate(
...
 	candidateMsgs := orderMessageCandidates(mq.Drain())
 
-	// run state transition to learn which messages are valid
-	vms := vm.NewStorageMap(w.blockstore)
-	results, err := w.processor.ApplyMessagesAndPayRewards(ctx, stateTree, vms, types.UnwrapSigned(candidateMsgs),
-		w.minerOwnerAddr, types.NewBlockHeight(blockHeight), ancestors)
-	if err != nil {
-		return nil, errors.Wrap(err, "generate apply messages")
-	}
+	// Dragons: ask something to select and order messages to include
```
ZenGround0: Good direction and I like how it removes a bunch of dependencies from the mining worker.

internal/pkg/mining/worker.go
```diff
@@ -72,8 +71,7 @@ type MessageSource interface {
 
 // A MessageApplier processes all the messages in a message pool.
 type MessageApplier interface {
-	// ApplyMessagesAndPayRewards applies all state transitions related to a set of messages.
-	ApplyMessagesAndPayRewards(ctx context.Context, st state.Tree, vms vm.StorageMap, messages []*types.UnsignedMessage, minerOwnerAddr address.Address, bh *types.BlockHeight, ancestors []block.TipSet) ([]*consensus.ApplyMessageResult, error)
+	// Dragons: add something back or remove
```
ZenGround0: To be consistent with your other Dragons, this should be replaced with an interface that asks for new messages and gets back valid messages (by running this on the implementation)

internal/pkg/testhelpers/consensus.go
```diff
@@ -190,29 +173,28 @@ func ApplyTestMessageWithActors(actors builtin.Actors, st state.Tree, store vm.S
...
 }
 
-func newMessageApplier(msg *types.UnsignedMessage, processor *consensus.DefaultProcessor, st state.Tree, storageMap vm.StorageMap,
-	bh *types.BlockHeight, minerOwner address.Address, ancestors []block.TipSet) (*consensus.ApplicationResult, error) {
-	amr, err := processor.ApplyMessagesAndPayRewards(context.Background(), st, storageMap, []*types.UnsignedMessage{msg}, minerOwner, bh, ancestors)
-	if err != nil {
-		return nil, err
-	}
+func newMessageApplier(msg *types.UnsignedMessage, processor *consensus.DefaultProcessor, st state.Tree, storageMap vm.StorageMap, bh *types.BlockHeight, minerOwner address.Address, ancestors []block.TipSet) (*consensus.ApplicationResult, error) {
+	// amr, err := processor.ApplyMessagesAndPayRewards(context.Background(), st, storageMap, []*types.UnsignedMessage{msg}, minerOwner, bh, ancestors)
```
ZenGround0: Dragons or TODO explaining what happens here?

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -65,15 +66,80 @@ type actorStorage struct {
...
+	if !ok {
+		panic("TODO")
+	}
+
+	return targetIDAddr
+}
+
 // implement VMInterpreter for VM
 
 var _ interpreter.VMInterpreter = (*VM)(nil)
```
ZenGround0: Saw this package for the first time, not this PR but note that `makeEpostMessage` and implicit apply is not how we're doing things so you can delete it.

internal/pkg/consensus/processor.go
```diff
@@ -64,28 +41,25 @@ type ApplyMessageResult struct {
 
 // DefaultProcessor handles all block processing.
 type DefaultProcessor struct {
-	validator     MessageValidator
-	blockRewarder BlockRewarder
-	actors        builtin.Actors
+	validator MessageValidator
```
ZenGround0: I don't see this getting called anywhere, is the validation done elsewhere in the new vm runtime?  If not I think this is a bug and we should introduce a call here.  If so I think this dependency can be deleted.

internal/pkg/consensus/validation.go
```diff
@@ -19,6 +19,21 @@ var errInsufficientGasCt *metrics.Int64Counter
 var errNonceTooLowCt *metrics.Int64Counter
 var errNonceTooHighCt *metrics.Int64Counter
 
+var (
+	// These errors are only to be used by ApplyMessage; they shouldn't be
```
ZenGround0: Since ApplyMessage is now deleted should these also be deleted?

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -295,12 +363,13 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize uint32, mi
```
ZenGround0: I'm not high context enough to start making incorrectness claims, but it would be worth taking a look at consensus/validation.go to make sure we are doing all the necessary checks here.  It looks like this covers most but not all of the checks we used to do on message validation.  Missing some checks allowing high nonce, negative value, block gas limit.  

No idea if this is more up to date, just flagging.

internal/pkg/consensus/processor_test.go
```diff
@@ -1,911 +1,911 @@
...
-	return addr1, act1, addr2, act2, st, vms, mockSigner
-}
-
-func mustCreateStorageMiner(ctx context.Context, t *testing.T, st state.Tree, vms vm.StorageMap, minerOwner address.Address) (cid.Cid, *actor.Actor, address.Address) {
-	miner, minerAddr := th.RequireNewMinerActor(ctx, t, st, vms, minerOwner, 1000, th.RequireRandomPeerID(t), types.ZeroAttoFIL)
-	stCid, err := st.Flush(ctx)
-	require.NoError(t, err)
-	return stCid, miner, minerAddr
-}
+// import (
```
ZenGround0: Would love a TODO or Dragons on us reinstating a lot of these test cases once we wire up the new runtime / get new actors in place.  Anything to tell us when we can expect to resurrect these.

internal/pkg/vm/actor/builtin/account/account.go
```diff
@@ -70,27 +67,27 @@ func (a *Actor) Method(id types.MethodID) (dispatch.Method, *dispatch.FunctionSi
 }
 
 // InitializeState for account actors does nothing.
-func (*Actor) InitializeState(storage runtime.LegacyStorage, initializerData interface{}) error {
-	state, ok := initializerData.(*State)
+func (*Actor) InitializeState(handle runtime.ActorStateHandle, initializerData interface{}) error {
```
ZenGround0: Question: is this code staying after we import new actor code from elsewhere?

internal/pkg/vm/internal/vmcontext/vmcontex2.go
```diff
@@ -65,15 +66,80 @@ type actorStorage struct {
...
+	// commit new actor state
+	if err := vm.state.Commit(context.Background()); err != nil {
+		return nil, err
+	}
+	// TODO: update state root (issue: #3718)
+
+	return ret, nil
+}
+
+func (vm *VM) normalizeFrom(from address.Address) address.Address {
```
ZenGround0: Question: do we plan to move to this from the init actor lookup stuff that used to live in the processor across the board?

internal/pkg/vm/testing.go
```diff
@@ -278,6 +278,15 @@ func (ts *testStorage) Get(cid cid.Cid, obj interface{}) bool {
 	return true
 }
 
+func (ts *testStorage) GetRaw(cid cid.Cid) ([]byte, bool) {
+	node, err := cbor.WrapObject(ts.state, types.DefaultHashFunction, -1)
```
ZenGround0: No matter what we do here it will need more attention so nbd.  But you could help me out a bit by grabbing bytes from encoding.Encode and then passing them to [`cbor.Decode`](https://github.com/ipfs/go-ipld-cbor/blob/master/node.go#L95) to make an ipld node directly around bytes.

tools/gengen/util/gengen.go
```diff
@@ -121,31 +119,32 @@ func GenGen(ctx context.Context, cfg *GenesisCfg, cst *hamt.CborIpldStore, bs bl
...
-	miners, err := setupMiners(st, storageMap, keys, cfg.Miners, pnrg)
+	miners, err := setupMiners(vm, st, keys, cfg.Miners, pnrg)
 	if err != nil {
 		return nil, err
 	}
 
-	stateRoot, err := st.Flush(ctx)
+	err = store.Flush()
 	if err != nil {
 		return nil, err
```
ZenGround0: This reordering is a nice catch. Weird this has always worked.  I guess small genesis size and hamt width has saved us so far?

internal/pkg/consensus/processor.go
icorderi: I'll delete it for now

internal/pkg/consensus/validation.go
icorderi: I deleted the unused, there are still references to this somewhere. 
Making a pass across the entire codebase to delete all the unsed/legacy code will take some effort

internal/pkg/vm/actor/builtin/account/account.go
icorderi: It's not relevant. 

The purpose of this got replaced by the constructor method on the spec. 
We still have the constructor wired to call this in our code since some tests and the gengen were calling this method directly. 

I haven't looked at the new actor code in the repo, so I can't tell you if the constructor method is building the state directly or calling another method to build the state. But it is now something internal to each actor and not needed by the VM.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: No, this is internal to the VM. 
The reason some places were doing the processor->init-lookup call to go between ID's was to load/set the state for an actor on the outside of the VM. 

The VM layer should provide a wrapper for the tree that lets you grab a view of the state of any actor. That wrapper can can figure out based on the address if it needs to normalize it or not. But that code lives inside the vm and nothing outside needs to worry about what address types we might be using to index the actors.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: we need to do a cleaning pass on all the circular dependencies of deprecated code that makes pieces look alive when they are not needed.

internal/pkg/vm/internal/vmcontext/vmcontex2.go
icorderi: This is nearly verbatim the spec up to dec 12th. 
There are some new changes now, so this needs to get revised. 

FWIW, any check not here and not in the spec I think its expected to happen BEFORE the messages are selected for execution. What I think of as "Message validation" happens way before here.

internal/pkg/vm/testing.go
icorderi: Let's talk offline

tools/gengen/util/gengen.go
icorderi: There are a few things in this file that fall under the "not sure how this is working" category.

cmd/go-filecoin/address_integration_test.go
```diff
@@ -44,6 +44,7 @@ func TestAddrsNewAndList(t *testing.T) {
 
 func TestWalletBalance(t *testing.T) {
 	tf.IntegrationTest(t)
+	t.Skip("not working")
```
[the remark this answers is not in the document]
icorderi: Long answer offline. 
Short answer, those are mostly all integration tests that can be turned back on after deleting the vmcontext and hooking tests helpers to use the new runtime.

cmd/go-filecoin/miner_daemon_test.go
```diff
@@ -263,54 +260,55 @@ func requireMinerCreate(ctx context.Context, t *testing.T, env *fastesting.TestE
...
-	d1.MineAndPropagate(time.Second, d)
-	wg.Wait()
-
-	expectedBlockReward := consensus.NewDefaultBlockRewarder().BlockRewardAmount()
-	expectedPrice := types.NewAttoFILFromFIL(333)
-	expectedGasCost := big.NewInt(100)
-	expectedBalance := expectedBlockReward.Add(expectedPrice.MulBigInt(expectedGasCost))
-	newBalance := queryBalance(t, d, miningMinerOwnerAddr)
-	assert.Equal(t, expectedBalance.String(), newBalance.Sub(startingBalance).String())
+	t.Skip("new runtime")
```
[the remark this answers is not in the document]
icorderi: For the most part it means rewrite or delete.
Some of this tests were using features we n longer have because the runtime no longer needs. 
We need to revisit most of them and case by case analyze WHAT is being tested, and then determine how to test that in the new world.

internal/app/go-filecoin/node/message_propagate_test.go
```diff
@@ -23,6 +23,7 @@ import (
 // pools of connected ndoes.
 func TestMessagePropagation(t *testing.T) {
 	tf.UnitTest(t)
+	t.Skip("its using the old vmcontext")
```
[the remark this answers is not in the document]
icorderi: this ones are mostly trivial, somewhere down the line they are using the old vmcontext through the vm/testhelpers. That will all go away on the next PR when I delete the legacy VM code.

internal/app/go-filecoin/plumbing/msg/waiter_test.go
icorderi: Casualty of deleting a lot of the old processor. 
Since the dedup of messages is now done in the VM interpreter I had deleted the "DedupMessages" method from the processor since it was no longer used.

I brought dedupMessages back as a private method on the waiter since nothing else uses that method.

The test has been re-enabled.

internal/pkg/consensus/genesis.go
icorderi: Short answer, most of this file needs to get re-written once we import the new actor code.

internal/pkg/mining/worker.go
icorderi: its test code that will get nuked on the next PR once the legacy vmcontext goes away.

internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -209,37 +208,40 @@ func (w *Waiter) receiptForTipset(ctx context.Context, ts block.TipSet, targetMs
 }
 
 func (w *Waiter) receiptByIndex(ctx context.Context, tsKey block.TipSetKey, targetCid cid.Cid, messages [][]*types.UnsignedMessage) (*types.MessageReceipt, error) {
-	receiptCid, err := w.chainReader.GetTipSetReceiptsRoot(tsKey)
```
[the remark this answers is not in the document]
icorderi: resurrected, long answer on the test file.

cmd/go-filecoin/miner_daemon_test.go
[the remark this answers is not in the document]
ZenGround0: Oh looking at this again it was already skipped with a description of how and when to address on line 261

internal/pkg/vm/internal/vmcontext/vmcontex2.go
ZenGround0: Yeah that naming suits me more too but consensus/validation.go contains the old version of these rules 🤷‍♂

PR #3737, opened 2020-02-03
Integrate storage market provider from go-fil-markets

### LINKED ISSUE #3719: Switch to shared address module
### Description
There's a new address module in https://github.com/filecoin-project/go-address intended to be shared by implementations. It's used by the shared components and we'll need to update to it in order to integrate them.

The code is very similar to our existing address type, with a few fixes from when it was copied to Lotus.

### Acceptance criteria
- Address type removed in favour of importing it

### Risks + pitfalls
- Interface may have changed slightly.

acruikshank:
### Motivation

The `go-fil-markets` project implements a spec compliant version of the storage market. We would like to use this code in go-filecoin to bring our storage market operations up to spec.

### Proposed changes

This PR brings in go-fil-markets and wires it into go-filecoin by implementing the storage provider API and storage client API. Wiring up the retrieval market or the retrieval client is out of scope for this PR, and some retrieval code has been commented out to make this PR work. Since we lack the actor support to make the storage market work, a fully operational storage market is also out of scope. Consequently, we skipped and commented out some tests to allow us to continue building prior to making the storage market functional. We would eventually like these tests to. work with the new code.

As a consequence of integrating go-fil-markets, we upgrade ipld-prime and graphsync. This has various ramifications throughout the code.

closes  #3719


internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -2,22 +2,105 @@ package submodule
...
+		return nil, err
+	}
+
+	fs, err := filestore.NewLocalFileStore(filestore.OsPath(pieceStagingPath))
+	if err != nil {
+		return nil, err
+	}
+
+	// ensure miner-address is in datastore because graphsyncimpl requires it
+	err = ds.Put(datastore.NewKey("miner-address"), minerAddr.Bytes())
```
ZenGround0: This seems quite bad.  If this is a necessary parameter for running a graphsync impl then why not pass this as a parameter to `NewGraphSyncDataTransfer`? 

At the very least the piece datastore should be separated to avoid overwrite deletion by other parts of the application which have no business knowing about this global requirement.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -2,22 +2,105 @@ package submodule
...
+	if err != nil {
+		return nil, err
+	}
+
+	graphsyncNetwork := gsnet.NewFromLibp2pHost(h)
+	bridge := ipldbridge.NewIPLDBridge()
+	loader := gsstoreutil.LoaderForBlockstore(bs)
+	storer := gsstoreutil.StorerForBlockstore(bs)
+	gsync := graphsync.New(ctx, graphsyncNetwork, bridge, loader, storer)
```
ZenGround0: I'm suspicious we are going to get into a lot of trouble starting two graphsync protocol instances on the same host and datastore.  Even if we don't do you think it would be better practice for us to use the same graphsync protocol?  Is this separation intended to prevent interference between chain sync and piece transfer?

internal/app/go-filecoin/node/block_propagate_test.go
```diff
@@ -42,7 +41,7 @@ func requireMineOnce(ctx context.Context, t *testing.T, minerNode *Node) *block.
 	worker, err := minerNode.CreateMiningWorker(ctx)
 	require.NoError(t, err)
 
-	// Miner should win first election as it has all the power so only
+	// Provider should win first election as it has all the power so only
```
ZenGround0: nit: change back to Miner which makes more sense in this context.  It's making blocks not deals.

internal/app/go-filecoin/node/block_propagate_test.go
```diff
@@ -154,8 +153,6 @@ func makeNodesBlockPropTests(t *testing.T, numNodes int) (address.Address, []*No
 	minerNode := builder.Build(ctx)
 	seed.GiveKey(t, minerNode, 0)
 	mineraddr, _ := seed.GiveMiner(t, minerNode, 0)
-	_, err := storage.NewMiner()
```
ZenGround0: unskip this test per comment 106!

ZenGround0: Also unskip `TestBlockPropsManyNodes` neither of these is expected to fail anymore.

internal/app/go-filecoin/node/node.go
```diff
@@ -431,12 +430,34 @@ func (node *Node) setupStorageMining(ctx context.Context) error {
...
+		node.Wallet.Wallet,
+		node.Host(),
+		node.Repo.Datastore(),
+		node.Blockstore.Blockstore,
+		repoPath,
+		node.PorcelainAPI.MinerGetWorkerAddress)
+	if err != nil {
+		return errors.Wrap(err, "error initializing storage protocol")
+	}
+
```
ZenGround0: If we're writing TODOs here another one is to separate the module creation out of the node setup mining function.  The module should be created in the builder and then started here.

internal/app/go-filecoin/node/node.go
```diff
@@ -518,7 +546,7 @@ func (node *Node) StopMining(ctx context.Context) {
 		node.BlockMining.MiningDoneWg.Wait()
 	}
 
-	// TODO: stop node.StorageProtocol.StorageMiner
+	// TODO: stop node.StorageProtocol.StorageProvider
```
ZenGround0: Shouldn't we keep both TODOs?  AFAICT the storage miner is a separate component that will need to be stopped independently of stopping the storage protocol provider.

internal/app/go-filecoin/node/node.go
```diff
@@ -568,7 +596,7 @@ func (node *Node) setupProtocols() error {
 	// DRAGONS: uncomment when we have a storage market
 	//smc := storage.NewClient()
 	//smcAPI := storage.NewAPI(smc)
-	//node.StorageProtocol.StorageAPI = &smcAPI
+	//node.StorageProtocol.StorageClientAPI = &smcAPI
```
ZenGround0: I am confused with the state we are leaving this in.  It looks like the commented out code contradicts the submodule initialization code above.  Ideally this inconsistency gets resolved in this PR by deciding on an approach and removing hints towards other approaches (i.e. this comment if it makes sense to).

internal/app/go-filecoin/node/node_test.go
```diff
@@ -157,8 +156,6 @@ func TestNodeStartMining(t *testing.T) {
 	seed.GiveMiner(t, minerNode, 0) // TODO: update to accommodate new go-fil-markets integration
 	// Start mining give error for fail to get miner actor from the heaviest tipset stateroot
 	assert.Contains(t, minerNode.StartMining(ctx).Error(), "failed to setup mining")
-	_, err := storage.NewMiner()
```
ZenGround0: Unskip this test and get it passing

cmd/go-filecoin/client.go
```diff
@@ -125,34 +122,36 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
...
-			return err
-		}
-
-		resp, err := GetStorageAPI(env).ProposeStorageDeal(req.Context, data, miner, askid, duration, allowDuplicates)
-		if err != nil {
-			return err
-		}
-
-		return re.Emit(resp)
+		panic("TODO: go-fil-markets integration")
```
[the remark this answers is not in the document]
acruikshank: Good point.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -2,22 +2,105 @@ package submodule
...
+
+	ma, err := address.NewFromBytes(minerAddr.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	ca, err := address.NewFromBytes(clientAddr.Bytes())
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
acruikshank: This is filecoin internal to go-address address translation. It should go away when we integrate go-address. (If I'm understanding your question correctly).

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: agreed. I'll see if I can get it changed in go-fil-markets. It's probably just a convenience coming from lotus.

acruikshank: https://github.com/filecoin-project/go-fil-markets/pull/108

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: This was just added here because it it inaccessible from the Syncer. That's not really a good excuse. I'll work on breaking it out so it can be shared.

internal/app/go-filecoin/storage_market_connector/client.go
```diff
@@ -0,0 +1,229 @@
...
+// Adapted from https://github.com/filecoin-project/lotus/blob/3b34eba6124d16162b712e971f0db2ee108e0f67/markets/storageadapter/client.go#L156
+func (s *StorageClientNodeConnector) ValidatePublishedDeal(ctx context.Context, deal storagemarket.ClientDeal) (uint64, error) {
+	// Fetch receipt to return dealId
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+	defer cancel()
+
+	var publishMsg *types.SignedMessage
+	var publishReceipt *types.MessageReceipt
+
+	err := s.waiter.Wait(waitCtx, *deal.PublishMessage, func(block *block.Block, signedMessage *types.SignedMessage, receipt *types.MessageReceipt) error {
```
ZenGround0: It feels like you actually want to use `Find` here.  You use 10 milliseconds which tells me this should error if the block with the deal message in it is not already synced by this node, which would make wait the wrong endpoint.

internal/app/go-filecoin/storage_market_connector/client.go
```diff
@@ -0,0 +1,229 @@
...
+	values, err := abi.DecodeValues(unsigned.Params, []abi.Type{abi.StorageDealProposals})
+	if err != nil {
+		return 0, err
+	}
+
+	msgProposals := values[0].Val.([]types.StorageDealProposal)
+
+	proposal := msgProposals[0] // TODO: Support more than one deal
+
+	// TODO: find a better way to do this
```
ZenGround0: Would `reflect.DeepEqual` help here?  If not seems like writing a custom `Equals` or `CompareForValidation` method on proposal would be nice.  Seems like it shouldn't have to wait for after this PR.

internal/app/go-filecoin/storage_market_connector/client.go
```diff
@@ -0,0 +1,229 @@
...
+		//proposal.Client == deal.Proposal.Client &&
+		//proposal.Provider == deal.Proposal.Provider &&
+		uint64(proposal.ProposalExpiration) == deal.Proposal.ProposalExpiration &&
+		uint64(proposal.Duration) == deal.Proposal.Duration &&
+		uint64(proposal.StoragePricePerEpoch) == deal.Proposal.StoragePricePerEpoch.Uint64() &&
+		uint64(proposal.StorageCollateral) == deal.Proposal.StorageCollateral.Uint64() &&
+		bytes.Equal([]byte(*proposal.ProposerSignature), deal.Proposal.ProposerSignature.Data)
+
+	if equals {
+		return leb128.ToUInt64(publishReceipt.Return[0]), nil // TODO: Is this correct?
```
ZenGround0: You should probably be using the abi package functions to decode this (though maybe I'm wrong that receipts are abi encoded?) or otherwise the encoding package.  Casting to leb128 varint is almost certainly wrong for where we are headed because big ints won't be encoded in leb128 anymore (and isn't this supposed to be a little int?)

internal/app/go-filecoin/storage_market_connector/client.go
ZenGround0: (If this was done to avoid annoying cid.Equals(cid) callback you could refactor to `Find` and `FindPredicate`)

internal/app/go-filecoin/storage_market_connector/client.go
```diff
@@ -0,0 +1,229 @@
+package storagemarketconnector
+
+import (
+	"bytes"
+	"context"
+	"time"
+
+	cborutil "github.com/filecoin-project/go-cbor-util"
```
ZenGround0: Call this something else to make it clear it is not gfc's internal cborutil but encoding functions that belong to external datatypes: perhaps `smcborutil`

anorth:
Drive-by opinion: I favour us moving away from porcelain, so if there's a path to deleting the duplicated concepts, 👍 for landing new ones.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: I've moved graphsync initialization into the network submodule so the syncer and storage protocol can share it.

internal/app/go-filecoin/node/block_propagate_test.go
acruikshank: These tests depend on new storage miner actor functionality that isn't currently integrated (i.e. they currently fail because we fail to retrieve valid SectorInfo structs from the miner actor). I changed the skip string to note this.

internal/app/go-filecoin/node/node_test.go
acruikshank: These tests depend on new storage miner actor functionality that isn't currently integrated (i.e. they currently fail because we fail to retrieve valid SectorInfo structs from the miner actor). I changed the skip string to note this.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: Graphsync construction has been moved to the network submodule and the graphsync exchange is now shared between the syncer and the storage protocol.

internal/app/go-filecoin/node/node.go
acruikshank: TODO added with issue #3738

internal/app/go-filecoin/node/node.go
acruikshank: Good catch. There was nothing really to stop for the storage provider. I added the code to stop the storage miner.

internal/app/go-filecoin/node/node.go
acruikshank: good call. Removed this code. May revisit later.

internal/app/go-filecoin/storage_market_connector/common.go
```diff
@@ -0,0 +1,290 @@
...
+		locked = spaabi.NewTokenAmount(0)
+	}
+
+	return storagemarket.Balance{
+		Available: tokenamount.FromInt(available.Int.Uint64()),
+		Locked:    tokenamount.FromInt(locked.Int.Uint64()),
+	}, nil
+}
+
+func (c *connectorCommon) GetMinerWorker(ctx context.Context, miner address.Address) (address.Address, error) {
```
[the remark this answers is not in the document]
acruikshank: In this case, we're actually using the plumbing method to actually get the worker. This method exists a) because it's needed to satisfy the storage provider and client api interfaces, and b) because we currently need to do address translation.

internal/app/go-filecoin/storage_market_connector/common.go
```diff
@@ -0,0 +1,290 @@
...
+
+		return receipt, nil
+	case err := <-errChan:
+		return nil, err
+	case <-ctx.Done():
+		return nil, xerrors.New("context ended prematurely")
+	}
+}
+
+func (c *connectorCommon) addFunds(ctx context.Context, from address.Address, addr address.Address, amount tokenamount.TokenAmount) error {
```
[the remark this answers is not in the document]
acruikshank: That's potentially true (note "addFunds" here actually means add storage collateral funds). This method simply puts a message on chain. It's not entirely clear if the CLI endpoint would go directly to this method or if it would interact with it indirectly through the storage miner. For that reason, it's probably best to wait before we try to share this code, but I agree in principle.

internal/app/go-filecoin/storage_market_connector/common.go
```diff
@@ -0,0 +1,290 @@
...
+	params, err := abi.ToEncodedValues(addr)
+	if err != nil {
+		return err
+	}
+
+	fromAddr, err := fcaddr.NewFromBytes(from.Bytes())
```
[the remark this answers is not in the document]
acruikshank: done

internal/app/go-filecoin/storage_market_connector/common.go
```diff
@@ -0,0 +1,290 @@
...
+	fcworker, err := c.workerGetter(ctx, fcMiner, c.chainStore.Head())
+	if err != nil {
+		return address.Undef, err
+	}
+
+	// Convert back to go-address
+	return address.NewFromBytes(fcworker.Bytes())
+}
+
+func (c *connectorCommon) getFCWorker(ctx context.Context, providerAddr address.Address) (fcaddr.Address, error) {
```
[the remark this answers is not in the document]
acruikshank: done.

internal/app/go-filecoin/storage_market_connector/client.go
acruikshank: Good point. I changed it to find.

internal/app/go-filecoin/storage_market_connector/client.go
acruikshank: done

anorth:
Thank you!

PR #3739, opened 2020-02-05
feat: remove legacy code 

icorderi:
### Motivation

During the writing of the new VM runtime, we left the old one in place.

### Proposed changes

Remove the legacy code.


internal/app/go-filecoin/plumbing/api.go
```diff
@@ -250,11 +250,14 @@ func (api *API) MessagePreview(ctx context.Context, from, to address.Address, me
 // it does not change any state. It is use to interrogate actor state. The from address
 // is optional; if not provided, an address will be chosen from the node's wallet.
 func (api *API) MessageQuery(ctx context.Context, optFrom, to address.Address, method types.MethodID, baseKey block.TipSetKey, params ...interface{}) ([][]byte, error) {
-	snapshot, err := api.actorState.Snapshot(ctx, baseKey)
-	if err != nil {
-		return [][]byte{}, err
-	}
-	return snapshot.Query(ctx, optFrom, to, method, params...)
+	// Dragons: delete
```
ZenGround0: Are you suggesting deleting the concept of message querying from plumbing (sounds like more trouble than its worth) or suggesting deleting the snapshot based implementation?

internal/app/go-filecoin/plumbing/msg/previewer.go
```diff
@@ -46,28 +42,32 @@ func NewPreviewer(chainReader previewerChainReader, cst *hamt.CborIpldStore, bs
 
 // Preview sends a read-only message to an actor.
 func (p *Previewer) Preview(ctx context.Context, optFrom, to address.Address, method types.MethodID, params ...interface{}) (types.GasUnits, error) {
-	encodedParams, err := abi.ToEncodedValues(params...)
-	if err != nil {
-		return types.NewGasUnits(0), errors.Wrap(err, "failed to encode message params")
-	}
+	// Dragons: delete
```
ZenGround0: Again is this suggesting doing away with message preview from plumbing or just this implementation of it?

internal/pkg/consensus/power_table_view.go
```diff
@@ -32,56 +27,66 @@ func NewPowerTableView(q ActorStateSnapshot) PowerTableView {
 
 // Total returns the total storage as a BytesAmount.
 func (v PowerTableView) Total(ctx context.Context) (*types.BytesAmount, error) {
-	rets, err := v.snapshot.Query(ctx, address.Undef, address.StoragePowerAddress, power.GetTotalPower)
-	if err != nil {
-		return nil, err
-	}
+	// Dragons: re-write without using query
```
ZenGround0: Why the blanket removal of snapshot.Query abstraction instead of fixing that abstraction to work with the new vm?

internal/pkg/testhelpers/consensus.go
```diff
@@ -143,41 +140,45 @@ func (ms testSigner) SignBytes(data []byte, addr address.Address) (types.Signatu
 }
 
 // RequireActorIDAddress looks up an actor address in the init actor and returns the associated id address
-func RequireActorIDAddress(ctx context.Context, t *testing.T, st state.Tree, store vm.StorageMap, addr address.Address) address.Address {
-	processor := consensus.NewConfiguredProcessor(builtin.DefaultActors)
-	params, err := abi.ToEncodedValues(addr)
-	require.NoError(t, err)
+func RequireActorIDAddress(ctx context.Context, t *testing.T, st state.Tree, store vm.Storage, addr address.Address) address.Address {
+	// Dragons: why is this needed? delete
```
ZenGround0: Since actor code testing is out of scope of gfc with the new project I suspect most of these helpers can be deleted.

internal/pkg/vm/internal/storagemap/storagemap_test.go
```diff
@@ -1,337 +0,0 @@
...
-	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/address"
-	internal "github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/errors"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
-
-func TestGetAndPutWithEmptyStorage(t *testing.T) {
```
ZenGround0: It would be nice to adapt some of these high level test conditions to the new untested storage.go stuff as there is a decent amount of conceptual overlap.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -1,718 +1,640 @@
...
-var _ ExportContext = (*VMContext)(nil)
+// Storage implements runtime.Runtime.
+func (vm *VM) Storage() runtime.Storage {
+	return actorStorage{inner: vm.store}
+}
 
-// Params implements ExportContext.
-func (ctx *VMContext) Params() []byte {
-	return ctx.message.Params
+// LegacyStorage implements runtime.Runtime.
```
ZenGround0: Do we still need this?

internal/app/go-filecoin/plumbing/api.go
icorderi: This got dropped a while a go. The actors do not have support for it. 
The way to access state is directly through each actors state object.

internal/pkg/vm/internal/vmcontext/vmcontext.go
icorderi: Yes, didn't want to peel off that layer of the onion on the same PR.

Honestly, it would be faster to hook the new actor code so we can delete all the current actors, then delete the rest of the legacy methods.

PR #3740, opened 2020-02-05
Czar/interop encode

ZenGround0:
### Motivation
We need to encode our data structures as specified in order to run our node.

### Proposed changes

This PR does not get us all the way there, we'll need to do a focused effort to interoperate datastructures with lotus and the spec field by field now that this is working, but this is the bulk of the work.

After a substantial effort to adapt refmt to our purposes we found that there is effectively no way to use it for cbor tuple encoding go structs.  We now consume [an MIT licensed cbor library](https://github.com/fxamacker/cbor) which enables tuple encoding of our data structures using a simple decorator field and tag.

The bulk of the work in this PR is cutting the ten thousand threads tying go-filecoin to refmt and go-ipld-cbor specifically as opposed to a cbor encoding library (like the one we're using, or lotus's cbg) generally.  Props to @icorderi for already getting us quite far on this months ago.  There was also a bit of work getting our graphsync code to handle the new cbor format.

Some parts of this that are non-ideal:  We are one go-ipld-cbor byte away from not needing the EncodableCid type at all to wrap cids using fxamacker/cbor's tag registration facility, but that is a PL-wide immovable object so we are stuck with it until we can find an unstoppable force.  Since cbg uses a non standard marshal/unmarshal interface that doesn't play nice with our default cbor library I have also hacked around HAMT (de)encoding in a serious way to get cids working correctly.

Closes issue #

### DIFF
--- a/internal/pkg/chain/car.go
+++ b/internal/pkg/chain/car.go
@@ -2,6 +2,7 @@ package chain
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	"github.com/ipfs/go-block-format"
@@ -29,6 +30,13 @@ type carStateReader interface {
 	ChainStateTree(ctx context.Context, c cid.Cid) ([]format.Node, error)
 }
 
+// Fields need to stay lower case to match car's default refmt encoding as
+// refmt can't handle arbitrary casing.
+type carHeader struct {
+	Roots   block.TipSetKey `cbor:"roots"`
+	Version uint64          `cbor:"version"`
+}
+
 // Export will export a chain (all blocks and their messages) to the writer `out`.
 func Export(ctx context.Context, headTS block.TipSet, cr carChainReader, mr carMessageReader, sr carStateReader, out io.Writer) error {
 	// ensure we don't duplicate writes to the car file. // e.g. only write EmptyMessageCID once.
@@ -40,10 +48,11 @@ func Export(ctx context.Context, headTS block.TipSet, cr carChainReader, mr carM
 	}
 
 	// Write the car header
-	chb, err := encoding.Encode(car.CarHeader{
-		Roots:   headTS.Key().ToSlice(),
+	ch := carHeader{
+		Roots:   headTS.Key(),
 		Version: 1,
-	})
+	}
+	chb, err := encoding.Encode(ch)
 	if err != nil {
 		return err
 	}
@@ -77,39 +86,39 @@ func Export(ctx context.Context, headTS block.TipSet, cr carChainReader, mr carM
 				return err
 			}
 
-			if !filter[hdr.Messages.SecpRoot] {
+			if !filter[hdr.Messages.SecpRoot.Cid] {
 				logCar.Debugf("writing message collection: %s", hdr.Messages)
 				if err := exportAMTSignedMessages(ctx, out, secpMsgs); err != nil {
 					return err
 				}
-				filter[hdr.Messages.SecpRoot] = true
+				filter[hdr.Messages.SecpRoot.Cid] = true
 			}
 
-			if !filter[hdr.Messages.BLSRoot] {
+			if !filter[hdr.Messages.BLSRoot.Cid] {
 				logCar.Debugf("writing message collection: %s", hdr.Messages)
 				if err := exportAMTUnsignedMessages(ctx, out, blsMsgs); err != nil {
 					return err
 				}
-				filter[hdr.Messages.BLSRoot] = true
+				filter[hdr.Messages.BLSRoot.Cid] = true
 			}
 
 			// TODO(#3473) we can remove MessageReceipts from the exported file once addressed.
-			rect, err := mr.LoadReceipts(ctx, hdr.MessageReceipts)
+			rect, err := mr.LoadReceipts(ctx, hdr.MessageReceipts.Cid)
 			if err != nil {
 				return err
 			}
 
-			if !filter[hdr.MessageReceipts] {
+			if !filter[hdr.MessageReceipts.Cid] {
 				logCar.Debugf("writing message-receipt collection: %s", hdr.Messages)
 				if err := exportAMTReceipts(ctx, out, rect); err != nil {
 					return err
 				}
-				filter[hdr.MessageReceipts] = true
+				filter[hdr.MessageReceipts.Cid] = true
 			}
 
 			if hdr.Height == 0 {
 				logCar.Debugf("writing state tree: %s", hdr.StateRoot)
-				stateRoots, err := sr.ChainStateTree(ctx, hdr.StateRoot)
+				stateRoots, err := sr.ChainStateTree(ctx, hdr.StateRoot.Cid)
 				if err != nil {
 					return err
 				}
@@ -165,6 +174,7 @@ type carStore interface {
 
 // Import imports a chain from `in` to `bs`.
 func Import(ctx context.Context, cs carStore, in io.Reader) (block.TipSetKey, error) {
+	fmt.Printf("about to load car\n")
 	header, err := car.LoadCar(cs, in)
 	if err != nil {
 		return block.UndefTipSet.Key(), err
--- a/internal/pkg/enccid/enc_cid.go
+++ b/internal/pkg/enccid/enc_cid.go
@@ -1,0 +1,122 @@
+package enccid
+
+import (
+	"encoding/json"
+	"fmt"
+
+	cbor "github.com/fxamacker/cbor"
+	cid "github.com/ipfs/go-cid"
+	ipldcbor "github.com/ipfs/go-ipld-cbor"
+)
+
+// Cid is a cid wrapper that implements UnmarshalCBOR and MarshalCBOR.
+// From ipld-cbor's perspective it is technically a pointer to a cid, because
+// it maps `cbor null-val` <==> `cid.Undef`
+type Cid struct {
+	cid.Cid
+}
+
+// NewCid creates an Cid struct from a cid
+func NewCid(c cid.Cid) Cid {
+	return Cid{c}
+}
+
+// MarshalCBOR converts the wrapped cid to bytes
+func (w Cid) MarshalCBOR() ([]byte, error) {
+	// handle undef cid by writing null
+	if w.Equals(cid.Undef) {
+		return []byte{0xf6}, nil
+	}
+
+	// tag = 42
+	tag0 := byte(0xd8)
+	tag1 := byte(0x2a)
+
+	raw, err := castCidToBytes(w.Cid)
+	if err != nil {
+		return nil, err
+	}
+	// because we need to do the cbor tag outside the byte string we are forced
+	// to write the cbor type-len value for a byte string of raw's length
+	cborLen, err := cbor.Marshal(len(raw), cbor.EncOptions{})
+	if err != nil {
+		return nil, err
+	}
+	cborLen[0] |= 0x40 // flip major type 0 to major type 2
+	prefixLen := len(cborLen) + 2
+
+	result := make([]byte, len(cborLen)+len(raw)+2, len(cborLen)+len(raw)+2)
+	result[0] = tag0
+	result[1] = tag1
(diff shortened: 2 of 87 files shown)


internal/pkg/chain/car.go
```diff
@@ -29,6 +30,13 @@ type carStateReader interface {
 	ChainStateTree(ctx context.Context, c cid.Cid) ([]format.Node, error)
 }
 
+// Fields need to stay lower case to match car's default refmt encoding as
+// refmt can't handle arbitrary casing.
+type carHeader struct {
+	Roots   block.TipSetKey `cbor:"roots"`
+	Version uint64          `cbor:"version"`
+}
```
anorth: This is a big red flag to me. Duplicating data structures declared in other packages so that we can annotate them is unsustainable. What if such a structure contains a CID? How is this to interact with structures that already define explicit CBOR encoding?

internal/pkg/enccid/enc_bigint.go
```diff
@@ -0,0 +1,31 @@
...
+// MarshalBinary returns the bytes of a big int
+func (b EncBigInt) MarshalBinary() ([]byte, error) {
+	return b.i.Bytes(), nil
+}
+
+// UnmarshalBinary parses a big int from bytes
+func (b *EncBigInt) UnmarshalBinary(x []byte) error {
+	b.i = big.NewInt(0).SetBytes(x)
+	return nil
+}
```
anorth: The specs-actors repo includes a big int type, and we should replace all our encodable uses with that.

internal/pkg/encoding/fxamacker_cbor.go
```diff
@@ -0,0 +1,192 @@
...
+	return encoder.encodeCbor(obj)
+}
+
+// EncodeMap encodes a map.
+func (encoder *FxamackerCborEncoder) EncodeMap(obj interface{}) error {
+	return encoder.encodeCbor(obj)
+}
+
+// EncodeStruct encodes a struct.
+func (encoder *FxamackerCborEncoder) EncodeStruct(obj interface{}) error {
```
anorth: If the struct implements CBORMarshal, I think we need a really good reason not to use it.

internal/pkg/types/intset.go
```diff
@@ -34,6 +34,22 @@ type IntSet struct {
...
+// UnmarshalBinary creates an intset out of an rleplus bitarray
+func (is *IntSet) UnmarshalBinary(bs []byte) error {
+	ints, err := rleplus.Decode(bs)
+	if err != nil {
+		return err
+	}
+	*is = NewIntSet(ints...)
+	return nil
+}
+
```
anorth: This will be extracted to a new repo shortly.

internal/pkg/vm/address/address.go
```diff
@@ -101,6 +101,21 @@ func (a Address) Empty() bool {
...
+// UnmarshalBinary unmarshals the address bytes into an address
+func (a *Address) UnmarshalBinary(b []byte) error {
+	addr, err := NewFromBytes(b)
+	if err != nil {
+		return err
+	}
+	*a = addr
+	return err
+}
+
```
anorth: This should have been replaced with the shared address package a long time ago.

ZenGround0: Agreed, and @acruikshank is on it!

internal/pkg/enccid/enc_bigint.go
ZenGround0: 👍 Will defer to the next PR I do addressing all of our big int technical debt.

internal/pkg/chain/car.go
ZenGround0: After talking online I think we agree that in general the above argument is valid but that it is not a blocker in this special case.

Rehashing some of the above questions for other readers:
> What if such a structure contains a CID

It will need to define its own encoding or use a wrapped cid type because raw cids are unfortunately not enc/decodable into ipld cbor without some kind of wrapper because the go-cid type's binary serialization is one byte off from ipld cbor's assumed binary serialization and go-cid can't know about ipld cbor.

> How is this to interact with structures that already define explicit CBOR encoding?

The answer is that if a type defines `MarshalCBOR() ([]byte, error)` and/or `UnmarshalCBOR([]byte) error` OR `MarshalBinary() ([]byte, error)` and or `UnmarshalBinary()` then our internal library handles automatically.

The right solution is for someone to PR either cbor marshaling functions onto the car header or struct tags like the above so that the type is portable across cbor libraries.  I view this as a nice to have because the duplication is so minor.

ZenGround0:
fyi to @icorderi, moving away from the great HAMT hack right now after talking to @anorth.  Expect changes soon.

internal/pkg/enccid/enc_bigint.go
```diff
@@ -0,0 +1,31 @@
+package enccid
+
+import (
+	"math/big"
+)
+
+// EncBigInt is an encodable big integer
+type EncBigInt struct {
```
icorderi: we use bigints in a lot of other types. I would move this to `types` so it can be reused. 
I would also drop the `Enc*` in the name.

internal/pkg/enccid/enc_cid.go
```diff
@@ -0,0 +1,114 @@
...
+	"encoding/json"
+	"fmt"
+
+	cbor "github.com/fxamacker/cbor"
+	cid "github.com/ipfs/go-cid"
+	ipldcbor "github.com/ipfs/go-ipld-cbor"
+)
+
+// EncodableCid is a cid wrapper that implements UnmarshalCBOR and MarshalCBOR.
+type EncodableCid struct {
```
icorderi: I would rename this package to **cid** and this type to `Cid`. 
Can we also leave a TODO to submit a PR to `go-cid` to add the un/marshaling methods?

internal/pkg/enccid/enc_hamt.go
```diff
@@ -0,0 +1,111 @@
+package enccid
```
icorderi: i would move this to an `ipld` package along with the ipld store you added.

internal/pkg/enccid/enc_hamt.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"github.com/ipfs/go-hamt-ipld"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+// HAMTNode is an encodable ipld hamt node
+type HAMTNode struct {
+	//	_        struct{}       `toarray`
```
icorderi: uncomment? delete?

internal/pkg/enccid/enc_hamt.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// CbgDeferred wraps cbor bytes directly
+type CbgDeferred struct {
+	Raw []byte
+}
+
+// HAMTKV is an encodable hamt kv
+type HAMTKV struct {
+	//	_     struct{} `toarray`
```
icorderi: uncomment, delete?

internal/pkg/encoding/encoding.go
```diff
@@ -63,15 +63,24 @@ type Decoder interface {
...
+type deprecatedEncoder = IpldCborEncoder
+type deprecatedDecoder = IpldCborDecoder
 
 // Encode encodes an object, returning a byte array.
 func Encode(obj interface{}) ([]byte, error) {
 	var encoder Encoder = &defaultEncoder{}
 	return encode(obj, reflect.ValueOf(obj), encoder)
 }
 
+// EncodeDeprecated uses the deprecated refmt cbor encoding
```
icorderi: leave a dragon to come back and delete this guy when you are done.

internal/pkg/encoding/fxamacker_cbor.go
```diff
@@ -0,0 +1,192 @@
...
+	switch v := obj.(type) {
+	case *hamt.Node:
+		var encHAMTNode e.HAMTNode
+		err := decoder.decodeCbor(&encHAMTNode)
+		if err != nil {
+			return err
+		}
+		encHAMTNode.FillIPLDHamtNode(v)
+		return nil
+	case *cbg.Deferred:
```
icorderi: where are we using cbg? and what for?

internal/pkg/types/channel_id.go
```diff
@@ -40,6 +40,18 @@ func (z ChannelID) MarshalJSON() ([]byte, error) {
 	return json.Marshal(z.val)
 }
 
+// MarshalBinary converts a ChannelID to bytes
```
icorderi: the idea was for this structures to implement `Encode/Decode` which can be then use with multiple encoders.

internal/pkg/vm/abi/abi.go
```diff
@@ -578,6 +578,7 @@ func Deserialize(data []byte, t Type) (*Value, error) {
 	case CommitmentsMap:
 		var m map[string]types.Commitments
 		if err := encoding.Decode(data, &m); err != nil {
+			fmt.Printf("error right here: %s for bad bytes %x\n", err, data)
```
icorderi: log, crash, delete?

internal/pkg/vm/actor/actor.go
```diff
@@ -67,14 +70,34 @@ func (a *Actor) IncrementSeqNum() {
...
+	obj, err := cbor.Decode(bs, types.DefaultHashFunction, -1)
+	if err != nil {
+		return cid.Undef, errors.Wrap(err, "failed to cast cbor bytes to ipld node")
+	}
 
 	return obj.Cid(), nil
 }
 
+// GetHead returns the cid of the actor's storage, or Undef if unset
+func (a *Actor) GetHead() cid.Cid {
```
icorderi: Why are we adding Getters with `Get*` we are not using the `Get*` pattern on other data structures.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -140,7 +141,7 @@ func (ctx *VMContext) LegacySend(to address.Address, method types.MethodID, valu
 
 	msg := types.NewUnsignedMessage(from, to, 0, value, method, paramData)
 	if msg.From == msg.To {
-		// TODO 3647: handle this
+		// TODO 3647: handle this˜µ
```
icorderi: that unicode at the end

tools/refmt-practice/main.go
```diff
@@ -0,0 +1,120 @@
+package main
```
icorderi: does this need to get merged?

internal/pkg/enccid/enc_cid.go
ZenGround0: We should sync offline about the systemic problems blocking the suggested approach

internal/pkg/encoding/fxamacker_cbor.go
ZenGround0: We should talk about this plus plan to remove duplicated hamt types offline

internal/pkg/vm/actor/actor.go
ZenGround0: Its an artifact of cid.Undef not being encodable / decodable forcing us to use datastructures with pointers to cid objects and therefore causing a need for non-nil dereferencing accessors.

tools/refmt-practice/main.go
ZenGround0: woops didn't mean to include this

internal/pkg/enccid/enc_bigint.go
ZenGround0: Removing in favor of shared type

internal/pkg/enccid/enc_hamt.go
ZenGround0: removed

internal/pkg/vm/actor/actor.go
ZenGround0: addressed

ZenGround0:
Blocking on this go-address PR: https://github.com/filecoin-project/go-address/pull/2

PR #3741, opened 2020-02-05
[WIP] Feat/implement retrieval market connectors

shannonwells:
### Motivation
integration of go-fil-market/retrieval_market with specs-actors


cmd/go-filecoin/client.go
```diff
@@ -6,7 +6,6 @@ import (
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit"
-	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
```
shannonwells: This stuff is sort of unrelated to the PR, but it needed to be fixed so my tests could run.

internal/pkg/piecemanager/storage_miner_back_end.go
```diff
@@ -122,3 +122,10 @@ func (s *StorageMinerBackEnd) LocatePieceForDealWithinSector(ctx context.Context
 
 	return 0, 0, 0, errors.Errorf("no encoded piece could be found corresponding to deal id: %d", dealID)
 }
+
+// GetSectorInfo produces information about a sector managed by this storage
+// miner, or an error if the miner does not manage a sector with the
+// provided identity.
+func(s *StorageMinerBackEnd) GetSectorInfo(sectorID uint64) (storage.SectorInfo, error) {
```
shannonwells: Needed to fulfill the interface requirements in this branch

internal/app/go-filecoin/retrieval_market_connector/client.go
```diff
@@ -1,27 +1,292 @@
...
+			return address.Undef, errors.New("not enough funds in wallet")
+		}
+
+		height, err := r.getBlockHeight()
+		if err != nil {
+			return address.Undef, err
+		}
+		validAt := height + 1 // valid almost immediately since a retrieval could theoretically happen in 1 block
+
+		to, err := GoAddrFromFcAddr(fcaddr.LegacyPaymentBrokerAddress)
```
acruikshank: This function should go away now that go-address is fully integrated in filecoin.

internal/app/go-filecoin/retrieval_market_connector/client.go
```diff
@@ -1,27 +1,292 @@
...
+func (r *RetrievalClientNodeConnector) getBlockHeight() (uint64, error) {
+	head := r.cs.GetHead()
+	ts, err := r.cs.GetTipSet(head)
+	if err != nil {
+		return 0, err
+	}
+	return ts.Height()
+}
+
+func (r *RetrievalClientNodeConnector) getPayerFromChannelID(paymentChannel address.Address) (address.Address, error) {
```
acruikshank: Nit: I would rename this to `getPayerForChannel`, and `paymentChannel` to `channelAddr`. The address is the unique identifier in this case (which may or may not be an ID address), but I expect actor addresses to be labeled as addresses, and sequence numbers stored in actor state to be labeled as IDs.

internal/app/go-filecoin/retrieval_market_connector/client.go
```diff
@@ -1,27 +1,292 @@
...
-	panic("TODO: go-fil-markets integration")
+	payer, err := r.getPayerFromChannelID(paymentChannel)
+	if err != nil {
+		return 0, err
+	}
+
+	if r.laneStore[payer] == nil {
+		r.laneStore[payer] = []types.AttoFIL{}
+	}
+	r.laneStore[payer] = append(r.laneStore[payer], types.NewAttoFILFromFIL(0))
```
acruikshank: This won't survive a restart. Is this temporary? What is the plan?

internal/app/go-filecoin/retrieval_market_connector/client.go
```diff
@@ -1,27 +1,292 @@
...
+		Amount:         amount,
+		MinCloseHeight: height + 1,
+	}
+
+	var buf bytes.Buffer
+	if err := v.MarshalCBOR(&buf); err != nil {
+		return nil, err
+	}
+
+	signingAddr, err := r.wal.GetDefaultWalletAddress()
```
acruikshank: I believe this voucher needs to be signed by the payer which is not necessarily the default wallet address.

internal/app/go-filecoin/retrieval_market_connector/provider.go
```diff
@@ -1,24 +1,72 @@
...
+
+	key, err := r.voucherStoreKeyFor(voucher)
+	if err != nil {
+		return tokenamt, err
+	}
+	_, ok := r.vs[key]
+	if ok {
+		return tokenamt, xerrors.New("voucher exists")
+	}
+	r.vs[key] = voucherEntry{
```
acruikshank: Unclear on the point of storing this in memory. Adding it to a datastore shouldn't be too much additional work.

internal/app/go-filecoin/retrieval_market_connector/retrieval_api_test.go
```diff
@@ -0,0 +1,205 @@
+package retrieval_market_connector_test
```
acruikshank: GFC prefers `fake` to `test` for implementations only used in tests. GFC convention is to put this in a file called `testing`.

internal/app/go-filecoin/retrieval_market_connector/client.go
shannonwells: Any of the stuff stored in memory is temporary. @hannahhoward said she wasn't sure if spec-actors had lanes (but they do) so this will go away.  Otherwise it would have been converted to save to a datastore.

internal/app/go-filecoin/retrieval_market_connector/client.go
shannonwells: That makes sense -- it was confusing to use channelIDs everywhere b/c that's an actual thing right now in go-filecoin, but it's changing with the new spec-actors.

PR #3743, opened 2020-02-06
feat: remove some more legacy vm code

icorderi:
### Motivation

The cleanup continues..

### Proposed changes


internal/app/go-filecoin/porcelain/miner.go
```diff
@@ -78,7 +77,8 @@ func MinerCreate(
 	var minerAddr address.Address
 	err = plumbing.MessageWait(ctx, smsgCid, func(blk *block.Block, smsg *types.SignedMessage, receipt *types.MessageReceipt) (err error) {
 		if receipt.ExitCode != uint8(0) {
-			return vmErrors.VMExitCodeToError(receipt.ExitCode, storagemarket.Errors)
+			// Dragons: do we want to have this back?
```
ZenGround0: Do you mean the mapping between exit codes and error values or something else?  If that's what you mean it seems nice to have but unimportant.

internal/pkg/consensus/processor.go
```diff
@@ -67,7 +67,7 @@ func (p *DefaultProcessor) ProcessTipSet(ctx context.Context, st state.Tree, vms
 
 	h, err := ts.Height()
 	if err != nil {
-		return nil, errors.FaultErrorWrap(err, "processing empty tipset")
+		return nil, fmt.Errorf("processing empty tipset")
```
ZenGround0: Don't do this, just return the error

internal/pkg/vm/actor/builtin/initactor/init.go
```diff
@@ -2,6 +2,7 @@ package initactor
 
 import (
 	"context"
+	"fmt"
```
ZenGround0: fyi not reviewing changes to actor code as it will all be replaced in a minute

internal/pkg/vm/errors/errors.go
```diff
@@ -1,239 +0,0 @@
-// Package errors implements error wrappers to signal the different
```
ZenGround0: Catharsis

internal/pkg/vm/internal/errors/vminternal.go
```diff
@@ -3,28 +3,8 @@
 // This contents can be slowly placed on the vm internal.
 package internal
 
-import (
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/errors"
-)
-
-// TODO fritz require actors to define their exit codes and associate
```
ZenGround0: @phritz we finally knocked this one out!

internal/pkg/vm/internal/vmcontext/export.go
```diff
@@ -80,7 +79,7 @@ func makeTypedExport(actor dispatch.ExecutableActor, method types.MethodID) (Exp
 	return func(ctx ExportContext) ([]interface{}, uint8, error) {
 		params, err := abi.DecodeValues(ctx.Params(), signature.Params)
 		if err != nil {
-			return nil, 1, errors.RevertErrorWrap(err, "invalid params")
+			return nil, 1, fmt.Errorf("invalid params")
```
ZenGround0: Please keep wrapping this error so we know how decodevalues failed

internal/pkg/vm/state/cached_tree.go
```diff
@@ -64,7 +64,7 @@ func (t *CachedTree) Commit(ctx context.Context) error {
 	for addr, actor := range t.cache {
 		err := t.st.SetActor(ctx, addr, actor)
 		if err != nil {
-			return errors.FaultErrorWrap(err, "Could not commit cached actors to state tree.")
+			return fmt.Errorf("Could not commit cached actors to state tree. %s", err)
```
ZenGround0: nit: would be slightly better to use standard wrapping than %s

internal/pkg/vm/internal/errors/vminternal.go
phritz: :)

PR #3746, opened 2020-02-08
feat: update ipld cbor

icorderi:
### Motivation

Pre-req to integrating new actor code 

### Proposed changes

update hamt/ipld-cbor library, deal with aftermath

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,7 +52,7 @@ require (
 	github.com/ipfs/go-ipfs-files v0.0.4
 	github.com/ipfs/go-ipfs-keystore v0.0.1
 	github.com/ipfs/go-ipfs-routing v0.1.0
-	github.com/ipfs/go-ipld-cbor v0.0.4
+	github.com/ipfs/go-ipld-cbor v0.0.5-0.20200204214505-252690b78669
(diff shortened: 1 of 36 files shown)


go.mod
```diff
@@ -52,7 +52,7 @@ require (
 	github.com/ipfs/go-ipfs-files v0.0.4
 	github.com/ipfs/go-ipfs-keystore v0.0.1
 	github.com/ipfs/go-ipfs-routing v0.1.0
-	github.com/ipfs/go-ipld-cbor v0.0.4
+	github.com/ipfs/go-ipld-cbor v0.0.5-0.20200204214505-252690b78669
```
anorth: I am surprised that this hasn't also upgraded go-hamt-ipld to match.

I am expecting something like the specs-actors deps:
	github.com/filecoin-project/go-amt-ipld/v2 v2.0.1-0.20200131012142-05d80eeccc5e
	github.com/ipfs/go-hamt-ipld v0.0.15-0.20200131012125-dd88a59d3f2e

anorth:
I didn't realise we could get away with only updating the interface without the hamt, but all good. We can upgrade the rest later.

PR #3747, opened 2020-02-08
feat: new dispatcher

icorderi:
### Motivation

The new actor code exports method differently.

### Proposed changes

Rewrite the dispatching.


internal/pkg/testhelpers/core.go
```diff
@@ -133,13 +133,14 @@ func RequireNewFakeActor(t *testing.T, vms vm.Storage, addr address.Address, cod
...
 // that its steps succeed.
 func RequireNewFakeActorWithTokens(t *testing.T, vms vm.Storage, addr address.Address, codeCid cid.Cid, amt types.AttoFIL) *actor.Actor {
-	act := actor.NewActor(codeCid, amt)
-	var err error
-	rawHead, err := (&actor.FakeActor{}).InitializeState(vms, &actor.FakeActorStorage{})
-	require.NoError(t, err)
-	act.Head = e.NewCid(rawHead)
-	require.NoError(t, vms.Flush())
-	return act
+	// act := actor.NewActor(codeCid, amt)
```
anorth: Nothing calls this, I think you can delete the methods

internal/pkg/vm/internal/dispatch/dispatch.go
```diff
@@ -1,30 +1,107 @@
...
+	"github.com/ipfs/go-cid"
 )
 
-// Method is a callable pointer to an executable method in an actor implementation.
-type Method interface {
+// Actor is the interface all actors have to implement.
+type Actor interface {
+	// Exports has a list of method available on the actor.
+	Exports() []interface{}
+}
```
anorth: This is abi.Invokee from specs-actors

internal/pkg/vm/internal/dispatch/signature.go
```diff
@@ -0,0 +1,46 @@
...
+func (ms *methodSignature) ReturnInterface(returnBytes []byte) (interface{}, error) {
+	// decode arg1 (this is the payload for the actor method)
+	t := ms.method.Type().Out(0)
+	v := reflect.New(t)
+	if err := encoding.Decode(returnBytes, v.Interface()); err != nil {
+		return nil, errors.Wrap(err, "failed to decode return bytes for method")
+	}
+
+	return v.Interface(), nil
+}
```
anorth: After we use the new actors, the args will all implement `runtime.CBORer` for en/decoding.

PR #3749, opened 2020-02-10
Feat/encodable int

ZenGround0:
### Motivation
We need data structures that serialize to specced bytes.  LEB128 is out and filecoin big ints / cbor ints are in so this PR moves the corresponding data types in the block header.

### Proposed changes
All block header fields are cbor int encoded or big int encoded.

Note: this PR does not remove the deprecated leb128 encoding Uint64 type from the actor sequence number in the actor data structure  and unsigned message struct.  We'll need to do this to get message processing interop and state machine interop.


ZenGround0:
This PR is blocked on a specs-actors PR [here](https://github.com/filecoin-project/specs-actors/pull/141).  It build and passes tests with a local go mod replace pointing at that branch.

internal/app/go-filecoin/internal/submodule/syncer_submodule.go
```diff
@@ -16,7 +14,9 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net/pubsub"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/proofs/verification"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/slashing"
+	fbig "github.com/filecoin-project/specs-actors/actors/abi/big"
```
anorth: FWIW, unless you have an actual conflict, I don't think it's worth renaming this import.

internal/pkg/block/block.go
```diff
@@ -154,14 +155,6 @@ func DecodeBlock(b []byte) (*Block, error) {
 	return &out, nil
 }
 
-// Score returns the score of this block. Naively this will just return the
-// height. But in the future this will return a more sophisticated metric to be
-// used in the fork choice rule
-// Choosing height as the score gives us the same consensus rules as bitcoin
-func (b *Block) Score() uint64 {
-	return uint64(b.Height)
-}
```
anorth: Wow, pre-history.

internal/pkg/block/epost_info.go
```diff
@@ -23,16 +23,16 @@ type EPoStInfo struct {
 type EPoStCandidate struct {
```
anorth: FYI we should align and take this type from specs-actors

internal/pkg/mining/worker_test.go
```diff
@@ -698,8 +699,8 @@ func TestGenerateSetsBasicFields(t *testing.T) {
 		Clock:         th.NewFakeClock(time.Unix(1234567890, 0)),
 	})
 
-	h := types.Uint64(100)
-	w := types.Uint64(1000)
+	h := uint64(100)
```
anorth: Note: we still need to resolve whether abi.ChainEpoch is signed or not.

ZenGround0: You probably already know this but FYI the uint64 vs int64 distinction doesn't matter for encoding purposes because cbor should encode both positive int64s and uint64s as major type 0.  This only becomes important for encoding if we ever want to allow a negative epoch number.

The whole specing out the semantics of algorithms in a consistent way with types that allow for intermediate negative values still applies of course.

internal/app/go-filecoin/internal/submodule/syncer_submodule.go
ZenGround0: I have a medium strength opinion that it is good for developers to immediately know which package is being referenced because all the operations are similar and the naming collisions can be pretty confusing.

Maybe the use of math/big will virtually vanish as we move to this type and then I'll agree more with your point.

PR #3753, opened 2020-02-12
feat: runtime integration

icorderi:
### Motivation

We neee to hook our vm with the actors runtime defined in `specs-actors`.


internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -71,7 +73,9 @@ func (ctx *invocationContext) invoke() interface{} {
 	// 6. create target state handle
 
 	// assert from address is an ID address.
-	runtime.Assert(ctx.msg.from.Protocol() == address.ID)
+	if ctx.msg.from.Protocol() != address.ID {
+		runtime.Abortf(exitcode.ErrForbidden, "Sender address MUST be an ID address at invocation time")
```
anorth: If the ctx is supposed to have the sender resolved already, it would be more appropriate to assert or return an explicit error here: it indicates a programming error that's probably not repeatable in other implementations.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -109,7 +115,7 @@ func (ctx *invocationContext) invoke() interface{} {
 
 	// 1. check caller was validated
 	if !ctx.isCallerValidated {
-		runtime.Abortf(exitcode.MethodAbort, "Caller MUST be validated during method execution")
+		runtime.Abortf(exitcode.SysErrInternal, "Caller MUST be validated during method execution")
```
anorth: IllegalActor

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -200,31 +206,31 @@ func (ctx *invocationContext) Runtime() runtime.Runtime {
...
 
 // ValidateCaller implements runtime.InvocationContext.
 func (ctx *invocationContext) ValidateCaller(pattern runtime.CallerPattern) {
 	if ctx.isCallerValidated {
-		runtime.Abortf(exitcode.MethodAbort, "Method must validate caller identity exactly once")
+		runtime.Abortf(exitcode.SysErrInternal, "Method must validate caller identity exactly once")
 	}
 	if !pattern.IsMatch((*patternContext2)(ctx)) {
-		runtime.Abortf(exitcode.MethodAbort, "Method invoked by incorrect caller")
+		runtime.Abortf(exitcode.SysErrInternal, "Method invoked by incorrect caller")
```
anorth: IllegalActor for both

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -200,31 +206,31 @@ func (ctx *invocationContext) Runtime() runtime.Runtime {
...
 	return ctx.stateHandle
 }
 
 // Send implements runtime.InvocationContext.
-func (ctx *invocationContext) Send(to address.Address, method types.MethodID, value types.AttoFIL, params interface{}) interface{} {
+func (ctx *invocationContext) Send(to address.Address, method types.MethodID, value abi.TokenAmount, params interface{}) interface{} {
 	// check if side-effects are allowed
 	if !ctx.allowSideEffects {
-		runtime.Abortf(exitcode.MethodAbort, "Calling Send() is not allowed during side-effet lock")
+		runtime.Abortf(exitcode.SysErrInternal, "Calling Send() is not allowed during side-effet lock")
```
anorth: IllegalActor

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -289,11 +295,11 @@ var _ runtime.ExtendedInvocationContext = (*invocationContext)(nil)
 func (ctx *invocationContext) CreateActor(actorID types.Uint64, code cid.Cid, constructorParams []byte) (address.Address, address.Address) {
 	// Dragons: code it over, there were some changes in spec, revise
 	if !isBuiltinActor(code) {
-		runtime.Abortf(exitcode.MethodAbort, "Can only create built-in actors.")
+		runtime.Abortf(exitcode.SysErrInternal, "Can only create built-in actors.")
```
anorth: SysErrorIllegalArgument for invalid parameters passed to runtime by the actor.

internal/pkg/vm/internal/vmcontext/runtime_adapter.go
```diff
@@ -0,0 +1,228 @@
...
+var _ specsruntime.Syscalls = (*syscallsWrapper)(nil)
+
+// VerifySignature implements Syscalls.
+func (w syscallsWrapper) VerifySignature(signature crypto.Signature, signer address.Address, plaintext []byte) bool {
+	panic("TODO")
+}
+
+// Hash_SHA256 implements Syscalls.
+func (w syscallsWrapper) Hash_SHA256(data []byte) []byte { // nolint: golint
+	// Review: why the underscore?
```
anorth: Debt. Should be `HashBlake2b`

internal/pkg/vm/state/cached_tree.go
```diff
@@ -60,6 +60,19 @@ func (t *CachedTree) GetOrCreateActor(ctx context.Context,
 	return actor, mappedAddr, nil
 }
 
+// DeleteActor remove the actor from the storage.
+// This method will NOT return an error if the actor was not found.
```
anorth: It almost certainly signals a bug if it's not found.

icorderi: No, and that is not the reason that method does not error if its not found. 
Depending on the storage used, the delete operation might have a way of determining if an object exists prior to deletion. This requires doing a get first which add a cost. 
Any user wanting to know if something exists before deletion can do a get himself.

anorth: Ok, there'll be exactly one caller in the runtime afaik, please make sure it sanity checks.

PR #3755, opened 2020-02-12
Remove some use of querymsg: miner asks and network name

anorth:
### Motivation
Asks are no longer in chain state at all, so removed related code. @acruikshank the CLI needs to be minimally hooked up to the market module.

### Proposed changes
Remove on-chain asks queries. Use state view for network name.


anorth:
Test failure is due to Go-filecoin using incorrect keys in the state root. I'll fix that in another PR to land before this one. This is still good for review.

anorth:
FYI I added some more deletions related to miner asks.

PR #3756, opened 2020-02-12
Hello Protocol up to spec

ZenGround0:
### Motivation
Interop with other nodes implementing spec

### Proposed changes
1. Hello message adds field and uses cbor tuple encoding
2. Flip order of receive and send so that stream opener sends and stream handler receives
3. Send and receive new latency message


ZenGround0:
FYI working on reading messages without hax right now so feel free to hold off on review.

internal/pkg/discovery/hello_protocol.go
```diff
@@ -112,36 +156,88 @@ func (h *HelloProtocolHandler) getOurHelloMessage() (*HelloMessage, error) {
...
 	}, nil
 }
 
-func (h *HelloProtocolHandler) receiveHello(ctx context.Context, p peer.ID) (*HelloMessage, error) {
-	s, err := h.host.NewStream(ctx, p, helloProtocolID(h.networkName))
+func (h *HelloProtocolHandler) receiveHello(ctx context.Context, s net.Stream) (*HelloMessage, error) {
+	var hello HelloMessage
+	// Read cbor bytes from stream into hello message
+	// Dragons: this handles ~9 head cids.  We should reduce assumptions
+	buf := make([]byte, 512)
```
anorth: Please factor out the 512 to a named constant with this note, and remove "dragons". Then either:
- make it bigger, and/or
- file an issue noting this

We don't need dragons for just "we could do this better".

internal/pkg/discovery/hello_protocol.go
```diff
@@ -112,36 +156,88 @@ func (h *HelloProtocolHandler) getOurHelloMessage() (*HelloMessage, error) {
...
-	return cbu.NewMsgWriter(s).WriteMsg(&msg)
+	msgRaw, err := encoding.Encode(msg)
+	if err != nil {
+		return err
+	}
+	n, err := s.Write(msgRaw)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("finished writing to stream\n")
```
anorth: Remove printf

internal/pkg/discovery/hello_protocol.go
```diff
@@ -160,44 +256,37 @@ func (hn *helloProtocolNotifiee) asHandler() *HelloProtocolHandler {
...
 		// add timeout
 		ctx, cancel := context.WithTimeout(context.Background(), helloTimeout)
 		defer cancel()
-
-		// receive the hello message
-		from := c.RemotePeer()
-		hello, err := hn.asHandler().receiveHello(ctx, from)
+		s, err := hn.asHandler().host.NewStream(ctx, c.RemotePeer(), helloProtocolID)
+		if err != nil {
+			fmt.Printf("erroring on open stream\n")
```
anorth: Log rather than print

internal/pkg/discovery/hello_protocol.go
```diff
@@ -160,44 +256,37 @@ func (hn *helloProtocolNotifiee) asHandler() *HelloProtocolHandler {
...
-		case err == nil:
-			// notify the local node of the new `block.ChainInfo`
-			hn.asHandler().peerDiscovered(ci)
-		// processing errors
-		case err == ErrBadGenesis:
-			log.Debugf("genesis cid: %s does not match: %s, disconnecting from peer: %s", &hello.GenesisHash, hn.asHandler().genesis, from)
-			genesisErrCt.Inc(context.Background(), 1)
-			_ = c.Close()
+		// now receive latency message
+		_, err = hn.asHandler().receiveLatency(ctx, s) // Dragons: we drop this on the floor.  We should use it.
```
anorth: No dragons for potential enhancements, maybe file an issue.

internal/pkg/discovery/hello_protocol.go
ZenGround0: @anorth Thanks for feedback!  Made it waaay better by just reading streamed cbor directly upon @Kubuxu's suggestion.  You might want to to take a quick look at latest commit to see diff

PR #3758, opened 2020-02-12
Fix encoding of address keys in the state root

anorth:
### Motivation
The top-level HAMT should be keyed by the bytes of the actor address, not the pretty string. I have confirmed this is what Lotus implements.

### Proposed changes
Replace as many uses of Address.String() as possible, especially when used as a HAMT key.

### DIFF
--- a/internal/app/go-filecoin/porcelain/client.go
+++ b/internal/app/go-filecoin/porcelain/client.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
-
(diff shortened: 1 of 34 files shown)


internal/app/go-filecoin/porcelain/client.go
```diff
@@ -6,11 +6,12 @@ import (
...
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 	"github.com/ipfs/go-cid"
 	"github.com/pkg/errors"
 
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
+
```
ZenGround0: nit: remove space
(and thanks for the dep order correction)

PR #3761, opened 2020-02-13
feat: use spec actors cid codes

icorderi:
### Motivation

Actor code ID's are now defined in the specs-actors repo.


internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -365,17 +366,26 @@ func (ctx *patternContext2) CallerAddr() address.Address {
...
 }
 
+// Dragons: move this to specs-actors
 func isSingletonActor(code cid.Cid) bool {
-	return code.Equals(types.StorageMarketActorCodeCid) ||
-		code.Equals(types.InitActorCodeCid)
+	return code.Equals(builtin.CronActorCodeID) ||
+		code.Equals(builtin.InitActorCodeID) ||
+		code.Equals(builtin.StorageMarketActorCodeID) ||
+		code.Equals(builtin.StoragePowerActorCodeID)
```
anorth: Yes please to both

PR #3762, opened 2020-02-13
Feat/msg format interop

ZenGround0:
### Motivation
Message and actor object correct serialization is needed for interop.

Using verbose deprecated types is technical debt we should clear.

### Proposed changes
1. Remove Uint64 type from message and actor `CallSeqNum` in favor of uint64.
2. Tuple encode unsiged message and signed message

### DIFF
--- a/internal/pkg/types/message.go
+++ b/internal/pkg/types/message.go
@@ -77,14 +77,14 @@ var (
 
 // UnsignedMessage is an exchange of information between two actors modeled
 // as a function call.
-// Messages are the equivalent of transactions in Ethereum.
 type UnsignedMessage struct {
+	_ struct{}	`cbor:",toarray"`
 	To   address.Address `json:"to"`
 	From address.Address `json:"from"`
 	// When receiving a message from a user account the nonce in
 	// the message must match the expected nonce in the from actor.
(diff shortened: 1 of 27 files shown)


internal/pkg/types/message_test.go
```diff
@@ -32,7 +32,7 @@ func TestMessageMarshal(t *testing.T) {
 
 	// This check requests that you add a non-zero value for new fields above,
 	// then update the field count below.
-	require.Equal(t, 8, reflect.TypeOf(*msg).NumField())
+	require.Equal(t, 9, reflect.TypeOf(*msg).NumField())
```
acruikshank: Probably worth documenting that phantom struct member makes the count go up here.

internal/pkg/types/message.go
```diff
@@ -77,14 +77,14 @@ var (
 
 // UnsignedMessage is an exchange of information between two actors modeled
 // as a function call.
-// Messages are the equivalent of transactions in Ethereum.
 type UnsignedMessage struct {
+	_    struct{}        `cbor:",toarray"`
```
acruikshank: Is this required for tuple encoding? This is confusing.

PR #3763, opened 2020-02-13
Cleanup/remove refmt atlas

ZenGround0:
### Motivation
We have recently moved away from refmt but needed to keep around lots of refmt types and initialization calls to keep things working.  This PR removes the dependence on refmt almost completely.  There are a few tests where we keep cbor.WrapObject and refmt atlases for convenience but could easily move these to encoding.Encode with a bit of work.

### Proposed changes
Use hamt raw setter getter methods to remove implicity refmt compatibility with gfc internal types
Remove refmt atlas + struct initialization 
Random cleanups

Note: This is blocking on go-hamt-ipld pr: https://github.com/ipfs/go-hamt-ipld/pull/43.  It won't build on ci because go.mod is pointing to my local dep on that branch, but tests pass locally with that change.


internal/pkg/types/signed_message.go
```diff
@@ -83,12 +84,31 @@ func (smsg *SignedMessage) Cid() (cid.Cid, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	c := cid.NewCidV1(cid.DagCBOR, hash)
+
+	blk, err := blocks.NewBlockWithCid(data, c)
+	if err != nil {
+		return nil, err
+	}
+	obj, err := cbor.DecodeBlock(blk)
```
anorth: Actually decoding it seems pretty roundabout, considering we have the data at line 90.

anorth:
Some of these files are deleted by #3791 .

Since you started this PR I have also added a couple of `atlas.go` files which you may also be able to remove.

internal/pkg/types/signed_message.go
ZenGround0: The names aren't great but this is the straightforward path.  It is the function that maps from a cbor binary/cided blocks.block to an ipld node.

PR #3772, opened 2020-02-17
Feat/token amount

ZenGround0:
### Motivation
Implement AttoFIL type as an alias of specs actors TokenAmount type.


internal/pkg/types/atto_fil_test.go
```diff
@@ -116,14 +115,9 @@ func TestAttoFILAddition(t *testing.T) {
 		assert.Equal(t, bStr, b.String())
 	})
 
-	t.Run("treats zero values as zero", func(t *testing.T) {
```
ZenGround0: FYI this behavior is no longer supported by the specs actors big int

internal/pkg/types/atto_fil.go
```diff
@@ -76,11 +39,11 @@ func (z *AttoFIL) UnmarshalBinary(b []byte) error {
 // Reasons for embedding a big.Int instead of *big.Int:
 //   - We don't have check for nil in every method that does calculations.
 //   - Serialization "symmetry" when serializing AttoFIL{}.
-type AttoFIL struct{ val big.Int }
+type AttoFIL = specsabi.TokenAmount
```
anorth: I would rather purge the type completely. Why is it still here?

internal/pkg/consensus/validation.go
```diff
@@ -120,8 +120,8 @@ func (v *DefaultMessageValidator) Validate(ctx context.Context, msg *types.Unsig
 // more value from the actor's balance.
 func canCoverGasLimit(msg *types.UnsignedMessage, actor *actor.Actor) bool {
 	// balance >= (gasprice*gasLimit + value)
-	gascost := specsbig.Mul(abi.NewTokenAmount(msg.GasPrice.AsBigInt().Int64()), abi.NewTokenAmount(int64(msg.GasLimit)))
-	expense := specsbig.Add(gascost, abi.NewTokenAmount(msg.Value.AsBigInt().Int64()))
+	gascost := specsbig.Mul(abi.NewTokenAmount(msg.GasPrice.Int.Int64()), abi.NewTokenAmount(int64(msg.GasLimit)))
```
anorth: If `msg.GasPrice` is a `big.Int` (or alias) then this should be just `Mul(msg.GasPrice, ...)`

(Note that GasUnits should become an `int64` too, making this yet cleaner).

internal/pkg/consensus/validation.go
```diff
@@ -120,8 +120,8 @@ func (v *DefaultMessageValidator) Validate(ctx context.Context, msg *types.Unsig
...
+	expense := specsbig.Add(gascost, abi.NewTokenAmount(msg.Value.Int.Int64()))
```
anorth: Ditto

internal/pkg/message/outbox.go
```diff
@@ -88,8 +88,8 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
 			msgSendErrCt.Inc(ctx, 1)
 		}
 		ob.journal.Write("Send",
-			"to", to.String(), "from", from.String(), "value", value.AsBigInt().Uint64(), "method", method,
-			"gasPrice", gasPrice.AsBigInt().Uint64(), "gasLimit", uint64(gasLimit), "bcast", bcast,
+			"to", to.String(), "from", from.String(), "value", value.Int.Uint64(), "method", method,
```
anorth: Would welcome a stringification method on big.Int in specs-actors

internal/pkg/types/atto_fil.go
```diff
@@ -19,55 +15,22 @@ func init() {
...
-func (z AttoFIL) MarshalJSON() ([]byte, error) {
-	return json.Marshal(z.String())
-}
-
-// MarshalBinary converts an AttoFIL to bytes
-func (z AttoFIL) MarshalBinary() ([]byte, error) {
-	return z.Bytes(), nil
-}
+var attoPower = 18
+var tenToTheEighteen = big.NewInt(10).Exp(big.NewInt(10), big.NewInt(18), nil)
```
anorth: `specsbig.Exp(specsbig.NewInt(10), specsbig.NewInt(18))`

Please avoid math/big

internal/pkg/types/atto_fil.go
```diff
@@ -130,105 +98,22 @@ func NewAttoFILFromFILString(s string) (AttoFIL, bool) {
...
+	return out, isErr
 }
 
 // DivCeil returns the minimum number of times this value can be divided into smaller amounts
 // such that none of the smaller amounts are greater than the given divisor.
 // Equal to ceil(z/y) if AttoFIL could be fractional.
 // If y is zero a panic will occur.
-func (z AttoFIL) DivCeil(y AttoFIL) AttoFIL {
-	value, remainder := big.NewInt(0).DivMod(&z.val, &y.val, big.NewInt(0))
+func DivCeil(z, y AttoFIL) AttoFIL {
```
anorth: Assuming we still need this for something, I'd welcome a `DivMod` method in specs-actors. But this looks to be unused.

internal/pkg/types/atto_fil_test.go
```diff
@@ -369,43 +269,43 @@ func TestString(t *testing.T) {
...
 		attoFIL, _ = NewAttoFILFromFILString("12345.0")
-		assert.Equal(t, BigIntFromString("12345000000000000000000"), attoFIL.val)
+		assert.Equal(t, BigIntFromString("12345000000000000000000"), *attoFIL.Int)
 
 		attoFIL, _ = NewAttoFILFromFILString("12345")
-		assert.Equal(t, BigIntFromString("12345000000000000000000"), attoFIL.val)
+		assert.Equal(t, BigIntFromString("12345000000000000000000"), *attoFIL.Int)
 	})
 
 	t.Run("rejects nonsense values", func(t *testing.T) {
```
anorth: We shouldn't need any tests because we shouldn't have any novel functionality regarding bigints. Except possibly formatting FIL values as human-scale strings.

internal/pkg/vm/actor/builtin/power/power.go
```diff
@@ -172,6 +172,7 @@ func (*impl) createStorageMiner(vmctx runtime.InvocationContext, params CreateSt
 				SectorSize:             params.SectorSize,
 			})
 			if err != nil {
+				fmt.Printf("here it is: %s\n", err)
```
anorth: Please remove the prints, even though this code will be trashed.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -303,7 +303,7 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize uint32, mi
 	// Dragons: temp until we remove legacy types
 	var msgGasLimit gas.Unit = gas.NewLegacyGas(msg.GasLimit)
 	var msgGasPrice abi.TokenAmount = abi.NewTokenAmount(int64(msg.GasLimit))
-	var msgValue abi.TokenAmount = abi.NewTokenAmount(msg.Value.AsBigInt().Int64())
+	var msgValue abi.TokenAmount = abi.NewTokenAmount(msg.Value.Int.Int64())
```
anorth:
```suggestion
	var msgValue abi.TokenAmount = msg.Value
```

internal/pkg/vm/testing.go
```diff
@@ -138,7 +138,7 @@ func (tc *FakeVMContext) Receiver() address.Address {
 // Note: the value is already been deposited on the actors account and is reflected on the balance.
 func (tc *FakeVMContext) ValueReceived() abi.TokenAmount {
 	// Dragons: temp until we remove legacy types
-	return abi.NewTokenAmount(tc.MessageValue.Value.AsBigInt().Int64())
+	return abi.NewTokenAmount(tc.MessageValue.Value.Int.Int64())
```
anorth:
```suggestion
	return tc.MessageValue.Value
```

anorth:
LMK if you want me to look again if you push this further, but also ok to follow up.

internal/pkg/types/atto_fil.go
ZenGround0: Id also like to use one thing but if we want to remove it completely and do it the right way I'm going to need to make PRs to specsactors (attofil from fil, string manipulation) that this will need to block on.  I figured this intermediate state was better than nothing and didn't want to get involved in a longer multi-project dependency waiting game (particularly because [I'm already in one of those](https://github.com/ipfs/go-hamt-ipld/pull/43).

Another thing: this name is significantly better than TokenAmount and since we need to leave functionality here I figured we could also leave the better name.

internal/pkg/message/outbox.go
ZenGround0: Good deal.  I'll file an issue with this and other proposals for shared stuff that should happen on that type (name change, FIL to AttoFIL conversion)

internal/pkg/consensus/validation.go
ZenGround0: #3773

internal/pkg/consensus/validation.go
ZenGround0: #3773

PR #3776, opened 2020-02-18
Windowed PoSt implementation

acruikshank:
### Motivation

In addition to submitting a PoSt with a block on every election won, the miner must also submit windowed PoSts on a regular basis. This functionality falls outside the purview of storage miner and block mining, so it still needs to be implemented

### Proposed changes

Add a `Poster` struct that listens for tipset changes, derives the proving set and random seed from chain state, computes PoSt candidates and a proofs and sends this as a PoSt submission to be mined on chain.

### Notes

The windowed PoSt may be triggered by a incoming tipset that may later be reorged out of the chain. If this happens, the chain randomness the PoSt is based on will be invalid. Some work was done on the `chainsampler` to support stopping in-progress PoSts and restarting with new randomness. 

Ultimately, though, the only significant call in PoSt creation is the sectorbuilder's `GenerateFallbackPoSt`. This method cannot be halted within the go process from which it is run. The solution would be to fork processes for PoSt creation. This solution was investigated, but ultimately left for future work in order to prioritze higher priority features for interop.


internal/pkg/chainsampler/height_threshold_scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+				m.heightListeners = append(m.heightListeners, heightListener)
+			case canceled := <-m.cancelListener:
+				listeners := []*HeightThresholdListener{}
+				for _, l := range listeners {
+					if l != canceled {
+						listeners = append(listeners, l)
+					}
+				}
+				m.heightListeners = listeners
+			case <-m.listenerDone:
```
ZenGround0: Looks like this should be called `schedulerDone`

ZenGround0: Looks like this should be `schedulerDone` or `listenersDone`

internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
```diff
@@ -38,53 +48,57 @@ func NewStorageMiningSubmodule(minerAddr, workerAddr address.Address, ds datasto
 
 	smbe := piecemanager.NewStorageMinerBackEnd(storageMiner, s)
 	sbbe := postgenerator.NewSectorBuilderBackEnd(s)
+	stateViewer := state.NewViewer(b.CborStore)
 
 	modu := &StorageMiningSubmodule{
 		PieceManager:  smbe,
 		PoStGenerator: sbbe,
+		minerNode:     minerNode,
```
ZenGround0: IIRC @laser left dependence on the concrete structs backing the interfaces off of the submodule so that we could more easily fake the proving/sealing system during test setup. I'm ok leaving it if you think its better but if you haven't already consider how this impacts testability.

internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
```diff
@@ -38,53 +48,57 @@ func NewStorageMiningSubmodule(minerAddr, workerAddr address.Address, ds datasto
...
+		storageMiner:  storageMiner,
+		outbox:        m.Outbox,
```
ZenGround0: The submodule should not hold onto things that only its fields need. I think the outbox, chain, sectorbuilder and minerAddr fields are all only used by the poster and should be removed.

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -59,72 +54,32 @@ var _ storage.NodeAPI = new(StorageMinerNodeConnector)
...
+	m.chainHeightScheduler.Stop()
 }
 
-func (m *StorageMinerNodeConnector) handleNewTipSet(ctx context.Context, previousHead block.TipSet) (block.TipSet, error) {
-	newHeadKey := m.chainStore.GetHead()
-	newHead, err := m.chainStore.GetTipSet(newHeadKey)
-	if err != nil {
-		return block.TipSet{}, err
-	}
+// CancelHeightListener removes on
```
ZenGround0: this looks misplaced

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -366,7 +322,42 @@ func (m *StorageMinerNodeConnector) GetSealSeed(ctx context.Context, preCommitMs
...
 
-		m.newListener <- chainsampler.NewHeightThresholdListener(h+interval, sc, ec, ic, dc)
+		m.chainHeightScheduler.AddListener(h + interval)
+	}()
+
+	// translate tipset key to seal seed handler
+	go func() {
+		for {
+			select {
+			case key := <-hc:
```
ZenGround0: Blocking: I don't understand how anything can get sent on this channel as it is not passed to any function, any field of m or returned from this function.

ZenGround0: I suspect this hc should be the newly created listener's hit channel above.

internal/pkg/poster/poster.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil
+	}
+
+	if uint64(provingPeriodStart) <= tipsetHeight {
+		// it's not time to PoSt
+		return nil
+	}
+
+	ctx, p.postCancel = context.WithCancel(ctx)
+	go func() {
```
ZenGround0: nit: stylistic preference to define this as a method on Poster i.e. `doPoSt(ctx, ppStart, stateView)` and then call it in a go routine here.  Suggesting only to keep function sizes down.

internal/pkg/poster/poster.go
```diff
@@ -0,0 +1,261 @@
...
+
+	minerID, err := address.IDFromAddress(minerIDAddr)
+	if err != nil {
+		return err
+	}
+
+	poStCandidates := make([]spaabi.PoStCandidate, len(candidates))
+	for i, candidate := range candidates {
+		poStCandidates[i] = spaabi.PoStCandidate{
+			RegisteredProof: spaabi.RegisteredProof_WinStackedDRG32GiBPoSt,
```
ZenGround0: nit: slight preference to just call it `abi` as spaabi is the only abi left.

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -31,6 +30,11 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/wallet"
 )
 
+type randomnessSampler interface {
```
ZenGround0: 👍

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -303,64 +305,67 @@ func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context) (storage.
...
-	}()
+		listener := m.chainHeightScheduler.AddListener(h + interval)
 
-	// translate tipset key to seal seed handler
-	go func() {
+		// translate tipset key to seal seed handler
 		for {
 			select {
-			case key := <-hc:
+			case key := <-listener.HitCh:
```
ZenGround0: 👍

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -303,64 +305,67 @@ func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context) (storage.
...
 				ts, err := m.chainState.GetTipSet(key)
 				if err != nil {
-					ec <- err
-					break
+					listener.ErrCh <- err
```
ZenGround0: hmm do we deadlock here as the same goroutine sends and receives on this channel?

internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -303,64 +305,67 @@ func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context) (storage.
...
+					continue
 				}
 
-				sc <- storage.SealSeed{
+				seedCh <- storage.SealSeed{
 					BlockHeight: tsHeight,
 					TicketBytes: randomness,
 				}
-			case <-dc:
+			case err := <-listener.ErrCh:
```
ZenGround0: Is some other goroutine listening on this and is that a problem?

ZenGround0: I think what you might want to do is return the listteners err, invalid and done channels, remove these cases (keeping the ctx.Done() one).

internal/app/go-filecoin/storage_miner_connector/connector.go
acruikshank: Yep, this should be errCh, not listener.ErrCh

internal/app/go-filecoin/storage_miner_connector/connector.go
acruikshank: The idea is that we're mapping all the channels from the listener to channels created in this method. This this method use the errCh as well as the listener. We can't return the listeners invalidCh and doneCh because this method should return immediately and those channels aren't available until after the wait.

PR #3779, opened 2020-02-19
go-storage-miner should pass state identifiers to methods which access chain state

laser:
## Why does this exist?

None of our getter methods accepted a state identifier; they were all implicitly scoped to the chain head. See [this convo](https://filecoinproject.slack.com/archives/CPLLE0ZQX/p1582128754110800?thread_ts=1582128521.110500&cid=CPLLE0ZQX) for more background information.

## Proposed changes

- add an implementation-nonspecific, opaque `TipSetToken`, used to identify a TipSet
- add `TipSetToken` parameter to `GetSealTicket` and `GetSealedCID` and `GetMinerWorkerAddress`
- add `GetChainHead` method
- drop address parameter from `GetMinerWorkerAddress`

## Notes

The go-filecoin node controls the lifecycle of the go-storage-miner `Miner` struct (which is provided a value which satisfies its "node API" interface). The node passes the `Miner` constructor the miner address X, so it is in a good position (in its connector code) to map from X to the worker address.

### DIFF
--- a/internal/app/go-filecoin/storage_miner_connector/connector.go
+++ b/internal/app/go-filecoin/storage_miner_connector/connector.go
@@ -130,7 +130,7 @@ func (m *StorageMinerNodeConnector) handleNewTipSet(ctx context.Context, previou
 
 // SendSelfDeals creates self-deals and sends them to the network.
 func (m *StorageMinerNodeConnector) SendSelfDeals(ctx context.Context, pieces ...abi.PieceInfo) (cid.Cid, error) {
-	waddr, err := m.GetMinerWorkerAddressFromChainHead(ctx, m.minerAddr)
+	waddr, err := m.getMinerWorkerAddress(ctx, m.chainState.Head())
 	if err != nil {
 		return cid.Undef, err
 	}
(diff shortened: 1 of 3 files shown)


internal/app/go-filecoin/storage_miner_connector/connector.go
```diff
@@ -336,10 +336,14 @@ func (m *StorageMinerNodeConnector) WaitForProveCommitSector(ctx context.Context
...
 }
 
-// GetSealTicket produces the seal ticket used when pre-committing a sector at
-// the moment it is called
-func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context) (storagenode.SealTicket, error) {
-	ts, err := m.chainStore.GetTipSet(m.chainStore.GetHead())
+// GetSealTicket produces the seal ticket used when pre-committing a sector.
+func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context, tok storagenode.TipSetToken) (storagenode.SealTicket, error) {
+	var tsk block.TipSetKey
+	if err := tsk.UnmarshalCBOR(tok); err != nil {
```
anorth: Prefer `encoding.Decode` which will end up calling this.

Issue #3781, opened 2020-02-19
Non-deterministic genesis state caused by BLS keys


anorth:
The https://github.com/filecoin-project/filecoin-ffi interface exposes only random BLS key generation, not deterministic or seeded generation. This makes it impossible to generate deterministic keys for inclusion in genesis state, so gengen produces a different output every time.

I'm going to disable our test for deterministic output while we wait for deterministic key generation. It's not clear to me that this is short-term fundamentally necessary, though it will make things like testing annoyingly difficult.

FYI @acruikshank @ZenGround0 @dignifiedquire @icorderi

anorth:
Also #3570

anorth:
I believe the FFI gives us a seeded BLS key generation now.

PR #3782, opened 2020-02-19
Use BLS for miner worker key in genesis

anorth:
### Motivation
Miner workers must use BLS keys in order to produce VRFs, and the miner actor now enforces this.

### Proposed changes
Change gengen to use BLS keys. This unfortunately causes them to be non-deterministic due to an unfriendly BLS interface in filecoin-ffi (see #3781)

I made the crypto interface a bit easier to use with some renames and returning KeyInfo from the crypto methods. This required moving some things to break a circular dependency.

### DIFF
--- a/internal/app/go-filecoin/node/init.go
+++ b/internal/app/go-filecoin/node/init.go
@@ -11,8 +11,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/cborutil"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/chain"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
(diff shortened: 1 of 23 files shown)


internal/app/go-filecoin/node/init.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/cborutil"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/chain"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
+	crypto2 "github.com/filecoin-project/go-filecoin/internal/pkg/crypto"
```
icorderi: clean the name

PR #3783, opened 2020-02-19
plumb start and end epochs for self-deal to DealProposal

laser:
### Motivation

Today, the self-deal start and end epochs are set to constants - and the wrong constants, at that.

### Proposed changes

This changeset pushes the choice of self-deal start and end-epochs into go-storage-miner, who uses existing deal information to choose self-deal start and end epochs (in the event that a to-be-sealed sector contains client data) or the provided "proving delay" and "duration" values otherwise.

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
@@ -40,7 +40,10 @@ func NewStorageMiningSubmodule(minerAddr address.Address, ds datastore.Batching,
 
 	// The amount of epochs we expect the storage miner to take to replicate and
 	// prove a sector. This value should be shared with the storage miner side
-	// of go-fil-markets.
+	// of go-fil-markets. The protocol specifies a maximum sealing duration (1)
+	// which could be used to improve the accuracy of provingDelay.
+	//
+	// 1: https://github.com/filecoin-project/specs-actors/commit/fa20d55a3ff0c0134b130dc27850998ffd432580#diff-5a14038af5531003ed825ab608d0dd51R21
 	//
 	// TODO: What is the correct value for proving delay given 32GiB sectors?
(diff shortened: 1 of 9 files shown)


internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
```diff
@@ -35,7 +37,20 @@ type StorageMiningSubmodule struct {
...
 func NewStorageMiningSubmodule(minerAddr address.Address, ds datastore.Batching, s sectorbuilder.Interface, c *ChainSubmodule, m *MessagingSubmodule, mw *msg.Waiter, w *WalletSubmodule, stateViewer *appstate.Viewer) (*StorageMiningSubmodule, error) {
 	minerNode := storageminerconnector.NewStorageMinerNodeConnector(minerAddr, c.ChainReader, c.State, m.Outbox, mw, w.Wallet, stateViewer)
-	storageMiner, err := storage.NewMiner(minerNode, ds, s, minerAddr)
+
+	// The amount of epochs we expect the storage miner to take to replicate and
+	// prove a sector. This value should be shared with the storage miner side
+	// of go-fil-markets.
+	//
+	// TODO: What is the correct value for proving delay given 32GiB sectors?
+	provingDelay := abi.ChainEpoch(2 * 60 * 24)
```
anorth: See the miner.MaxSealDuration table in the actor code. It's used in `verifySeal` in the miner actor. 

I think that table is only for the delay between pre- and post-commit, so this might be a larger number. To leave room for versioning of that number it's probably best that it's recapitulated here though, but maybe leave a pointer to the actor policy.

laser: @anorth - Very good. Thanks for the pointer to `miner.MaxSealDuration`. I'll leave a pointer to that value in this code.

PR #3785, opened 2020-02-19
feat: hook some of the new syscalls

icorderi:

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/syscalls.go
+++ b/internal/pkg/vm/internal/vmcontext/syscalls.go
@@ -1,0 +1,48 @@
+package vmcontext
+
+import (
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/crypto"
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	specscrypto "github.com/filecoin-project/specs-actors/actors/crypto"
+	specsruntime "github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/ipfs/go-cid"
+	"github.com/minio/blake2b-simd"
+)
+
+type syscalls struct {
+	gasTank *GasTracker
+}
+
+var _ specsruntime.Syscalls = (*syscalls)(nil)
+
+// VerifySignature implements Syscalls.
+func (sys syscalls) VerifySignature(signature specscrypto.Signature, signer address.Address, plaintext []byte) bool {
+	return crypto.IsValidSignature(plaintext, signer, signature)
+}
+
+// HashBlake2b implements Syscalls.
+func (sys syscalls) HashBlake2b(data []byte) [32]byte {
+	return blake2b.Sum256(data)
+}
+
+// ComputeUnsealedSectorCID implements Syscalls.
+// Review: why is this returning an error instead of aborting? is this failing recoverable by actors?
+func (sys syscalls) ComputeUnsealedSectorCID(sectorSize abi.SectorSize, pieces []abi.PieceInfo) (cid.Cid, error) {
+	panic("TODO")
+}
+
+// VerifySeal implements Syscalls.
+func (sys syscalls) VerifySeal(sectorSize abi.SectorSize, info abi.SealVerifyInfo) bool {
+	panic("TODO")
+}
+
+// VerifyPoSt implements Syscalls.
+func (sys syscalls) VerifyPoSt(sectorSize abi.SectorSize, info abi.PoStVerifyInfo) bool {
+	panic("TODO")
+}
+
+// VerifyConsensusFault implements Syscalls.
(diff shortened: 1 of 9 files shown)


internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/runtime"
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/crypto"
+	specsruntime "github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	"github.com/ipfs/go-cid"
+	"github.com/minio/blake2b-simd"
+)
+
+// Dragons: have the VM take a SysCalls object on construction (it will need a wrapper to charge gas)
```
anorth: Done?

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+func (sys syscalls) ComputeUnsealedSectorCID(sectorSize abi.SectorSize, pieces []abi.PieceInfo) (cid.Cid, error) {
+	panic("TODO")
+}
+
+// VerifySeal implements Syscalls.
+func (sys syscalls) VerifySeal(sectorSize abi.SectorSize, info abi.SealVerifyInfo) bool {
+	verifier := verification.NewFFIBackedProofVerifier()
+
+	commD := to32bytes(info.UnsealedCID.Bytes())
+	commR := to32bytes(info.OnChain.SealedCID.Bytes())
```
anorth: Blocking: use the https://github.com/filecoin-project/go-fil-commcid library to convert CIDs to the raw commitments.

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+
+	minerAddr, err := address.NewIDAddress(uint64(info.Miner))
+	if err != nil {
+		runtime.Abortf(exitcode.SysErrInternal, "failed to build ID address. %w", err)
+	}
+	miner := to32bytes(minerAddr.Bytes())
```
anorth: This doesn't look right to me. I've asked in https://filecoinproject.slack.com/archives/CEGB67XJ8/p1582158323462200

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+
+	ticket := to32bytes([]byte(info.Randomness))
+	seed := to32bytes([]byte(info.InteractiveRandomness))
+	// Review: why the "ProofBytes" field? are we expecting other fields?
```
anorth: This is probably left over from when there were other fields. I'll attempt to remove it.

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		runtime.Abortf(exitcode.SysErrInternal, "failed to validate PoRep: %w", err)
+	}
+
+	return ok
+}
+
+// VerifyConsensusFault implements Syscalls.
+func (sys syscalls) VerifyConsensusFault(h1, h2 []byte) bool {
+	panic("TODO")
```
anorth: FYI this one is defined in the spec, we should be able to implement it. It's note delegated to ffi.

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
...
+
+// VerifyPoSt implements Syscalls.
+func (sys syscalls) VerifyPoSt(sectorSize abi.SectorSize, info abi.PoStVerifyInfo) bool {
+	verifier := verification.NewFFIBackedProofVerifier()
+
+	sectorInfo := ffi.SortedPublicSectorInfo{}
+
+	randomness := to32bytes(info.Randomness)
+	// Dragons: where is this coming from?
+	challengeCount := uint64(21)
```
anorth: Apparently the go-sector-builder has some of these answers. Maybe we should be invoking that instead.

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -0,0 +1,92 @@
+package vmcontext
+
+import (
+	ffi "github.com/filecoin-project/filecoin-ffi"
```
anorth: We shouldn't depend on ffi directly here, but an indirection so we can fake it for testing.

This interface already exists as verification.Verifier, hooked up to FFI and with a fake for testing. You can trace existing call paths into here to answer some of the questions about how to marshal parameters.

PR #3791, opened 2020-02-20
Delete a large number of unused, deprecated types

anorth:
### Motivation
We don't use any of these any more.

### Proposed changes
Remove them! This was almost entirely mechanical.

@ZenGround0 I left types/fixed_point behind because it's used in chain weight. Can we remove it? If not, we should move it next to the only code that uses it.


ZenGround0:
@anorth we can and should remove that shortly.  I'd like to do it as part of the weight function upgrade which I'm planning to get to next week when we hear an update about consensus from research people.

You are welcome to remove pre-emptively here and now but there is likely some tweaking of the syncer integration test that goes along with changing the parent weight number from fixed point to big int.

PR #3794, opened 2020-02-20
Replace types.BytesAmount with abi.SectorSize or abi.StoragePower as appropriate.

anorth:
### Motivation
The spec disambiguates these (and doesn't need a bigint for sector size), as well as un/padded piece size types.

### Proposed changes
Replace ours with the new stuff.


internal/pkg/types/sector_size.go
```diff
@@ -1,9 +1,11 @@
 package types
 
+import "github.com/filecoin-project/specs-actors/actors/abi"
+
 // OneKiBSectorSize contain 1024 bytes after sealing. These sectors are used
 // when the network is in TestProofsMode.
-var OneKiBSectorSize = NewBytesAmount(1024)
+var OneKiBSectorSize = abi.SectorSize(1024)
```
acruikshank: 💯

PR #3797, opened 2020-02-21
misc: cleanup legacy fixtures

icorderi:
### Motivation

Some legacy files leftover in `./fixtures` after a refactor.

### Proposed changes

- remove legacy files
- add a `.gitignore` so people don't push them accidentally back in


fixtures/gen.json
```diff
@@ -1,34 +0,0 @@
-{
```
frrist: Noting the most recent version of these live in `fixtures/live/` and `fixtures/test/`.

frrist: Noting the most recent version of these live in `fixtures/live/` and `fixtures/test/`.

PR #3798, opened 2020-02-21
misc: remove legacy code

icorderi:
### Motivation

Remove dead code


internal/pkg/vm/actor/actor.go
```diff
@@ -71,43 +68,18 @@ func (a *Actor) IncrementSeqNum() {
...
-func (a *Actor) Unmarshal(b []byte) error {
-	return fxamackercbor.Unmarshal(b, a)
-}
-
-// Marshal the actor into bytes.
-func (a *Actor) Marshal() ([]byte, error) {
-	return fxamackercbor.Marshal(a, fxamackercbor.EncOptions{})
-}
-
 // UnmarshalCBOR must implement cbg.Unmarshaller to insert this into a hamt.
```
anorth: This comment might be false after @ZenGround0's recent work on HAMT

ZenGround0: I tried to remove this during that HAMT work but specs actors adts used in the state view ended up depending on this too.  One day we should clean it up but its very low priority right now.

anorth: Ok. It would be possible to change the state view to not use the ADT abstractions at the top level and hence not rely on this.

PR #3800, opened 2020-02-21
Implement chain randomness for actors, to spec.

anorth:
### Motivation
The randomness VM syscall has recently been defined with domain tags and entropy, and spec PR https://github.com/filecoin-project/specs/pull/883 defines how to draw a seed from the chain and mix entropy to produce randomness. 

### Proposed changes
This PR implements the new randomness functions and hooks them up to the VM.

@sternhenri I made some inferences based on my comments in 833 and our discussion. Please look at the chain sampling and randomness computation code and point out if I've done something inconsistent with what you intend for the spec.

@icorderi based on your addition of the randomness input to `ProcessTipSetMessages` I assumed that the design direction is toward a stateless re-usable, rather than one-shot, VM, so I removed the randomness source from the VM's state to the invocation context. The randomness source implicitly curries in the head tipset (identifying the chain to sample). Alternatively we could move in the stateful but one-shot direction (which is the only way it's currently used anyway and move this state to construction instead. (The epoch should move with it to cement this approach).

@ZenGround0 FYI.

This change _does not_ fix up the other places that sample/compute randomness, so they're
currently inconsistent. A follow-up will replace use of the old `sampling` package with this code. (The spec has yet to fully define the parameters to all these calls).


internal/pkg/chain/sampler.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, err
+		}
+		ticket, err = tip.MinTicket()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// Sampling for the genesis block.
+		ticket.VRFProof = []byte{}
+	}
```
sternhenri: I think this might be wrong. Specifically, the genesis block will have an explicit genesis ticket. This does mean that `findTipsetAtEpoch` should be special cased so as to return the genesis tipset for any epoch `e <= 0`

internal/pkg/chain/sampler.go
```diff
@@ -0,0 +1,80 @@
...
+
+	buf := bytes.Buffer{}
+	buf.Write(ticket.VRFProof)
```
sternhenri: don't use the `VRFProof` here, use the VRF digest (a hash of the VRF proof)... Annoying I know, but this is how VRFs are defined such that were we to change VRFs the interface would stay the same.

internal/pkg/chain/sampler.go
```diff
@@ -0,0 +1,80 @@
...
+		ts = iterator.Value()
+		h, err = ts.Height()
+		if err != nil {
+			return
+		}
+		if h <= epoch {
+			break
+		}
+	}
+	// If the iterator completed, ts is the genesis tipset.
```
sternhenri: great

internal/pkg/crypto/randomness.go
```diff
@@ -0,0 +1,66 @@
...
+
+// A randomness source for use when computing genesis state (the state that the genesis block points to as parent state).
+// There is no chain to sample a seed from.
+type GenesisRandomnessSource struct{}
+
+func (g *GenesisRandomnessSource) Randomness(tag crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) (abi.Randomness, error) {
+	if epoch > 0 {
+		return nil, fmt.Errorf("invalid use of genesis randomness source for epoch %d", epoch)
+	}
+	seed := []byte{}
```
sternhenri: again, I think this is wrong.

internal/pkg/chain/sampler.go
anorth: Just checking: another blake2 hash is taken just a couple of lines further down (of the buffer including the epoch). Do you still maintain the proof must be hashed separately first?

internal/pkg/chain/sampler.go
sternhenri: Yeah, my understanding from Lotus' plan which seems good to me is that this will be the same indeed (ie the initial ticket value, which will be used for the first LE). That value will be specified closer to launch. You can coordinate with Lotus on what to use for Painnet, but best have it as a placeholder either way.

internal/pkg/chain/sampler.go
sternhenri: I do alas.

This would make it easier for us to swap out this VRF for another without changing interfaces should we ever choose (Zen can help motivate this if useful)

internal/pkg/crypto/randomness.go
sternhenri: Insofar as I understand it, we will want to have a specific pseudo-random value for use so the initial set of miners can preSeal their sectors for genesis. I don't see a reason why it couldn't be the one you define above (though there may be some reason to keep it secret until a chosen moment?). Just need to make sure it's the same as w lotus...

PR #3804, opened 2020-02-25
integrate chain-val tipset application

frrist:
### Motivation
Test TipSet application using chain-validation


internal/pkg/types/signed_message.go
```diff
@@ -110,8 +110,17 @@ func (smsg *SignedMessage) VerifySignature() error {
 }
 
 // OnChainLen returns the amount of bytes used to represent the message on chain.
+// Dragons: verify this is the correct way to determine a signed messages size on chain.
```
frrist: I am unsure the below implementation is correct, but need this method to not panic else it borks the chain-validation tests.

anorth: Just encode and measure the size of the whole message. Remove the dragons, but leave a todo that we could save this redundant encoding if we plumbed the size through from when the message was originally decoded from the network.

internal/pkg/vm/internal/vmcontext/testing.go
```diff
@@ -53,9 +57,15 @@ func (f *Factories) NewKeyManager() vstate.KeyManager {
...
+}
+
+func (r fakeRandSrc) Randomness(_ context.Context, _ acrypto.DomainSeparationTag, _ abi.ChainEpoch, _ []byte) (abi.Randomness, error) {
+	panic("implement me")
+}
+
 func (f *Factories) NewRandomnessSource() vstate.RandomnessSource {
-	// TODO implement in a follow on
-	return nil
+	return &fakeRandSrc{}
```
anorth: I have something coming soon. But I think for chain validation tests, we'll need this to be provided by the chain val code as part of the fake execution context that it provides. Otherwise different implementations will fake this differently.

internal/pkg/vm/internal/vmcontext/testing.go
```diff
@@ -302,6 +301,72 @@ func (a *ValidationApplier) ApplyMessage(context *vtypes.ExecutionContext, state
...
+			ours[i].SECPMessages = append(ours[i].SECPMessages, oursecp)
+		}
+	}
+	return ours
+}
+
+func (a *ValidationApplier) ApplyTipSetMessages(state vstate.VMWrapper, blocks []vtypes.BlockMessagesInfo, epoch abi.ChainEpoch, rnd vstate.RandomnessSource) ([]vtypes.MessageReceipt, error) {
+	st := state.(*ValidationVMWrapper)
+
+	// XXX: unsure if this is redundant
```
anorth: It's not; remove the comment.

internal/pkg/vm/internal/vmcontext/testing.go
```diff
@@ -302,6 +301,72 @@ func (a *ValidationApplier) ApplyMessage(context *vtypes.ExecutionContext, state
...
+	st.vm.currentEpoch = epoch
+
+	ourBlkMsgs := toOurBlockMessageInfoType(blocks)
+	receipts, err := st.vm.ApplyTipSetMessages(ourBlkMsgs, epoch, rnd)
+	if err != nil {
+		return nil, err
+	}
+
+	// XXX: unsure if this is redundant.
```
anorth: Please do figure this out before landing. If you can't work it out, I guess leave the code (but not the XXX) here

internal/pkg/vm/internal/vmcontext/testing.go
```diff
@@ -302,6 +301,72 @@ func (a *ValidationApplier) ApplyMessage(context *vtypes.ExecutionContext, state
...
+	if err := st.PersistChanges(); err != nil {
+		return nil, err
+	}
+
+	theirReceipts := make([]vtypes.MessageReceipt, len(receipts))
+	for i, r := range receipts {
+		theirReceipts[i] = vtypes.MessageReceipt{
+			ExitCode:    r.ExitCode,
+			ReturnValue: r.ReturnValue,
+			GasUsed:     big.Int(r.GasUsed),
```
anorth: GasUnits should be an int64. This suggests the chain validation code is using the wrong type.

frrist: Correct, chain validation is using the wrong type. I will update it and fix here in a follow on.

PR #3806, opened 2020-02-26
refactor: cleanup vm state 

icorderi:
### Motivation

The actor object cache could cause nasty bugs with the actor posters running loose. 

### Proposed changes

- remove tree loader
- remove cached tree 
- rename tree to state


internal/pkg/state/view.go
```diff
@@ -180,7 +180,7 @@ func (v *View) loadPowerClaim(ctx context.Context, powerState *power.State, mine
 		return nil, err
 	}
 	if !found {
-		return nil, errors.Errorf("no registered miner %v", miner)
+		return nil, types.ErrNotFound
```
anorth: This is an unfortunate loss of information. Does anything actually catch and check this error? If not, consider reverting.

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -128,9 +128,14 @@ func (ctx *invocationContext) invoke() interface{} {
 		return id
 	})
 
+	// 3. update actor state
+	// Dragons: this is wrong, we need to load it back up, change the head, and store it other we risk losing/fabricating money
```
anorth: Idea for your follow-up: a compare-and-set operation for actors.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -334,8 +352,11 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize uint32, rn
 	}
 
 	// Dragons: change this to actor, ok, error ok for found or not, err are non-recoverable IO errors and such
```
anorth: Remove this comment?

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -368,6 +389,10 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize uint32, rn
 	// - sender balance withheld
 	// - (if did not previously exist, the new auto-created target actor)
 
+	if err := vm.state.SetActor(vm.context, msg.From, fromActor); err != nil {
+		panic(err)
+	}
+
```
anorth: This behaviour was reverted from the spec, so the comment above (copied from a prior version) is suggesting the wrong thing. The spec interpreter now says:

```
// At this point, construct compTreePreSend as a state snapshot which includes
// the sender paying gas, and the sender's CallSeqNum being incremented;
// at least that much state change will be persisted even if the
// method invocation subsequently fails.
```

anorth: cc @frrist for messages tests, can you check that the implicit destination actor is not created in this case?

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -462,8 +487,11 @@ func (vm *VM) transfer(debitFrom address.Address, creditTo address.Address, amou
 	ctx := context.Background()
 
 	// retrieve debit account
-	fromActor, err := vm.state.GetActor(ctx, debitFrom)
+	fromActor, found, err := vm.state.GetActor(ctx, debitFrom)
 	if err != nil {
+		panic(err)
```
anorth: FWIW I think it would be more helpful to propagate an error from this method, deferring any panic or handling to higher up.

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+type State struct {
+	store    cbor.IpldStore
+	dirty    bool
+	root     Root
+	rootNode *hamt.Node
+}
+
+// NewState creates a new VM state.
+func NewState(store cbor.IpldStore) *State {
+	// Review: we should probably start the tree as dirty given that the current root does not match the state it will persist
```
anorth: I think this is only used for testing and genesis creation. An alternative would be for this method to write the empty hamt root to the store, the side-effect explicit with a name like `CreateEmptyState()`.

But otherwise, yes dirty.

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+	return newState(store, cid.Undef, hamt.NewNode(store, hamt.UseTreeBitWidth(TreeBitWidth)))
+}
+
+// LoadState creates a new VMStorage.
+func LoadState(ctx context.Context, store cbor.IpldStore, root Root) (*State, error) {
+	// Review: do we want to special case root=cid.Undef to be a new state?
```
anorth: I'd prefer to fail.

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	"github.com/pkg/errors"
+)
+
+// Review: can we get rid of this?
+type Tree interface {
```
anorth: The interface may be handy to keep, but could probably be declared in consumers (Go style) rather than here (Java style). Your call.

However,  the term "state tree" is a widely used and useful term for the top level of the state. Assuming you delete or move this interface elsewhere, I suggest renaming `State` below to `Tree`, which is a much more descriptive name to people who carry the state tree idea in their head already. `State` is already overloaded.

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// Commit will flush the state tree into the backing store.
+// The new root is returned.
+func (st *State) Commit(ctx context.Context) (Root, error) {
+	if err := st.rootNode.Flush(ctx); err != nil {
+		return cid.Undef, err
+	}
+
+	// Review: is this put needed because it has not been flushed or because flush does not return the root cid?
```
anorth: I believe the root node is not `Put` in the store on `Flush`, only children are (so Flush doesn't yet have the CID).

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+// GetAllActorsResult is the struct returned via a channel by the GetAllActors
+// method. This struct contains only an address string and the actor itself.
+type GetAllActorsResult struct {
+	Key   actorKey
+	Actor *actor.Actor
+	Error error
+}
+
+// GetAllActors returns a channel which provides all actors in the StateTree.
+func (st *State) GetAllActors(ctx context.Context) <-chan GetAllActorsResult {
```
anorth: I think this is way too deep to use a channel and introduce asynchrony. A callback would work just as well here. You can copy/paste from specs-actors [adt/map](https://github.com/filecoin-project/specs-actors/blob/ad3d104d72523b18d6eccd9db140f43a00eaea3c/actors/util/adt/map.go#L97) which already implements this. In fact in the state view I directly used the Map ADT [to inspect the top of the state tree](https://github.com/filecoin-project/go-filecoin/blob/86ddd4e6d4278943d579a5e7362c3c16395e6a37/internal/pkg/state/view.go#L237). (Change the direct use of CBORMarshal if you wish).

(A channel higher up close to the CLI is a bit better, though a callback or iterator would still be better. I'm not suggesting you change that - just adapt to the channel higher up).

internal/pkg/vm/state/state.go
```diff
@@ -0,0 +1,204 @@
...
+	out := make(chan GetAllActorsResult)
+	go func() {
+		defer close(out)
+		st.getActorsFromPointers(ctx, out, st.rootNode.Pointers)
+	}()
+	return out
+}
+
+// NOTE: This extracts actors from pointers recursively. Maybe we shouldn't recurse here.
```
anorth: I think the depth is very restricted (log N) so it's ok.

Even if you challenge me on using a channel, please use the existing HAMT ForEach method rather than re-implement this deep inspection.

internal/pkg/vm/state/testing.go
```diff
@@ -8,138 +8,48 @@ import (
...
-	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/dispatch"
 )
 
-// MustFlush flushes the StateTree or panics if it can't.
-func MustFlush(st Tree) cid.Cid {
-	cid, err := st.Flush(context.Background())
+// NewFromString sets a state tree based on an int.  TODO: this indirection
+// can be avoided when we are able to change cborStore to an interface and then
+// making a test implementation of the cbor store that can map test cids to test
+// states.
```
anorth: It is an interface now, so maybe update to "We could avoid this indirection by ..."

internal/pkg/vm/internal/vmcontext/vmcontext.go
icorderi: Yes, there are a few places now that should change to error instead of panic. 
I'll see to this on another PR.

internal/pkg/vm/state/state.go
icorderi: this is not new code, git didn't detect the file as changed.

internal/pkg/vm/state/state.go
icorderi: inherited code, it's also only used in 1 test and to list the actors in CLI.

PR #3807, opened 2020-02-26
Hook up new randomness to seals and proofs

anorth:
### Motivation
The new specced randomness generation routines should replace our prior chain sampling mechanisms.

### Proposed changes
Replace all uses of `sampling/SampleChainRandomness` with new randomness. Along the way changed the interface to randomness consumed by the processor to enable sharing implementation.

Note that the election ticket generation is done differently and still needs to be reconciled with this approach.

### DIFF
--- a/internal/pkg/chain/sampler.go
+++ b/internal/pkg/chain/sampler.go
@@ -12,7 +12,20 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/crypto"
 )
 
-// A sampler draws randomness seeds from the chain.
+// Creates a new sampler for the chain identified by `head`.
+func NewSamplerAtHead(reader TipSetProvider, head block.TipSetKey) *SamplerAtHead {
+	return &SamplerAtHead{
+		sampler: NewSampler(reader),
+		head:    head,
+	}
+}
+
+// A sampler draws randomness seeds from the chain. The seed is computed from the minimum ticket of the tipset
+// at or before the requested epoch, mixed with the epoch itself (and is thus unique per epoch, even when they are
+// empty).
+//
+// This implementation doesn't do any caching: it traverses the chain each time. A cache that could be directly
+// indexed by epoch could speed up repeated samples from the same chain.
 type Sampler struct {
 	reader TipSetProvider
 }
@@ -48,7 +61,8 @@ func (s *Sampler) Sample(ctx context.Context, head block.TipSetKey, epoch abi.Ch
 	}
 
 	buf := bytes.Buffer{}
-	buf.Write(ticket.VRFProof)
+	vrfDigest := blake2b.Sum256(ticket.VRFProof)
+	buf.Write(vrfDigest[:])
 	err := binary.Write(&buf, binary.BigEndian, epoch)
 	if err != nil {
 		return nil, err
@@ -77,4 +91,15 @@ func (s *Sampler) findTipsetAtEpoch(ctx context.Context, start block.TipSet, epo
 	}
 	// If the iterator completed, ts is the genesis tipset.
 	return
+}
+
+///// A chain sampler with a specific head tipset key. /////
+
+type SamplerAtHead struct {
+	sampler *Sampler
+	head    block.TipSetKey
+}
+
+func (s *SamplerAtHead) Sample(ctx context.Context, epoch abi.ChainEpoch) (crypto.RandomSeed, error) {
+	return s.sampler.Sample(ctx, s.head, epoch)
 }
--- a/internal/pkg/chain/sampler.go
+++ b/internal/pkg/chain/sampler.go
@@ -1,21 +1,18 @@
 package chain
 
 import (
-	"bytes"
 	"context"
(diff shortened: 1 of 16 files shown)


internal/pkg/chain/sampler.go
```diff
@@ -48,7 +61,8 @@ func (s *Sampler) Sample(ctx context.Context, head block.TipSetKey, epoch abi.Ch
 	}
 
 	buf := bytes.Buffer{}
-	buf.Write(ticket.VRFProof)
+	vrfDigest := blake2b.Sum256(ticket.VRFProof)
```
anorth: By request of @sternhenri , https://github.com/filecoin-project/go-filecoin/pull/3800#discussion_r383325303

internal/pkg/chain/sampler.go
```diff
@@ -12,7 +12,20 @@ import (
...
+// Creates a new sampler for the chain identified by `head`.
+func NewSamplerAtHead(reader TipSetProvider, head block.TipSetKey) *SamplerAtHead {
+	return &SamplerAtHead{
+		sampler: NewSampler(reader),
+		head:    head,
+	}
+}
+
+// A sampler draws randomness seeds from the chain. The seed is computed from the minimum ticket of the tipset
+// at or before the requested epoch, mixed with the epoch itself (and is thus unique per epoch, even when they are
```
ZenGround0: I wouldn't mind something more precise: "randomness seeds are the black2b hash of the vrf digest of the minimum ticket of the tipset at or before the requested epoch concatenated with the big endian bytes of the epoch value"

ZenGround0: This is particularly useful because it makes it clear entropy bytes and dsts don't make it into the sampled bytes.

internal/pkg/crypto/randomness.go
```diff
@@ -13,22 +14,22 @@ import (
...
 }
 
 // A randomness source that seeds computations with a sample drawn from a chain epoch.
 type ChainRandomnessSource struct {
 	Sampler ChainSampler
 }
 
-func (c *ChainRandomnessSource) Randomness(tag crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) (abi.Randomness, error) {
-	seed, err := c.Sampler.Sample(epoch)
+func (c *ChainRandomnessSource) Randomness(ctx context.Context, tag crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) (abi.Randomness, error) {
```
ZenGround0: nit: It's unclear why we separate "entropy blending" from "epoch blending".  Do we ever use the raw output of sampler?  It seems like blendEntropy should be a subfunction on the sampler and should do epoch mixing along with entropy bytes and tag mixing.

internal/pkg/chain/sampler.go
```diff
@@ -43,19 +55,11 @@ func (s *Sampler) Sample(ctx context.Context, head block.TipSetKey, epoch abi.Ch
...
-		ticket.VRFProof = []byte{}
-	}
-
-	buf := bytes.Buffer{}
-	buf.Write(ticket.VRFProof)
-	err := binary.Write(&buf, binary.BigEndian, epoch)
-	if err != nil {
-		return nil, err
+		// Sampling for the genesis state or genesis tipset.
+		ticket = s.genesisTicket
```
anorth: FYI @sternhenri a genesis ticket is now used instead of empty, and that ticket still mixed with the epoch number. The same ticket is used whenever there is no parent block: both pre-genesis state and the evaluation of the genesis block.

internal/pkg/crypto/randomness.go
anorth: This is how I understand it from @sternhenri . I agree with you from a software org point of view it seems weird to first hash the ticket, then hash (that + epoch), and then hash (that + tag + entropy).

ZenGround0: small thing: first hashing the ticket makes sense from a crypto standpoint because sign then hash is one conceptual operation.

Agree about the next two hashes being weird but I am totally ok with taking this requirement from the spec.

Even taking the two hashes from the spec from our software point of view it still makes more sense to me for one thing/place to be doing all of the levels of hashing.  I don't think there is anything that consumes partial outputs, so why leave multiple layers of indirection here?

anorth: Thanks. We can fix the organization later if it proves possible.

> I don't think there is anything that consumes partial outputs

I haven't hooked up ticket generation yet and am pretty worried it's different.

internal/pkg/chain/sampler.go
anorth: I'll improve the comment, but sampling has nothing at all to do with entropy or DSTs. Those words do not appear in the chain package. I think mentioning them would be more confusing, and certainly prone to rot.

anorth: Actually because MakeRandomSeed moved to `crypto`, I'll put the comment there.

PR #3808, opened 2020-02-26
Retrieval market: Retrieval provider connector

shannonwells:
### Motivation
This is a subset of changes to implement retrieval market for go-filecoin, pulled from draft PR #3741 in order to make brain-friendly PRs

### Proposed changes
* RetrievalProviderConnector implementation
* Instantiate RetrievalMarketSubmodule in the node, with only a provider connector
* tests for the provider
* uses **stable branch of go-fil-markets**
* uses **latest master of specs-actors** (as of this writing)

### For follow-up PRs
* update to use latest go-fil-markets with persisted deal store (sectorbuilder will need updates also to use it)
* RetrievalClientConnector
* RetrievalPeerResolverConnector
* PaymentChannelManager, to manage persisting to payment channel store and posting messages on chain
* end to end Retrieval integration test


internal/app/go-filecoin/paymentchannel/types.go
```diff
@@ -0,0 +1,19 @@
+package paymentchannel
```
shannonwells: the paymentchannel dir is where the paymentchannel manager will live.

internal/app/go-filecoin/retrieval_market_connector/network.go
```diff
@@ -1,29 +0,0 @@
-package retrievalmarketconnector
```
shannonwells: this connector actually isn't needed.

internal/pkg/state/view.go
```diff
@@ -52,7 +52,7 @@ func NewView(store cbor.IpldStore, root cid.Cid) *View {
 	}
 }
```
shannonwells: To please the linters

internal/app/go-filecoin/node/node.go
```diff
@@ -465,12 +470,28 @@ func (node *Node) setupStorageMining(ctx context.Context) error {
 		return errors.Wrap(err, "error initializing storage protocol")
 	}
 
-	// TODO: Retrieval Market Integration
```
shannonwells: this was left in the func to set up storage mining.  instead put the initialization in its own func, which will be called in a later PR.

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -35,7 +35,7 @@ func (sys syscalls) HashBlake2b(data []byte) [32]byte {
 
 // ComputeUnsealedSectorCID implements Syscalls.
 // Review: why is this returning an error instead of aborting? is this failing recoverable by actors?
-func (sys syscalls) ComputeUnsealedSectorCID(sectorSize abi.SectorSize, pieces []abi.PieceInfo) (cid.Cid, error) {
```
shannonwells: for compatibility with latest specs-actors

internal/app/go-filecoin/retrieval_market_connector/provider.go
```diff
@@ -1,28 +1,117 @@
...
+// UnsealSector unseals the sector given by sectorId and offset with length `length`
+// It rejects offsets > int size and length > int64 size; the interface wants
+// uint64s. This would return a bufio overflow error anyway, but the check
+// is provided as a debugging convenience for the consumer of this function.
+func (r *RetrievalProviderConnector) UnsealSector(ctx context.Context, sectorID uint64,
+	offset uint64, length uint64) (io.ReadCloser, error) {
+	// reject anything that's a real uint64 rather than trying to get cute
+	// and offset that much or copy into a buf that large
+	intSz := reflect.TypeOf(0).Size()*8 - 1
+	maxOffset := uint64(1 << intSz)
```
acruikshank: I'm not feeling great about using `reflect.TypeOf(0).Size()*8 - 1`. Here's a simpler maxint version I got off the internet `const MaxInt = int(^uint(0) >> 1)`.

internal/app/go-filecoin/retrieval_market_connector/provider.go
```diff
@@ -1,28 +1,117 @@
...
+	if length >= math.MaxInt64 {
+		return nil, xerrors.New("length overflows int64")
+	}
+
+	unsealedSector, err := r.unsealer.UnsealSector(ctx, sectorID)
+	if err != nil {
+		return nil, err
+	}
+
+	fname := fmt.Sprintf("retrieval_market_%d_%d_%d", sectorID, offset, length)
```
acruikshank: It's possible to use a LimitedReader to avoid buffering to a temporary file:
``` go
bufr := bufio.NewReader(unsealedSector)
_, err := bufr.Discard(int(offset))
if err != nil {
  return nil, err
}
return &io.LimitedReader{R: bufr, N: int64(length)}, nil
```
Is there a reason not to?

shannonwells: No, I just didn't know about LimitedReader.

internal/app/go-filecoin/retrieval_market_connector/provider.go
shannonwells: Thanks, your google fu is better than mine

internal/app/go-filecoin/retrieval_market_connector/provider.go
```diff
@@ -1,28 +1,117 @@
...
+		return nil, xerrors.Errorf("could not seek to beginning")
+	}
+
+	return res, nil
+}
+
+// SavePaymentVoucher stores the provided payment voucher with the payment channel actor
+func (r *RetrievalProviderConnector) SavePaymentVoucher(_ context.Context, paymentChannel address.Address, voucher *paych.SignedVoucher, proof []byte, expected abi.TokenAmount) (abi.TokenAmount, error) {
+
+	_, err := r.paychMgr.GetPaymentChannelInfo(paymentChannel)
```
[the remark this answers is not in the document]
shannonwells: getting channel info doesn't alter state. it's to error early if the info doesn't exist.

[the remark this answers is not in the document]
acruikshank: That should probably be reflected in a comment or in the error message then.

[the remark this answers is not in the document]
shannonwells: I added a comment.

Issue #3457, opened 2019-09-18
Replace VMContext ancestors with interface to chain


acruikshank:
#### Problem

Actors need access to previous chain state in order to derive randomness from the chain. Our current solution to this is to pull some number of tipsets out of the chain, and put them in memory in the `VMContext` to supply data to the context's `SampleChainRandomness` function. 

Since our one use case for this is `SubmitPoSt` and submit PoSt needs to be able to verify late PoSts from the previous proving period, we always need to cache `proving period + proving window` tipsets (minus null blocks). This is currently set to 450. If there are no late PoSts, the most we will need to look back is the proving window (150).

Since our message calling logic doesn't not know whether or not `SampleChainRandomness` will be called, we need to generate this cache for every block we process. Since every message in a tipset is at the same height, we only need to generate the tipset cache once per tipset, but we throw it away and recalculate it each time after that. 

Generating the tipset cache would be way too expensive for query messages or preview messages. Consequently, submitPoSt cannot currently be previewed. This is a big gap in the previewing system. It would be nice to be able to query whether or not a PoSt is valid. We might have other needs for querying chain randomness in the future.

Finally, since we only provide an interface for chain randomness, keeping the entire tipset in memory is wasteful. We should only cache what we need to compute the randomness.

#### Proposal

We should replace the ancestors slice we pass to all the messaging logic with an interface to the chain. This interface should be to a cache that is persistent in memory. It should be able to pull chain state when it needs to, but maintain a memory efficient cache for chain randomness. This cache may be updated on heaviest tipset, or it may pull from the chain whenever randomness is requested at a higher block height (reorgs will still need to be accounted for).

anorth:
👍 on the problem description.

It's not clear what "interface to the chain" looks like.
- One possibility is a fairly direct interface that talks in terms of blocks and tipsets etc, exposing the data structures directly. 
- Another is to expose very tight methods that return narrow pieces of data pre-processed for use by actors, e.g. `SampleTicket(height) []byte`.

Of these, my initial preference is for the latter. It exposes the minimum surface area, removes expensive logic from actors, gives implementation flexibility and will be easiest to fake out for testing.

acruikshank:
Agreed that the interface should expose tight method. That may result in better optimizations (e.g. initially a cache would only need to store heights and tickets (and maybe tipSet keys).

anorth:
See also #3025

ZenGround0:
> Of these, my initial preference is for the latter

[...]

PR #3810, opened 2020-02-27
Use new randomness for ticket and epost challenge

### LINKED ISSUE #2223: EC Upgrade 2: randomness sampling from ticket chain

anorth:
The spec (in https://github.com/filecoin-project/specs/pull/883/) defines a common randomness generation routine for ticket and EPoSt challenge derivation. Go-filecoin already implements this for  fallback PoSt challenge derivation and other uses. This PR replaces old ticket and challenge derivation with the new stuff. The key difference from the other randomness calls is that the result of sampling the chain (+ entropy) is then signed as a VRF proof, and a digest of this proof taken for use as an EPoSt challenge seed.

The existing consensus and mining tests are now a bit awkward - the new randomness implementation fully abstracts the chain away from the election mechanisms, but the current test structure can't really make use of that. I haven't confirmed whether or not the skipped tests will work, so it's possible this change introduces another barrier. @ZenGround0 please look closely.

Closes #2223, #2225, #3457, #3025


internal/pkg/mining/worker.go
```diff
@@ -343,10 +338,25 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
...
 	return consensus.NewPowerTableView(view), nil
 }
+
+type ePoStChallengeSeedEntropy struct {
+	// control field for encoding struct as an array
+	_     struct{} `cbor:",toarray"`
+	Epoch abi.ChainEpoch
+	Miner address.Address
+}
+
```
sternhenri: Thiis is no longer necessary. 

Given the changes we agreed to wrt passing in the epoch as part of drawing a ticket, we no longer need it here. Only the miner address

internal/pkg/consensus/testing.go
```diff
@@ -239,14 +242,14 @@ func SeedFirstWinnerInNRounds(t *testing.T, n int, ki *crypto.KeyInfo, networkPo
...
-	signer := types.NewMockSigner([]crypto.KeyInfo{*ki})
-	wAddr, err := ki.Address()
-	require.NoError(t, err)
+func winsAtEpoch(t *testing.T, epoch abi.ChainEpoch, ticket block.Ticket, miner address.Address, networkPower, numSectors, sectorSize uint64, sectorInfos ffi.SortedPublicSectorInfo) bool {
 	em := ElectionMachine{}
 
 	// use ticket to seed postRandomness
-	postRandomness, err := em.GeneratePoStRandomness(ticket, wAddr, signer, epoch)
+	rnd := crypto.FakeRandomnessSource{Ticket: ticket}
+	entropy, err := mining.MakeElectionPoStChallengeSeedEntropy(epoch, miner)
```
sternhenri: no need to take epoch here. entropy is just miner id address (cbor serialized)

internal/pkg/crypto/testing.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+)
+
+type FakeRandomnessSource struct {
+	Ticket block.Ticket
+}
+
+func (f *FakeRandomnessSource) Randomness(_ context.Context, tag crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) (abi.Randomness, error) {
+	buffer := bytes.Buffer{}
+	// Real randomness is more complicated than just writing the ticket and epoch
```
sternhenri: I'm not sure if i should comment below, but it is wrong indeed

internal/pkg/mining/worker.go
```diff
@@ -208,32 +204,31 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
...
-		outCh <- Output{Err: err}
-		return
-	}
-	electionTicket, err := sampling.SampleNthTicket(int(miner.ElectionLookback-1), ancestors)
+	// Look-back for the election ticket.
+	// The parameter is interpreted as: lookback=1 means parent tipset. Subtract one here because the base from
+	// which the lookback is counted is already the parent, rather than "current" tipset.
+	// The sampling code will handle this underflowing past the genesis.
+	targetEpoch := baseEpoch - (miner.ElectionLookback - 1) + abi.ChainEpoch(nullBlkCount)
+	entropy, err := MakeElectionPoStChallengeSeedEntropy(targetEpoch, w.minerAddr)
```
sternhenri: epoch is used as part of samplinig randomness below now, no need for it in entropy here

internal/pkg/mining/worker.go
```diff
@@ -332,7 +327,7 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
 		post = postOut
 	}
 
-	postInfo := block.NewEPoStInfo(post, postRandomness, block.FromFFICandidates(winners...)...)
+	postInfo := block.NewEPoStInfo(post, block.VRFPi(postRandomness), block.FromFFICandidates(winners...)...)
```
sternhenri: this is right, but you want the VRF digest, not the VRFPi (the digest is a hash of the Pi)

internal/pkg/consensus/testing.go
```diff
@@ -239,14 +242,14 @@ func SeedFirstWinnerInNRounds(t *testing.T, n int, ki *crypto.KeyInfo, networkPo
...
+	require.NoError(t, err)
+	postRandomness, err := rnd.Randomness(context.Background(), acrypto.DomainSeparationTag_ElectionPoStChallengeSeed, epoch, entropy)
```
ZenGround0: Going by the interface this won't work as you need a signing key to generate the post vrf proof.

Continuing slack conversation here: PostRandomness is a bad name because it made you think this was the way to get it.  We should consider PRing a change to the spec to indicate that it is a VRF Proof derived off of the ticket chain with a miner sk, not something you can generate with publicly known values.

internal/pkg/consensus/testing.go
```diff
@@ -239,14 +242,14 @@ func SeedFirstWinnerInNRounds(t *testing.T, n int, ki *crypto.KeyInfo, networkPo
...
 	require.NoError(t, err)
 
 	// does this postRandomness create a winner?
```
ZenGround0: The call to GenerateCandidates below consumes the postRandomness VRF proof.  Right now go-sectorbuilder interface says that it wants to take in an abi.PoStRandomness.  I’m not sure if that type name matches the spec like go-filecoin’s code so you’ll want to investigate to see if you actually want to draw randomness from the vrf proof you have here before passing in.

internal/pkg/mining/worker.go
```diff
@@ -208,32 +204,31 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
...
+	targetEpoch := baseEpoch - (miner.ElectionLookback - 1) + abi.ChainEpoch(nullBlkCount)
+	entropy, err := MakeElectionPoStChallengeSeedEntropy(targetEpoch, w.minerAddr)
 	if err != nil {
-		log.Warnf("Worker.Mine couldn't read parent ticket %s", err)
+		log.Errorf("Worker.Mine failed to encode entropy %s", err)
 		outCh <- Output{Err: err}
 		return
 	}
-	postRandomness, err := w.election.GeneratePoStRandomness(electionTicket, workerAddr, w.workerSigner, nullBlkCount)
+	postRandomness, err := w.api.SampleChainRandomness(ctx, base.Key(),
```
ZenGround0: Same issue as above

internal/pkg/mining/worker.go
ZenGround0: No you do want the VRFPi here, this is what goes on chain which needs to be the proof for verification reasons.

ZenGround0: One thing that I was meaning to do if I ended up doing this change is realign our types to correctly capture the difference between VRF proofs and digests.  I think it is worth doing to keep it clear which data is a proof (signature output) vs which is a digest (hash of signature output).

internal/pkg/consensus/testing.go
ZenGround0: postVRFProof is a nice boring and precise name

ZenGround0: postVRFProof is a nice boring and precise name

internal/pkg/mining/worker.go
```diff
@@ -208,32 +204,31 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
...
 		outCh <- Output{Err: err}
 		return
 	}
-	ancestors, err := w.getAncestors(ctx, base, baseHeight+(abi.ChainEpoch(nullBlkCount+1)))
-	if err != nil {
-		log.Warnf("Worker.Mine couldn't get ancestorst %s", err)
-		outCh <- Output{Err: err}
-		return
-	}
-	electionTicket, err := sampling.SampleNthTicket(int(miner.ElectionLookback-1), ancestors)
```
ZenGround0: To make this PR right I believe you should sample `electionRandomness` (no longer a ticket) and then pass that election randomness into the existing postProofGeneration code (modulo changing ticket type to randomness type).

There is another step maybe done by the sectorbuilder or maybe we have to do it that transforms the postRandomness VRF pi into a post challenge seed for use in the GenerateCandidates function.  We'll need to consult the spec but that transformation is at least another hash to get vrf digest and maybe it also mixes in epoch / other entropy.

cc @sternhenri to verify that the order of operations to get epost candidates is: 1) draw lookback randomness 2) create post randomness vrf Pi off of that randomness 3) draw randomness (vrf digest / maybe mising) from post randomness for generatecandidates.

internal/pkg/mining/worker.go
anorth: I was misled by this line, which explicitly mentions epoch for the epost challenge seed, even though all cases already have the epoch mixed in. https://github.com/filecoin-project/specs/pull/883/files#diff-760dd431bb3ed63837eb875d1b328bfcR95

internal/pkg/consensus/testing.go
anorth: I still need to investigate this.

anorth:
@ZenGround0 @sternhenri please take another look for me. I haven't fixed the tests etc yet, just the mining worker.

Discussion with @sternhenri made it clear that I was closer than I thought, the main thing missing being to sign the output of the existing chain randomness generation code already written. I believe now correctly using the vrfproof and its digest properly.

anorth:
Also now implements tickets exactly the same way, but with a different DST.

ZenGround0:
Overall looks good.  I'll pour through the details once this is out of draft and ready for review

internal/pkg/consensus/testing.go
anorth: There's nothing in the Go side that takes the digest. Lotus also takes a digest (SHA :-( ) before invoking the sectorbuilder method. 

There are numerous other divergences in Lotus from the chain sampling and randomness generation code  I've written, FYI @sternhenri, though the general structures now match.

internal/pkg/chain/traversal.go
```diff
@@ -117,3 +119,68 @@ func CollectTipsToCommonAncestor(ctx context.Context, store TipSetProvider, oldH
 	newTips, err = CollectTipSetsOfHeightAtLeast(ctx, newIter, commonHeight+1)
 	return
 }
+
+// ErrNoCommonAncestor is returned when two chains assumed to have a common ancestor do not.
+var ErrNoCommonAncestor = errors.New("no common ancestor")
```
anorth: Changes here are just a vanilla move from the deleted ancestors file. No need to review (ditto for the tests).

internal/pkg/consensus/election.go
```diff
@@ -1,37 +1,38 @@
...
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/util/convert"
 )
 
-// ElectionMachine generates and validates PoSt partial tickets and PoSt
-// proofs.
-type ElectionMachine struct{}
+// ElectionMachine generates and validates PoSt partial tickets and PoSt proofs.
+type ElectionMachine struct {
+	chain ChainRandomness
```
anorth: The huge argument lists were getting really old, and this now abstracts the chain from callers of the methods on this object. We could probably move the signer here too.

internal/pkg/consensus/expected_test.go
```diff
@@ -72,57 +73,20 @@ func TestExpected_RunStateTransition_validateMining(t *testing.T) {
...
-		exp := consensus.NewExpected(cistore, bstore, th.NewFakeProcessor(), &views, th.BlockTimeTest, mockElection, mockTicketGen, &proofs.ElectionPoster{})
-
-		nextBlocks := requireMakeNBlocks(t, 3, pTipSet, nextRoot, types.EmptyReceiptsCID, miners, m2w, mockSigner)
-		tipSet := th.RequireNewTipSet(t, nextBlocks...)
-
-		emptyBLSMessages, emptyMessages := emptyMessages(len(nextBlocks))
-		_, _, err = exp.RunStateTransition(ctx, tipSet, emptyBLSMessages, emptyMessages, ancestors, nextBlocks[0].ParentWeight, nextRoot, nextBlocks[0].MessageReceipts.Cid)
-		assert.NoError(t, err)
-	})
+	// TODO: test that the correct tickets are processed for election and ticket generation
```
anorth: This test (and use of the mock ticket/election machines) was too hard to keep since tickets are no longer simply generated from each other.

internal/pkg/consensus/testing.go
```diff
@@ -196,62 +197,57 @@ func NFakeSectorInfos(numSectors uint64) ffi.SortedPublicSectorInfo {
...
-// MakeFakeTicketForTest().
+// first such seed.
 //
 // Note that there are no guarantees that this function will terminate on new
 // inputs as miner power might be so low that winning a ticket is very
 // unlikely.  However runtime is deterministic so if it runs fast once on
 // given inputs is safe to use in tests.
-func SeedFirstWinnerInNRounds(t *testing.T, n int, ki *crypto.KeyInfo, networkPower, numSectors, sectorSize uint64) block.Ticket {
-	tm := TicketMachine{}
+func SeedFirstWinnerInNRounds(t *testing.T, n int, miner address.Address, ki *crypto.KeyInfo, networkPower, numSectors, sectorSize uint64) *FakeChainRandomness {
```
anorth: This stuff was brain-wracking to adapt to the new world. The tests are skipped so I'm not 100% sure this works.

internal/app/go-filecoin/internal/submodule/syncer_submodule.go
```diff
@@ -66,7 +66,10 @@ func NewSyncerSubmodule(ctx context.Context, config syncerConfig, repo chainRepo
 
 	// set up consensus
 	stateViewer := consensus.AsPowerStateViewer(state.NewViewer(blockstore.CborStore))
-	nodeConsensus := consensus.NewExpected(blockstore.CborStore, blockstore.Blockstore, chn.Processor, &stateViewer, config.BlockTime(), consensus.ElectionMachine{}, consensus.TicketMachine{}, postVerifier)
+	elections := consensus.NewElectionMachine(chn.State)
+	tickets := consensus.NewTicketMachine(chn.State)
+	nodeConsensus := consensus.NewExpected(blockstore.CborStore, blockstore.Blockstore, chn.Processor, &stateViewer,
+		config.BlockTime(), elections, tickets, postVerifier)
```
ZenGround0: Why not wrap postverifier in with election machine?

internal/pkg/chainsync/internal/syncer/syncer_test.go
```diff
@@ -417,7 +417,7 @@ func newPoisonValidator(t *testing.T, headerFailure, fullFailure uint64) *poison
 	return &poisonValidator{headerFailureTS: headerFailure, fullFailureTS: fullFailure}
 }
 
-func (pv *poisonValidator) RunStateTransition(_ context.Context, ts block.TipSet, _ [][]*types.UnsignedMessage, _ [][]*types.SignedMessage, _ []block.TipSet, _ fbig.Int, _ cid.Cid, _ cid.Cid) (cid.Cid, []vm.MessageReceipt, error) {
+func (pv *poisonValidator) RunStateTransition(ctx context.Context, ts block.TipSet, blsMessages [][]*types.UnsignedMessage, secpMessages [][]*types.SignedMessage, parentWeight fbig.Int, stateID cid.Cid, receiptRoot cid.Cid) (cid.Cid, []vm.MessageReceipt, error) {
```
ZenGround0: Do you have a stylistic objection to blanking out unused arguments?  To me it makes code clearer as you can immediately tell what doesn't get used when satisfying some interface with a stub.

internal/pkg/consensus/election.go
ZenGround0: If you want to run with making these types stateful I'm also not opposed to wrapping actor state accessors like the new view up in these and expanding their functionality to do, for example, power table lookups.

internal/pkg/consensus/election.go
```diff
@@ -92,23 +99,51 @@ func (em ElectionMachine) VerifyPoSt(ep verification.PoStVerifier, allSectorInfo
...
-		VRFProof: vrfPi.Data,
-	}, nil
+	return crypto.ValidateBlsSignature(randomness, worker, ticket.VRFProof)
 }
 
-// ValidateTicket verifies that the ticket's proof of randomness and delay are
-// valid with respect to its parent.
-func (tm TicketMachine) ValidateTicket(parent, ticket block.Ticket, signerAddr address.Address) error {
-	return crypto.ValidateBlsSignature(parent.VRFProof[:], signerAddr, ticket.VRFProof)
+func computeVrf(ctx context.Context, chain ChainRandomness, base block.TipSetKey, epoch abi.ChainEpoch, tag acrypto.DomainSeparationTag,
```
ZenGround0: nit: I'd prefer the name to reflect that this does randomness sampling work along with computing a VRF.  But nbd as the arguments mostly tell the story.

internal/pkg/consensus/expected.go
```diff
@@ -145,12 +131,13 @@ func (c *Expected) BlockTime() time.Duration {
 // RunStateTransition applies the messages in a tipset to a state, and persists that new state.
 // It errors if the tipset was not mined according to the EC rules, or if any of the messages
 // in the tipset results in an error.
-func (c *Expected) RunStateTransition(ctx context.Context, ts block.TipSet, blsMessages [][]*types.UnsignedMessage, secpMessages [][]*types.SignedMessage, ancestors []block.TipSet, parentWeight fbig.Int, parentStateRoot cid.Cid, parentReceiptRoot cid.Cid) (root cid.Cid, receipts []vm.MessageReceipt, err error) {
+func (c *Expected) RunStateTransition(ctx context.Context, ts block.TipSet, blsMessages [][]*types.UnsignedMessage, secpMessages [][]*types.SignedMessage,
```
ZenGround0: I see you doing this lately.  Is it the new style approach to long function signatures that I should adopt?

internal/pkg/consensus/protocol.go
```diff
@@ -29,7 +29,7 @@ import (
 type Protocol interface {
 	// RunStateTransition returns the state root CID resulting from applying the input ts to the
 	// prior `stateID`.  It returns an error if the transition is invalid.
-	RunStateTransition(ctx context.Context, ts block.TipSet, blsMsgs [][]*types.UnsignedMessage, secpMsgs [][]*types.SignedMessage, ancestors []block.TipSet, parentWeight fbig.Int, parentStateRoot cid.Cid, parentReceiptRoot cid.Cid) (cid.Cid, []vm.MessageReceipt, error)
```
ZenGround0: FYI I've been meaning to delete this forever.  In the future you can get rid of this instead of fixing.

internal/pkg/consensus/testing.go
ZenGround0: Yeah looks like it.  Thanks for heads up.

internal/pkg/chain/traversal.go
```diff
@@ -117,3 +119,68 @@ func CollectTipsToCommonAncestor(ctx context.Context, store TipSetProvider, oldH
...
+	var h abi.ChainEpoch
+	for ; !iterator.Complete(); err = iterator.Next() {
+		if err != nil {
+			return nil, err
+		}
+		h, err = iterator.Value().Height()
+		if err != nil {
+			return nil, err
+		}
+		if h < minHeight {
```
sternhenri: Should you not have it be `h < minHeight || h == 0` so the genesis is returned if needed?

internal/pkg/consensus/election.go
```diff
@@ -41,20 +42,26 @@ func (em ElectionMachine) GenerateCandidates(poStRand []byte, sectorInfos ffi.So
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to encode entropy")
+	}
+	randomness, err := em.chain.SampleChainRandomness(ctx, base, acrypto.DomainSeparationTag_ElectionPoStChallengeSeed, epoch, entropy)
+	if err != nil {
+		return errors.Wrap(err, "failed to generate epost randomness")
+	}
 
-	return crypto.ValidateBlsSignature(buf[:], candidateAddr, rand)
+	return crypto.ValidateBlsSignature(randomness, worker, vrfProof)
```
sternhenri: wfm. Note that the actual randomness value is the digest but checking this is fine, so long as the digest is what is passed to the proofs code for proof verification

internal/pkg/consensus/expected.go
```diff
@@ -284,7 +254,8 @@ func (c *Expected) validateMining(
 		if err != nil {
 			return errors.Wrapf(err, "failed to read sector infos from power table")
 		}
-		valid, err := c.VerifyPoSt(c.postVerifier, allSectorInfos, uint64(sectorSize), blk.EPoStInfo.PoStRandomness, blk.EPoStInfo.PoStProof, blk.EPoStInfo.Winners, blk.Miner)
+		valid, err := c.VerifyPoSt(c.postVerifier, allSectorInfos, uint64(sectorSize), blk.EPoStInfo.PoStRandomness,
```
sternhenri: blocking: `VerifyPoSt()` needs to take in the VRF digest, not the VRFProof, so you need to hash the VRFProof that you pass in here.

internal/pkg/consensus/testing.go
sternhenri: Yikes. Let me look into it some more on lotus end.

sternhenri:
I made a slight change to https://github.com/filecoin-project/specs/pull/883 to get closer to lotus (and save us a hash). Will commit a change here to match @anorth . Sorry!

internal/app/go-filecoin/internal/submodule/syncer_submodule.go
anorth: I dunno, why not? Let's do it in follow-up.

internal/pkg/chainsync/internal/syncer/syncer_test.go
anorth: These names were restored by an automated refactor. I've blanked them out again.

internal/pkg/consensus/expected.go
anorth: Yes, I won't specify a specific line length, and the long arg list usually signals a deeper problem, but wrap long lines somewhere. Chopping down to one arg per line is also fine.

internal/pkg/consensus/protocol.go
anorth: I had a quick go. Some re-wiring is needed to get the `BlockTime()` to the right place, so left for later.

internal/pkg/chain/traversal.go
anorth: I don't think I follow. AFAICT this code will do the right thing whether the last block in the iterator (the genesis, if so constructed) has height 1, 0, or -50. We don't need to harcode any constants here. But please describe a concrete case that I'm not thinking of.

internal/pkg/consensus/election.go
```diff
@@ -1,37 +1,38 @@
...
+	return &ElectionMachine{chain: chain}
+}
 
-	signature, err := signer.SignBytes(buf[:], candidateAddr)
-	if err != nil {
-		return nil, err
-	}
-	return signature.Data, nil
+// GenerateEPoStVrfProof computes the Election PoSt challenge seed for a target epoch after a base tipset.
+func (em ElectionMachine) GenerateEPoStVrfProof(ctx context.Context, base block.TipSetKey,
```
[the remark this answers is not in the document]
anorth: I found the spec naming of this following the same pattern as other randonmess to be terribly confusing. This is different to, say, the windowed post randomness seed, and it took me nearly a day to figure that out.

internal/pkg/consensus/expected.go
anorth: Fixed

internal/pkg/chain/traversal.go
sternhenri: The value i had in mind was a negative one. I don't understand how the iterator could go < 0, but I may be missing context. If you think it looks good, that works for me.

internal/pkg/consensus/election.go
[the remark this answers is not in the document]
sternhenri: Noted and understood. Thank you!

internal/pkg/consensus/expected.go
```diff
@@ -241,10 +212,9 @@ func (c *Expected) validateMining(
 			}
 		}
 
-		// Verify PoStRandomness
-		nullBlkCount := blk.Height - prevHeight - 1
-		if err := c.VerifyPoStRandomness(blk.EPoStInfo.PoStRandomness, electionTicket, workerAddr, uint64(nullBlkCount)); err != nil {
-			return errors.Wrap(err, "PoStRandomness invalid")
+		// Verify EPoSt VRF proof ("PoSt randomness")
+		if err := c.VerifyEPoStVrfProof(ctx, blk.Parents, blk.Height, blk.Miner, workerAddr, blk.EPoStInfo.PoStRandomness); err != nil {
```
ZenGround0: nit: it would be great to change the name of the postrandomness field to epostvrfproof (maybe keeping the json tag to match spec until changes there merge)

internal/pkg/mining/worker.go
```diff
@@ -194,46 +194,28 @@ func (w *DefaultWorker) Mine(ctx context.Context, base block.TipSet, nullBlkCoun
...
+
+	postVrfProof, err := w.election.GenerateEPoStVrfProof(ctx, base.Key(), targetEpoch, w.minerAddr, workerAddr, w.workerSigner)
 	if err != nil {
-		log.Errorf("Worker.Mine failed to generate post randomness %s", err)
+		log.Errorf("Worker.Mine failed to generate epost postVrfProof %s", err)
 		outCh <- Output{Err: err}
 		return
 	}
-	powerTable, err := w.getPowerTable(ctx, base.Key())
+	postVrfProofDigest := blake2b.Sum256(postVrfProof)
```
ZenGround0: nit: we should use either vrfpi digest or the crypto make randomness seed function

internal/pkg/mining/worker_test.go
```diff
@@ -53,33 +53,24 @@ func TestLookbackElection(t *testing.T) {
...
+	rnd := &consensus.FakeChainRandomness{Seed: 0}
 	minerAddr := addrs[3]      // addr4 in sharedSetup
 	minerOwnerAddr := addrs[4] // addr5 in sharedSetup
 
 	messages := chain.NewMessageStore(bs)
 
 	t.Run("Election sees ticket lookback ancestors back", func(t *testing.T) {
-		electionTicket, err := ancestors[lookback-1].MinTicket()
-		require.NoError(t, err)
-		mem := consensus.NewMockElectionMachine(func(ticket block.Ticket) {
```
ZenGround0: Was the fix skipped for expediency (which is ok with me)?  If not I'm suspicious it should be straightforward to do mock machines in the new model but maybe I'm missing something there.  Since it looks like we plan to come back we should skip this louder as its not testing anything right now.

internal/pkg/mining/worker_test.go
```diff
@@ -102,20 +92,15 @@ func TestLookbackElection(t *testing.T) {
 		go worker.Mine(ctx, head, 0, outCh)
 		r := <-outCh
 		assert.NoError(t, r.Err)
+		// TODO: make an assertion about the epost/ticket produced
 	})
 
 	t.Run("Ticket gensees ticket 1 ancestor back", func(t *testing.T) {
-		genTicket, err := ancestors[0].MinTicket()
```
ZenGround0: same comment as above make sure we skip louder.

internal/pkg/crypto/randomness.go
```diff
@@ -23,31 +23,21 @@ type ChainSampler interface {
...
-	err := binary.Write(&buf, binary.BigEndian, epoch)
-	if err != nil {
-		return nil, err
-	}
-
-	bufHash := blake2b.Sum256(buf.Bytes())
-	return bufHash[:], nil
+// Computes a random seed from raw ticket bytes.
+// A randomness seed is the VRF digest of the minimum ticket of the tipset at or before the requested epoch
+func MakeRandomSeed(rawVRFProof []byte) (RandomSeed, error) {
```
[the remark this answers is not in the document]
anorth: Yes, I made a start on that but pulled out as it spread. Can follow-up.

internal/pkg/consensus/expected.go
anorth: Will follow up

internal/pkg/mining/worker_test.go
anorth: The test is still skipped because it depends on code that was commented-out during the VM integration, but I've restored code that checks this (that may or may not work).

PR #3811, opened 2020-02-27
Feat/fake power

ZenGround0:
### Motivation
Setup genesis state with miners that have configurable sector commitments 1) for setting up preseal genesis block 2) for setting up genesis block for test 

Note: this doesn't build only because I am stuck in dependency hell trying to get an updated version of go-car that doesn't break fil-markets and storage miner.  I am pursuing a few options right now to get unblocked on this but its at the point where I want to parallelize review too.  If you want to check build you can replace the go-car module with [the one line go-car change here](https://github.com/ipfs/go-car/pull/23). 

### Proposed changes
1) Genesis sets up upgraded specs actors state
2) tweak genesis config to better suit use cases
3) Refactor gengen for maintainability and fix tests (pretty easy since so much is skipped)


internal/app/go-filecoin/node/node_test.go
```diff
@@ -233,3 +235,13 @@ func TestNodeConfig(t *testing.T) {
 		Address: "/ip4/127.0.0.1/tcp/0",
 	}, cfg.Swarm)
 }
+
+func TestGenerateTestComms(t *testing.T) {
```
ZenGround0: This is how I got cids for setup.json.  Suggestions for where to put this?  I'm thinking gengen binary maybe?

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -307,11 +308,16 @@ func (ctx *invocationContext) Send(toAddr address.Address, methodNum abi.MethodN
 		}
 	}()
 
+	// replace nil params with empty value
+	if params == nil {
```
ZenGround0: @icorderi Added this to handle a specs actor method that returned nil.  Maybe we should enforce this in actor code instead.

tools/gengen/util/gengen.go
```diff
@@ -114,203 +118,74 @@ type RenderedMinerInfo struct {
...
-		if err != nil {
-			return nil, err
-		}
-
-		var pid peer.ID
-		if m.PeerID != "" {
-			p, err := peer.Decode(m.PeerID)
-			if err != nil {
-				return nil, err
+	ignoreWalkFunc := func(nd format.Node) (out []*format.Link, err error) {
```
ZenGround0: This is why we need an update to go-car, the current dependency can't do this for us.

ZenGround0:
FYI @acruikshank if I can get go-car dep resolved today and get this past review we can use gfc gengen to drive the preseal workstream home if you haven't already gone another route.

tools/gengen/util/generator.go
```diff
@@ -0,0 +1,651 @@
...
+		a.Head = e.NewCid(headCid)
+	}
+	if err := g.stateTree.SetActor(context.Background(), addr, &a); err != nil {
+		return nil, fmt.Errorf("failed to create actor during genesis block creation")
+	}
+
+	return &a, nil
+}
+
+func (g *GenesisGenerator) setupDefaultActors(ctx context.Context) error {
```
ZenGround0: This is copypasted from consensus/genesis.go.  It 100% belongs here and consensus/genesis.go utils should either be deleted entirely or become a thin wrapper around gengen.  Right now I'm deferring that move to another PR but will do it here if reviewers prefer.

tools/gengen/util/generator.go
```diff
@@ -0,0 +1,651 @@
...
+		return nil, err
+	}
+
+	ret := out.(*market.PublishStorageDealsReturn)
+
+	return ret.IDs, nil
+}
+
+func (g *GenesisGenerator) setupPost(ctx context.Context, addr, mIDAddr address.Address) error {
+	mAct, found, err := g.stateTree.GetActor(ctx, mIDAddr)
```
ZenGround0: There is a bit of code duplication in getting and setting actors.  It didn't seem like enough to warrant creating a utility (especially because I don't know of any other places apart from genesis construction this is used) but I will do this if reviewers think duplication rises to that level.

ZenGround0:
Dependencies fixed

fixtures/setup.json
```diff
@@ -10,7 +10,17 @@
   ],
   "miners": [{
     "owner": 0,
-    "numCommittedSectors": 1
+    "committedSectors": [{
+      "commR": {"/":"bafy2bzaceb4n26dwrtbzx7l6ifj32xmxkl5lc6lfzq4qlcoft3t2jzzs7anfi"},
+      "commD": {"/":"bafy2bzacec3p63bwba2ph3stn27jbiyadowkpdmtg2m7aysfciuevzrjpyndo"},
+      "sectorID": 0,
```
anorth: Nit: actors call this SectorNum (the ID also includes the miner addr)

internal/app/go-filecoin/node/node_test.go
anorth: sounds good

internal/app/go-filecoin/node/testing.go
```diff
@@ -202,23 +200,27 @@ var PeerKeys = []crypto.PrivKey{
...
-	Keys:       2,
-	Miners: []*gengen.CreateStorageMinerConfig{
-		{
-			Owner:               0,
-			NumCommittedSectors: 100,
-			PeerID:              mustPeerID(PeerKeys[0]).Pretty(),
-			SectorSize:          constants.DevSectorSize,
+// MakeTestGenCfg returns a genesis configuration used for tests.
+func MakeTestGenCfg(t *testing.T) *gengen.GenesisCfg {
+	commCfgs, err := gengen.MakeNCommitCfgs(100)
```
anorth: Could we describe the intended configuration in english too? E.g. "One miner and two accounts (the first is the miner's owner/worker) with 100 sectors. the accounts both have 10000 FIL"

Take the 100 as a parameter if possible. Parameters for more things would help decouple the tests from each other, otherwise various tests will lock in parts of this setup that are then make other tests difficult. (low priority)

internal/app/go-filecoin/node/node_test.go
ZenGround0: moved to a tiny binary in gengen package

internal/pkg/vm/actor/builtin/default.go
```diff
@@ -21,5 +22,6 @@ var DefaultActors = dispatch.NewBuilder().
 	Add(specs.MultisigActorCodeID, &multisig.Actor{}).
 	Add(specs.PaymentChannelActorCodeID, &paych.Actor{}).
 	Add(specs.StoragePowerActorCodeID, &power.Actor{}).
+	Add(specs.StorageMarketActorCodeID, &market.Actor{}).
```
anorth: Eek, could you please add all the actors for specs-actors builtin/codes.go. There are 10 total. We can fix the debt that this is likely to rot later.

internal/pkg/vm/internal/vmcontext/invocation_context.go
anorth: This should definitely be enforced in actor code, the tests are just insufficient. I don't love this fallback, but ok if you must.

Please propose a fix for the actor in question

internal/pkg/vm/internal/vmcontext/syscalls.go
```diff
@@ -18,6 +18,13 @@ var _ specsruntime.Syscalls = (*syscalls)(nil)
 
 // VerifySignature implements Syscalls.
 func (sys syscalls) VerifySignature(signature specscrypto.Signature, signer address.Address, plaintext []byte) error {
+	// Dragons: this lets all id addresses off the hook -- we need to remove this
+	// once market actor code actually checks proposal signature.  Depending on how
+	// that works we may want to do id address to pubkey address lookup here or we
+	// might defer that to VM
```
anorth: Intention is to do the lookup here, I believe (where we can cache it). This would make the call non-pure function though. I'll confirm with lotus.

ID addresses are the *only* addresses I expect to actually see here.

tools/gengen/util/gengen.go
```diff
@@ -50,14 +32,36 @@ type CreateStorageMinerConfig struct {
...
 	// SectorSize is the size of the sectors that this miner commits, in bytes.
 	SectorSize abi.SectorSize
 }
 
+// CommitConfig carries all information needed to get a sector commitment in the
+// genesis state.
+type CommitConfig struct {
+	CommR    cid.Cid
+	CommD    cid.Cid
+	SectorID uint64
```
anorth: `abi.SectorNumber`?

tools/gengen/util/gengen.go
```diff
@@ -50,14 +32,36 @@ type CreateStorageMinerConfig struct {
...
+	DealCfg  *DealConfig
+}
+
+// DealConfig carries the information needed to specify a self-deal committing
+// power to the market while initializing genesis miners
+type DealConfig struct {
+	CommP     cid.Cid
+	PieceSize uint64
```
anorth: `abi.PaddedPieceSize`

tools/gengen/util/gengen.go
```diff
@@ -50,14 +32,36 @@ type CreateStorageMinerConfig struct {
...
+	// Client and Provider are miner worker and miner address
+
+	// StartEpoch is 0
+	EndEpoch uint64
```
anorth: `abi.chainEpoch`, or at least use `int64` to be consistent.

tools/gengen/util/gengen.go
anorth: That's fine AFAIC. @frrist solved a similar problem in chain export - check for a re-usable piece of code there.

tools/gengen/util/testing.go
```diff
@@ -0,0 +1,40 @@
+package gengen
+
+import (
+	"fmt"
+
+	"github.com/filecoin-project/go-filecoin/internal/pkg/constants"
+)
+
+// MakeNCommitCfgs creates n gengen commit configs, casting strings to cids.
+func MakeNCommitCfgs(n int) ([]*CommitConfig, error) {
```
anorth: `N` doesn't add much to the name, suggest just `MakeCommitCfgs()`.

internal/app/go-filecoin/node/testing.go
ZenGround0: Sounds great

internal/pkg/vm/actor/builtin/default.go
ZenGround0: great idea

internal/pkg/vm/internal/vmcontext/invocation_context.go
ZenGround0: This is the culprit: https://github.com/filecoin-project/specs-actors/pull/214.  I'll patch go-filecoin up once it lands.

PR #3812, opened 2020-02-27
Feat/presealed sectors

acruikshank:
### Motivation

We need to be able to import sealed sectors into filecoin.

### Proposed changes

Sectorbuilder already supports the ability to read sector information from an existing database. Wire this into go-filecoin init.

### Note

Reading a sector datastore generated by lotus requires that we upgrade badger, which, in this case, means using ipfs/go-ds-badger2 instead of ipfs/go-ds-badger.


cmd/go-filecoin/init.go
```diff
@@ -90,6 +98,47 @@ var initCmd = &cmds.Command{
...
+				WorkerThreads: 1,
+				Paths:         sectorbuilder.SimplePath(presealedSectorDir),
+			}, namespace.Wrap(oldMetaDs, datastore.NewKey("/sectorbuilder")))
+			if err != nil {
+				return xerrors.Errorf("failed to open up preseal sectorbuilder: %w", err)
+			}
+
+			path, err := rep.Path()
+			if err != nil {
+				return xerrors.Errorf("failed find filecoin path: %w", err)
```
ZenGround0: nit: failed to find

acruikshank:
> I would l ❤️ve to see a daemon test proving that this works

Good call. I added one that runs the preseal import off a fixture and checks that the files have been installed.

anorth:
Or a non-deamon in-process test? Just dreamin'....

PR #3813, opened 2020-02-28
feat: add new gas costs

icorderi:
### Motivation

Gas accounting changed on the new spec

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/actor_store.go
+++ b/internal/pkg/vm/internal/vmcontext/actor_store.go
@@ -1,0 +1,45 @@
+package vmcontext
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/gascost"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/internal/storage"
+	specsruntime "github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/ipfs/go-cid"
+)
+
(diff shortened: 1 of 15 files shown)


internal/pkg/vm/internal/vmcontext/actor_store.go
```diff
@@ -0,0 +1,45 @@
...
+	specsruntime "github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/ipfs/go-cid"
+)
+
+// actorStorage hides the storage methods from the actors and turns the errors into runtime panics.
+type actorStorage struct {
+	context context.Context
+	inner   *storage.VMStorage
+	gasTank *GasTracker
+}
```
frrist: something like `meteredStorage` would be more descriptive.

internal/pkg/vm/internal/vmcontext/actor_store.go
```diff
@@ -0,0 +1,45 @@
...
+		panic(fmt.Errorf("could not put object in store. %s", err))
+	}
+	s.gasTank.Charge(gascost.OnIpldPut(ln))
+	return cid
+}
+
+func (s actorStorage) Get(cid cid.Cid, obj specsruntime.CBORUnmarshaler) bool {
+	ln, err := s.inner.Get(s.context, cid, obj)
+	if err == storage.ErrNotFound {
+		return false
```
anorth: Should this really charge zero? It has a real execution cost.

PR #3815, opened 2020-03-02
update chain-validation and add test skipper

frrist:
### Motivation
Automatically run latest chain-validation test when updating the library without needing to manually edit test files.

### Proposed changes
Update to latest chain-validation library and add test skipper for tests that do not pass (yet).


frrist:
This code has already landed in master.

PR #3816, opened 2020-03-02
Init miner address and various testnet 3 compatibility fixes.

acruikshank:
### Motivation

In order to properly initialize a filecoin bootstrap node, we need to be able to set the miner address to that of the miner configured in the genesis block, and import it's worker keys.

### Proposed changes

This PR adds the changes to init to create a fully initialized bootstrap node. 

This PR is the first time we've been able to exercise significant sections of filecoin code (including mining start) in a while. While testing the new functionally, we found many small issues that have been addressed in this PR


cmd/go-filecoin/init.go
```diff
@@ -43,9 +44,11 @@ var initCmd = &cmds.Command{
 	Options: []cmdkit.Option{
 		cmdkit.StringOption(GenesisFile, "path of file or HTTP(S) URL containing archive of genesis block DAG data"),
 		cmdkit.StringOption(PeerKeyFile, "path of file containing key to use for new node's libp2p identity"),
+		cmdkit.StringOption(WalletKeyFile, "path of file containing keys to import into the wallet on initialization"),
```
acruikshank: These two options are required to set up a bootstrap miner with no extra configuration between `init` and `daemon`

cmd/go-filecoin/init_daemon_test.go
```diff
@@ -73,12 +73,12 @@ func TestImportPresealedSectors(t *testing.T) {
 		"--symlink-imported-sectors",
 	)).Start()
 
-	staging, err := ioutil.ReadDir(path.Join(td.RepoDir(), "staging"))
+	staging, err := ioutil.ReadDir(path.Join(td.SectorDir(), "staging"))
```
acruikshank: Prior to this PR, the sector import had gone to the wrong directory.

internal/pkg/block/block_test.go
```diff
@@ -54,7 +54,10 @@ func TestTriangleEncoding(t *testing.T) {
 		types.AssertHaveSameCid(t, exp, &cborJSONRoundTrip)
 	}
 	t.Run("encoding block with zero fields works", func(t *testing.T) {
-		testRoundTrip(t, &blk.Block{})
+		testRoundTrip(t, &blk.Block{
+			BlockSig:        crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
+			BLSAggregateSig: crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte{}},
```
acruikshank: New specs-actors signature encoding requires that signatures be non-nil.

internal/pkg/version/protocol_versions.go
```diff
@@ -13,6 +13,9 @@ const DEVNET4 = "devnet4"
 // LOCALNET is the network name of localnet
 const LOCALNET = "localnet"
 
+// TESTNET3 is the network name of the third testnet
+const TESTNET3 = "localnet-270a8688-1b23-4508-b675-444cb1e6f05d"
```
acruikshank: Required to interop with testnet genesis file. This will be fixed in a subsequent PR.

internal/pkg/block/block_test.go
ZenGround0: This requirement will be removed if we merge: https://github.com/filecoin-project/specs-actors/pull/223 which will not error when unmarshalling zero valued signatures.

anorth: FWIW I don't think it's a bad requirement, and would prefer we move toward maintaining it.

internal/pkg/consensus/election.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/binary"
 	"math/big"
 
+	"github.com/filecoin-project/go-filecoin/internal/pkg/util/hasher"
```
anorth: Import grouping (some other files too)

internal/pkg/state/view.go
```diff
@@ -78,6 +80,20 @@ func (v *View) InitResolveAddress(ctx context.Context, a addr.Address) (addr.Add
 	return state.ResolveAddress(StoreFromCbor(ctx, v.ipldStore), a)
 }
 
+// Returns public key address if id address is given
+func (v *View) AccountSignerAddress(ctx context.Context, a addr.Address) (addr.Address, error) {
+	if a.Protocol() != addr.ID {
```
anorth: Check for equality with SECP or BLS to avoid pass through of an actor temporary address.

internal/pkg/state/view.go
```diff
@@ -237,14 +253,26 @@ func (v *View) loadPowerActor(ctx context.Context) (*power.State, error) {
...
+	var state account.State
+	err = v.ipldStore.Get(ctx, actr.Head.Cid, &state)
+	return &state, err
+}
+
 func (v *View) loadActor(ctx context.Context, address addr.Address) (*actor.Actor, error) {
-	tree := v.asMap(ctx, v.root)
-	var actr actor.Actor
-	found, err := tree.Get(adt.AddrKey(address), &actr)
+	st, err := treestate.LoadState(ctx, v.ipldStore, v.root)
```
anorth: Sigh, if you must. Was this necessary?

ZenGround0:
@anorth yeah we can remove that change if you think it's the right thing.  I don't agree for reasons in the comments of that PR but don't feel too strongly.

internal/pkg/state/view.go
ZenGround0: This was something I chanced upon while trying to debug an earlier interop spike (that I then shared with @acruikshank ).  I was using this pattern to inspect actor state which was not loading into the adt and then found that it worked to replace the functionality of the method.  I did not investigate why this was so it may not be necessary, it was only expedient.  However there was a problem with treating the lotus block state as an adt map.

acruikshank: Yeah, I'm reverting the change. Seems to be fine.

PR #3817, opened 2020-03-03
demonstrate serialization failure

frrist:
### Motivation
Demonstrate that something in serialization is borked


ZenGround0:
Hey @frrist I think this was caused by the zero valued signature issue that @acruikshank removed from tests and that I attempted to remove entirely by PRing into specs actors (might be removed).  I merged latest master into this branch and `TestAcceptHeavierFork` (and all other tests) passed with no panic.

We'll reopen this PR/line of inquiry if we see this behavior again.  Thanks for checking in about it.

PR #3818, opened 2020-03-03
feat: add syscalls gas cost 

icorderi:


internal/pkg/vm/internal/gascost/pricelistV0.go
```diff
@@ -114,3 +128,36 @@ func (pl *pricelistV0) OnCreateActor() gas.Unit {
...
+}
+
+// OnHashing
+func (pl *pricelistV0) OnHashing(dataSize int) gas.Unit {
+	return pl.hashingBase + gas.Unit(dataSize)*pl.hashingPerByte
+}
+
+// OnComputeUnsealedSectorCid
+func (pl *pricelistV0) OnComputeUnsealedSectorCid(proofType abi.RegisteredProof, pieces *[]abi.PieceInfo) gas.Unit {
+	// Dragons: this needs more cost tunning, check with @lotus
```
anorth: I don't think we need to invoke dragons for this. It's just something we know we need to do, unlikely to have any effects elsewhere.

PR #3819, opened 2020-03-03
Move VRFPi type from block to crypto package

anorth:
### Motivation
Moving the type here allows using it for better safety in the crypto randomness methods.

Note: we can probably move types.signer here too later.


anorth:
This is much less clear a win after the integration of `abi.PostRandomness`. Perhaps we can rename that type in specs-actors, and even include the Digest method? Or we can locally replace it and adapt only at the FFI/actor boundaries?

ZenGround0:
@anorth both strategies SGTM with slight preference for locally replacing.

PR #3830, opened 2020-03-05
Fake proofs

ZenGround0:
### Motivation
We want mining - syncing end to end integration testing without the trouble of storing real sector data.

More details:
Allowing node construction with fake election post generation and verification simplifies setup needed for creating integration tests.  Instead of requiring on disk sectors to be setup correctly and linked to genesis state integration tests now only need to use fake sector commitment values during genesis construction.  Partial ticket values are faked at the data level but using gengen the state machine is setup with a real power table (with power claims not tied to storage) so we can now exercise the true behavior of leader election.

### Proposed changes
- Give node and builder ability to set up fake proof gen and verification
- Reinstate integration tests for empty (no messages) block mining and syncing
- Use tests to drive last mile fixes to empty block mine and sync
  * Randomness sampling off by one
  * gengen state construction bugs
  * handling specs-actors edge case values in runtime

Closes issue #


internal/pkg/consensus/expected.go
```diff
@@ -216,7 +216,7 @@ func (c *Expected) validateMining(ctx context.Context,
 		}
 
 		// Verify EPoSt VRF proof ("PoSt randomness")
-		if err := c.VerifyEPoStVrfProof(ctx, blk.Parents, blk.Height, blk.Miner, workerSignerAddr, blk.EPoStInfo.VRFProof); err != nil {
+		if err := c.VerifyEPoStVrfProof(ctx, blk.Parents, blk.Height-1, blk.Miner, workerSignerAddr, blk.EPoStInfo.VRFProof); err != nil {
```
acruikshank: This -1 could use some explanation in a comment or assign it to a variable that names it  (`parentHeight`?).

go.sum
```diff
@@ -172,6 +172,7 @@ github.com/filecoin-project/specs-actors v0.0.0-20200226225703-1fa643f1847f/go.m
 github.com/filecoin-project/specs-actors v0.0.0-20200226225703-1fa643f1847f/go.mod h1:0HAWYrvajFHDgRaKbF0rl+IybVLZL5z4gQ8koCMPhoU=
 github.com/filecoin-project/specs-actors v0.0.0-20200226233436-635911a01775 h1:vA732y5yFNiOx1DBWsG5QSHXD46rCwZkhrjtIYaalrk=
 github.com/filecoin-project/specs-actors v0.0.0-20200226233436-635911a01775/go.mod h1:0HAWYrvajFHDgRaKbF0rl+IybVLZL5z4gQ8koCMPhoU=
+github.com/filecoin-project/specs-actors v0.0.0-20200304172711-1cfc2ba9f4a6 h1:CYGe21swMwnnIeK9jmGj495qdA9e+Xmtkg0owouw83Y=
```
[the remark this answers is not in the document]
ZenGround0: Nope and `go mod tidy` removes.  We should consider adding that to our lint step maybe. Thanks

internal/pkg/consensus/expected.go
ZenGround0: yeah I should really use the election post lookback constant here

internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -317,10 +317,15 @@ func (ctx *invocationContext) Send(toAddr address.Address, methodNum abi.MethodN
 		}
 	}()
 
+	// DRAGONS: remove these once specs actors project enforces Send params
```
anorth: FYI This will go the other way, with nil being the correct value for "no params"

Issue #3124, opened 2019-07-24
Enable message signing for pubsub


frrist:
### Description
Message signing is currently disabled in libp2p pubsub, this is something we may want moving forward. v0.1.0 of pubsub now requires that it be enabled in order for strict signature verification to work.

### Where to begin
https://github.com/filecoin-project/go-filecoin/blob/33ffd9ab8a9f13310990121b1c30ea165ffa0739/node/node.go#L416-L418

anorth:
@ZenGround0 is this done?

ZenGround0:
No and this might become an interop hurdle.  Thanks for flagging.

Issue #3701, opened 2019-12-06
Genesis construction technical debt


ZenGround0:
### Description
We have two nearly equivalent ways of constructing genesis state / block.  We could reduce the surface area of code we need to fiddle with by removing most of the consensus / genesis logic to operate over a gengen config transformed into genesis state / block by the gengen library.

Genesis state setup isn't even close to being stable (i.e. the spec's mandate to have a *genesis chain* rather than a block which is going to require lots of code).  It is not too late to save ourselves time.

### Where to begin
functions operating over consensus Config should instead operate over genesis config

slightly different interface between callers of consensus genesis are probably what has kept us in this place for so long.  We need to address that head on and decide how to resolve differences, either by change consensus/genesis callers' guarantees (i.e. test keys instead of test addresses) or change gengen config format (i.e. allow direct address setting instead of enforcing all addresses belong with keys)

anorth:
@ZenGround0 did your recent work resolve this?

ZenGround0:
No, we gave gengen ownership of setupDefaultActors through copy-paste but the consensus function still exists.  The integration test work I'm on now might include this to get our tests back.

PR #3831, opened 2020-03-05
Sign Pubsub Messages & Simplify gengen for pubsub test

ZenGround0:
### Motivation
We need to sign pubsub messages for interop.  And want some test coverage that pubsub block and message channels still propagate.

Secondarily we have been living with duplicated genesis state generation for a long time and have needed to get rid of it.

### Proposed changes
The first commit is very easy and simply passes the correct flag to libp2p.

The second commit unfolded while trying to unskip the pubsub msg topic test in the node package.  It became apparent it was relying on the duplicated consensus genesis function machinery.  It probably would have been ok to just patch that up but I have so much gengen and node test context lately from recent work that this seemed only a little bit harder and way way nicer.

Closes issue #3124 #3701


internal/pkg/consensus/genesis.go
```diff
@@ -43,334 +19,8 @@ type GenesisInitFunc func(cst cbor.IpldStore, bs blockstore.Blockstore) (*block.
 // and *also* the ticket value used when computing the genesis state (the parent state of the genesis block).
 var GenesisTicket = block.Ticket{VRFProof: []byte{0xec}}
 
-var (
-	defaultAccounts map[address.Address]abi.TokenAmount
```
anorth: ❤️

PR #3833, opened 2020-03-06
Upgrade actors and chain validation

anorth:
### Motivation
Changes in the new actors:
- fixed signature for the consensus fault syscall
- actors return nil instead of empty value
- reward actor requires winning ticket count to compute reward

The latest chain validation tests are all skipped because of gas calculation incompatibilities (which affect balances). We can restore them if the configuration to skip gas calculations can also skip sender balance checks. Tipset tests are skipped due to different expectations of the new actor address method.

### Proposed changes
The only significant change is handling nil return values (and parameters) in the VM.


internal/pkg/vm/internal/dispatch/dispatch.go
```diff
@@ -50,28 +50,26 @@ func (d *actorDispatcher) Dispatch(methodNum abi.MethodNum, ctx interface{}, arg
 
 	// build args to pass to the method
 	args := []reflect.Value{
-		// the ctx will be automaticall coerced
+		// the ctx will be automatically coerced
 		reflect.ValueOf(ctx),
 	}
 
 	// Dragons: simplify this to arginterface
-	if raw, ok := arg1.([]byte); ok {
```
frrist: switch could be cleaner here imo

internal/pkg/vm/internal/message/result.go
```diff
@@ -41,9 +30,13 @@ func Ok() Receipt {
...
 func Value(obj interface{}) Receipt {
-	aux, err := encoding.Encode(obj)
-	if err != nil {
-		return Receipt{ExitCode: exitcode.SysErrSerialization}
+	var aux []byte
+	if obj != nil {
+		var err error
+		aux, err = encoding.Encode(obj)
+		if err != nil {
+			return Receipt{ExitCode: exitcode.SysErrSerialization}
```
frrist: is the caller responsible for adding the other fields -- gas and return value?

internal/pkg/vm/internal/vmcontext/validation_test.go
```diff
@@ -1,36 +1,77 @@
...
-	vsuites.TestMessageApplicationEdgecases(t, f)
+	for _, testCase := range suites.MessageTestCases() {
+		// All skipped due to gas accounting causing divergent balance calculations.
+		if true || TestSuiteSkipper.Skip(testCase) {
+			continue
+		}
+		t.Run(caseName(testCase), func(t *testing.T) {
+			testCase(t, f)
+		})
+	}
```
frrist: :star:

internal/pkg/vm/internal/message/result.go
anorth: Cleaned up

PR #3836, opened 2020-03-06
Resolve account signer addresses when validating message signatures

### LINKED ISSUE #3838: Message signature validation resolves ID addresses
A message `From` address may be an ID address, which doesn't directly contain the required public key. Resolve the actor's address to a key in the state before validation.

anorth:
### Motivation
The message `From` address may be an ID address of an account actor, with the signing address held in the account actor's state. We need to resolve ID addresses to signers before validation.

### Proposed changes
Removes the `VerifySignature` method from `SignedMessage` and adds a `SignatureValidator` type that uses a state view to resolve account addresses. Also move the BLS aggregate validation code inside the same type. The block signature, ticket, and epost signature are all already explicitly using a resolved signer address, but future work could move those methods to the same place.

Closes #3838

We still need to fix the outbox to be able to sign messages that carry ID addresses.


internal/pkg/consensus/expected.go
```diff
@@ -173,7 +173,9 @@ func (c *Expected) validateMining(ctx context.Context,
 	parentWeight fbig.Int,
 	parentReceiptRoot cid.Cid) error {
 
-	powerTable := NewPowerTableView(c.state.StateView(parentStateRoot))
+	stateView := c.state.StateView(parentStateRoot)
+	sigValidator := NewSignatureValidator(stateView)
```
ZenGround0: nit: Ideally we would would use this abstraction everywhere we need this translation.  I believe it's needed in the mining worker and the storage and market connectors.

internal/pkg/consensus/signing.go
```diff
@@ -0,0 +1,59 @@
+package consensus
```
ZenGround0: nit: a slight name change would be nice as this doesn't do any signing, only signature validation

internal/pkg/consensus/signing_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestSignMessageOk(t *testing.T) {
+	tf.UnitTest(t)
+	ctx := context.Background()
+
+	idAddress := vmaddr.RequireIDAddress(t, 1)
+	ms, kis := types.NewMockSignersAndKeyInfo(1)
+	keyAddr, err := kis[0].Address()
+	require.NoError(t, err)
+
+	{
```
ZenGround0: Maybe use t.Run to separate these into separate failures?

internal/pkg/consensus/signing_test.go
```diff
@@ -0,0 +1,135 @@
...
+			Message:   *msg,
+			Signature: sig,
+		}
+
+		assert.NoError(t, v.ValidateMessageSignature(ctx, smsg))
+	}
+}
+
+// Signature is valid but signer does not match From Address.
+func TestBadFrom(t *testing.T) {
```
ZenGround0: It would be nice to see a test case exercising the id address translation similar to the one above to verify.

internal/pkg/types/signed_message.go
```diff
@@ -30,6 +29,8 @@ type SignedMessage struct {
 
 // NewSignedMessage accepts a message `msg` and a signer `s`. NewSignedMessage returns a `SignedMessage` containing
 // a signature derived from the serialized `msg` and `msg.From`
+// NOTE: this method can only sign message with From being a public-key type address, not an ID address.
+// We should deprecate this and move to more explicit signing via an address resolver.
```
ZenGround0: 👍

internal/pkg/consensus/expected.go
anorth: Mining worker signs things, but doesn't validate signatures as far as I can tell (it assumes the pool provides only valid messages).

anorth: Done for the module connector common (but with unfortunate type signature mismatch)

internal/pkg/consensus/signing.go
anorth: Done, also moved out of consensus package to state package (alongside view).

PR #3845, opened 2020-03-07
Implement consensus fault checker and hook it up to the syscall.

anorth:
### Motivation
We need to implement the consensus fault report syscall.

### Proposed changes
Implemented a checker in a distinct object (for easy testing) and connected through to VM. The VM is now required to provide the chain head tipset key to seed the chain traversal that ensures that at least one of the reported headers is in the current chain.

I identified numerous problems with the consensus fault spec while building this, and there's one more that I've commented with `Dragons`: I've implemented checking of the signatures in the state referenced by the offending block headers, but there's no guarantee that our node has that parent state for both headers. I'm going to try to clarify what to do here with specs.

I haven't written any tests yet, but it works AFAIK so please review so I can get any changes in early. I may merge before testing just to avoid rebase pain while resolving the spec.


internal/pkg/slashing/check.go
```diff
@@ -0,0 +1,189 @@
...
+		fault = &runtime.ConsensusFault{
+			Target: b1.Miner,
+			Epoch:  b2.Height,
+			Type:   runtime.ConsensusFaultDoubleForkMining,
+		}
+	}
+	// Time-offset mining fault: two blocks with the same parent but different epochs.
+	// The height check is redundant at time of writing, but included for robustness to future changes to this method.
+	// The blocks have a common ancestor by definition (the parent).
+	if b1.Parents.Equals(b2.Parents) && b1.Height != b2.Height {
```
ZenGround0: @anorth I think the following is a spec regression, not a code problem but FYI:

These two checks don't seem to cover all provable forking which is a regression from previous versions of these validation rules.  Let null blocks be dashes, letter X be a tipset mined by X and A be the common ancestor of the fork in the diagrams below.  I believe these two rules cover:

Rule 1
A B
A B
and
A C D B
A T R B
etc

Rule 2
A - - - B
A - - B
etc

Neither cover
A C - B
A - - - B
because the two blocks mined by B above don't have the same parent and don't have the same epoch, however they were provable mined concurrently and we should be able to slash them.

In the past we've had forms of the slashing rule that do calculations on the parent height and block height to show an intersection in mining epoch range.  If you agree with this problem I can follow up in specs or will leave this to you as you prefer.

ZenGround0: To handle this properly I think the interface to the syscall would need to change, for example by including both parent headers in the extra bytes.

anorth: Please raise an issue or discussion with the spec.

ZenGround0: Done: https://github.com/filecoin-project/specs/issues/891

anorth:
I've adjusted this a little before landing to evaluate the miner signatures in the state of the fault report, rather than the parent state of the block header (which may not be available).

Issue #2224, opened 2019-03-09
EC Upgrade 3: weight change


ZenGround0:
### Description
The weighting function has changed.  See [the spec](https://github.com/filecoin-project/specs/blob/master/expected-consensus.md#chain-weighting)

### Acceptance criteria
Bring the implementation in line with the spec.

anorth:
Closed by #3851

PR #3851, opened 2020-03-09
Update weight function

ZenGround0:
### Motivation
Weight function needs to be up to spec for protocol correctness

### Proposed changes
Update weight function to match spec and lotus

Closes issue #2224


internal/pkg/consensus/chain_selector.go
```diff
@@ -76,12 +67,14 @@ func (c *ChainSelector) Weight(ctx context.Context, ts block.TipSet, pStateID ci
 	if err != nil {
 		return fbig.Zero(), err
 	}
-	roughLogTotalBytes := new(big.Float).SetInt64(int64(totalBytes.BitLen()))
-	innerTerm.Add(innerTerm, roughLogTotalBytes)
 
-	w.Add(w, innerTerm)
+	wPowerFactor := fbig.Mul(wPrecision, log2b(totalBytes))
+	wBlocksFactorNum := fbig.Mul(wRatioNum, fbig.Mul(log2b(totalBytes), fbig.NewInt(int64(ts.Len()))))
+	wBlocksFactorDen := fbig.Mul(fbig.NewInt(int64(expectedLeadersPerEpoch)), wRatioDen)
```
anorth: Put numerator/denominator consistently first or last in the product.

anorth: Factor out `log2b(totalBytes)` so it's computed only once.

PR #3852, opened 2020-03-09
gengen with presealed data

acruikshank:
### Motivation

A necessary step toward a functional network is ensuring that we can create a bootstrap node using presealed sectors so that it can mine messages until the other nodes gain power. The goal of this PR is to allow us to generate a genesis block using go-filecoin's gengen tool that is compatible with imported presealed data created with the Lotus seed tool, and to use that genesis block to start a node and mine blocks (using `mine once`).

### Proposed changes

1. Make some small changes to gengen to support presealed sectors.
2. Import presealed sectors into fixtures with a compatible setup.json file.
3. Modify functional test framework to support the configuration necessary for a bootstrap node.
3. Add a functional test for mine once using presealed sectors that ensures the PoSt generation and verification work against those sectors.


fixtures/setup.json
```diff
@@ -1,23 +1,35 @@
...
-    "owner": 0,
+    "owner": 5,
+    "sectorSize": 2048,
     "committedSectors": [{
-      "commR": {"/":"bafy2bzaceb4n26dwrtbzx7l6ifj32xmxkl5lc6lfzq4qlcoft3t2jzzs7anfi"},
-      "commD": {"/":"bafy2bzacec3p63bwba2ph3stn27jbiyadowkpdmtg2m7aysfciuevzrjpyndo"},
-      "sectorNum": 0,
+      "commR": {"/":"bafk4ehzaqugph7z7yajugw6wz6nn7x3akr3e32wecwexllfgqndnjt3nsaya"},
+      "commD": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
+      "sectorNum": 3,
```
anorth: This is a bit weird, I'd expect numbers starting from zero. SectorNum is per-miner. 

Since we can't put documentation in the JSON files, could you please add a README for the fixtures that explains how this set-up all works? And how to re-generate it.

fixtures/setup.json
```diff
@@ -1,23 +1,35 @@
 {
   "seed": 4,
   "keysToGen": 5,
+  "importKeys": [{"privateKey":"/8UoXvzGIU2H/TgTAmswx96m2dVCXkRqnIIlDhJrGCo=","sigType":2}],
```
anorth: Move this line after preAllocGenKeys since the doc elsewhere says these keys come after the gen keys. Or change so that these keys are ordered first.

fixtures/setup.json
```diff
@@ -1,23 +1,35 @@
...
   "preAllocGenKeys": [
     "10000000000",
     "10000000000",
     "10000000000",
     "10000000000",
+    "10000000000",
```
anorth: `keysToGen` is still 5, so what does this do? Or is this allocating to the non-gen keys too (in which case please rename `preAllocGenKeys`)?

fixtures/setup.json
acruikshank: Agreed. The presealed data is starts at 3 for this miner, so this is required. I'll add the README.

fixtures/setup.json
acruikshank: It is allocating to all keys. I'll rename.

acruikshank: renamed to `PreallocatedFunds`.

fixtures/setup.json
acruikshank: I've renamed the other field to `preallocatedFunds` to clarify it doesn't actually generate the keys and that it applies to both generated and imported keys. Given that, I think ordering makes the most sense the way it is.

fixtures/genesis-sectors/README.md
```diff
@@ -0,0 +1,26 @@
+## Generating genesis-sectors
+
+Genesis sectors are generated using the [Lotus](https://docs.lotu.sh/en+setup-local-dev-net) seed tool.
+Follow the directions there to install lotus and then run:
+```
+./lotus-seed pre-seal --sector-size 2048 --num-sectors 2 --miner-addr=t0106
+```
+The sector-size must be one the sizes in use by the proof system. And the miner address should be the address
```
ZenGround0: nit one of the sizes

fixtures/genesis-sectors/README.md
```diff
@@ -0,0 +1,26 @@
...
+of the bootstrap miner to be created with gengen. This will create a `.genesis-sectors` directory in your home directory.
```
ZenGround0: Might be nice to leave a hint that the inlined `t0106` is right: "by default in gengen this address starts at `t0106` and the id address number is incremented for every additional miner"

fixtures/setup.json
ZenGround0: Does the sector num get mixed in during sealing?  Does anything break if we just change this number here?  Keep in mind genesis sectors don't require verification of a PoRep they just get set in the miner state.

functional-tests/mine_once_test.go
```diff
@@ -0,0 +1,61 @@
...
+	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
+)
+
+func TestBootstrapMineOnce(t *testing.T) {
+	tf.FunctionalTest(t)
+
+	ctx := context.Background()
+	root := project.Root()
+
+	tns, err := iptbtester.NewTestNodes(t, 1, nil)
```
ZenGround0: What's the reason to use this instead of fast?  Do other functional tests use this?  Is fast setup no good for our purposes here?  Fast is too advanced a tool for what we want?  Only asking because I've never seen this integration testing framework used directly like this and we already have a proliferation of integration testing methods (in process tools, test daemon, fast)

functional-tests/mine_once_test.go
```diff
@@ -0,0 +1,61 @@
...
+	node0.MustRunCmdJSON(ctx, &blocks, "go-filecoin", "chain", "ls")
+	require.Equal(t, 1, len(blocks))
+	assert.Equal(t, address.Undef, blocks[0].Miner)
+
+	// Mine once
+	node0.MustRunCmd(ctx, "go-filecoin", "mining", "once")
+
+	// Assert that the chain head has now been mined by the miner
+	node0.MustRunCmdJSON(ctx, &blocks, "go-filecoin", "chain", "ls")
+	require.Equal(t, 1, len(blocks))
```
ZenGround0: Blocking: shouldn't this be 2?

internal/pkg/testhelpers/iptbtester/genesis.go
```diff
@@ -5,30 +5,49 @@ import (
...
-	"github.com/stretchr/testify/require"
-
 	commands "github.com/filecoin-project/go-filecoin/cmd/go-filecoin"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/constants"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 	gengen "github.com/filecoin-project/go-filecoin/tools/gengen/util"
+	"github.com/stretchr/testify/require"
 )
 
 // GenesisInfo chains require information to start a single node with funds
```
ZenGround0: I'm guessing you checked if the testdaemon and fast don't give you this explicitly (esp since I think fast uses this) but wanted to double check because the necessity of doing work on this integration test framework directly means there are 4 places to worry about which I would love to avoid in the future.

fixtures/setup.json
acruikshank: Things do break if we change this number. First of all the sector number is in the file name of all the pre-sealed sectors on disk. We could rename those, but I think I tried that and something still broke.

functional-tests/mine_once_test.go
acruikshank: TBH, I started with the first functional test in the functional tests directory and it didn't happen to be a fast test. I'll look into what it would take to convert it.

functional-tests/mine_once_test.go
acruikshank: This isn't 2 because chain ls returns newline delmited JSON and MustRunCmdJSON only accepts the first line. This is actually the number of blocks in the tipset (which should be 1) and the require just ensures the following assert doesn't ugly panic.

functional-tests/mine_once_test.go
acruikshank: I looked into it. It would require quite a bit of fast modifications to get its initialization to support presealed sectors. I'm going to merge this and revisit for the next functional test.

PR #3854, opened 2020-03-10
Updates as preface to payment channel manager implementation

shannonwells:
### Motivation
This is a set of changes to set up for the payment channel manager PR, to try to make each PR easier to review.

### Proposed changes
* A change to the ChannelInfo struct and an Equal function
* Add a PaychActorParties function to the state viewer
* types_cbor_gen.go
* Add ChannelExists to the PaychMrgAPI
* update tests and test apis as needed
* additional tests in the follow-on PR exercise this code more.


internal/app/go-filecoin/paymentchannel/types.go
```diff
@@ -1,18 +1,43 @@
...
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/specs-actors/actors/builtin/paych"
 )
 
+//go:generate cbor-gen-for ChannelInfo VoucherInfo
+
+// The key for the store is the payment channel's "robust" or "unique" address
+
 // ChannelInfo is the primary payment channel record
+// IDAddr: internal ID address for payment channel actor
```
acruikshank: This should be unnecessary since you can use `state.View.InitResolveAddress` to retrieve IDAddr from UniqueAddr.

internal/app/go-filecoin/paymentchannel/types.go
```diff
@@ -1,18 +1,43 @@
...
+	LastLane                     uint64
+	Vouchers                     []*VoucherInfo // All vouchers submitted for this channel
+}
+
+// ChannelInfoUndef is an empty ChannelInfo
+var ChannelInfoUndef = &ChannelInfo{}
+
+// Equal is a comparator for two ChannelInfos
+func (ci *ChannelInfo) Equal(ci2 *ChannelInfo) bool {
+	if ci.UniqueAddr == ci2.UniqueAddr &&
```
acruikshank: simplify ```
if x { 
  return true 
} 
return false``` to  ```return x```

internal/app/go-filecoin/paymentchannel/types.go
```diff
@@ -1,18 +1,43 @@
...
+		ci.To == ci2.To &&
+		ci.From == ci2.From &&
+		ci.LastLane == ci2.LastLane &&
+		reflect.DeepEqual(ci.Vouchers, ci2.Vouchers) {
```
acruikshank: This seems expensive. Is it really necessary? Is it going to be used in high throughput code?

internal/app/go-filecoin/paymentchannel/types.go
shannonwells: hah, yeah...

internal/app/go-filecoin/paymentchannel/types.go
shannonwells: Good point; it's only used to check against a blank channel info, so I'll change this func to reflect this and make it faster.

PR #3856, opened 2020-03-10
Test/msg integration

ZenGround0:
### Motivation
Get some test coverage on the user message processing state transition code.

### Proposed changes
Add a test that sends a value from one account to another.  Includes improvement to mining once behavior that propagates syncing errors instead of dropping.


internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -396,16 +395,24 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd c
 
 	// 5. Increment sender CallSeqNum
 	fromActor.IncrementSeqNum()
+	// update actor
```
ZenGround0: @icorderi out of band you mentioned there was a way to do the same fix by only moving two lines.  Please point it out in the comments here if that's still the case.

ZenGround0: Disregard I see it here https://github.com/filecoin-project/go-filecoin/pull/3857/files

icorderi: I have this fixes and a few others waiting to merge in https://github.com/filecoin-project/go-filecoin/pull/3857

I would rebase this PR as soon as that one merges.

internal/app/go-filecoin/node/block_propagate_test.go
```diff
@@ -108,6 +115,102 @@ func TestChainSync(t *testing.T) {
...
+	// genesis has two accounts
+	genCfg := &gengen.GenesisCfg{}
+	require.NoError(t, gengen.MinerConfigs(MakeTestGenCfg(t, 1).Miners)(genCfg))
+	require.NoError(t, gengen.GenKeys(3)(genCfg))
+	require.NoError(t, gengen.GenKeyPrealloc(1, "100000")(genCfg))
+	require.NoError(t, gengen.GenKeyPrealloc(2, "100")(genCfg))
+	require.NoError(t, gengen.NetworkName(version.TEST)(genCfg))
+	cs := MakeChainSeed(t, genCfg)
+	fakeClock := th.NewFakeClock(time.Unix(1234567890, 0))
+	blockTime := 100 * time.Millisecond
```
anorth: Given the use of a fake clock (:100), set a large block time like 30s so there's no chance that wall-clock time is actually involved.

PR #3857, opened 2020-03-10
fix: misc vm bugs during validation testing

icorderi:


internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -397,16 +397,16 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd c
...
-
 	// update actor
 	if err := vm.state.SetActor(vm.context, msg.From, fromActor); err != nil {
 		panic(err)
 	}
 
+	// 6. Deduct gas limit funds from sender first
+	// Note: this should always succeed, due to the sender balance check above
+	// Note: after this point, we nede to return this funds back before exiting
+	vm.transfer(msg.From, builtin.BurntFundsActorAddr, gasLimitCost)
```
ZenGround0: Doesn't the `newInvocationConext` call need an accurate view of the fromActor's balance?  If we just do this without reloading then vm.transfer's changes won't make it into the fromActor object passed in on line 434.

icorderi: You are technically correct, I've been wanting to get rid of the fromActor from that signature since it's purpose got reduced in recent spec changes. 
Currently, the fromActor is only used for its code cid.

icorderi:
cc: @frrist

Issue #3867, opened 2020-03-12
Multiple consumers from HeaviestTipSetCh will not see all events


anorth:
The Poster (which schedules windowed PoSt) consumes from the same channel as the message queue/pool handler, This means each consumer will only see some of the events. There must be a single consumer of the channel, which can multiplex to multiple concrete handers.

See also #2309

PR #3869, opened 2020-03-12
use handler method instead of channel to check PoSt

acruikshank:
### Motivation

The poster and message handling are currently using the chain.HeaviestTipsetCh which means one or the other will not see tipset events.

### Proposed changes

Introduce a HandleNewHead method on the storage mining submodule and the poster to replace the channel listening go routine previously in poster. This method is called by `node.handleNewChainHeads`.

Closes issue #3867


anorth:
Thanks, much better. I'll land this now since it may have a role in the chain sync failures we're observing.

Issue #3839, opened 2020-03-06
Basic message filtering in block production


anorth:
Beyond syntactic validation, there are some easily-checkable properties of messages we should do before including them in a block. The spec identifies a set of conditions under which the miner will be penalised for including a message that couldn't possibly succeed (e.g. because the sending actor doesn't exist).

Issue #3566, opened 2019-10-16
Message Validation Done Right


ZenGround0:
### Description
#3559 is a step in the right direction for pubsub propagation validation.  However that PR takes the shortcut of validating messages twice.  This will likely become quite costly and must be avoided eventually.  

At the time of writing the spec is also quite silent on what exactly should and should not be validated about a message before acceptance into the message pool or before repropagation on pubsub. 

### Acceptance criteria
 This issue tracks the work of 
(1) incorporating properly speced out validation rules into the node
(2) removing all duplication of message validation.  Rules should be checked exactly once and the validation condition should propagate from this computation to all the places it needs to be checked.

### Risks + pitfalls
Message validation is spread across disparate areas and so consolidating and registering a deduplicated computation callback on libp2p is not going to be trivial.

### Where to begin
(1) wait for / push for clarification from spec
(2) very likely block validation is going to get this right first.  We can draw inspiration from that work

anorth:
We must validate locally-generated messages too (e.g. for too-low gas limit)

PR #3870, opened 2020-03-12
Feat/mining message filter

ZenGround0:
### Motivation
We want to prevent blocks mined from including messages that will penalize the miner.

### Proposed changes
1) Reorganize message validation rules into 
  a) syntax validation -- checks that can be done without any actor state   
  b) signature validation
  c) penalty checks -- checks that reference actor state and are marked as causing miner penalty in the spec

2) Reorganize where validation is called.  Mpool Add, pubsub replay and inbox add do syntax validation.  Signature validation is now out of mpool / inbox add and only in pubsub message topic receipt and pubsub message topic replay.  Unfortunately there is still duplication in these two validations. Removing duplication here would have required veering far from the intention of this PR. So #3566 is still open.  Finally we now do message penalty checking validation in the mining worker to filter messages included in blocks.

This omits new tests to get this out the door as errors related to the new penalty checking feature should be non-critical and fairly easy to track down.  But I'll write a mining worker test or an integration mining once test showing we filter out bad messages if that's preferred.

Closes issue #3839

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/messaging_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/messaging_submodule.go
@@ -9,6 +9,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/message"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/net/pubsub"
+	"github.com/pkg/errors"
 )
 
 // MessagingSubmodule enhances the `Node` with internal messaging capabilities.
@@ -23,7 +24,8 @@ type MessagingSubmodule struct {
 	MessageTopic *pubsub.Topic
 	MessageSub   pubsub.Subscription
 
-	MsgPool *message.Pool
+	MsgPool   *message.Pool
+	MsgSigVal *consensus.MessageSignatureValidator
 }
 
 type messagingConfig interface {
@@ -36,10 +38,17 @@ type messagingRepo interface {
 
 // NewMessagingSubmodule creates a new discovery submodule.
 func NewMessagingSubmodule(ctx context.Context, config messagingConfig, repo messagingRepo, network *NetworkSubmodule, chain *ChainSubmodule, wallet *WalletSubmodule) (MessagingSubmodule, error) {
-	msgPool := message.NewPool(repo.Config().Mpool, consensus.NewIngestionValidator(chain.State, repo.Config().Mpool))
+	msgSyntaxValidator := consensus.NewMessageSyntaxValidator()
+	msgSignatureValidator := consensus.NewMessageSignatureValidator(chain.State)
+	msgPool := message.NewPool(repo.Config().Mpool, msgSyntaxValidator)
 	inbox := message.NewInbox(msgPool, message.InboxMaxAgeTipsets, chain.ChainReader, chain.MessageStore)
 
 	// setup messaging topic.
+	// register block validation on pubsub
+	mtv := net.NewMessageTopicValidator(msgSyntaxValidator, msgSignatureValidator)
+	if err := network.pubsub.RegisterTopicValidator(mtv.Topic(network.NetworkName), mtv.Validator(), mtv.Opts()...); err != nil {
+		return MessagingSubmodule{}, errors.Wrap(err, "failed to register message validator")
+	}
 	topic, err := network.pubsub.Join(net.MessageTopic(network.NetworkName))
 	if err != nil {
 		return MessagingSubmodule{}, err
@@ -49,13 +58,14 @@ func NewMessagingSubmodule(ctx context.Context, config messagingConfig, repo mes
 	outboxPolicy := message.NewMessageQueuePolicy(chain.MessageStore, message.OutboxMaxAgeRounds)
 	msgPublisher := message.NewDefaultPublisher(pubsub.NewTopic(topic), msgPool)
 	signer := wallet.Wallet
-	outbox := message.NewOutbox(signer, consensus.NewOutboundMessageValidator(), msgQueue, msgPublisher, outboxPolicy, chain.ChainReader, chain.State, config.Journal().Topic("outbox"))
+	outbox := message.NewOutbox(signer, msgSyntaxValidator, msgQueue, msgPublisher, outboxPolicy, chain.ChainReader, chain.State, config.Journal().Topic("outbox"))
 
 	return MessagingSubmodule{
 		Inbox:        inbox,
 		Outbox:       outbox,
 		MessageTopic: pubsub.NewTopic(topic),
 		// MessageSub: nil,
-		MsgPool: msgPool,
+		MsgPool:   msgPool,
+		MsgSigVal: msgSignatureValidator,
 	}, nil
 }
--- a/internal/app/go-filecoin/node/node.go
+++ b/internal/app/go-filecoin/node/node.go
@@ -687,7 +687,7 @@ func (node *Node) CreateMiningWorker(ctx context.Context) (*mining.DefaultWorker
 
 		MessageSource: node.Messaging.Inbox.Pool(),
 		MessageStore:  node.chain.MessageStore,
-		Processor:     node.Chain().Processor,
+		PenaltyChecker: consensus.NewMessagePenaltyChecker(node.Chain().State),
 		Blockstore:    node.Blockstore.Blockstore,
 		Clock:         node.ChainClock,
 		Poster:        node.StorageMining.PoStGenerator,
--- a/internal/app/go-filecoin/node/node.go
+++ b/internal/app/go-filecoin/node/node.go
@@ -685,12 +685,12 @@ func (node *Node) CreateMiningWorker(ctx context.Context) (*mining.DefaultWorker
 		TicketGen:      consensus.NewTicketMachine(node.PorcelainAPI),
 		TipSetMetadata: node.chain.ChainReader,
 
-		MessageSource:  node.Messaging.Inbox.Pool(),
-		MessageStore:   node.chain.MessageStore,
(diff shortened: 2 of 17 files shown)


internal/app/go-filecoin/internal/submodule/messaging_submodule.go
```diff
@@ -36,10 +38,17 @@ type messagingRepo interface {
...
 // NewMessagingSubmodule creates a new discovery submodule.
 func NewMessagingSubmodule(ctx context.Context, config messagingConfig, repo messagingRepo, network *NetworkSubmodule, chain *ChainSubmodule, wallet *WalletSubmodule) (MessagingSubmodule, error) {
-	msgPool := message.NewPool(repo.Config().Mpool, consensus.NewIngestionValidator(chain.State, repo.Config().Mpool))
+	msgSyntaxValidator := consensus.NewMessageSyntaxValidator()
+	msgSignatureValidator := consensus.NewMessageSignatureValidator(chain.State)
+	msgPool := message.NewPool(repo.Config().Mpool, msgSyntaxValidator)
 	inbox := message.NewInbox(msgPool, message.InboxMaxAgeTipsets, chain.ChainReader, chain.MessageStore)
 
 	// setup messaging topic.
+	// register block validation on pubsub
```
ZenGround0: We actually weren't doing replay validation before this PR, woop

internal/pkg/chainsync/fetcher/graphsync_fetcher.go
```diff
@@ -436,9 +433,6 @@ func (gsf *GraphSyncFetcher) loadAndVerifyFullBlock(ctx context.Context, key blo
 				return err
 			}
 
-			if err := gsf.validator.ValidateUnsignedMessagesSyntax(ctx, messages); err != nil {
```
ZenGround0: These have been calling unimplemented stubs for a long time.  I didn't try to plug in the right thing here but can file an issue for following up on this with actually message validation.

internal/pkg/consensus/validation.go
```diff
@@ -47,54 +47,41 @@ func init() {
...
+// causing miner penality.  It treats any miner penalty condtion as an error.
+func (v *MessagePenaltyChecker) PenaltyCheck(ctx context.Context, msg *types.UnsignedMessage) error {
+	fromActor, err := v.api.GetActorAt(ctx, v.api.Head(), msg.From)
+	if err != nil {
+		return err
 	}
-
-	if msg.GasPrice.LessThanEqual(types.ZeroAttoFIL) {
-		return errGasPriceZero
+	// Sender should not be an empty actor
```
ZenGround0: This is a departure from previous practice that I got from the spec "the From actor does not exist in the state (miner penalized)"

internal/pkg/consensus/validation.go
```diff
@@ -47,54 +47,41 @@ func init() {
...
+// penaltyCheckerAPI allows the validator to access latest state
+type penaltyCheckerAPI interface {
+	Head() block.TipSetKey
+	GetActorAt(ctx context.Context, tipKey block.TipSetKey, addr address.Address) (*actor.Actor, error)
 }
 
-// NewOutboundMessageValidator creates a new default validator for outbound messages. This
-// validator matches the default behaviour but allows nonces higher than the actor's current nonce
-// (allowing multiple messages to enter the mpool at once).
-func NewOutboundMessageValidator() *DefaultMessageValidator {
```
ZenGround0: I've done away with this abstraction because we're not doing *any* high nonce validation on messages that we put in the mpool after this PR.  I am probably least sure about the ramifications here so please take a look and let me know if this is something I need to think more about.

internal/app/go-filecoin/node/node.go
```diff
@@ -679,12 +679,12 @@ func (node *Node) CreateMiningWorker(ctx context.Context) (*mining.DefaultWorker
...
 
-		MessageSource: node.Messaging.Inbox.Pool(),
-		MessageStore:  node.chain.MessageStore,
-		Processor:     node.Chain().Processor,
-		Blockstore:    node.Blockstore.Blockstore,
-		Clock:         node.ChainClock,
-		Poster:        node.StorageMining.PoStGenerator,
+		MessageSource:  node.Messaging.Inbox.Pool(),
+		MessageStore:   node.chain.MessageStore,
+		PenaltyChecker: consensus.NewMessagePenaltyChecker(node.Chain().State),
```
anorth: I'd maybe say call it `MessageQualifier` or `MessageSelecter` or something. Scope will very likely expand in the future.

internal/pkg/chainsync/fetcher/graphsync_fetcher.go
anorth: Thanks, please do. I do believe we need something here else catchup sync messages are going unvalidated.

Can you reference https://github.com/filecoin-project/go-filecoin/issues/3312 here rather than leave no trace?

internal/pkg/consensus/validation.go
```diff
@@ -47,54 +47,41 @@ func init() {
 	errNonceTooHighCt = metrics.NewInt64Counter("consensus/msg_nonce_high_err", "Number of messages with nonce too high")
 }
 
-// DefaultMessageValidator validates incoming signed messages.
-type DefaultMessageValidator struct {
-	allowHighNonce bool
+// MessageSelectionChecker checks for miner penalties on signed messages
+type MessagePenaltyChecker struct {
```
anorth: Mismatched name/comment.

I would rename to something more positive like message qualifier or something, not specifically tied to the penalties we know happen in the VM. It's role is to select messages that make sense to include in a block, with reference to the parent state.

anorth: I would argue that this does not belong in the consensus package, but closer to block production. It needs access to state, but has nothing else to do with consensus.

internal/pkg/consensus/validation.go
```diff
@@ -47,54 +47,41 @@ func init() {
...
 }
 
-// Validate checks that a message is semantically valid for processing, returning any
-// invalidity as an error.
-func (v *DefaultMessageValidator) Validate(ctx context.Context, msg *types.UnsignedMessage, fromActor *actor.Actor) error {
-	if msg.From == msg.To {
-		return errSelfSend
+// PenaltyCheck checks that a message is semantically valid for processing without
+// causing miner penality.  It treats any miner penalty condtion as an error.
+func (v *MessagePenaltyChecker) PenaltyCheck(ctx context.Context, msg *types.UnsignedMessage) error {
```
anorth: `error` is no longer a good return type to use here, IMO. It's not an error in any of the usual senses, just a signal that this message is not appropriate for this state (but may be fine later). I would say use bool, but it's handy to get a reason string too. Maybe (bool, string)? Your call if you want to stay with error, but please document that it's a misuse of error.

internal/pkg/consensus/validation.go
```diff
@@ -47,54 +47,41 @@ func init() {
...
-
-	if msg.GasPrice.LessThanEqual(types.ZeroAttoFIL) {
-		return errGasPriceZero
+	// Sender should not be an empty actor
+	if fromActor == nil || fromActor.Empty() {
+		return errEmptySender
 	}
 
 	// Sender must be an account actor, or an empty actor which will be upgraded to an account actor
 	// when the message is processed.
```
anorth: The second half of this comment is wrong.

internal/pkg/consensus/validation.go
```diff
@@ -127,32 +114,59 @@ func canCoverGasLimit(msg *types.UnsignedMessage, actor *actor.Actor) bool {
...
+
+	if msg.Value.LessThan(specsbig.Zero()) {
+		log.Debugf("Cannot transfer negative value: %s from actor: %s", msg.Value, msg.From)
+		errNegativeValueCt.Inc(ctx, 1)
+		return errNegativeValue
+	}
+
+	if msg.From == msg.To {
+		return errSelfSend
+	}
```
anorth: This is now allowed by the spec. I assume that your goal was not to fully implement message syntax validation since there are other divergences too, and that's fine. But please ensure an issue remains to complete this.

internal/pkg/consensus/validation.go
anorth: We're going to need something here. "Validation" isn't the right term, but we need something to limit mpool spam, and evict lower-likelihood messages before better-looking ones. An existant sender and a limit on the number of distinct nonces from each sender will be important parts of that.

Needn't block this PR, but it's an important thing to get back. Please file an issue.

internal/pkg/net/validators.go
```diff
@@ -72,7 +72,7 @@ type MessageTopicValidator struct {
 
 // NewMessageTopicValidator returns a MessageTopicValidator using `mv` for
 // message validation
```
anorth: Comment is out of date.

internal/pkg/consensus/validation.go
ZenGround0: #3872

internal/pkg/consensus/validation.go
ZenGround0: This is wrapping the consensus rules which the mining messagequalifier uses but is not restricted to.  In the future I imagine the qualifier will own an instance of this type instead of being this type.  I am under the impression we are going to need to use these rules somewhere during validation (i.e. consensus) not just in mining message selection in order to actually extract the penalty.  

Odds are good I'm getting something wrong here but I'm going to move to leave things more obviously wrong now and fix them once I'm back up to speed.

internal/pkg/consensus/validation.go
ZenGround0: Not sure I agree but I've file an issue anyway for getting this part of the system working and we can work it out there: #3873

internal/pkg/consensus/validation.go
ZenGround0: Ah didn't realize, I'll remove

PR #3871, opened 2020-03-12
update chain-validation and specs-actors

frrist:
### Motivation
Update the version of chain-validation and specs-actors. Chain-validation now includes tests for invalid from addresses and miner creation, commitment, and slashing.


anorth:
Please hold off on this for a moment until we get the problems experienced by https://github.com/filecoin-project/go-filecoin/pull/3868 worked out. I think this actor update require us to implement fast catchup block production, because otherwise the bootstrap miner will be immediately slashed.

cc @ZenGround0

anorth:
I rebased and this didn't break our tests, but I think the testing flow of MineOnce is too different from the real mining cycle. It uses a different fake clock in order to produce a correct timestamp. So I think this is still liable to break in production. Keep holding off.

anorth:
Ok I think we can land this after all. When deploying a devnet we will need to ensure it starts within a short window after the configured genesis timestamp though, otherwise the miner will slash itself.

functional-tests/mining_sync_test.go
```diff
@@ -43,7 +43,12 @@ func TestBootstrapMineOnce(t *testing.T) {
 	// Check the miner's initial power corresponds to 2 2kb sectors
 	var status porcelain.MinerStatus
 	node0.MustRunCmdJSON(ctx, &status, "go-filecoin", "miner", "status", minerAddress.String())
-	assert.Equal(t, uint64(2*3072), status.Power.Uint64())
+
+	kib := uint64(1024)
+	// expected miner power is 2 2kib sectors
+	expectedMinerPower := (2 * kib) * 2
```
anorth: Use `constants.DevSectorSize`

PR #3874, opened 2020-03-13
circleci: append BASH_ENV to shell profile

travisperson:
### Motivation

https://github.com/circleci/circleci-docs/issues/1650

BASH_ENV is not being correctly loaded into the environment which is resulting in improper configuration.

### Proposed changes

Source the BASH_ENV from the shell profile script so its provided in each shell.


anorth:
Folded into #3868 (I used `bashrc`)

PR #3881, opened 2020-03-16
Feat/network protocol interop

ZenGround0:
### Motivation
Interop with lotus on message data over the wire

### Proposed changes
- Tag hello protocol structures properly so we can de/encode them and speak hello
- Reorder message fields to match new spec PR / lotus
- Change gas units serialization to that of big int instead of int64 to match lotus / spec.  We keep the internal representation of int64 which should be good (@icorderi @anorth could use your help confirming this doesn't put is in hot water with the spec)
- Remove deprecated gas unit type and use the new one everywhere

Note that you need to do a bit more hacking (specifically using a hardcoded network name different than that in gen block) to actually speak pubsub enough to see this working.

### DIFF
--- a/cmd/go-filecoin/main.go
+++ b/cmd/go-filecoin/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/paths"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/repo"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/gas"
 )
 
 const (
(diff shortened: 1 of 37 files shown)


cmd/go-filecoin/main.go
```diff
@@ -354,29 +355,29 @@ var priceOption = cmdkit.StringOption("gas-price", "Price (FIL e.g. 0.00013) to
 var limitOption = cmdkit.Uint64Option("gas-limit", "Maximum GasUnits this message is allowed to consume")
 var previewOption = cmdkit.BoolOption("preview", "Preview the Gas cost of this command without actually executing it")
 
-func parseGasOptions(req *cmds.Request) (types.AttoFIL, types.GasUnits, bool, error) {
+func parseGasOptions(req *cmds.Request) (types.AttoFIL, gas.Unit, bool, error) {
 	priceOption := req.Options["gas-price"]
 	if priceOption == nil {
-		return types.ZeroAttoFIL, types.GasUnits(0), false, errors.New("gas-price option is required")
+		return types.ZeroAttoFIL, gas.NewGas(0), false, errors.New("gas-price option is required")
```
anorth: Nit: `gas.Unit(0)` is shorter and stutters less. We should delete `NewGas` entirely.

internal/pkg/consensus/validation_test.go
```diff
@@ -136,7 +137,7 @@ func TestMessageSyntaxValidator(t *testing.T) {
 	})
 
 	t.Run("block gas limit fails", func(t *testing.T) {
-		msg, err := types.NewSignedMessage(*newMessage(t, alice, bob, 100, 5, 1, uint64(types.BlockGasLimit)+1), signer)
+		msg, err := types.NewSignedMessage(*newMessage(t, alice, bob, 100, 5, 1, int64(types.BlockGasLimit)+1), signer)
```
anorth: Change BlockGasLimit to `gas.Unit`?

cmd/go-filecoin/main.go
anorth: Also for this case, `gas.Zero`.

internal/pkg/types/message.go
```diff
@@ -24,14 +23,11 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/constants"
 	e "github.com/filecoin-project/go-filecoin/internal/pkg/enccid"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/gas"
 )
 
-// GasUnits represents number of units of gas consumed
-// This type is signed by design; it is possible for operations to consume negative gas.
```
anorth: Please preserve this comment by moving it to `gas.Unit`.

internal/pkg/version/protocol_versions.go
```diff
@@ -8,7 +8,7 @@ import (
 const USER = "alpha2"
 
 // DEVNET4 is the network name of devnet
-const DEVNET4 = "devnet4"
+const DEVNET4 = "interop"
```
anorth: Rename the constant too

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -330,7 +330,7 @@ func (vm *VM) applyImplicitMessage(imsg internalMessage, rnd crypto.RandomnessSo
 // applyMessage applies the message to the current state.
 func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd crypto.RandomnessSource) (message.Receipt, minerPenaltyFIL, gasRewardFIL) {
 	// Dragons: temp until we remove legacy types
-	var msgGasLimit gas.Unit = gas.Unit(msg.GasLimit)
+	var msgGasLimit gas.Unit = msg.GasLimit
```
anorth: Remove

PR #3884, opened 2020-03-17
Move block timestamp-setting logic from MineOnce to the block producer

anorth:
### Motivation
Blocks must specify a timestamp in the window derived from the genesis timestamp and epoch. We had this logic in MineOnce (which is only used by tests) but not the primary mining flow. We want to change behaviour when a node discovers that clock time is ahead of the chain it can observe to attempt to mine blocks from the past. This timestamp-fudging is necessary for such blocks to be valid.

### Proposed changes
I moved the timestamp-setting logic to the block producer so that it's reached in both testing and production flows. A future change will adjust the scheduler to fill a gap at the end of the chain, rather than leave it there.

Also moved the fake clock into the `clock` package out from the `testhelpers` dumping ground.


functional-tests/mining_chain_test.go
```diff
@@ -87,7 +86,8 @@ func TestSyncFromSingleMiner(t *testing.T) {
 	presealPath := filepath.Join(wd, "..", "fixtures/genesis-sectors")
 	genTime := int64(1000000000)
 	blockTime := 30 * time.Second
-	fakeClock := th.NewFakeClock(time.Unix(genTime, 0))
+	// The clock is set some way ahead of the genesis block.
+	fakeClock := clock.NewFake(time.Unix(genTime, 0).Add(4 * time.Hour))
```
ZenGround0: I don't think this we should add this edge case as the default to our syncing tests, there should be one test that covers this behavior.

internal/pkg/clock/testing.go
```diff
@@ -1,27 +1,30 @@
...
 	Advance(d time.Duration)
-	// BlockUntil will block until the FakeClock has the given number of
+	// BlockUntil will block until the Fake has the given number of
 	// sleepers (callers of Sleep or After)
 	BlockUntil(n int)
 }
 
-// NewFakeClock returns a FakeClock initialised at the given time.Time.
-func NewFakeClock(n time.Time) FakeClock {
+// NewFakeClock returns a Fake initialised at the given time.Time.
```
ZenGround0: nit: comment mismatch

ZenGround0: It would be nice if we could configure the linter to force consistency between function name and first comment word without forcing there to be a comment for all functions.

anorth: The Go recommendation to redundantly start comments with the symbol being commented is silly. My preferred solution is to ignore this recommendation and write comments that won't rot.

functional-tests/mining_chain_test.go
anorth: I've put this back to the "right" time, with advances when mining. Note that the clock is pretty much irrelevant to block production here since the MineOnce call avoids the scheduler and just sets the "right" block time. But the validating nodes chain sync will still reject blocks in the future.

It would be better if we could avoid MineOnce, or get it down to an entry into the scheduler rather than avoiding it.

PR #3886, opened 2020-03-17
add tipset identifier to more chain-querying methods

laser:
### Motivation

We need to provide a tipset identifier to chain-querying methods to ensure a consistent view of the chain across queries.

### Proposed changes

This PR updates go-fil-markets to use the latest and greatest, which also necessitates a few tipset token changes.


internal/app/go-filecoin/connectors/retrieval_market/provider.go
```diff
@@ -104,9 +104,8 @@ func (wrc limitedOffsetReadCloser) Close() error {
 
 // SavePaymentVoucher stores the provided payment voucher with the payment channel actor
```
shannonwells: NB: this comment is wrong and I have fixed it in my branch.

PR #3893, opened 2020-03-19
Update NewActorAddress computation.

### LINKED ISSUE #3826: NewActorAddress runtime implementation computes wrong address
The [spec defines](https://filecoin-project.github.io/specs/#vmruntime-implementation) the `NewActorAddress()` method to be implemented as follows:
```go
func (rt *VMContext) NewActorAddress() addr.Address {
	addrBuf := new(bytes.Buffer)

	err := rt._immediateCaller.MarshalCBOR(addrBuf)
	util.Assert(err == nil)
	err = binary.Write(addrBuf, binary.BigEndian, rt._toplevelSenderCallSeqNum)
	util.Assert(err != nil)
	err = binary.Write(addrBuf, binary.BigEndian, rt._internalCallSeqNum)
	util.Assert(err != nil)

	newAddr, err := addr.NewActorAddress(addrBuf.Bytes())
	util.Assert(err == nil)
	return newAddr
}
```

The current implementation incorrectly marshals creatorAddress as bytes instead of the expected CBOR and the omits `internalCallSeqNum` in the calculation:
https://github.com/filecoin-project/go-filecoin/blob/bfa563adeed5aa1daded4518850ecdb7b824f9ac/internal/pkg/vm/internal/vmcontext/runtime_adapter.go#L104-L111

https://github.com/filecoin-project/go-filecoin/blob/593634357fd67221ee9a8f3d39a5b77b4b2a0a6d/internal/pkg/vm/internal/vmcontext/invocation_context.go#L440-L452

anorth:
### Motivation
New actor address computation is given by https://github.com/filecoin-project/specs/issues/894

### Proposed changes
- Use the originator's stable (pubkey) address, not the immediate caller' ID address
- Use a counter of calls to NewActorAddress rather than all VM calls

FYI @arajasek, @whyrusleeping, @frrist 
Closes #3826

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/vmcontext.go
+++ b/internal/pkg/vm/internal/vmcontext/vmcontext.go
@@ -57,12 +57,11 @@ type minerPenaltyFIL = abi.TokenAmount
 type gasRewardFIL = abi.TokenAmount
 
 type internalMessage struct {
-	from          address.Address
-	to            address.Address
-	value         abi.TokenAmount
-	method        abi.MethodNum
-	params        interface{}
-	callSeqNumber uint64
+	from   address.Address
+	to     address.Address
+	value  abi.TokenAmount
+	method abi.MethodNum
+	params interface{}
 }
 
 // NewVM creates a new runtime for executing messages.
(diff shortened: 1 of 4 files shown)


internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -424,18 +449,23 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd c
 	// 2. build invocation context
 	// 3. process the msg
 
+	topLevel := topLevelContext{
+		originatorStableAddress: senderState.Address,
+		originatorCallSeq:       0,
```
icorderi: `originatorCallSeq: fromActor.CallSeqNum`?

not sure if you want it before or after the `fromActor.IncrementSeqNum()`

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -308,7 +307,26 @@ func (vm *VM) applyImplicitMessage(imsg internalMessage, rnd crypto.RandomnessSo
 		return nil, fmt.Errorf("implicit message `from` field actor not found, addr: %s", imsg.from)
 	}
 
-	imsg.callSeqNumber = fromActor.CallSeqNum
+	var originator address.Address
```
icorderi: comment block

internal/pkg/vm/internal/vmcontext/vmcontext.go
anorth: Oops, thanks. Used `msg.CallSeqNum`

PR #3895, opened 2020-03-19
Bug/use signer address for windowed post

acruikshank:
### Motivation

When used in a running node, the Poster doesn't work. This PR addresses some (possibly not all the issues).

1. Poster sets the message "From" to the miner worker address. This address is an ID address that can't be used to sign the message.
2. The number of candidates provided by sectorbuilder is always less than the minimum required by the miner actor.
3. The Poster is not providing enough gas for the PoSt message.

### Proposed changes

1. Find the signing address for the miner worker and use that as the message From.
2. ~Upgrade specs-actors where the limit has been relaxed.~ This has already been done on master.
3. Set the gas units to 5000 for the SubmitPoSt message.

### Note

This PR contains work towards a functional test of the PoSter (through which these issues were discovered and fixed). Unfortunately the test (or the Poster) is not quite working yet, so it can't really be included in this PR.

### DIFF
--- a/internal/pkg/poster/poster.go
+++ b/internal/pkg/poster/poster.go
@@ -82,7 +82,6 @@ func (p *Poster) startPoStIfNeeded(ctx context.Context, newHead block.TipSet) er
 		return nil
 	}
 
-	log.Infof("PoSter new head %+v", newHead.At(0))
 	tipsetHeight, err := newHead.Height()
 	if err != nil {
 		return err
@@ -98,8 +97,6 @@ func (p *Poster) startPoStIfNeeded(ctx context.Context, newHead block.TipSet) er
 	if err != nil {
 		return nil
 	}
-
-	log.Infof("PoSter test for start start, %d <=? height, %d", provingPeriodStart, tipsetHeight)
 
 	if provingPeriodStart > tipsetHeight {
 		// it's not time to PoSt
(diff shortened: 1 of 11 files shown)


internal/pkg/poster/poster.go
```diff
@@ -160,13 +165,18 @@ func (p *Poster) sendPoSt(ctx context.Context, stateView *appstate.View, candida
...
-	_, _, err = p.outbox.Send(
+	signerAddr, err := stateView.AccountSignerAddress(ctx, workerAddr)
+	if err != nil {
+		return err
+	}
+
+	mcid, _, err := p.outbox.Send(
 		ctx,
-		workerAddr,
+		signerAddr,
```
ZenGround0: To save onchain space / gas we may want to use the workerAddr but give the outbox the capability to lookup the corresponding signing address and key.

acruikshank: Thought of that. This is a bit of a quick fix. The Outbox is already accessing actor state, so it's not much of an architectural shift to have it look up signing addresses. I'll take a stab at it.

anorth: Fine to file an issue for it and come back later.

tools/gengen/util/generator.go
```diff
@@ -526,7 +526,11 @@ func (g *GenesisGenerator) setupPost(ctx context.Context, addr, mIDAddr address.
 	}
 
 	// set provingperiod start state field
-	minerActorState.PoStState.ProvingPeriodStart = miner.ProvingPeriod
+	ppStart := abi.ChainEpoch(miner.ProvingPeriod)
+	if provingPeriodStart != nil {
```
anorth: Minor preference for using 0 or -1 for the sentinel rather than introducing nil pointer possibilities here.

internal/pkg/poster/poster.go
acruikshank: #3900

PR #3897, opened 2020-03-19
Serialize signable block data with proper dummy value for signature 

ZenGround0:
### Motivation
To interop with lotus we need to sign and verify consistent block bytes.

There is some unspecced (issue forthcoming) ambiguity in how to serialize the bytes for signing and verifying.  We are adopting lotus's strategy of serializing sig field as null because it makes a bit more sense than our current strategy

### Proposed changes
- Turn sig field into pointer so that we can use the block datastructure to get this encoding automatically (could use a different structure if reviewers prefer)
- Add non-nil sig validation to syntax val to prevent users from crashing nodes

Closes issue #

### DIFF
--- a/internal/pkg/block/block_test.go
+++ b/internal/pkg/block/block_test.go
@@ -56,7 +56,7 @@ func TestTriangleEncoding(t *testing.T) {
 	}
 	t.Run("encoding block with zero fields works", func(t *testing.T) {
 		testRoundTrip(t, &blk.Block{
-			BlockSig:        crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
+			BlockSig:        &crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
 			BLSAggregateSig: crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte{}},
 		})
 	})
@@ -77,7 +77,7 @@ func TestTriangleEncoding(t *testing.T) {
 			ParentWeight:    fbig.NewInt(1000),
 			StateRoot:       e.NewCid(types.CidFromString(t, "somecid")),
 			Timestamp:       1,
-			BlockSig: crypto.Signature{
+			BlockSig: &crypto.Signature{
 				Type: crypto.SigTypeBLS,
 				Data: []byte{0x3},
 			},
@@ -129,7 +129,7 @@ func TestDecodeBlock(t *testing.T) {
 			Messages:        e.NewCid(cM),
 			StateRoot:       e.NewCid(c2),
 			MessageReceipts: e.NewCid(cR),
-			BlockSig:        crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
+			BlockSig:        &crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
 			BLSAggregateSig: crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte{}},
 		}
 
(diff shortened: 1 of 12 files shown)


internal/pkg/consensus/expected.go
```diff
@@ -202,7 +203,9 @@ func (c *Expected) validateMining(ctx context.Context,
 			return errors.Wrapf(err, "failed to convert address, %s, to a signing address", workerAddr.String())
 		}
 		// Validate block signature
-		if err := crypto.ValidateSignature(blk.SignatureData(), workerSignerAddr, blk.BlockSig); err != nil {
+		sigData := blk.SignatureData()
+		fmt.Printf("sigData: %x\n", sigData)
```
acruikshank: Debug code? Probably should remove.

internal/pkg/consensus/expected.go
```diff
@@ -202,7 +203,9 @@ func (c *Expected) validateMining(ctx context.Context,
...
+		if err := crypto.ValidateSignature(blk.SignatureData(), workerSignerAddr, *blk.BlockSig); err != nil {
```
acruikshank: This will panic if `blk.BlockSig` is nil. Are we guaranteed it won't be at this point? Some validation would make me feel better.

internal/pkg/slashing/check.go
```diff
@@ -130,7 +130,7 @@ func verifyBlockSignature(ctx context.Context, view FaultStateView, blk block.Bl
 	if err != nil {
 		panic(errors.Wrapf(err, "failed to inspect miner addresses"))
 	}
-	err = state.NewSignatureValidator(view).ValidateSignature(ctx, blk.SignatureData(), worker, blk.BlockSig)
+	err = state.NewSignatureValidator(view).ValidateSignature(ctx, blk.SignatureData(), worker, *blk.BlockSig)
```
acruikshank: `*blk.BlockSig` will panic if nil. Is this guaranteed not to be?

internal/pkg/consensus/expected.go
ZenGround0: That's why I added [this](https://github.com/filecoin-project/go-filecoin/pull/3897/files#diff-339af0e38c9bd41224f91100b0f74b51R116)

We can be paranoid and do it here if you think that's a good idea.  However we are already in such a bad state that panicing the node is probably the right response if we are doing this validation on a block that fails syntactic validation

internal/pkg/slashing/check.go
ZenGround0: I didn't follow up properly on this one, I'll take a look

internal/pkg/consensus/expected.go
ZenGround0: I'm just going to be paranoid and check for nil here too

internal/pkg/slashing/check.go
ZenGround0: turns out we really need that nil check nice catch

internal/pkg/block/block_test.go
```diff
@@ -56,7 +56,7 @@ func TestTriangleEncoding(t *testing.T) {
 	}
 	t.Run("encoding block with zero fields works", func(t *testing.T) {
 		testRoundTrip(t, &blk.Block{
-			BlockSig:        crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
+			BlockSig:        &crypto.Signature{Type: crypto.SigTypeSecp256k1, Data: []byte{}},
 			BLSAggregateSig: crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte{}},
```
anorth: Should this also be a pointer? What happens if there are no BLS messages? 

I guess that's what this test is exercising, and Lotus also uses a non-pointer here.

ZenGround0: > Should this also be a pointer?

That *would* be more consistent with the block sig convention and save us bytes I think.  We'd need to make some implementation changes but probably worth it.  At the very least we need to specify whether this can be serialized as nil or not.  

> Lotus also uses a non-pointer here

Yeah  I'm viewing this as fine tuned spec work that's probably valuable in the long run but not immediately necessary since we seem to be bytewise consistent across impls.

ZenGround0: > save us bytes
a byte

Issue #3903, opened 2020-03-19
Validate message CID lists in blocksub payload


anorth:
The blocksub payload includes lists of BLS and SECP message CIDs. The topic validator should check that these lists match the AMT roots in the accompanying block header.

That validation is skipped at present because those lists are unused in chain sync (#3902).

PR #3904, opened 2020-03-20
Include message CIDs in block pubsub payloads.

### LINKED ISSUE #3882: Interop Block pubsub message
### Description
I misread lotus code a few weeks back: it turns out the data sent over the gossipsub block topic is not a raw block header but a header with message cids.  I haven't found a good spec description 😶 but this is what it should look like:
```golang
type BlockPubSubData struct { 
  _ struct{} `cbor:",toarray"`
  Header block.Block
  BLSMsgCids []e.Cid
  SecpMsgCids []e.Cid
}
```
The cid ordering must be the same as the amt ordering (i.e. how our message store holds them).  I've gone through the trouble of writing in terms of encoding types and struct directives so this can be copy-pasted into a file.

We need to 
1) accept pubsub data assuming this format.  Introduce a new type, unmarshal into it, extract the header for sending to syncing system
2) create messages in this format: either pass cid lists along with new header in return value of mining worker OR read message arrays from message store in the block pubsub message sender (slight preference for this)

### Acceptance criteria
Block message interops with lotus

### Risks + pitfalls
This issue doesn't track making use or sense of these messages on receipt.  That's feature work I consider out of scope until optimizing for security and speed in painnet

### Where to begin
1) new type -> net validator unmarshal into it -> throw away message cids
2) choose your approach then make modifications to mining system and or pubsub message send in node/block.go

anorth:
### Motivation
Issue #3882 calls for including the included message CIDs in the block-pubsub ("blocksub") message payload. 

### Proposed changes
This hooks up the encoding and decoding of the message CID lists in the publisher, subscriber, and validator. The CID lists are not validated (since we ignore them), #3903.

- [x] Land #3901 and rebase

Closes #3882.

### DIFF
--- a/internal/pkg/net/blocksub/topic.go
+++ b/internal/pkg/net/blocksub/topic.go
@@ -2,6 +2,13 @@ package blocksub
 
 import (
 	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+	e "github.com/filecoin-project/go-filecoin/internal/pkg/enccid"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 )
 
 // BlockTopic returns the network pubsub topic identifier on which new blocks are announced.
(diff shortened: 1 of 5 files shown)


internal/pkg/net/blocksub/topic.go
```diff
@@ -2,17 +2,48 @@ package blocksub
...
-//	SECPMsgCids []e.Cid
-//}
+type Payload struct {
+	_           struct{} `cbor:",toarray"`
+	Header      block.Block
+	BLSMsgCids  []e.Cid
+	SECPMsgCids []e.Cid
+}
+
+func MakePayload(header *block.Block, BLSMessages, SECPMessages []*types.SignedMessage) ([]byte, error) {
```
ZenGround0: It would be slightly better to pass BLSMessages as unsigned and SECPMessages as signed since they are already separated and the the bls message signatures are unused

ZenGround0: You could also argue that only the two cid slices should be output from the mining worker

anorth: On the latter point I consider the abstraction cleaner by returning the messages, and it doesn't cost anything extra. The question is where should the knowledge about how to extract pubsub-relevant info from the messages go? I don't think the block producer should know/care.

PR #3906, opened 2020-03-20
Transfer balance of deleted actor to beneficiary.

anorth:
### Motivation
https://github.com/filecoin-project/specs-actors/pull/246 introduces a beneficiary for the deleted actor's balance.

### Proposed changes
This PR does the work to transfer the balance so that subsequent actor upgrade is easy.

### DIFF
--- a/internal/pkg/vm/state/state.go
+++ b/internal/pkg/vm/state/state.go
@@ -113,6 +113,8 @@ func (st *State) SetActor(ctx context.Context, key actorKey, a *actor.Actor) err
 // DeleteActor remove the actor from the storage.
 //
 // This method will NOT return an error if the actor was not found.
+// This behaviour is based on a principle that some store implementations might not be able to determine
+// whether something exists before deleting it.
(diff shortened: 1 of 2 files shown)


internal/pkg/vm/state/state.go
```diff
@@ -111,15 +111,12 @@ func (st *State) SetActor(ctx context.Context, key actorKey, a *actor.Actor) err
 }
 
 // DeleteActor remove the actor from the storage.
-//
-// This method will NOT return an error if the actor was not found.
```
anorth: This has only one caller, which now loads the actor in question before attempting to delete it, so it's appropriate to indicate if not found.

icorderi: This was always the case, the actor being deleted is the actor that is running. We already know it exists. 

The reason deleteActor here does not leak found/not-found out is for callers not make any assumptions about it. 
Not all implementations of stores can determine if something exists before deleting, most often that not, you do not want to do the check due to performance reasons. 

Please keep the looser requirement.

PR #3907, opened 2020-03-20
Message cid signing

### LINKED ISSUE #3885: Sign and Validate over message cid data
### Description
Message cids should be the target for signature creation and validation, not message data itself.  This is how lotus currently operates and @whyrusleeping will soon be PRing this protocol change to the spec.

We need to update go-filecoin to follow this rule.  

I've confirmed on interop net that if we do change signature operations in this way then go-filecoin will be able to validate lotus messages appearing on the message pubsub topic.

### Acceptance criteria

1) BLS and secp message signatures are made off of cids
2) validation of message signatures is with respect to message cid

### Risks + pitfalls

The right way to do this is on the ipld cbor encoding of a cid however lotus uses cid bytes.  The upcoming spec PR will have the final say but it isn't in yet so there is some ambiguity.

We're probably going to have to do this for block signatures and other other signatures too, but this issue is only for messages

### Where to begin

Follow signature creation and validation pathways to find the right place to insert these changes.  It should be at a level above the crypto package.

ZenGround0:
### Motivation
See #3885

### Proposed changes

Closes #3885

### DIFF
--- a/internal/pkg/state/sigval.go
+++ b/internal/pkg/state/sigval.go
@@ -34,30 +34,33 @@ func (v *SignatureValidator) ValidateSignature(ctx context.Context, data []byte,
 }
 
 func (v *SignatureValidator) ValidateMessageSignature(ctx context.Context, msg *types.SignedMessage) error {
-	data, err := msg.Message.Marshal()
+	mCid, err := msg.Message.Cid()
 	if err != nil {
-		return errors.Wrapf(err, "failed to encode message to check signature")
+		return errors.Wrapf(err, "failed to take cid of message to check signature")
 	}
+	data := mCid.Bytes()
+
 	return v.ValidateSignature(ctx, data, msg.Message.From, msg.Signature)
 }
 
 func (v *SignatureValidator) ValidateBLSMessageAggregate(ctx context.Context, msgs []*types.UnsignedMessage, sig crypto.Signature) error {
 	pubKeys := [][]byte{}
-	encodedMsgs := [][]byte{}
+	encodedMsgCids := [][]byte{}
 	for _, msg := range msgs {
 		signerAddress, err := v.state.AccountSignerAddress(ctx, msg.From)
 		if err != nil {
 			return errors.Wrapf(err, "failed to load signer address for %v", msg.From)
 		}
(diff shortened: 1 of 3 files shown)


internal/pkg/state/sigval.go
```diff
@@ -34,30 +34,33 @@ func (v *SignatureValidator) ValidateSignature(ctx context.Context, data []byte,
...
 			return errors.Wrapf(err, "failed to load signer address for %v", msg.From)
 		}
 		pubKeys = append(pubKeys, signerAddress.Payload())
-		msgBytes, err := msg.Marshal()
+		mCid, err := msg.Cid()
 		if err != nil {
 			return err
 		}
-		encodedMsgs = append(encodedMsgs, msgBytes)
+		msgBytes := mCid.Bytes()
```
anorth: Inline

internal/pkg/state/sigval_test.go
```diff
@@ -55,8 +55,9 @@ func TestSignMessageOk(t *testing.T) {
 		}}
 		v := NewSignatureValidator(state)
 		msg := types.NewMeteredMessage(idAddress, idAddress, 1, types.ZeroAttoFIL, builtin.MethodSend, nil, types.NewGasPrice(0), 0)
-		msgData, err := msg.Marshal()
+		msgCid, err := msg.Cid()
 		require.NoError(t, err)
+		msgData := msgCid.Bytes()
```
anorth: Inline

internal/pkg/types/signed_message.go
```diff
@@ -32,11 +32,12 @@ type SignedMessage struct {
 // NOTE: this method can only sign message with From being a public-key type address, not an ID address.
 // We should deprecate this and move to more explicit signing via an address resolver.
 func NewSignedMessage(msg UnsignedMessage, s Signer) (*SignedMessage, error) {
-	msgData, err := msg.Marshal()
+	msgCid, err := msg.Cid()
 	if err != nil {
 		return nil, err
 	}
 
+	msgData := msgCid.Bytes()
```
anorth: Inline

PR #3909, opened 2020-03-20
Wrap wallet with state.Signer that can retrieve a signing addresses

### LINKED ISSUE #3900: Outbox should accept messages sent from ID addreses
### Description

The Outbox must sign messages using the message's from address. ID addresses are not valid for signing but are valid "from" addresses (preferable, actually, because they're shorter). Either the Outbox or the Signer that is passed into it should convert ID addresses to signing addresses (by looking up account actor state on chain) in order to sign the message.

### Acceptance criteria

All messages sent from at least the miner worker address should be sent over the wire with an ID address.

acruikshank:
### Motivation

Some addresses can sign data and some can't. The ones that can't however should reference an account actor that contains an address that can sign. It's not always possible to understand which address is necessary in higher level code, and in some cases we really need both. The proposed solution is to move address resolution logic closer to signing so other code doesn't need to worry about signing vs. non-signing addresses

### Proposed changes

1. Modify the `types.Signer` interface to add a `HasAddress` method and add a context to the `SignBytes`.
2. Create `state.Signer` that implements the above interface with built in signer address resolution.
3. Add an instance of `state.Signer` to the Wallet submodule.
4. Replace all direct usages of `wallet.Wallet` that only deal with signing and checking addresses with `type.Signer` and populate it with `state.Signer`.
5. Stop implementing signing address resolution anywhere outside of `state.Signer`
6. Remove some unused wallet plumbing methods.

Closes #3900


internal/app/go-filecoin/internal/submodule/wallet_submodule.go
```diff
@@ -27,5 +30,6 @@ func NewWalletSubmodule(ctx context.Context, repo walletRepo) (WalletSubmodule,
 
 	return WalletSubmodule{
 		Wallet: fcWallet,
+		Signer: state.NewSigner(chain.ActorState, chain.ChainReader, fcWallet),
```
anorth: I don't think a dependency from wallet to chain is acceptable. That resolution should happen higher level.

internal/pkg/state/signer.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/crypto"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/wallet"
+)
+
+type chainHeadTracker interface {
+	GetHead() block.TipSetKey
```
anorth: This is the crux of the problem. Implicit head state access plus asynchrony can cause inconsistent state to be inspected by different bits of the code at different times. We're aiming to move all state access to use explicit keys. That means that this resolution cannot be implicit down deep in the system, but must be at time of use. Something like the SignatureValidator.

I would suggest leaving the old types.Signer interface alone and introducing a new one, with `Context`.

internal/pkg/types/signed_message.go
```diff
@@ -31,13 +32,13 @@ type SignedMessage struct {
 // a signature derived from the serialized `msg` and `msg.From`
 // NOTE: this method can only sign message with From being a public-key type address, not an ID address.
 // We should deprecate this and move to more explicit signing via an address resolver.
-func NewSignedMessage(msg UnsignedMessage, s Signer) (*SignedMessage, error) {
+func NewSignedMessage(ctx context.Context, msg UnsignedMessage, s Signer) (*SignedMessage, error) {
```
anorth: Note the comment above. I don't want to lean into this interface further.

internal/app/go-filecoin/internal/submodule/wallet_submodule.go
acruikshank: I agree that wallet to chain is a bad dependency. This is the wallet submodule. It's already created after the chain submodule in the builder, and every submodule that depends on it also depends on the chain submodule. It could be renamed to something that more obviously depends on chain.

internal/pkg/state/signer.go
acruikshank: I agree that implicit head state access is bad in general, but account actor signer addresses are immutable. Even if they weren't the question of what chain epoch am I signing this data at is somewhat non-nonsensical. I strongly doubt, if we were to ensure sub-components could provide a tip set key for signing, that they would ever use anything other than head, and that that key would every be used for another chain state operation.

So using the head tipset key retrieved only for this operation is a bad practice that we don't want to propagate, but in this case it has no significant consequences and helps eliminate a source of rather fiddly errors and overall complexity of the app. I didn't make this trade-off lightly.

internal/pkg/types/signed_message.go
acruikshank: That's fine. This was really a refactoring tool, that can go away now.

internal/pkg/state/signer.go
anorth: > the question of what chain epoch am I signing this data at

This has clear answers sometimes, like when signing a block header.

anorth:
I'm going to land this so it can be included in a new devnet deploy

PR #3910, opened 2020-03-20
Update chain-validation tests

anorth:
### Motivation
We fixed lots of stuff so I expected to be able to unskip a bunch of tests

### Proposed changes
Updated to master (after fixing filenames). Unfortunately I can't unskip anything. The commit history of chain-validation seems to imply that gas values should match go-filecoin, but they don't.

A bunch of tests seem to fail due to not finding an expected gas file. The test case not knowing what gas value to expect seems like a pretty weak reason to fail a test on us! I don't know what the expected workflow is for regenerating them.

I tried unskipping some other tests but no dice. The comments aren't quite enough to know what to do. Going forward, @frrist could you please link each skipped test to an issue describing what needs to change in go-filecoin?


frrist:
> We fixed lots of stuff...

Would this affect any of the state root CID calculations? If so, this is the cause of the gas value lookup failures. 

The gas used for applying a message is keys on the state tree CID (before message application) and the message CID. The map containing this information was generated using go-filecoin and if go-filecoin has changed much it will need to be regenerated.

> The test case not knowing what gas value to expect seems like a pretty weak reason to fail a test on us! 

Agreed, the tests wrongly expect that an implementation generates a consistent set of state transitions when looking up gas. I will remove the logic that causes a test to fail when this lookup fails.

PR #3918, opened 2020-03-21
Fix VMContext transfer to persist debit when receiver is sender.

anorth:
### Motivation
Sending to self shouldn't magically create money!

- [x] Land #3917 and rebase PR

### Proposed changes
Persist the debit account before loading the credit account, in case they are the same actor.

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/validation_test.go
+++ b/internal/pkg/vm/internal/vmcontext/validation_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/filecoin-project/chain-validation/suites"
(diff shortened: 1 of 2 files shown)


internal/pkg/vm/internal/vmcontext/validation_test.go
```diff
@@ -34,8 +33,6 @@ var TestSuiteSkipper TestSkipper
 func init() {
 	// initialize the test skipper with tests being skipped
 	TestSuiteSkipper = TestSkipper{testSkips: []suites.TestCase{
-		// Incorrect behavior on self-send.
-		message.TestValueTransferAdvance,
```
frrist: :tada:

PR #3923, opened 2020-03-24
increment bls seed when generating bls address

frrist:
Gas checking for tipset tests has been disabled as the bls seed bug caused the wrong values to be generated. Corresponding PR in chain-validation: https://github.com/filecoin-project/chain-validation/pull/139


internal/pkg/vm/internal/vmcontext/validation_test.go
```diff
@@ -55,7 +55,7 @@ func TestChainValidationMessageSuite(t *testing.T) {
 
 func TestChainValidationTipSetSuite(t *testing.T) {
 	f := NewFactories(&ValidationConfig{
-		trackGas:         true,
```
anorth: Do not merge this.

frrist: I have updated the chan-validation version and set this back to `true`.

PR #3926, opened 2020-03-24
fix: empty arrays not playing nice with deserializing into adt.EmptyValue

icorderi:
Issue with deserializing empty arrays into arguments.


internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -273,6 +273,7 @@ func (ctx *invocationContext) resolveTarget(target address.Address) (*actor.Acto
 	targetIDAddr, err := state.ResolveAddress(ctx.rt.ContextStore(), target)
 	// Dragons: move this logic to resolve address
 	notFound := (targetIDAddr == target && target.Protocol() != address.ID)
+	created := false
 	if err != nil || notFound {
```
anorth: As an aside, swallowing this error looks bad. Could we test against `ErrAddressNotFound` instead (and panic otherwise)?

icorderi: that is why the dragon is there for, i don't think we have a true notfound because of how resolve address was written.

anorth: Maybe that was true before, but now it returns `ErrAddressNotFound`

PR #3928, opened 2020-03-25
Update specs actors

ZenGround0:
### Motivation
This update helps chain syncing along by introducing needed caller validation to changing peer id

### Proposed changes

Closes #


anorth:
This is probably obsolete since #3929

PR #3930, opened 2020-03-26
pledge-sector

acruikshank:
### Motivation

This implementation has had no way to seal sectors and gain power since the old add-piece functionality was deprecated and removed. This PR adds a `mining pledge-sector` command and fixes numerous issue that prevented the untested (until now) PledgeSector functionality from working.

### Proposed changes

1. Expose PledgeSector through StorageAPI
2. Add `pledge-sector` command that calls PledgeSector.
3. Add `TestMiningSealSector` functional test.
4. Numerous modifications here and in go-storage-miner to make the above work. These are called out inline.


cmd/go-filecoin/daemon.go
```diff
@@ -163,7 +164,10 @@ func RunAPIAndWait(ctx context.Context, nd *node.Node, config *config.APIConfig,
 		inspectorAPI:   NewInspectorAPI(nd.Repo),
 		porcelainAPI:   nd.PorcelainAPI,
 		//retrievalAPI:   nd.RetrievalProtocol.RetrievalClient,
-		//storageAPI:     nd.StorageProtocol.StorageClient,
+	}
+
+	if nd.StorageProtocol != nil {
```
acruikshank: Nodes don't necessarily have storage protocols in test environments. Add this only if it does.

cmd/go-filecoin/mining.go
```diff
@@ -149,30 +149,17 @@ var stringEncoderMap = cmds.EncoderMap{
 	}),
 }
 
-// MiningAddPieceResult is a wrapper around the uint64 sectorID
-type MiningAddPieceResult struct {
```
acruikshank: I probably could have left Add-Piece as is, but it's not doing much good like this. It might be worth implementing it next.

go.mod
```diff
@@ -24,14 +24,13 @@ require (
 	github.com/filecoin-project/go-paramfetch v0.0.2-0.20200218225740-47c639bab663
 	github.com/filecoin-project/go-sectorbuilder v0.0.2-0.20200319233206-2a7435bc45d9
 	github.com/filecoin-project/go-statestore v0.1.0
-	github.com/filecoin-project/go-storage-miner v0.0.0-20200320161144-fd9286326083
+	github.com/filecoin-project/go-storage-miner v0.0.0-20200325225107-781a37a4b1b4
```
acruikshank: This fixes a lot of problems with sealing.

internal/app/go-filecoin/connectors/storage_miner/connector.go
```diff
@@ -85,7 +87,24 @@ func (m *StorageMinerNodeConnector) handleNewTipSet(ctx context.Context, previou
...
+		return cid.Undef, err
+	}
+
+	_, waddr, err := view.MinerControlAddresses(ctx, m.minerAddr)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	// ensure miner is in escrow table
+	err = m.establishEscrowBalanceIfNeeded(ctx, view, waddr, m.minerAddr)
```
acruikshank: We are (per go-storage-miner) still self-dealing to fill sectors. This requires the miner and miner work have an entry (which can be zero) in the markets escrow table.

internal/app/go-filecoin/connectors/storage_miner/connector.go
```diff
@@ -204,7 +225,7 @@ func (m *StorageMinerNodeConnector) SendPreCommitSector(ctx context.Context, sec
 	}
 
 	params := miner.SectorPreCommitInfo{
-		RegisteredProof: abi.RegisteredProof_StackedDRG32GiBSeal,
+		RegisteredProof: proofType,
```
acruikshank: This now ultimately comes from sectorbuilder.

internal/app/go-filecoin/connectors/storage_miner/connector.go
```diff
@@ -304,11 +326,7 @@ func (m *StorageMinerNodeConnector) GetSealTicket(ctx context.Context, tok stora
 		return storagenode.SealTicket{}, err
 	}
 
-	entropy, err := encoding.Encode(m.minerAddr)
-	if err != nil {
-		return storagenode.SealTicket{}, err
-	}
-	r, err := m.chainState.SampleChainRandomness(ctx, tsk, crypto.DomainSeparationTag_SealRandomness, h-miner.ChainFinalityish, entropy)
+	r, err := m.chainState.SampleChainRandomness(ctx, tsk, crypto.DomainSeparationTag_SealRandomness, h, nil)
```
acruikshank: No entropy in the miner-actor, no entropy here.

internal/app/go-filecoin/connectors/storage_miner/connector.go
```diff
@@ -347,38 +365,22 @@ func (m *StorageMinerNodeConnector) GetSealSeed(ctx context.Context, preCommitMs
...
-				}
-
-				entropy, err := encoding.Encode(m.minerAddr)
-				if err != nil {
-					ec <- storagenode.NewGetSealSeedError(err, storagenode.GetSealSeedFatalError)
-					break
-				}
 				randomness, err := m.chainState.SampleChainRandomness(ctx, key,
-					crypto.DomainSeparationTag_InteractiveSealChallengeSeed, tsHeight, entropy)
+					crypto.DomainSeparationTag_InteractiveSealChallengeSeed, seedEpoch, nil)
```
acruikshank: no entropy here either.

internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
```diff
@@ -78,7 +78,7 @@ func NewStorageMiningSubmodule(minerAddr address.Address, ds datastore.Batching,
 		PoStGenerator:    postGen,
 		minerNode:        minerNode,
 		storageMiner:     storageMiner,
-		heaviestTipSetCh: c.HeaviestTipSetCh,
+		heaviestTipSetCh: make(chan interface{}),
```
acruikshank: Nothing was ever coming out of this channel because its already being consumed. Create a new channel here and feed it in HandleNewHead method.

internal/app/go-filecoin/node/node.go
```diff
@@ -447,11 +447,11 @@ func (node *Node) setupStorageMining(ctx context.Context) error {
 		PoStProofType: postProofType,
 		SealProofType: sealProofType,
 		Miner:         minerAddr,
-		WorkerThreads: 1,
+		WorkerThreads: 2,
```
acruikshank: Sectorbuilder behaves differently depending on whether it's given 1 or more threads. It won't seal if only given 1.

internal/app/go-filecoin/node/node.go
```diff
@@ -447,11 +447,11 @@ func (node *Node) setupStorageMining(ctx context.Context) error {
...
 		Paths: []fs.PathConfig{
 			{
 				Path:   sectorDir,
-				Cache:  false,
+				Cache:  true,
```
acruikshank: At least one cache path is required for sealing.

internal/app/go-filecoin/porcelain/miner.go
```diff
@@ -208,6 +211,14 @@ func MinerGetStatus(ctx context.Context, plumbing minerStatusPlumbing, minerAddr
 	if err != nil {
 		return MinerStatus{}, err
 	}
+	sectorCount := 0
```
acruikshank: I think sector count is a intrinsically useful statistic, but it's the only artifact on chain that indicates sealing has been successful. It's used in the TestMiningSealSector test.

internal/pkg/chainsampler/height_threshold_scheduler.go
```diff
@@ -91,12 +91,17 @@ func (m *HeightThresholdScheduler) HandleNewTipSet(ctx context.Context, previous
 		return block.TipSet{}, err
 	}
 
-	_, newTips, err := chain.CollectTipsToCommonAncestor(ctx, m.chainStore, previousHead, newHead)
-	if err != nil {
-		return block.TipSet{}, err
+	var newTips []block.TipSet
+	if previousHead.Defined() {
```
acruikshank: This code had an initialization bug.

internal/pkg/protocol/storage/api.go
```diff
@@ -1,8 +1,24 @@
...
+	"github.com/filecoin-project/go-fil-markets/storagemarket"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/piecemanager"
 )
 
 // API is the storage API for the test environment
-type API = iface.StorageClient
+type API struct {
+	storage      storagemarket.StorageClient
+	piecemanager piecemanager.PieceManager
+}
```
acruikshank: This API is now a composite of the storage client and the piece manager.

internal/pkg/state/view.go
```diff
@@ -121,6 +122,18 @@ func (v *View) MinerSectorSize(ctx context.Context, maddr addr.Address) (abi.Sec
 	return minerState.Info.SectorSize, nil
 }
 
+// MinerSectors returns all the miner's onchain sectors
+func (v *View) MinerSectorsForEach(ctx context.Context, maddr addr.Address, f func(miner.SectorOnChainInfo) error) error {
```
acruikshank: This allows us to count a miner's sectors.

internal/pkg/state/view.go
```diff
@@ -170,6 +183,18 @@ func (v *View) MinerGetPrecommittedSector(ctx context.Context, maddr addr.Addres
 	return minerState.GetPrecommittedSector(StoreFromCbor(ctx, v.ipldStore), abi.SectorNumber(sectorNum))
 }
 
+// MarketEscrowBalance looks up a token amount in the escrow table for the given address
+func (v *View) MarketEscrowBalance(ctx context.Context, addr addr.Address) (found bool, amount abi.TokenAmount, err error) {
```
acruikshank: This is needed to test if we need to create and escrow balance in the market actor for a deal participant.

tools/gengen/util/generator.go
```diff
@@ -334,9 +334,12 @@ func (g *GenesisGenerator) setupMiners(ctx context.Context) ([]*RenderedMinerInf
 		}
 
 		// Add configured deals to the market actor with miner as provider and worker as client
-		dealIDs, err := g.commitDeals(ctx, wAddr, mIDAddr, m)
-		if err != nil {
-			return nil, err
+		dealIDs := []abi.DealID{}
+		if len(m.CommittedSectors) > 0 {
```
acruikshank: I forget why this blows up if the miner has no committed sectors, but clearly we don't need to commitDeals if we have no sectors.

functional-tests/plege_sector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			fakeClock.Advance(blockTime)
+			_, err := node.BlockMining.BlockMiningAPI.MiningOnce(ctx)
+			require.NoError(t, err)
+		}
+		time.Sleep(time.Second)
```
anorth: Why do you need to sleep? If the fake clock is working properly there should be no need to sleep here. This could be more obvious by setting the blockTIme to be high.

A miniscule sleep might be necessary for chain sync to process the new block.

internal/app/go-filecoin/connectors/storage_miner/connector.go
anorth: Thanks, please link https://github.com/filecoin-project/specs-actors/issues/260

internal/app/go-filecoin/connectors/storage_miner/connector.go
anorth: Ditto

internal/app/go-filecoin/internal/submodule/storage_mining_submodule.go
anorth: ugh, we have to get rid of these channels.

internal/pkg/state/view.go
anorth: This unfortunately couples the caller to the `SectorOnChainInfo` type, preventing versioning of it. If we only need the count right now, I suggest you have the method compute that directly. Otherwise replace the single struct parameter with multiple parameters, each being a field of the struct, but only as many of them as a caller today actually needs.

internal/app/go-filecoin/connectors/storage_miner/connector.go
acruikshank: linked

internal/app/go-filecoin/connectors/storage_miner/connector.go
acruikshank: linked

internal/pkg/state/view.go
acruikshank: Converted to MinerSectorCount

functional-tests/plege_sector_test.go
acruikshank: It's not necessary. Removed.

PR #3933, opened 2020-03-26
rewrite windowed post test as in-process test (that works)

acruikshank:
### Motivation

We have not, up until now, tested that windowed PoSt's generated by the poster will actually be accepted by the miner-actor. The new-ish in-process test framework makes this relatively trivial.

### Proposed changes

1. Write TestBootstrapWindowedPoSt functional test as an in-process test.
2. Refactor some common functional test functions into a common.go file.

### DIFF
--- a/functional-tests/mining_sync_test.go
+++ b/functional-tests/mining_sync_test.go
@@ -133,6 +133,7 @@ func TestBootstrapWindowedPoSt(t *testing.T) {
 			return
 		}
 
+		// If we mine too many blocks before the post is sent we could miss our window. Add some friction here.
 		time.Sleep(2 * time.Second)
 	}
 	t.Fatal("Timouut wating for windowed PoSt")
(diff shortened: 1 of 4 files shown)


functional-tests/mining_sync_test.go
```diff
@@ -74,53 +77,63 @@ func TestBootstrapMineOnce(t *testing.T) {
...
+
+		status, err := miner.PorcelainAPI.MinerGetStatus(ctx, maddr, requireChainHead(t, miner))
+		require.NoError(t, err)
+
+		if status.ProvingPeriodStart > 1 {
 			return
 		}
-		time.Sleep(5 * time.Second)
+
+		time.Sleep(2 * time.Second)
```
anorth: Do we still really need a non-trivial sleep here? Please jot a comment about what it's waiting for (the PoSt computation?)

anorth:
Love it

PR #3934, opened 2020-03-26
plumb tipset identifier to Node* methods

laser:
### Motivation

A few more chain accessing methods need a tipset identifier. This will be the final round.

### Proposed changes

This is the final PR in a chain of other PRs which have also added a tipset identifier parameter to methods on the various `Node*` interfaces.


internal/app/go-filecoin/connectors/retrieval_market/client.go
```diff
@@ -139,12 +139,12 @@ func (r *RetrievalClientConnector) getBlockHeight(tok shared.TipSetToken) (abi.C
 }
 
 func (r *RetrievalClientConnector) getBalance(ctx context.Context, account address.Address, tok shared.TipSetToken) (types.AttoFIL, error) {
-	ts, err := r.getTipSet(tok)
-	if err != nil {
+	var tsk block.TipSetKey
+	if err := tsk.UnmarshalCBOR(tok); err != nil {
```
ZenGround0: This will work but its idiomatic to use `encoding.Decode(tok, &tsk)` assuming tok is a byte slice underneath the hood

internal/app/go-filecoin/connectors/storage_market/client.go
```diff
@@ -58,8 +59,8 @@ func (s *StorageClientNodeConnector) AddFunds(ctx context.Context, addr address.
 }
 
 // EnsureFunds checks the current balance for an address and adds funds if the balance is below the given amount
-func (s *StorageClientNodeConnector) EnsureFunds(ctx context.Context, addr, walletAddr address.Address, amount abi.TokenAmount) error {
-	balance, err := s.GetBalance(ctx, addr)
+func (s *StorageClientNodeConnector) EnsureFunds(ctx context.Context, addr, walletAddr address.Address, amount abi.TokenAmount, tok shared.TipSetToken) error {
+	balance, err := s.GetBalance(ctx, addr, tok)
```
ZenGround0: It seems to me that we should be deserializing the toks (which I think are just cbor bytes of a tipset key) as soon as they cross into the go-filecoin part of the code base, rather than letting the token move between function calls internal to the connector.

I can see this case might be unavoidable because GetBalance is on the API boundary.  However the internal `listDeals` shouldn't be.

laser: Makes sense!

internal/app/go-filecoin/connectors/retrieval_market/client.go
laser: This makes sense. I've been given this direction before; spaced out and used the wrong method this time. Will update.

PR #3937, opened 2020-03-27
Fix invocation rollback to revert to the right prior state

anorth:
### Motivation
State was being rolled back too far when an inner call failed. Additionally, state was not being rolled back at all when a top-level call failed for receipt or return value reasons.

### Proposed changes
Take a checkpoint at the top of invoke() and remember the prior root, to roll back to it explicitly when aborting.

The previous behaviour checkpointed at the end of a successful send, and rolled back on failure to the _last_ checkpoint (not nec. correct). This new behaviour checkpoints at the _top_ of an invocation, and implicitly (via function scope) tracks the call tree, rather than linearised list of calls, so it can roll back to the right place.

With this change, we pass all chain validation tests!


anorth:
The second commit adds annotations to gas charges for debugging, and bumps some gas limits to get the TestPledgeSector passing,

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -501,6 +500,16 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd c
 		receipt.ReturnValue = []byte{}
 	}
 
+	// Roll back all state if the receipt's exit code is not ok.
+	// This is required in addition to rollback within the invocation context since top level messages can fail for
+	// more reasons that internal ones. Invocation context still needs its own rollback so actors can recover and
```
acruikshank: that -> than

internal/pkg/vm/internal/vmcontext/validation_test.go
```diff
@@ -33,7 +32,7 @@ var TestSuiteSkipper TestSkipper
 func init() {
 	// initialize the test skipper with tests being skipped
 	TestSuiteSkipper = TestSkipper{testSkips: []suites.TestCase{
-		message.TestNestedSends,
+		// None!
```
icorderi: woot woot!

PR #3948, opened 2020-03-31
Change per-message gas transfer to go direct to reward actor rather than staging in burnt funds.

anorth:
### Motivation
We implemented this correctly from the spec but Lotus took this latter approach. It shouldn't matter in the final state, but per-message chain validation tests inspect this state so the difference caused tests to fail.

Transferring direct to the reward actor saves a transfer per block.

### Proposed changes
Stage gas prepayments in the reward actor rather than the burnt funds actor.

- [x] Pull in corresponding chain validation expectations: https://github.com/filecoin-project/chain-validation/pull/152


anorth:
The tests will pass when I update this to point at the matching expectation files from chain-validation.

PR #3956, opened 2020-04-01
Spec-compliant message syntax validation.

anorth:
### Motivation
Message syntax validation to match the spec. This was specifically motivated by changing the min gas cost to allow zero (allowed by the spec), which is most useful for testing and debugging.

See #3312 (which is really about using this in graphsync).

### Proposed changes
Message validation to match [the spec](https://filecoin-project.github.io/specs/#systems__filecoin_vm__message).

See also https://github.com/filecoin-project/specs/issues/906

### DIFF
--- a/internal/pkg/consensus/validation.go
+++ b/internal/pkg/consensus/validation.go
@@ -26,7 +26,6 @@ var dropNonceTooHighCt *metrics.Int64Counter
 
 var invReceiverUndefCt *metrics.Int64Counter
 var invSenderUndefCt *metrics.Int64Counter
-var invCallSeqNegativeCt *metrics.Int64Counter
 var invValueAboveMaxCt *metrics.Int64Counter
 var invParamsNilCt *metrics.Int64Counter
 var invGasPriceNegativeCt *metrics.Int64Counter
@@ -50,7 +49,6 @@ func init() {
 
 	invReceiverUndefCt = metrics.NewInt64Counter("consensus/msg_undef_receiver", "Count of")
 	invSenderUndefCt = metrics.NewInt64Counter("consensus/msg_undef_sender", "Count of")
-	invCallSeqNegativeCt = metrics.NewInt64Counter("consensus/msg_callseqno_negative", "Count of")
 	invValueAboveMaxCt = metrics.NewInt64Counter("consensus/msg_value_max", "Count of")
 	invParamsNilCt = metrics.NewInt64Counter("consensus/msg_params_nil", "Count of")
 	invGasPriceNegativeCt = metrics.NewInt64Counter("consensus/msg_gasprice_negative", "Count of")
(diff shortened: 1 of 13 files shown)


internal/pkg/consensus/validation.go
```diff
@@ -118,25 +130,66 @@ func NewMessageSyntaxValidator() *MessageSyntaxValidator {
...
+	if msg.Value.LessThan(big.Zero()) {
+		invNegativeValueCt.Inc(ctx, 1)
+		return fmt.Errorf("negative value %s: %s", msg.Value, msg)
+	}
+	if msg.Value.GreaterThan(msgMaxValue) {
+		invValueAboveMaxCt.Inc(ctx, 1)
+		return fmt.Errorf("value %s exceeds max %s: %s", msg.Value, msgMaxValue, msg)
 	}
+	// The spec calls for validating a non-negative method num, but by the time it's decoded into a uint64
+	// we can't do that any more.
```
ZenGround0, line 172: It's good to call this out here to help the reader see that this check is happening elsewhere.  I think the spec should express this opinion as a requirement that this field is serialized as an unsigned integer (major type 0) and never a negative integer (major type 1). I think the standard way to do this is by changing `CallSeqNum` [defined in this section](https://filecoin-project.github.io/specs/#systems__filecoin_vm__actor) to a specs "uint64".  Then this comment can be removed.

If we don't want to do this then in my view the current approach is still working as intended, its just an implementation detail that we cover this validation rule on the decode step.  The comment can change a bit to reflect this "we can't do that any more" => "we have already checked this"

Finally if there is some reason that we want to allow negative CallSeqNums that fail validation then we can change the field type to int64 and add a check here.

internal/pkg/message/outbox.go
```diff
@@ -99,6 +99,11 @@ func (ob *Outbox) Send(ctx context.Context, from, to address.Address, value type
 	if err != nil {
 		return cid.Undef, nil, errors.Wrap(err, "invalid params")
 	}
+	// Our encoder can't distinguish between producing a nil and producing an empty byte array,
```
ZenGround0, line 102: nit: this comment is a bit confusing.  I think the situation is the opposite: our encoder does distinguish between serializing nil (as `0xf6`) and and empty byte slice (as `0x40`).

I'd prefer referencing the filecoin spec requirement here if we want a comment making the motivation clear.  i.e. "The spec's message syntax validation rules restricts empty parameters to be encoded as an empty byte string not cbor null"

anorth: I'm saying our encoder cannot produce zero bytes as an output. If you encode an object that implements CBORMarshaler and writes zero bytes, then this encode will return a `nil`, not `[]byte{}`. This is different to encoding a `[]byte{}`, which produces one byte.

That nil is then the wrong value to put in this structure, which is itself then encoded.

PR #3957, opened 2020-04-01
bug fix 512 mib sectors

acruikshank:
### Motivation

Almost all non-interop testing for go-filecoin to date has been using the development 2KiB sector size for proofs. This branch reflects bug fixes required to get 512MiB sector election and windowed PoSt generation and verification to work in a one-off test (which is not included in this PR for size and test performance reasons).

### Proposed changes

see inline comments.

### DIFF
--- a/fixtures/setup.json
+++ b/fixtures/setup.json
@@ -17,6 +17,7 @@
       "commR": {"/":"bafk4ehzaqugph7z7yajugw6wz6nn7x3akr3e32wecwexllfgqndnjt3nsaya"},
       "commD": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
       "sectorNum": 3,
+      "proofType": 4,
       "dealCfg": {
         "commP": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
         "pieceSize": 2048,
@@ -26,6 +27,7 @@
       "commR": {"/":"bafk4ehza24k4j256kulsu34e4szjvmsbytibrx7kjzajc6c6rlcn2jfosada"},
       "commD": {"/":"bafk4chzaqfni3wpij32pzgv2vuf63utnufburrjzrcoud75oz3pxkfisligq"},
       "sectorNum": 4,
+      "proofType": 4,
       "dealCfg": {
         "commP": {"/":"bafk4chzaqfni3wpij32pzgv2vuf63utnufburrjzrcoud75oz3pxkfisligq"},
         "pieceSize": 2048,
--- a/functional-tests/mining_sync_test.go
+++ b/functional-tests/mining_sync_test.go
@@ -2,6 +2,7 @@ package functional
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"testing"
 	"time"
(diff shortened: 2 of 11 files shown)


fixtures/setup.json
```diff
@@ -17,6 +17,7 @@
       "commR": {"/":"bafk4ehzaqugph7z7yajugw6wz6nn7x3akr3e32wecwexllfgqndnjt3nsaya"},
       "commD": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
       "sectorNum": 3,
+      "proofType": 4,
```
acruikshank, line 20: The proof type must be added to the SectorInfo on chain, so I've added it to the genesis config.

functional-tests/mining_sync_test.go
```diff
@@ -84,6 +84,10 @@ func TestBootstrapWindowedPoSt(t *testing.T) {
 	wd, _ := os.Getwd()
 	genCfgPath := filepath.Join(wd, "..", "fixtures/setup.json")
 	presealPath := filepath.Join(wd, "..", "fixtures/genesis-sectors")
+	// setup presealed sectors and uncomment to run test against sectors with larger sector size
+	//genCfgPath := filepath.Join("./512", "setup.json")
+	//presealPath := "./512"
```
acruikshank, line 89: This is how I ran the 512MiB test. It's probably too much data for a fixture, and makes for longer functional tests (although still less than a minute on my laptop).

internal/app/go-filecoin/node/init.go
```diff
@@ -146,14 +149,13 @@ func importInitKeys(w *wallet.Wallet, importKeys []*crypto.KeyInfo) error {
...
 func ImportPresealedSectors(rep repo.Repo, srcPath string, sectorSize abi.SectorSize, symlink bool) error {
-	badgerOptions := badger.Options{
-		GcDiscardRatio: badger.DefaultOptions.GcDiscardRatio,
-		GcInterval:     badger.DefaultOptions.GcInterval,
-		GcSleep:        badger.DefaultOptions.GcSleep,
-		Options:        badger.DefaultOptions.Options,
-	}
-	badgerOptions.ReadOnly = true
-	oldMetaDs, err := badger.NewDatastore(filepath.Join(srcPath, "badger"), &badgerOptions)
+	nextSecnum, err := findNextSecnum(srcPath)
```
acruikshank, line 152: Lotus's seed tool no longer provides a badger directory in the genesis-sectors. We were only using the badger database to set the sectornum, so we can do that directly. Checking the highest sectornum found in a file path works. Not this isn't production code.

internal/pkg/consensus/election.go
```diff
@@ -40,15 +40,30 @@ func (em ElectionMachine) GenerateEPoStVrfProof(ctx context.Context, base block.
...
+	minerID, err := address.IDFromAddress(maddr)
+	if err != nil {
+		return nil, err
+	}
+
 	candidates := make([]abi.PoStCandidate, len(candidatesWithTicket))
 	for i, candidateWithTicket := range candidatesWithTicket {
 		candidates[i] = candidateWithTicket.Candidate
+		candidates[i].RegisteredProof = proofTypeBySectorNumber[candidates[i].SectorID.Number]
+		candidates[i].SectorID.Miner = abi.ActorID(minerID)
```
acruikshank, line 66: These need to be set for generate and validate to work (not sure why this was working before).

tools/gengen/util/generator.go
```diff
@@ -638,7 +638,7 @@ func (g *GenesisGenerator) putSectors(ctx context.Context, comm *CommitConfig, m
 
 	newSectorInfo := &miner.SectorOnChainInfo{
 		Info: miner.SectorPreCommitInfo{
-			RegisteredProof: abi.RegisteredProof_StackedDRG2KiBSeal, // default to 2kib, TODO set based on sector size
+			RegisteredProof: comm.ProofType,
```
acruikshank, line 641: This lets genesis work with sector sizes other than 2KiB.

tools/gengen/util/testing.go
```diff
@@ -39,6 +40,7 @@ func MakeCommitCfgs(n int) ([]*CommitConfig, error) {
 			CommD:     commD,
 			SectorNum: uint64(i),
 			DealCfg:   dealCfg,
+			ProofType: abi.RegisteredProof_StackedDRG2KiBPoSt,
```
acruikshank, line 43: This is the default genesis config, which we want to be 2KiB.

functional-tests/mining_sync_test.go
ZenGround0: fancy laptop

internal/pkg/consensus/election.go
ZenGround0, line 66: What about sectorID.Number?  Does it not matter?

acruikshank: These candidates are being generated by sectorbuilder with the sectorID.Number. For some reason they do not come back with the proof type or miner id. It's not entirely clear whether this is necessary, but it differs from the way the candidate data is presented to the verifier.

Issue #3959, opened 2020-04-02
Single-miner chain sync halts


anorth:
We've observed quite a few times that a single mining node with all the power will stop extending the chain after some time. The miner keeps producing blocks, but all with the same parent and an increasing number of null blocks.

It appears like the miner's chain sync system is not accepting its own blocks. There are no log messages indicating validation failure, though. Restarting the node fixes the issue temporarily, and the node once again accepts its own blocks.

Another possibility is that the goroutine processing the `incoming` channel is getting wedged and stops draining the queue, so the new blocks don't even reach validation.

FYI @ZenGround0

anorth:
Log here: https://gist.github.com/anorth/383f9a332132354495cee4eead22cd11

The last thing the dispatcher says is

```
2020-04-02T03:09:41.619Z        ESC[35mDEBUGESC[0m      chainsync.dispatcher    dispatcher/dispatcher.go:184    processing {{QmdSftL5s53Zb8Hkp8F5k1sGNA6MpJ6FEuSYVmNfQyFydT QmdSftL5s53Zb8Hkp8F5k1sGNA6MpJ6FEuSYVmNfQyFydT { bafy2bzacedjfwhdmrxr2ywyxe6bw4z
```

Subsequent messages about failing to dial self are common (but also something we should fix).

anorth:
Due to naming confusing, the error return from HandleNewTipSet was being dropped

anorth:
I think the problem is the HeaviestTipSetCh that is the subscription to the chain store's HeadEvents cskr/pubsub thingy. The last log message we see in the thread of control is right before the chain store publishes the new head to the pubsub. I think it's blocking because the channel underlying at least one of the subscriptions has reached capacity (128) and is not being drained.

anorth:
Ref #2309

anorth:
Ok, I think I have a smoking gun. The height threshold listener used by storage mining creates an unbuffered channel, to which it pushes new heads received via HandleNewHead, but when the listener exits (due to context cancellation) it does not unsubscribe so subsequent call to HandleNewHead blocks.

This listener exist immediately that storage mining starts up, due to context cancellation. That's probably a bug of using the wrong context. We should fix these independently though so that if the listener ever does stop it can't wedge the chain sync.


```
[...]

PR #3967, opened 2020-04-03
Add logs for heaviest tipset handling

anorth:
### Motivation
Debugging chain sync halt (#3959). Hypothesis is that the new heads pubsub channel is not being drained.

### Proposed changes
Add log messages around the subscribers to the new head. Clean up unsub handling.

These messages revealed that the problem is the storage mining module's height listener shutting down and then blocking in `HandleNewHead`.

### DIFF
--- a/internal/pkg/chain/store.go
+++ b/internal/pkg/chain/store.go
@@ -87,6 +87,8 @@ type Store struct {
 	// Successive published tipsets may be supersets of previously published tipsets.
 	// TODO: rename to notifications.  Also, reconsider ordering assumption depending
 	// on decisions made around the FC node notification system.
+	// TODO: replace this with a synchronous event bus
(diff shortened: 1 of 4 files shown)


internal/pkg/chain/store.go
```diff
@@ -101,7 +103,7 @@ func NewStore(ds repo.Datastore, cst cbor.IpldStore, sr Reporter, genesisCid cid
 	return &Store{
 		stateAndBlockSource: newSource(cst),
 		ds:                  ds,
-		headEvents:          pubsub.New(128),
+		headEvents:          pubsub.New(12),
```
anorth, line 106: A large buffer here just masked the problem for longer.

anorth:
Subsumed by #3968

miguel1942:
#file:///storage/emulated/0/Download/codebeautify.json

PR #3968, opened 2020-04-03
Remove channels and goroutine from HeightThresholdScheduler and storage mining module.

### LINKED ISSUE #3959: Single-miner chain sync halts

anorth:
### Motivation
Fewer channels = fewer problems. All the asynchrony here was unnecessary. This fixes a few bugs.
- the context being passed to Start() was not long-running, so the scheduler's goroutine exited immediately
- after the goroutine exited, the module HandleNewHead blocked on the unbuffered heaviestTipSet channel
- the argument to HandleNewTipSet was named and treated as the previous head, rather than new head; the chain store was inspected for head which might race with other writes to it
- CancelListener iterated over an empty slice and always forgot all listeners

I have no idea how things worked at all here.

This will probably fix issue #3959, but I'll wait to close it until we observe better behaviour in devnets.

### Proposed changes
Remove all the channel control flow from HeightThresholdScheduler along with the unbuffered channel in the module, and just call HandleNewTipSet directly. Track the previous head inside the schedule for correct CollectTipsToCommonAncestor call (like the message pool handler does).

### DIFF
--- a/internal/pkg/chain/store.go
+++ b/internal/pkg/chain/store.go
@@ -87,6 +87,8 @@ type Store struct {
 	// Successive published tipsets may be supersets of previously published tipsets.
 	// TODO: rename to notifications.  Also, reconsider ordering assumption depending
 	// on decisions made around the FC node notification system.
+	// TODO: replace this with a synchronous event bus
+	// https://github.com/filecoin-project/go-filecoin/issues/2309
 	headEvents *pubsub.PubSub
 
 	// Tracks tipsets by height/parentset for use by expected consensus.
@@ -101,7 +103,7 @@ func NewStore(ds repo.Datastore, cst cbor.IpldStore, sr Reporter, genesisCid cid
 	return &Store{
 		stateAndBlockSource: newSource(cst),
 		ds:                  ds,
-		headEvents:          pubsub.New(128),
+		headEvents:          pubsub.New(12),
 		tipIndex:            NewTipIndex(),
 		genesis:             genesisCid,
 		reporter:            sr,
--- a/internal/pkg/chainsampler/height_threshold_scheduler.go
+++ b/internal/pkg/chainsampler/height_threshold_scheduler.go
@@ -4,11 +4,13 @@ import (
 	"context"
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
-	"github.com/prometheus/common/log"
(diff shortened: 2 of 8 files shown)


internal/pkg/chainsampler/height_threshold_scheduler.go
```diff
@@ -2,108 +2,81 @@ package chainsampler
...
-				if err != nil {
-					log.Warn("failed to handle new tipset")
-				} else {
-					previousHead = head
-				}
-			case heightListener := <-m.newListener:
-				m.heightListeners = append(m.heightListeners, heightListener)
-			case canceled := <-m.cancelListener:
-				listeners := []*HeightThresholdListener{}
-				for _, l := range listeners {
```
anorth, line 49: This was iterating the wrong slice, hence always dropping all listeners.

internal/pkg/chainsampler/height_threshold_scheduler.go
```diff
@@ -2,108 +2,81 @@ package chainsampler
...
-					if l != canceled {
-						listeners = append(listeners, l)
-					}
-				}
-				m.heightListeners = listeners
-			case <-m.schedulerDone:
-				return
-			case <-ctx.Done():
```
anorth, line 57: The wrong context being passed in cause this case to trigger immediately

internal/pkg/chain/store.go
```diff
@@ -101,7 +103,7 @@ func NewStore(ds repo.Datastore, cst cbor.IpldStore, sr Reporter, genesisCid cid
 	return &Store{
 		stateAndBlockSource: newSource(cst),
 		ds:                  ds,
-		headEvents:          pubsub.New(128),
+		headEvents:          pubsub.New(12),
```
anorth, line 106: The large buffer just masked problems.

PR #3971, opened 2020-04-06
drand client integration with associated infrastructure

acruikshank:
### Motivation

go-filecoin needs to access public drand networks to get randomness for leader election. Drand clients require configuration. This PR aims to provide a client and associated infrastructure to manage its configuration.

### Proposed changes

1. Introduce GRPC drand client to go-filecoin.
2. Add client configuration to node config.
3. Create a DrandAPI for use in commands and the client.
4. Introduce a commands to fetch drand configuration that updates the client in place and stores configuration in config.
5. Introduce functional test that is skipped pending a reliable source of drand.


cmd/go-filecoin/main.go
```diff
@@ -132,6 +132,8 @@ NETWORK COMMANDS
   go-filecoin ping <peer ID>...      - Send echo request packets to p2p network members
   go-filecoin swarm                  - Interact with the swarm
   go-filecoin stats                  - Monitor statistics on your network usage
+  go-filecion drand configure        - Configure drand srever connection
```
ZenGround0, line 135: nit server

functional-tests/drand_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/drand"
+	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDrandPublic(t *testing.T) {
+	tf.FunctionalTest(t)
+	t.Skip(("requires local drand setup"))
```
ZenGround0, line 20: I'm fine with this for now.  I've been advocating for having functional tests hit the default (i.e. public drand network) but maybe that's not a good idea.  cc @anorth who I've talked to about this.

internal/pkg/config/config.go
```diff
@@ -127,6 +128,34 @@ func newDefaultWalletConfig() *WalletConfig {
...
+		Addresses: []string{
+			"localhost:8080",
+			"localhost:8081",
+			"localhost:8082",
+			"localhost:8083",
+			"localhost:8084",
+		},
+		Secure: false,
+		DistKey: []string{
+			"8d98373363b8b62da5d5ef24accd81a35254a37c2c706989ff425c7a0ac24abee024721db6e03478831523fb7bf02cad",
```
ZenGround0, line 152: Where does this key come from?  A local test network like the addresses? The public network's real key?

internal/pkg/config/config.go
```diff
@@ -127,6 +128,34 @@ func newDefaultWalletConfig() *WalletConfig {
...
+	// Secure is whether or not the drand address are secure (e.g. TLS)
+	Secure bool `json:"secure"`
+	// DistKey is the distributed public key of the server group expressed as hex encoded coefficients
+	DistKey []string `json:"distKey"`
+}
+
+func newDefaultDrandConfig() *DrandConfig {
+	return &DrandConfig{
+		Addresses: []string{
+			"localhost:8080",
```
ZenGround0, line 144: Eventually the public network's addresses go here right?  When will we switch over?

internal/pkg/drand/drand_grpc.go
```diff
@@ -0,0 +1,156 @@
...
+			Signature: crypto.Signature{
+				Type: crypto.SigTypeBLS,
+				Data: pub.GetSignature(),
+			},
+		}, nil
+	}
+	return nil, errors.New("Could not retrieve drand randomess from any address")
+}
+
+// VerifyEntry always returns true without error
```
ZenGround0, line 74: nit function comment needs changing

internal/pkg/drand/drand_grpc.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return &GRPC{
+		addresses: addresses,
+		client:    core.NewGrpcClient(),
+		key:       distKey,
+	}, nil
+}
+
+// ReadEntry immediately returns a drand entry with a signature equal to the
+// round number
```
ZenGround0, line 53: nit: function comment needs changing

cmd/go-filecoin/main.go
acruikshank: 🙏

functional-tests/drand_test.go
acruikshank: Yeah, I played around a bit with getting this to work against the existing public network, but it's just too far out of date.

internal/pkg/config/config.go
acruikshank: It's a local network. It's actually an old local network. I'll remove these for now.

internal/pkg/config/config.go
acruikshank: When we know what addresses we'll be using.

PR #3975, opened 2020-04-06
remove text encoders from commands

### LINKED ISSUE #287: Define standard default output format
### Description

From discussion in #274, @phritz wrote: 'Could you please propose a standard default output format in the contributing guide, maybe informed by a short convo with why? I think we want something that is not json but is pretty printy.'

I'm creating this issue as  TODO for myself so I don't lose track when #274 merges.

I discussed this with @whyrusleeping. Tentative conclusions:

+ This has to be excellent and work correctly eventually, but is not yet top priority.
+ Getting this right is tricky.
+ It would be best to follow IPFS if possible.

@laser's comment from #274 may therefore be relevant:

> It is my belief that we want to do things like IPFS does as a general rule. Some go-ipfs commands have a notion of "format", which allows the caller to provide a string containing some tokens used to create command output. See [here](https://github.com/ipfs/go-ipfs/blob/11ce0445cb24b6b99613223fc6d8128c6648a4df/core/commands/files.go#L103) and [here](https://github.com/ipfs/go-ipfs/blob/11ce0445cb24b6b99613223fc6d8128c6648a4df/core/commands/files.go#L173) for an example.

### Proposal

+ For now, just don't implement text-encoders at all. Better to get this right than insert a stopgap. There's always JSON available from the CLI.
+ For each missing text encoder, spawn a new issue to track the deficit.
+ When implementing the first of these, just focus on defining a standard (non-customizable) format with emphasis on human-readability.
+ Prefer defining standard textual outputs for each necessary type independently and composing these representations -- as opposed to crafting entirely novel representations for each composite data-type or result.
+ Do include appropriate human-readable scaffolding as necessary for readability, on a case-by-case basis. (@whyrusleeping is this right?)
+ Define such textual renderers such that they can be reused between commands' text encoders.
+ Once this pattern has been established, when completing the issue associated with a command whose output warrants it, implement a text-substitution based approach (like shown in the links above) to allow for user-supplied formatting.

It follows that we should defer working on the most complicated renderers until some groundwork has been covered in the simpler ones.

While this description completely fails to fully define a standard default output format, it does provide a default strategy (do nothing), and an incremental approach to designing a correct and general solution.

For this to work, we should to some extent anticipate and allow for 'design in review' when the relevant PRs come up. This is because getting the details right is tricky and can probably only be discussed well in the presence of concrete examples.

### LINKED ISSUE #599: there's a funky smell and it's coming from command output encoding
OK I'm sorry / not sorry to have to hand this off as I go on vacation. As part of https://github.com/filecoin-project/go-filecoin/issues/587 which you can see in https://github.com/filecoin-project/go-filecoin/pull/598 I uncovered some funky smells in our output encoding. See in particular https://github.com/filecoin-project/go-filecoin/pull/598/files#diff-e77b7ad9d6713425351aae9703d0d36bR16

First, take note of https://github.com/filecoin-project/go-filecoin/issues/266 which is almost certainly related to these problems. The issues specifically are:

1. Block.Miner doesn't make it through the nutty series of encodings in dag_daemon_test. The only reason this test passes is that the genesis block doesn't have a miner address set. Anyway, this call returns an error that is presently ignored, if you capture and assert no error on it you can see it: https://github.com/filecoin-project/go-filecoin/blob/master/commands/dag_daemon_test.go#L45

2. This test also embodies the following pattern. It gets a block two ways and then checks that they are equal. It gets the block first via `Block => json => Block`: it calls chain ls which json encodes and outputs a block from memory. Then it dag gets the cid from that block which results in `Block => IPLD node => CBOR => IPLD node => json => IPLD node => Block`. Walking you through that: the block gets turned into an ipld node which is written as cbor in the block store (`Block => IPLD node => CBOR`) and then dag get CBOR decodes it into an IPLD node and json encodes it and outputs it (`Block => IPLD node => CBOR => IPLD node => json`). Then the test json decodes back to the ipld node (`Block => IPLD node => CBOR => IPLD node => json => IPLD node`) and then decodes the ipld node as a block (bringing us all the way to `Block => IPLD node => CBOR => IPLD node => json => IPLD node => Block`). There was an issue with this test when implemented Uint64 so I wanted to write a test in block_test that specifically checked whether the block survived this sequence of transformations. I wrote the test, but it fails with the same error as above. However! The dag_daemon_test which does the should-be-logically-identical transformation passes. So, that's mystery number 2, possibly related to mystery 1.

3. I mentioned we json encode Uint64's to strings. We *have* to do this because of the way that dag_daemon_test's (old, outdated) cbor implementation works: https://github.com/ipfs/go-ipld-cbor/blob/master/node.go#L82. It uses json for a reason I don't understand, it was probably expedient. The ipld node contains bytes for Uint64's so unless we define custom json marshaling it will fail because without custom Uint64 marshaling it expects that Uint64's are marshaled as numbers (not bytes). It's a bummer to have to json encode Uint64's to a string (the LEB128 bytes are base64-encoded and js-quoted) because that makes the json output of Uint64's in commands opaque. Ick. 

4. I'd like to understand what we're shooting for output encoding wise. The sensible thing to me and i think(?) what we have been shooting for in general is to have exactly two types of output encoding for commands. 
  * human-readable output. This is a json encoding of the model object (eg, Block), meant for consumption by humans, *and should never be read by machines* because of that
  * machine-readable output. This should be something close to if not identical to the storage format, so call it raw CBOR. (I'm not sure why we json-encode it? seems like that's a transport concern).

Is that right? If so we need a pass over the code to make sure it is what we're doing.

OK, whew, popping back up. I think the place to start is with https://github.com/filecoin-project/go-filecoin/issues/266, now that it is clear that things are not well in this area of the code. Note that everything that is not dag_daemon_test uses this fork of go-ipld-cbor: https://github.com/filecoin-project/go-ipld-cbor. Once dag_daemon_test has been moved to use QmRiRJhn427YVuufBEHofLreKWNw7P7BWNq86Sb9kzqdbd like the rest of the codebase then tackling 1 and 2 above makes sense. Then once they are fixed or understood it would be amazing to look at 3, whether we can json marshal to a number (sees like we should since the new version of go-ipld-cbor uses refmt and not some hacky series of json marshals).

cc @whyrusleeping @laser @aboodman 

Edit: clarified 2 based on feedback from @travisperson in https://github.com/filecoin-project/go-filecoin/pull/598#pullrequestreview-131829176

acruikshank:
### Motivation

We want a JSON-only command API, so we need to remove all the existing text encoders.

### Proposed changes

1. Remove text encoders.
2. Fix tests that depend on them.

Closes #287, closes #599

### DIFF
--- a/cmd/go-filecoin/address_integration_test.go
+++ b/cmd/go-filecoin/address_integration_test.go
@@ -163,15 +163,13 @@ func TestWalletExportImportRoundTrip(t *testing.T) {
 	_, cmdClient, done := builder.BuildAndStartAPI(ctx)
 	defer done()
 
-	lsJSON := cmdClient.RunSuccess(ctx, "address", "ls").ReadStdoutTrimNewlines()
 	var lsResult commands.AddressLsResult
-	err := json.Unmarshal([]byte(lsJSON), &lsResult)
-	require.NoError(t, err)
+	cmdClient.RunMarshaledJSON(ctx, &lsResult, "address", "ls")
 	require.Len(t, lsResult.Addresses, 1)
 
-	exportJSON := cmdClient.RunSuccess(ctx, "wallet", "export", lsResult.Addresses[0].String(), "--enc=json").ReadStdout()
+	exportJSON := cmdClient.RunSuccess(ctx, "wallet", "export", lsResult.Addresses[0].String()).ReadStdout()
 	var exportResult commands.WalletSerializeResult
-	err = json.Unmarshal([]byte(exportJSON), &exportResult)
+	err := json.Unmarshal([]byte(exportJSON), &exportResult)
 	require.NoError(t, err)
 
 	wf, err := os.Create("walletFileTest")
@@ -184,8 +182,10 @@ func TestWalletExportImportRoundTrip(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, wf.Close())
 
-	maybeAddr := cmdClient.RunSuccess(ctx, "wallet", "import", wf.Name()).ReadStdoutTrimNewlines()
-	assert.Equal(t, lsJSON, maybeAddr)
+	var importResult commands.AddressLsResult
+	cmdClient.RunMarshaledJSON(ctx, &importResult, "wallet", "import", wf.Name())
+	assert.Len(t, importResult.Addresses, 1)
+	assert.Equal(t, lsResult.Addresses[0], importResult.Addresses[0])
 }
 
 // MustDecodeCid decodes a string to a Cid pointer, panicking on error
--- a/cmd/go-filecoin/address_integration_test.go
+++ b/cmd/go-filecoin/address_integration_test.go
@@ -2,9 +2,11 @@ package commands_test
 
 import (
 	"context"
+	"encoding/json"
 	"os"
-	"strings"
 	"testing"
+
+	commands "github.com/filecoin-project/go-filecoin/cmd/go-filecoin"
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/specs-actors/actors/builtin"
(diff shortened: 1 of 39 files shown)


cmd/go-filecoin/address_integration_test.go
```diff
@@ -161,41 +163,29 @@ func TestWalletExportImportRoundTrip(t *testing.T) {
 	_, cmdClient, done := builder.BuildAndStartAPI(ctx)
 	defer done()
 
-	dw := cmdClient.RunSuccess(ctx, "address", "ls").ReadStdoutTrimNewlines()
+	lsJSON := cmdClient.RunSuccess(ctx, "address", "ls").ReadStdoutTrimNewlines()
+	var lsResult commands.AddressLsResult
+	err := json.Unmarshal([]byte(lsJSON), &lsResult)
```
anorth, line 168: It's probably worth folding the JSON unmarshal into something like `RunSuccess(ctx, &lsResult, "address", "ls")`. It will be ubiquitous now.

cmd/go-filecoin/config_integration_test.go
```diff
@@ -74,7 +74,7 @@ func TestConfigDaemon(t *testing.T) {
 		jsonOut := op1.ReadStdout()
 		bootstrapConfig := config.NewDefaultConfig().Bootstrap
 		bootstrapConfig.Addresses = []string{"fake1", "fake2"}
-		someJSON, err := json.MarshalIndent(bootstrapConfig, "", "\t")
+		someJSON, err := json.Marshal(bootstrapConfig)
```
anorth, line 77: Nit: I would actually prefer that pretty JSON is the default, with a flag for compressed JSON. Can we acheive that easily?

cmd/go-filecoin/address_integration_test.go
acruikshank: I called it `RunMarshaledJSON`. RunSuccess is still convenient when you don't care about the result type.

cmd/go-filecoin/config_integration_test.go
acruikshank: Yes. Pretty easily. I added a new 'pretty-json' command encoder and made 'pretty-json' the default for the `--enc` option. `--enc=json` still works and outputs condensed json.

PR #3978, opened 2020-04-07
Update to latest go-fil-markets w/ enhanced Graphsync

hannahhoward:
### Motivation

Update graphsync, data-transfer, and go-fil-markets.
Enable features from latest graphsync.

### Background

Graphsync has some new enhancements:
- Passing extensions in protocol requests to enable custom behavior
- Hooks on requests and responses
- The ability to lock down graphsync so it rejects all incoming requests unless a hook validates the request based on an extension (or other request data)
- The ability to load from other block stores based on extensions

### Proposed changes

- Remove IPLD bridge from network initialization (IPLDBridge was a legacy concept that's been removed from graphsync)
- Turn on default request rejection for Graphsync when constructing graphsync instance in network submodule 
- In graphsync fetcher, define a ChainsyncProtocolExtension to Graphsync (just a name -- no data for now), pass this with all outgoing requests
- In syncer initialize add hook to graphsync instance to validate chainsync requests so they are accepted

### Out Of Scope

- loading from a different block store (need guidance on how you all want to do this)
- more rigorous request validation of chainsync requests (needs discussion among implementations)
Add Graphsync extension for chainsync to identify chainsync requests.

### For Discussion

Is the syncer submodule the right place to register the hook? Maybe it should live inside the fetcher itself?

### FYI

should be merged around the same time as https://github.com/filecoin-project/lotus/pull/1520 which locks down graphsync in Lotus


cumtlzhy:
Hi, how can I find my FIL?

PR #3979, opened 2020-04-08
restore miner create

acruikshank:
### Motivation

The miner create is the first real step towards creating functioning non-bootstrap miner. Unfortunately, it hasn't been working since the move to specs actors.

### Proposed changes

This PR resolves it's issues and adds an integration test. In the process, it introduces some new test infrastructure, `MustCreateNodesWithBootstrap`, that makes it simple to create a new in-process integration test that allows a node to send messages and receive mined blocks containing their receipts.

The bug fixes are:
1. The test needs to send sufficient gas.
2. CreateMiner is an operation on the power actor, but we were sending the message to the market actor.
3. The CreateMiner return type is now a struct containing the miner id address and the robust address instead of a single address.

I wanted to test that the miner's state actually appeared on chain, but the miner address stored in config is the robust address, not the miner id (for good reason). Actor addresses did not work in the state viewer. I added address resolution logic to fix this.

### DIFF
--- a/internal/app/go-filecoin/node/test/setup.go
+++ b/internal/app/go-filecoin/node/test/setup.go
@@ -46,20 +46,24 @@ func MustCreateNodesWithBootstrap(ctx context.Context, t *testing.T, additionalN
 
 	_, _, err := initNodeGenesisMiner(t, bootstrapMiner, seed, genCfg.Miners[0].Owner, presealPath, genCfg.Miners[0].SectorSize)
 	require.NoError(t, err)
-	bootstrapMiner.Start(ctx)
+	err = bootstrapMiner.Start(ctx)
+	require.NoError(t, err)
 
 	nodes[0] = bootstrapMiner
 
 	// create addtiional nodes
 	for i := uint(0); i < additionalNodes; i++ {
-		nodes[i+1] = NewNodeBuilder(t).
+		node := NewNodeBuilder(t).
 			WithGenesisInit(seed.GenesisInitFunc).
 			WithBuilderOpt(node.FakeProofVerifierBuilderOpts()...).
 			WithBuilderOpt(node.ChainClockConfigOption(chainClock)).
 			Build(ctx)
-		addr := seed.GiveKey(t, nodes[i+1], int(i+1))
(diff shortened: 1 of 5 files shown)


internal/app/go-filecoin/node/test/setup.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
+	gengen "github.com/filecoin-project/go-filecoin/tools/gengen/util"
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/stretchr/testify/require"
+)
+
+func MustCreateNodesWithBootstrap(ctx context.Context, t *testing.T, additionalNodes uint) ([]*node.Node, context.CancelFunc) {
```
anorth, line 20: Please add a brief comment of what this does for the caller

internal/app/go-filecoin/node/test/setup.go
```diff
@@ -0,0 +1,117 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	nodes := make([]*node.Node, 1+additionalNodes)
+
+	// set up paths and fake clock.
+	presealPath := project.Root("fixtures/genesis-sectors")
+	genCfgPath := project.Root("fixtures/setup.json")
+	genTime := int64(1000000000)
+	blockTime := 1 * time.Second
```
anorth, line 28: I'd prefer this to be 30s so it's clear that time is fake. I would guess we will later need to return the fake clock from this, for external manipulation.

internal/app/go-filecoin/node/test/setup.go
```diff
@@ -0,0 +1,117 @@
...
+	fakeClock := clock.NewFake(time.Unix(genTime, 0))
+
+	// Load genesis config fixture.
+	genCfg := loadGenesisConfig(t, genCfgPath)
+	genCfg.Miners = append(genCfg.Miners, &gengen.CreateStorageMinerConfig{
+		Owner:      1,
+		SectorSize: 2048,
```
anorth, line 35: constants.DevSectorSize

PR #3983, opened 2020-04-10
fix parent grinding fault and missing check, remove blockInChain, rem…

sternhenri:
…ove earliest

### Motivation

Working in parallel on https://github.com/filecoin-project/lotus/pull/1534, found a few issues in this code.

### Proposed changes

- remove need to check whether blocks are in miner's chain (they can be in any chain and the consensus fault is still valid)
- fix parent-grinding check (per https://github.com/filecoin-project/specs/pull/912) so it accounts for null blocks (both in ensuring A and C are siblings, and in ignoring any null blocks used to grind B)
- add check for extra so parent-grinding check doesn't error out.

### DIFF
--- a/internal/pkg/slashing/check.go
+++ b/internal/pkg/slashing/check.go
@@ -10,7 +10,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/chain"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/state"
 )
@@ -50,12 +49,7 @@ func (s *ConsensusFaultChecker) VerifyConsensusFault(ctx context.Context, h1, h2
 	if innerErr != nil {
 		return nil, errors.Wrapf(innerErr, "failed to decode h2")
 	}
-	if len(extra) > 0 {
-		innerErr = encoding.Decode(extra, &b3)
-		if innerErr != nil {
-			return nil, errors.Wrapf(innerErr, "failed to decode extra")
-		}
-	}
+
 	// Block syntax is not validated. This implements the strictest check possible, and is also the simplest check
 	// possible.
 	// This means that blocks that could never have been included in the chain (e.g. with an empty parent state)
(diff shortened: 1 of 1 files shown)


internal/pkg/slashing/check.go
```diff
@@ -92,21 +87,28 @@ func (s *ConsensusFaultChecker) VerifyConsensusFault(ctx context.Context, h1, h2
...
-			Type:   runtime.ConsensusFaultParentGrinding,
+	// The provable case is that two blocks are mined and the later one does not include the
+	// earlier one as a parent even though it could have.
+	// B3 must prove that the higher block (B2) could have been included in B1's tipset.
+	if len(extra) > 0 {
+		innerErr = encoding.Decode(extra, &b3)
+		if innerErr != nil {
+			return nil, errors.Wrapf(innerErr, "failed to decode extra")
+		}
+		if b1.Height == b3.Height && b3.Parents.Equals(b1.Parents) && !b2.Parents.Has(b1.Cid()) && b2.Parents.Has(b3.Cid()) {
```
anorth, line 98: it makes sense that something should change to account for null blocks, but `b1.Height == b3.Height` is a straight up double-fork mining fault, so doesn't seem right yet.

These changes seem to change the role of blocks, so I think I'll need a new diagram to understand it.

internal/pkg/slashing/check.go
```diff
@@ -116,10 +118,6 @@ func (s *ConsensusFaultChecker) VerifyConsensusFault(ctx context.Context, h1, h2
 	if err != nil {
 		return nil, err
 	}
-	err = verifyOneBlockInChain(ctx, s.chain, head, b1, b2, earliest)
-	if err != nil {
-		return nil, err
-	}
```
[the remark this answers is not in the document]
sternhenri: I'm happy to discuss it if you'd like. The main goal is that someone who provably commits such a fault should be slashed on every chain, not just which on that they try to pull this off...

internal/pkg/slashing/check.go
sternhenri: So it isn't a double-fork mining fault because b1 and b3 need not be from the same miner (if they were then I agree it is a double-fork mining fault). The case in question is shown below: the case where there is a null block mined between the parent block and the faulty block (whiich omitted the parent block). 

That is, in the below B2 should include B1 as a parent but does not. This is a provable fault. In the prior check by looking at `B2.height == b1.height + 1`, we would miss this. As is, we are ok with null blocks on either side (ancestor to B1, B3 or ahead of B2). `B3.height == b1.height && b3.Parents == b1.Parents`` are the checks to prove that they should have been in the same tipset.

```
             ┌ ─ ─ ─ ─      ┌────────┐      ┌ ─ ─ ─ ─       ┌────────┐
                      │     │B3,     │               │      │B2,     │
             │              │MinerC  │      │               │MinerA  │
     ┌──────▶         │────▶│        │─────▶         │─────▶│        │
     │       │              │        │      │               │        │
┌────────┐    ─ ─ ─ ─ ┘     └────────┘       ─ ─ ─ ─ ┘      └────────┘
│B0,     │
│MinerAny│
│        │
│        │
└────────┘   ┌ ─ ─ ─ ─      ┌────────┐
     │                │     │B1,     │
     │       │              │MinerA  │
     └──────▶         │─────▶        │
             │              │        │
              ─ ─ ─ ─ ┘     └────────┘
```

anorth: Very helpful, thank you.

What about if there's no null before B3, but there still is before B1, so B1.Height <= B3.Height, but still B2.Height > B1.Height? Given the parent linkages, maybe the only height check we need is B2 > B1?

sternhenri: It's a very good question. Here's a try:

I think the check does have to include `B1.height == B3.height` in that this guarantees they could have been in the same tipset. 

In the case where `B1.height < B3.height` (so take B1 is a child of B0 and B3 has a null block as in the diagram), you could imagine that the miner A puts out B1, then hears about a new Tipset that includes B3 (and is heavier than the one off B1), and so logically chooses to miner atop that, yielding B2. Valid

Conversely, if  `B1.height > B3.height` (so the reverse, where B3 is a direct child of B0, B1 isn't), you could imagine that MinerA simply didn't hear about B3's tipset in time and so mined B1, before receiving B3's tipset and mining B2 atop it.
With a very strict definition of network synchrony (that assumes that max delay is less than a block time), that would be a provable fault, but I think it's a dangerous assumption (yields attacks where I can delay someone's block reception and then slash them).

So I don't think `B2 > B1` suffices for provable deviations. What do you think of the above reasoning?

anorth: SGTM.

anorth:
@sternhenri were you waiting for me to land this?

sternhenri:
Yeah, sorry I assumed you'd land it. Thanks for doing so.

PR #3985, opened 2020-04-14
client propose-deal

acruikshank:
### Motivation

Although the storage market has been integrated into go-filecoin, it has not been fully tested because the CLI commands have never been connected and the necessary chain interactions have only recently been completed. This PR restores the `client propose-deal` and `client query` commands and fixes multiple minor issues with the integration so that storage miners and storage clients can now make deals.

### Proposed changes

1. Restore the `client propose-deal` and `client query` commands.
2. Introduce a new integration test to test this functionality. This is an in-process test that tests a deal proposal from first contact, through multiple actor interactions up to the point where the client sees the miner has accepted the deal. This test actually works through staging and sealing until the deal is active, but the sealing part of it isn't appropriate for an integration test.
3. I reworked the storage submodule so that it can be mostly initialized when the node is built. Only its storage provider component is constructed when mining is started. This allows the node to act as a storage client without needing to start mining.
4. Fix signature encodings in the client to match the expectations in go-fil-market.
5. Implement the unimplemented `OnDealSectorCommitted`
6. Adjust gas limits and resolve addresses where needed.

### DIFF
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -98,14 +98,9 @@ var ClientProposeStorageDealCmd = &cmds.Command{
 		Tagline:          "Propose a storage deal with a storage miner",
 		ShortDescription: `Sends a storage deal proposal to a miner`,
 		LongDescription: `
-Send a storage deal proposal to a miner. IDs provided to this command should
-represent valid asks. Existing asks can be listed with the following command:
+Send a storage deal proposal to a miner.
 
-$ go-filecoin client list-asks
-
-See the miner command help text for more information on asks.
-
-Duration should be specified with the number of blocks for which to store the
+Start and end should be specified with the number of blocks for which to store the
 data. New blocks are generated about every 30 seconds, so the time given should
 be represented as a count of 30 second intervals. For example, 1 minute would
 be 2, 1 hour would be 120, and 1 day would be 2880.
@@ -116,7 +111,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 		cmdkit.StringArg("data", true, false, "CID of the data to be stored"),
 		cmdkit.StringArg("start", true, false, "Chain epoch at which deal should start"),
 		cmdkit.StringArg("end", true, false, "Chain epoch at which deal should end"),
-		cmdkit.StringArg("price", true, false, "Price of deal in FIL (e.g. 0.01)"),
+		cmdkit.StringArg("price", true, false, "Storage price per epoch of all data in FIL (e.g. 0.01)"),
 		cmdkit.StringArg("collateral", true, false, "Collateral of deal in FIL (e.g. 0.01)"),
 	},
 	Options: []cmdkit.Option{
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-fil-markets/storagemarket"
-	"github.com/filecoin-project/go-fil-markets/storagemarket/network"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/constants"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 	"github.com/filecoin-project/specs-actors/actors/abi"
@@ -200,7 +199,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 
 		return re.Emit(resp)
 	},
-	Type: network.Response{},
+	Type: storagemarket.ProposeStorageDealResult{},
 }
 
 var ClientQueryStorageDealCmd = &cmds.Command{
@@ -228,7 +227,7 @@ format is specified with the --enc flag.
 
 		return re.Emit(deal)
 	},
-	Type: network.SignedResponse{},
+	Type: storagemarket.ClientDeal{},
 }
 
 // VerifyStorageDealResult wraps the success in an interface type
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -27,7 +27,7 @@ var clientCmd = &cmds.Command{
 		"cat":                  clientCatCmd,
 		"import":               clientImportDataCmd,
 		"propose-storage-deal": ClientProposeStorageDealCmd,
-		"query-storage-deal":   clientQueryStorageDealCmd,
+		"query-storage-deal":   ClientQueryStorageDealCmd,
 		"verify-storage-deal":  clientVerifyStorageDealCmd,
 		"list-asks":            clientListAsksCmd,
 	},
@@ -208,7 +208,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 	Type: network.Response{},
 }
 
-var clientQueryStorageDealCmd = &cmds.Command{
+var ClientQueryStorageDealCmd = &cmds.Command{
 	Helptext: cmdkit.HelpText{
 		Tagline: "Query a storage deal's status",
 		ShortDescription: `
@@ -221,19 +221,17 @@ format is specified with the --enc flag.
 		cmdkit.StringArg("id", true, false, "CID of deal to query"),
 	},
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		panic("not implemented pending full storage market integration")
+		dealCID, err := cid.Decode(req.Arguments[0])
+		if err != nil {
+			return errors.Wrap(err, "could not decode deal cid")
+		}
 
-		//propcid, err := cid.Decode(req.Arguments[0])
-		//if err != nil {
-		//	return err
-		//}
-		//
-		//resp, err := GetStorageAPI(env).QueryStorageDeal(req.Context, propcid)
-		//if err != nil {
-		//	return err
-		//}
-		//
-		//return re.Emit(resp)
+		deal, err := GetStorageAPI(env).GetStorageDeal(req.Context, dealCID)
+		if err != nil {
+			return err
+		}
+
+		return re.Emit(deal)
 	},
 	Type: network.SignedResponse{},
 }
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -16,6 +16,7 @@ import (
 	cmdkit "github.com/ipfs/go-ipfs-cmdkit"
 	cmds "github.com/ipfs/go-ipfs-cmds"
 	files "github.com/ipfs/go-ipfs-files"
+	p2pcore "github.com/libp2p/go-libp2p-core/peer"
 )
 
 var clientCmd = &cmds.Command{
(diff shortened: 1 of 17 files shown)


cmd/go-filecoin/client.go
```diff
@@ -107,48 +114,101 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 	Arguments: []cmdkit.Argument{
 		cmdkit.StringArg("miner", true, false, "Address of miner to send storage proposal"),
 		cmdkit.StringArg("data", true, false, "CID of the data to be stored"),
-		cmdkit.StringArg("ask", true, false, "ID of ask for which to propose a deal"),
-		cmdkit.StringArg("duration", true, false, "Time in blocks (about 30 seconds per block) to store data"),
+		cmdkit.StringArg("start", true, false, "Chain epoch at which deal should start"),
+		cmdkit.StringArg("end", true, false, "Chain epoch at which deal should end"),
+		cmdkit.StringArg("price", true, false, "Price of deal in FIL (e.g. 0.01)"),
```
anorth, line 119: It would be good to clarify whether this is total, per-byte-epoch, or what other units.

cmd/go-filecoin/client.go
```diff
@@ -107,48 +114,101 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
...
+			req.Context,
+			addr,
+			providerInfo,
+			data,
+			abi.ChainEpoch(start),
+			abi.ChainEpoch(end),
+			price,
+			collateral,
+			// proof version (not circuit or size) should be all that is important here
+			constants.DevRegisteredPoStProof,
```
anorth, line 200: I am very surprised this is here at all. We should see what we can do to get rid of it (later). cc @hannahhoward

cmd/go-filecoin/daemon.go
```diff
@@ -216,18 +216,13 @@ func RunAPIAndWait(ctx context.Context, nd *node.Node, config *config.APIConfig,
...
-
 	return &Env{
 		blockMiningAPI: nd.BlockMining.BlockMiningAPI,
 		drandAPI:       nd.DrandAPI,
 		ctx:            ctx,
 		inspectorAPI:   NewInspectorAPI(nd.Repo),
 		porcelainAPI:   nd.PorcelainAPI,
 		retrievalAPI:   nd.RetrievalProtocol,
-		storageAPI:     storageAPI,
+		storageAPI:     storage.NewAPI(nd.StorageProtocol),
```
anorth, line 226: Thanks. Now that you've improved this we should just init and expose the StorageAPI on the node, like the other.s

internal/app/go-filecoin/connectors/storage_market/client.go
```diff
@@ -191,7 +201,7 @@ func (s *StorageClientNodeConnector) ValidatePublishedDeal(ctx context.Context,
 
 // SignProposal uses the local wallet to sign the given proposal
 func (s *StorageClientNodeConnector) SignProposal(ctx context.Context, signer address.Address, proposal market.DealProposal) (*market.ClientDealProposal, error) {
-	buf, err := encoding.Encode(proposal)
+	buf, err := cborutil.Dump(&proposal)
```
anorth, line 204: Why this change? The encoding library should work here (tho passing a pointer looks good)? If it doesn't, we have a deeper problem that should not be skimmed over.

internal/app/go-filecoin/connectors/storage_market/client.go
```diff
@@ -229,5 +239,50 @@ func (s *StorageClientNodeConnector) GetChainHead(ctx context.Context) (shared.T
...
+		return err
+	}
+
+	err = s.waiter.WaitPredicate(ctx, func(msg *types.SignedMessage, c cid.Cid) bool {
+		if msg.Message.To != provider {
+			return false
+		}
+
+		// prove commit is method 8
+		if msg.Message.Method != abi.MethodNum(8) {
```
anorth, line 260: Blocking: Eeek, this is bound to rot. Use the constant defined in `builtin`.

internal/app/go-filecoin/connectors/storage_market/client.go
```diff
@@ -229,5 +239,50 @@ func (s *StorageClientNodeConnector) GetChainHead(ctx context.Context) (shared.T
...
+	}
+
+	resolvedProvider, err := view.InitResolveAddress(ctx, provider)
+	if err != nil {
+		cb(err)
+		return err
+	}
+
+	err = s.waiter.WaitPredicate(ctx, func(msg *types.SignedMessage, c cid.Cid) bool {
+		if msg.Message.To != provider {
```
anorth, line 255: This doesn't do as you'd hope. The on-chain message could have either the resolved or the unresolved address. You need to check both. Or don't at all.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -51,40 +87,45 @@ func NewStorageProtocolSubmodule(
...
+	sm.StorageProvider, err = impl.NewProvider(smnetwork.NewFromLibp2pHost(h), ds, bs, fs, piecestore.NewPieceStore(ds), dt, pnode, minerAddr, sealProofType)
+	return err
+}
 
-	return &StorageProtocolSubmodule{
-		StorageClient:   client,
-		StorageProvider: provider,
-	}, nil
+func (sm *StorageProtocolSubmodule) Provider() iface.StorageProvider {
+	return sm.StorageProvider
```
anorth, line 122: Since this may be nil, suggest returning an error or ok boolean to alert the caller.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -51,40 +87,45 @@ func NewStorageProtocolSubmodule(
...
+}
+
+func (sm *StorageProtocolSubmodule) Client() iface.StorageClient {
+	return sm.StorageClient
+}
+
+func (sm *StorageProtocolSubmodule) PieceManager() piecemanager.PieceManager {
+	return sm.pieceManager
```
anorth, line 130: Ditto.

Long term, perhaps we should re-arrange this into two modules.

internal/app/go-filecoin/node/node.go
```diff
@@ -405,6 +405,19 @@ func (node *Node) SetupMining(ctx context.Context) error {
...
+	}
+
+	if err := node.StorageProtocol.StorageProvider.Start(ctx); err != nil {
+		fmt.Printf("error starting storage provider: %s\n", err)
+	}
+
+	// TODO: Retrieval Market Integration
+	//if err := node.RetrievalProtocol.RetrievalProvider.Start(); err != nil {
+	//	fmt.Printf("error starting retrieval provider: %s\n", err)
+	//}
```
anorth, line 419: @shannonwells is there still something missing here?

internal/app/go-filecoin/node/test/setup.go
```diff
@@ -99,6 +99,22 @@ func MustCreateNodesWithBootstrap(ctx context.Context, t *testing.T, additionalN
...
+	emitter := &testEmitter{}
+	err = cmd.Run(req, emitter, commands.CreateServerEnv(ctx, nd))
+	if err != nil {
+		return nil, err
+	}
+	if emitter.err != nil {
+		return nil, err
+	}
+	return emitter.value, nil
+}
```
anorth, line 116: Interesting, this is an alternative way to run the command. I think all-up I would prefer to stick to the existing way via `RunNodeAPI` (which is still in-process, and maybe you didn't notice). 
- exercises more code, including the CLI de/serialization
- this test emitter doesn't handle mutliple emissions

See protocol_cmd_test. Then un-export the commands.

If there's some reason you need to leave it, please move to a different file from `setup`, maybe 'direct_cmds' or something, so it's more obvious.

internal/app/go-filecoin/connectors/storage_market/client.go
```diff
@@ -229,5 +239,50 @@ func (s *StorageClientNodeConnector) GetChainHead(ctx context.Context) (shared.T
...
+			return false
+		}
+
+		// that's enough for us to check chain state
+		view, err = s.chainStore.StateView(s.chainStore.Head())
+		if err != nil {
+			return false
+		}
+
+		dealIDs, err := view.MarketGetDealIDs(ctx, resolvedProvider)
```
anorth, line 270: Nearly-blocking: This is going to be a really big collection. Since all you want it for is to look for a value, I suggest pushing that down into the state view as a `MarketGetDeal()` so it can iterate without instantiating the whole collection.

internal/pkg/state/view.go
```diff
@@ -198,6 +198,24 @@ func (v *View) MarketEscrowBalance(ctx context.Context, addr addr.Address) (foun
...
+	byParty := market.AsSetMultimap(StoreFromCbor(ctx, v.ipldStore), marketState.DealIDsByParty)
+	var providerDealIds []abi.DealID
+	if err = byParty.ForEach(addr, func(i abi.DealID) error {
+		providerDealIds = append(providerDealIds, i)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return providerDealIds, err
+}
```
anorth, line 217: Giant. I would prefer not to expose this so callers don't think it's reasonable. You could expose a ForEach instead, or just the particular behaviour of finding one deal that motivated this.

internal/app/go-filecoin/node/node.go
shannonwells: No, that comment is left from much earlier TODOs and can be removed. There is no "Start," only setup.

internal/app/go-filecoin/connectors/storage_market/client.go
acruikshank: It was unnecessary. The pointer was the fix. I switched it back.

internal/pkg/state/view.go
acruikshank: It's MarketHasDealID now and does the traversal internally.

cmd/go-filecoin/daemon.go
acruikshank: done.

internal/app/go-filecoin/connectors/storage_market/client.go
acruikshank: It now resolves both addresses

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: Now returns an error if uninitialized.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
acruikshank: now returns an error if uninitialized

internal/app/go-filecoin/node/node.go
acruikshank: deleted

internal/app/go-filecoin/connectors/storage_market/client.go
acruikshank: Replaced with MarketHasDealID

internal/app/go-filecoin/node/test/setup.go
acruikshank: I removed this and replaced with RunNodeAPI (which I was unaware of).

internal/app/go-filecoin/connectors/storage_market/client.go
acruikshank: replaced with builtin constant.

PR #3986, opened 2020-04-14
Sync gfc drand integration with latest drand changes.

acruikshank:
### Motivation

We need to have full drand integration working soon, so it's important that our drand network client stays up-to-date with changes that are happening on the drand side. This PR updates our drand dependency and fixes the integration to work with the new api.

### Proposed changes

1. Update drand dependency.
2. Change distkey from hex string slice to double byte slice.
3. Modify verification to take parent round as part of the message.
4. Modify (still skipped) functional test to use new test drand network.
5. Extend test a bit to cover verification.


functional-tests/drand_test.go
```diff
@@ -42,16 +43,20 @@ func TestDrandPublic(t *testing.T) {
 	defer nd.Stop(ctx)
 
 	err = nd.DrandAPI.Configure([]string{
-		"localhost:8080",
-		"localhost:8081",
-		"localhost:8082",
-		"localhost:8083",
-		"localhost:8084"}, false, true)
+		"drand-test3.nikkolasg.xyz:5003",
```
ZenGround0, line 46: 🎉

internal/pkg/protocol/drand/drand_api.go
```diff
@@ -81,3 +81,8 @@ func (api *API) Configure(addrs []string, secure bool, overrideGroupAddrs bool)
 func (api *API) GetEntry(ctx context.Context, round drand.Round) (*drand.Entry, error) {
 	return api.drand.ReadEntry(ctx, round)
 }
+
+// GetEntry retrieves a entry from the drand server
```
ZenGround0, line 85: nit: fix comment

PR #3993, opened 2020-04-15
Power Table handles consensus faults

### LINKED ISSUE #3976: DRAND -- Power table and sector set read with lookback parameter
### Description
The sector set used to mine and validate election (soon to be winning) post and the miner and total power calculations must be read with respect to a past stateroot, not the parent.

In light of other upcoming DRAND changes and future refactors this will involve adding a chain state accessor to consensus and mining modules.  We then need to extract a method from the sampler and make it public so that we can access the lookback block for computing the power table.

ZenGround0:
### Motivation
Power calculations need to handle the case where consensus faults have slashed a miner in between the current head and power table lookback.

### Proposed changes
Power table now takes in two different state view and checks that a miner exists in the new `faultsState` before returning the power claim in the power state.  

Closes #3976

### DIFF
--- a/internal/pkg/chainsync/internal/syncer/syncer_integration_test.go
+++ b/internal/pkg/chainsync/internal/syncer/syncer_integration_test.go
@@ -6,11 +6,7 @@ import (
 	"time"
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/specs-actors/actors/abi"
-	fbig "github.com/filecoin-project/specs-actors/actors/abi/big"
-	"github.com/ipfs/go-cid"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
-	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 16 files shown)


internal/pkg/chainsync/internal/syncer/syncer_integration_test.go
```diff
@@ -120,193 +112,6 @@ func TestLoadFork(t *testing.T) {
 	assert.NoError(t, offlineSyncer.HandleNewTipSet(ctx, block.NewChainInfo("", "", left.Key(), heightFromTip(t, left)), false))
 }
 
-// Power table weight comparisons impact syncer's selection.
-// One fork has more blocks but less total power.
-// Verify that the heavier fork is the one with more power.
-// All blocks in this test follow protocol version 1 upgrade weighting rules.
-func TestSyncerWeighsPower(t *testing.T) {
```
ZenGround0, line 127: While modifying the state view this broke.  It would have been simple to fix but this test has been skipped for a long time and reflecting on its purpose leads me to believe that we are better deleting it.

internal/pkg/consensus/power_table_view_test.go
```diff
@@ -24,62 +25,114 @@ import (
 
 func TestTotal(t *testing.T) {
 	tf.UnitTest(t)
-	t.Skip("requires new init actor address resolution")
```
acruikshank, line 27: 🎉

PR #3994, opened 2020-04-15
replace go-sectorbuilder with v24 storage-fsm and sectors-storage

laser:
This PR was go-authored by @laser and @acruikshank.

### Motivation

The go-sectorbuilder and go-storage-miner projects have been deprecated and replaced with storage-fsm, sector-storage, and specs-storage. V25 proofs are built into the latest versions of _those_ projects (not go-sectorbuilder), so in order to get those new, fancy proofs, we must make the switch.

### Proposed changes

This changeset removes go-filecoin's dependency on go-storage-miner and sector-storage, replacing with storage-fsm, sector-storage, and specs-storage.

### Things to note

1. The storage-fsm project relies on `github.com/libp2p/go-libp2p-core` version 0.5.0, but go-filecoin's stack uses 0.3.0. To work around this, we pinned (via a go.mod `replace`) to 0.3.0. Somebody (@ZenGround0 @acruikshank - who should this be?) should remove that `replace` and update go-filecoin to 0.5.0 soon.
1. The sector-storage mechanism for initializing a node with pre-sealed sectors involves mounting a directory containing those sectors (and their metadata) whereas the go-sectorbuilder _copied_ the contents of a pre-sealed sector directory (or created a symlink).
1. The directory containing pre-sealed sectors is expected to contain a metadata file (which communicates things like read-only flag). 
1. The `setup.json` file contains pre-sealed sector size. The node `init` command no longer accepts `--presealed-sector-size` (because the size is baked into `setup.json`).


cmd/go-filecoin/chain_integration_test.go
```diff
@@ -35,8 +35,8 @@ func TestChainHead(t *testing.T) {
 
 func TestChainLs(t *testing.T) {
 	tf.IntegrationTest(t)
-	ctx := context.Background()
 	t.Skip("DRAGONS: fake post for integration test")
+	ctx := context.Background()
```
acruikshank, line 39: Lint didn't like this context being before the skip

internal/app/go-filecoin/plumbing/dag/dag_test.go
```diff
@@ -33,7 +33,7 @@ func TestDAGGet(t *testing.T) {
 		dag := NewDAG(dserv)
 
 		_, err := dag.GetNode(ctx, "awful")
-		assert.EqualError(t, err, "invalid 'ipfs ref' path")
+		assert.EqualError(t, err, "invalid path \"awful\": selected encoding not supported")
```
acruikshank, line 36: This appears to be a change in error handling due to a dag upgrade.

internal/pkg/testhelpers/iptbtester/wrapper.go
```diff
@@ -114,10 +117,25 @@ func (tn *TestNode) MustInit(ctx context.Context, args ...string) *TestNode {
...
 	// Did IPTB fail to function correctly?
 	if err != nil {
+		stdout := new(bytes.Buffer)
+		_, err := stdout.ReadFrom(out.Stdout())
+		require.NoError(tn.T, err)
+		tn.T.Logf("STDOUT:\n%s\n\n", stdout.String())
+		stderr := new(bytes.Buffer)
+		_, err = stderr.ReadFrom(out.Stderr())
+		require.NoError(tn.T, err)
+		tn.T.Logf("STDERR:\n%s\n\n", stderr.String())
```
acruikshank, line 127: This is here to make debugging `init` reasonable.

internal/pkg/testhelpers/net.go
```diff
@@ -86,7 +87,7 @@ func (fd *FakeDialer) StopNotify(inet.Notifiee)                             { pa
 
 // fakeStream is a test inet.Stream
 type fakeStream struct {
-	_   smux.Stream
+	_   mux.MuxedStream
```
acruikshank, line 90: smux.Stream was deprecated.

cmd/go-filecoin/init_daemon_test.go
```diff
@@ -63,26 +60,3 @@ func TestDownloadGenesis(t *testing.T) {
 
 	td.ShutdownSuccess()
 }
-
-func TestImportPresealedSectors(t *testing.T) {
```
acruikshank, line 67: This test causes CI to hang for some reason. It's a pretty poor test of presealed sectors anyway. The functionality is covered in functional tests like `TestBootstrapMineOnce`.

internal/app/go-filecoin/node/node.go
```diff
@@ -693,6 +664,11 @@ func (node *Node) CreateMiningWorker(ctx context.Context) (*mining.DefaultWorker
 		return nil, err
 	}
 
+	poster := node.BlockMining.PoStGenerator
+	if poster == nil {
+		poster = node.StorageMining.PoStGenerator
+	}
```
acruikshank, line 670: The PoStGenerator in BlockMining sometimes gets set to fake proofs, but is not always defined in test code. The PoStGenerator in StorageMining is always defined with real proofs. The worker panics with an undefined poster, but we want it to use fake proofs if we're testing with fake proofs.

This solution is suboptimal, but I think a better solution would require re-working the BlockMining and StorageMining submodule.

PR #3995, opened 2020-04-16
Fix/production drand post rebase

ZenGround0:
### Motivation
Get drand branch running after rebasing new changes from master.  Align production drand component with learnings from drand integration

### Proposed changes
- Drand GRPC component implements needed interface functions
- Basic caching added
- More configuration added and grabbed from drand group file
- Fix tests to use fake drand
Closes #

### DIFF
--- a/internal/pkg/drand/drand_grpc.go
+++ b/internal/pkg/drand/drand_grpc.go
@@ -222,6 +222,9 @@ func (d *GRPC) RoundsInInterval(ctx context.Context, startTime, endTime time.Tim
 		// wait on network to determine which rounds exist.  If maxRound is
 		// skipped this will error or block indefinitely.
 		next, err = d.ReadEntry(ctx, maxRound)
+		if err != nil {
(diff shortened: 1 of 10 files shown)


internal/pkg/drand/drand_grpc.go
```diff
@@ -59,17 +82,30 @@ func (d *GRPC) ReadEntry(ctx context.Context, drandRound Round) (*Entry, error)
...
 }
 
+func (d *GRPC) updateLocalState(entry *Entry) {
+	if d.latestEntry == nil {
+		d.latestEntry = entry
+	}
+	if entry.Round > d.latestEntry.Round {
+		d.latestEntry = entry
+	}
+	d.cache[entry.Round] = entry
```
acruikshank, line 106: This leaks. An LRU read-through cache should be really effective here. Maybe not the best time to deal with it, but an issue and TODO wouuld be good.

PR #4004, opened 2020-04-21
Make the block structure compatible with Lotus

arajasek:
The first commit simply removes the (now unused) ElectionPoSt field.

The second commit does the following:

- Puts the `ElectionProof` bytes in a struct (this is what Lotus does, and is preferred for future changes)
- Simplifies the signature field in Drand entries to just be a byte array (since I believe Drand is using a different version of BLS than "everything else", and so it isn't meaningfully associated with type BLS anyway).

With this, Lotus and go-filecoin agree on the serialisation of the new block headers.

### DIFF
--- a/internal/pkg/block/block.go
+++ b/internal/pkg/block/block.go
@@ -31,7 +31,7 @@ type Block struct {
 	Ticket Ticket `json:"ticket"`
 
 	// ElectionProof is the vrf proof giving this block's miner authoring rights
-	ElectionProof crypto.VRFPi
+	ElectionProof crypto.ElectionProof
 
 	// DrandEntries contain the verifiable oracle randomness used to elect
 	// this block's author leader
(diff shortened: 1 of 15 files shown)


internal/pkg/block/block.go
```diff
@@ -31,10 +31,7 @@ type Block struct {
 	Ticket Ticket `json:"ticket"`
 
 	// ElectionProof is the vrf proof giving this block's miner authoring rights
-	ElectionProof crypto.VRFPi
-
-	// EPoStInfo wraps all data for verifying this block's Election PoSt
-	EPoStInfo EPoStInfo `json:"ePoStInfo"`
```
ZenGround0, line 37: This LGTM but don't we need a winning PoSts field to fully implement the protocol?  No need to add here as I'm working on this in another PR, but curious how we can interop with lotus blocks without post proofs.

PR #4007, opened 2020-04-21
Update Ticket creation and validation 

ZenGround0:
### Motivation
The latest edition of the filecoin protocol specifies that ticket creation should sample drand entries and resample a previous ticket in the case that there are no new drand entries in a block

```
If newPeriod == false:
   RandomInput = MinerID || 
   ChainEpoch||ElectionSeed||GetRandomTicketFromTipset(ChainEpoch-1)
If newPeriod == true:
   RandomInput = MinerID || ChainEpoch||ElectionSeed
```

### Proposed changes
Move to the above by mixing old tickets in with miner id as part of entropy bytes.


internal/pkg/consensus/election.go
```diff
@@ -168,20 +169,40 @@ func (tm TicketMachine) MakeTicket(ctx context.Context, base block.TipSetKey, ep
...
+	_, err = entropyBuf.Write(minerEntropy)
+	if err != nil {
+		return nil, err
+	}
+	if !newPeriod { // resample previous ticket and add to entropy
+		ticketSeed, err := tm.sampler.Sample(ctx, base, epoch)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to sample previous ticket")
+		}
+		_, err = entropyBuf.Write(ticketSeed)
```
ZenGround0, line 197: The way that I'm doing this (putting ticket bytes in entropy buffer) doesn't require moving from `SampleChainRandomness` to `Sample`.  Since the closest thing we have to a spec specifies a different ordering of bytes and speccing is not very strong I am very suspicious that the arrangement of bytes will change during interop testing so its possible this gain in flexibility will help us.

I'm fining moving back to `SampleChainRandomness` if reviewer has a strong preference.

PR #4013, opened 2020-04-22
Winning post mined and validated

ZenGround0:
### Motivation
Winning PoSt

### Proposed changes
This finishes winning post integration by mining the post into a block and verifying that it is correct during validation.

Closes #


internal/app/go-filecoin/node/node.go
```diff
@@ -416,13 +416,13 @@ func (node *Node) SetupMining(ctx context.Context) error {
 func registeredProofsFromSectorSize(ss abi.SectorSize) (registeredSealProof abi.RegisteredProof, registeredPoStProof abi.RegisteredProof, err error) {
 	switch ss {
 	case constants.DevSectorSize:
-		return constants.DevRegisteredPoStProof, constants.DevRegisteredSealProof, nil
+		return constants.DevRegisteredWindowPoStProof, constants.DevRegisteredSealProof, nil
```
anorth, line 419: This looks like it was already wrong, but the order of values here doesn't match the names of the out parameters. Please check the call sites and fix up one way or the other.

internal/pkg/consensus/election.go
```diff
@@ -62,12 +62,22 @@ func (em ElectionMachine) GenerateWinningPoSt(ctx context.Context, allSectorInfo
...
-	challengeIndexes, err := ffiwrapper.ProofVerifier.GenerateWinningPoStSectorChallenge(ctx, allSectorInfos[0].RegisteredProof, minerID, poStRandomness, uint64(len(allSectorInfos)))
+	rp, err := allSectorInfos[0].RegisteredProof.RegisteredWinningPoStProof()
+	if err != nil {
+		return nil, err
+	}
+	challengeIndexes, err := ffiwrapper.ProofVerifier.GenerateWinningPoStSectorChallenge(ctx, rp, minerID, poStRandomness, uint64(len(allSectorInfos)))
 	if err != nil {
 		return nil, err
 	}
 	challengedSectorInfos := filterSectorInfosByIndex(allSectorInfos, challengeIndexes)
```
anorth, line 73: A pretty important scaling follow-up is to remove the allSectorInfos parameter and load sector info only for the very few sectors actually challenged. Please file an issue.

ZenGround0: Thank you so much for flagging this: #4018

PR #4024, opened 2020-04-25
Deterministic BLS key generation

### LINKED ISSUE #3781: Non-deterministic genesis state caused by BLS keys

anorth:
### Motivation
Determinism by default, for stability of testing. Randomness optional, most specifically for the wallet backend.

Closes #3781

### DIFF
--- a/tools/gengen/util/gengen_test.go
+++ b/tools/gengen/util/gengen_test.go
@@ -71,13 +71,12 @@ func TestGenGenLoading(t *testing.T) {
 	assert.Contains(t, stdout, builtin.InitActorCodeID.String())
 }
 
-func TestGenGenDeterministicBetweenBuilds(t *testing.T) {
-	tf.UnitTest(t)
-	t.Skip("Non-deterministic BLS key generation https://github.com/filecoin-project/go-filecoin/issues/3781")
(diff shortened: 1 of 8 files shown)


tools/gengen/util/gengen_test.go
```diff
@@ -71,13 +71,12 @@ func TestGenGenLoading(t *testing.T) {
 	assert.Contains(t, stdout, builtin.InitActorCodeID.String())
 }
 
-func TestGenGenDeterministicBetweenBuilds(t *testing.T) {
-	tf.UnitTest(t)
-	t.Skip("Non-deterministic BLS key generation https://github.com/filecoin-project/go-filecoin/issues/3781")
```
ZenGround0, line 76: 🎉

PR #4036, opened 2020-04-28
Upgrade to specs-actors v0.2.0 and associated dependencies

anorth:
There was a bit of upstream change coupled with the v0.2 actors that I had to adapt too as well.

FYI @hannahhoward @shannonwells @frrist


internal/app/go-filecoin/connectors/retrieval_market/client.go
```diff
@@ -49,28 +49,32 @@ func NewRetrievalClientConnector(
...
-			return address.Undef, errors.New("not enough funds in wallet")
+			return address.Undef, cid.Undef, errors.New("not enough funds in wallet")
 		}
 		return r.paychMgr.CreatePaymentChannel(clientAddress, minerAddress, clientFundsAvailable)
 	}
-	return chinfo.UniqueAddr, nil
+	// TODO: I think this is supposed to return the message CID from the creation of the payment channel.
+	// If there was one already, what CID should this return?
+	// What happens if that message is subsequently re-orged out of the chain and never replayed?
+	// https://github.com/filecoin-project/go-filecoin/issues/4034
```
acruikshank, line 76: It looks like go-fil-markets ignores the cid if the address is defined. For the second question, our message wait logic could possibly use a confidence interval. It could be implemented with the chain scheduler.

internal/app/go-filecoin/paymentchannel/manager.go
```diff
@@ -170,10 +170,10 @@ func (pm *Manager) CreatePaymentChannel(clientAddress, minerAddress address.Addr
 
 	err = pm.waiter.Wait(pm.ctx, msgCid, handleResult)
```
acruikshank, line 171: I believe this Wait is redundant. It waits here, and then go-fil-markets has it wait again (which works because the waiter looks backwards in the chain then waits). Not a problem with this PR obviously. There's no reason to return a message cid that's already been found in the chain.

internal/app/go-filecoin/connectors/retrieval_market/client.go
```diff
@@ -127,6 +131,16 @@ func (r *RetrievalClientConnector) CreatePaymentVoucher(ctx context.Context, pay
 	return &v, nil
 }
 
+func (r *RetrievalClientConnector) WaitForPaymentChannelCreation(messageCID cid.Cid) (address.Address, error) {
+	// TODO https://github.com/filecoin-project/go-filecoin/issues/4034
+	panic("implement me")
```
acruikshank, line 136: It's a little disturbing that these aren't tested.

internal/app/go-filecoin/connectors/retrieval_market/client.go
anorth: I'll leave this comment here with the issue for @shannonwells to confirm if it's ok as is.

I am very uncomfortable with any logic that watches for message CIDs instead of observing the state changes directly. At least unless that logic also handles reverting of that message up to finality in the future. A confidence interval < finality doesn't really help: if/when the interval *is* exceeded, will the module be in a legal state at all?

I'm increasingly aware that the modules use this pattern and am worried.

internal/app/go-filecoin/paymentchannel/manager.go
anorth: FYI @shannonwells , I'm not confident to remove it.

shannonwells: This wait is changed in my current branch to be more like what's  in lotus.

internal/app/go-filecoin/connectors/retrieval_market/client.go
shannonwells: this is Coming Soon via my current branch

internal/app/go-filecoin/connectors/retrieval_market/client.go
hannahhoward: @anorth yes, to be clear, all modules including storage miner wait for messages, and the main change here is that we are no longer have the wait for message hidden inside the CreatePaymentChannel call itself. This allows use to track whether we have at least attempted to post to chain already, and separate the confirmation phase form the posting phase.

We are absolutely working towards the change you describe -- handling reverts up to finality, and handling timeouts in the case of the message never appearing on chain. We have a PR merged in our `development` branch that does much of these this on the storage side. Getting these waiting calls separated from the original posting functions is what this PR does.

PR #4037, opened 2020-04-28
Feat/window post

acruikshank:
### Motivation

Poster needs to be updated to the new Window PoSt construction. 

### Proposed changes

Implement Window PoSt and reinstate TestBootstrapWindowedPoSt to confirm it works.

### DIFF
--- a/internal/pkg/poster/poster.go
+++ b/internal/pkg/poster/poster.go
@@ -1,22 +1,34 @@
 package poster
 
 import (
+	"bytes"
 	"context"
 	"sync"
 
-	sectorstorage "github.com/filecoin-project/sector-storage"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/gas"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	logging "github.com/ipfs/go-log"
 
 	"github.com/filecoin-project/go-address"
+	sectorstorage "github.com/filecoin-project/sector-storage"
+
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/plumbing/cst"
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/plumbing/msg"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/message"
 	appstate "github.com/filecoin-project/go-filecoin/internal/pkg/state"
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	acrypto "github.com/filecoin-project/specs-actors/actors/crypto"
 )
+
+var log = logging.Logger("poster")
+
+// epochs to wait past proving period to avoid invalid posts due to reorgs
+var confidenceInterval uint64 = 7
 
 // Poster listens for changes to the chain head and generates and submits a PoSt if one is required.
 type Poster struct {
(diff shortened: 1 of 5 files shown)


internal/pkg/poster/poster.go
```diff
@@ -70,148 +81,146 @@ func (p *Poster) StopPoSting() {
...
+		return
+	}
+
+	randomness, err := p.chain.SampleChainRandomness(ctx, head, acrypto.DomainSeparationTag_WindowedPoStChallengeSeed, challengeAt, buf.Bytes())
+	if err != nil {
+		log.Errorf("could not sample chain randomness at %d: %s", challengeAt, err)
+		return
+	}
+
+	proofs, err := p.mgr.GenerateWindowPoSt(ctx, abi.ActorID(minerID), sectors, abi.PoStRandomness(randomness))
```
anorth, line 165: You'll need to pass the on-chain miner faults and recoveries bitfields in here (or maybe a prior call when `sectors` is calculated) in order to calculate the right proof. I think that's worth doing now even though this impl doesn't yet report faults. The sector info for any faulty sector must be replaced by the info for the first non-faulty sector (see the actor code which does the same).

internal/pkg/poster/poster.go
```diff
@@ -70,148 +81,146 @@ func (p *Poster) StopPoSting() {
...
+		return err
 	}
 
-	copy(challengeSeed[:], randomness)
-	return challengeSeed, nil
+	return nil
+}
+
+func (p *Poster) getProvingSet(ctx context.Context, stateView *appstate.View) ([]abi.SectorInfo, error) {
+	return consensus.NewPowerTableView(stateView, stateView).SortedSectorInfos(ctx, p.minerAddr)
```
anorth, line 223: Thankfully, this method seems to be unused. We should only load the sector info for sectors being proven.

internal/pkg/poster/poster.go
```diff
@@ -70,148 +81,146 @@ func (p *Poster) StopPoSting() {
...
+	index, _, _, challenge, err := stateView.MinerDeadlineInfo(ctx, p.minerAddr, tipsetHeight)
+	if err != nil {
+		return nil
+	}
+
+	// check that we haven't already seen this deadline
+	if int64(index) == p.poStDeadlineIndex {
+		return nil
+	}
+	p.poStDeadlineIndex = int64(index)
```
anorth, line 112: I infer that this logic at the moment won't handle a re-org that causes a different challenge seed for the post. It'll compute the (now-incorrect) proof, submit it, but not notice that it should retry. 

Ok for now if so, but could you please file an issue outlining the known shortcomings?

internal/pkg/poster/poster.go
```diff
@@ -70,148 +81,146 @@ func (p *Poster) StopPoSting() {
...
+		return err
+	}
+
+	root, err := p.chain.GetTipSetStateRoot(ctx, newHead.Key())
+	if err != nil {
+		return err
+	}
+
+	stateView := p.stateViewer.StateView(root)
+	index, _, _, challenge, err := stateView.MinerDeadlineInfo(ctx, p.minerAddr, tipsetHeight)
```
anorth, line 103: This will need to handle the incomplete proving period right after genesis, when no proofs are due. The actors code now specifies a `ProvingPeriodStart` which, if it's in the future, means nothing to do.

acruikshank: I'm unclear what this means. Currently the miner starts with no sectors in its deadlines. Those don't get added until the cron job sets up the next proving period. With no sectors due, the miner doesn't PoSt in that incomplete proving period. Is this not the correct behavior? Does the genesis block start with miners that have sectors due that are supposed to be ignored?

internal/pkg/poster/poster.go
```diff
@@ -70,148 +81,146 @@ func (p *Poster) StopPoSting() {
...
+	sectors, err := stateView.MinerSectorInfoForDeadline(ctx, p.minerAddr, deadlineIndex, partitions)
+	if err != nil {
+		log.Errorf("error retrieving sector info for miner %s partitions at index %d: %s", p.minerAddr, deadlineIndex, err)
+		return
+	}
+
+	// if no sectors, we're done
+	if len(sectors) == 0 {
+		return
+	}
```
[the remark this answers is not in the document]
acruikshank: removed

internal/pkg/poster/poster.go
acruikshank: We now do the set logic in the state view. This is just copied from the actor. It's a lot of logic and probably should be a method on miner state.

internal/pkg/poster/poster.go
acruikshank: removed

internal/pkg/poster/poster.go
acruikshank: I added some logic to restart the PoSt if the chain randomness changes within the same deadline (implying a reorg). This plus the confidence interval, should make reorg issues unlikely.

internal/pkg/state/view.go
```diff
@@ -145,7 +146,142 @@ func (v *View) MinerSectorCount(ctx context.Context, maddr addr.Address) (int, e
...
+
+	partitions := make([]uint64, partitionCount)
+	for i := uint64(0); i < partitionCount; i++ {
+		partitions[i] = start + i
+	}
+
+	return partitions, err
+}
+
+// nerSectorInfoForPartitions retrieves sector info for sectors needed to be proven over for the given proving window partitions
```
ZenGround0, line 201: nit Miner

internal/pkg/state/view.go
```diff
@@ -145,7 +146,142 @@ func (v *View) MinerSectorCount(ctx context.Context, maddr addr.Address) (int, e
 	return count, err
 }
 
-// DeadlineInfo returns information about the next proving period
+// MinerDeadlineInfo returns all partitions that need to be proven for the
```
ZenGround0, line 149: nit: finish sentence

internal/pkg/poster/poster.go
```diff
@@ -70,154 +80,159 @@ func (p *Poster) StopPoSting() {
...
+	if tipsetHeight < open {
+		return nil
+	}
+
+	randomness, err := p.getChallenge(ctx, newHead.Key(), challengeAt)
+	if err != nil {
+		return err
+	}
+
+	// If we have already seen this randomness, either the deadline has changed or
```
ZenGround0, line 117: nit: or or

internal/pkg/poster/poster.go
```diff
@@ -70,154 +80,159 @@ func (p *Poster) StopPoSting() {
...
+	// or the chain as reorged to a point prior to the challenge. Either way,
+	// it is time to start a new PoSt.
+	if bytes.Equal(p.challenge, randomness) {
```
ZenGround0, line 120: I'm not sure I see how the comment and code line up.  It seems like "we have already seen this randomness" would correspond to this if statement condition, but in that case we *do not* start a new PoSt.

internal/pkg/poster/poster.go
```diff
@@ -70,154 +80,159 @@ func (p *Poster) StopPoSting() {
...
-	//	return nil
-	//}
-	//
-	//ctx, p.postCancel = context.WithCancel(ctx)
-	//go p.doPoSt(ctx, stateView, provingPeriodStart, newHead.Key())
-	//
+	p.postMutex.Lock()
+	defer p.postMutex.Unlock()
+
+	if p.postCancel != nil {
```
ZenGround0, line 86: Where does this get reset to nil?

internal/pkg/poster/poster.go
```diff
@@ -70,154 +80,159 @@ func (p *Poster) StopPoSting() {
...
+		log.Errorf("Error retrieving partitions for address %s at index %d: %s", p.minerAddr, deadlineIndex, err)
+		return
+	}
+
+	// if no partitions, we're done
+	if len(partitions) == 0 {
+		return
+	}
+
+	// Some day we might want to choose a subset of partitions to prove at one time. Today is not that day.
```
ZenGround0, line 154: sounds like an issue

internal/pkg/poster/poster.go
acruikshank: We call `p.cancelPost()` instead of `p.postCancel()`. It gets set to nil in that method.

internal/pkg/poster/poster.go
acruikshank: good call. I added a "not".

internal/pkg/poster/poster.go
acruikshank: I'm not sure. The only real reason I can think to do this is parallelization of PoSts within a single deadline. Miners are going to have to get pretty big before they'll even have more than one partition, so this seems like a pretty distant concern.

internal/pkg/poster/poster.go
anorth: You're right, the staging of genesis sectors in NewSectors handles this for us.

PR #4040, opened 2020-04-29
fix: update invoke actor state after transfer

frrist:
The actor state references held by `invocationContext` were outdated after calls to transfer. This prevented reward funds from being distributed to miners by the reward actor.

### DIFF
--- a/internal/pkg/vm/internal/vmcontext/invocation_context.go
+++ b/internal/pkg/vm/internal/vmcontext/invocation_context.go
@@ -189,7 +189,9 @@ func (ctx *invocationContext) invoke() (ret returnWrapper, errcode exitcode.Exit
 	ctx.toActor, ctx.msg.to = ctx.resolveTarget(ctx.msg.to)
 
 	// 3. transfer funds carried by the msg
-	ctx.rt.transfer(ctx.msg.from, ctx.msg.to, ctx.msg.value)
+	if !ctx.msg.value.Nil() && !ctx.msg.value.IsZero() {
+		ctx.toActor, ctx.fromActor = ctx.rt.transfer(ctx.msg.from, ctx.msg.to, ctx.msg.value)
+	}
 
 	// 4. if we are just sending funds, there is nothing else to do.
 	if ctx.msg.method == builtin.MethodSend {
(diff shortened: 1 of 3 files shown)


internal/pkg/vm/internal/vmcontext/invocation_context.go
```diff
@@ -191,6 +191,17 @@ func (ctx *invocationContext) invoke() (ret returnWrapper, errcode exitcode.Exit
...
+	var found bool
+	ctx.toActor, found, err = ctx.rt.state.GetActor(ctx.rt.context, ctx.msg.to)
+	if !found {
+		runtime.Abort(exitcode.ErrNotFound)
+	}
+
+	ctx.fromActor, found, err = ctx.rt.state.GetActor(ctx.rt.context, ctx.msg.from)
+	if !found {
+		runtime.Abort(exitcode.ErrNotFound)
+	}
```
anorth, line 203: Both the actor states are loaded in the `transfer` call. Return them from that instead of loading them again here.

frrist: They are loaded iff transfer is called with a non-zero amount, else it bails:
https://github.com/filecoin-project/go-filecoin/blob/ecfbd2bf822f835c825809fadec7c04a0f13a013/internal/pkg/vm/internal/vmcontext/vmcontext.go#L525-L534

frrist: couple ideas:
1. Load them before the `if amount.Nil() || amount.IsZero()` check. seems equally as bad
2. Return a boolean in addition to the actors indicating if state changed (ie transfer amount was non-zero)
3. Larger refactor than is here

anorth: Ok lets do the check outside of transfer. One of the three call sites already does this, and the other one will never transfer zero.

internal/pkg/vm/internal/vmcontext/vmcontext.go
```diff
@@ -511,28 +514,27 @@ func (vm *VM) applyMessage(msg *types.UnsignedMessage, onChainMsgSize int, rnd c
 	// 2. settle gas money around (unused_gas -> sender)
 	receipt.GasUsed = gasTank.GasConsumed()
 	refundGas := msg.GasLimit - receipt.GasUsed
-	vm.transfer(builtin.RewardActorAddr, msg.From, refundGas.ToTokens(msg.GasPrice))
+	amount := refundGas.ToTokens(msg.GasPrice)
+	if !amount.Nil() && !amount.IsZero() {
```
anorth, line 518: The nil checks here aren't necessary, but no harm either.

PR #4042, opened 2020-04-29
Upgrade to specs-actors v0.3.0

anorth:
This is fairly painless. Tests won't pass until https://github.com/filecoin-project/chain-validation/pull/179 lands and this PR also consumes the appropriate chain validation tests.

It would be beneficial to also consume compatible upstream repos at the same time, though in this case I think the changes are small enough that it's also safe not to.
- [ ] https://github.com/filecoin-project/filecoin-ffi/pull/81
- [ ] https://github.com/filecoin-project/specs-storage/pull/8
- [x] https://github.com/filecoin-project/go-fil-markets/pull/207
- [ ] https://github.com/filecoin-project/storage-fsm/pull/15
- [x] https://github.com/filecoin-project/sector-storage/pull/21

### DIFF
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -157,7 +157,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 			Address:    maddr,
 			Owner:      status.OwnerAddress,
 			Worker:     status.WorkerAddress,
-			SectorSize: uint64(status.SectorSize),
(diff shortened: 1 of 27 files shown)


anorth:
Apologies, it seems I hadn't traced through all the impact of the upgrade, possibly because the build was prevented by chain validation. I've changed more now.

I'm going to land it despite being somewhat different to what you reviewed, @acruikshank, so we can progress with interop, but would appreciate another quick look and any comments.

PR #4044, opened 2020-04-29
Feat/interop bug fix

ZenGround0:
### Motivation
Drand config setup bug.  Interop header field miscommunication handled.

### Proposed changes

Closes #

### DIFF
--- a/internal/app/go-filecoin/node/builder.go
+++ b/internal/app/go-filecoin/node/builder.go
@@ -201,6 +201,23 @@ func (b *Builder) build(ctx context.Context) (*Node, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to build node.Chain")
 	}
+	if b.drand == nil {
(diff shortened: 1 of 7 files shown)


internal/app/go-filecoin/node/builder.go
```diff
@@ -201,6 +201,23 @@ func (b *Builder) build(ctx context.Context) (*Node, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to build node.Chain")
 	}
+	if b.drand == nil {
```
acruikshank, line 204: Consider moving this block into the `Drand()` method itself (making it a lazy initializer). This seems like too much logic to include in this already long function.

ZenGround0: In retrospect this isn't worth it yet.  This PR is brining the drand setup back to where it was first introduced because moving it to sync module was a bug.  Doing this right requires more untangling than meets the eye and untangling the build process should be approached wholistically and at too low priority to do right now.

PR #4047, opened 2020-04-30
Only charge gas in CreateActor once we've made a state change

arajasek:
There's no real reason to charge gas until an actual state change occurs. More context can be found [here](https://filecoinproject.slack.com/archives/CEHHJNJS3/p1588216005182400).


anorth:
A minor tweak: I suggest we charge immediately before SetActor. The general rationale for this is that we should check that sufficient gas limit is available _before_ doing the expensive work, else malicious messages can cause a lot of work to happen without paying for it. It doesn't make a big difference in this case, but would prefer we model the pattern consistently.

In general I think the desired pattern is:
- do all the cheap precondition checks possible
- check sufficient gas for the expensive operation (by charging it)
- do the expensive operation, confident it will be paid for

arajasek:
Change made, thanks! :)

Issue #4053, opened 2020-05-01
Simplify DRAND processing code now the protocol has no gaps


anorth:
When we first implemented DRAND integration the protocol had gaps (like Filecoin null rounds) if the network was not available. Now it's changed to not have gaps, but rush catch-up instead.

We have a bunch of code for gap-handling that's no longer necessary. We should remove it from the codebase to prevent confusion and possible bugs in now-unnecessary code.

PR #4054, opened 2020-05-01
Update to drand 0.8.1

### LINKED ISSUE #4048: Drand Upgrade
### Description
Biggest change is that we will never skip a round which simplifies things drastically.
Other change is that we no longer get previous round from response

anorth:
### Motivation
DRAND protocol changed to remove the parent round and have no gaps.

This is the minimum change to consume the new protocol. The changes mean we now have a bunch of unnecessary catch-up logic; clean-up is deferred to #4053

### Proposed changes
Remove use of the DRAND previous round number. Assume rounds are contiguous.

Closes #4048


internal/pkg/block/block.go
```diff
@@ -33,9 +33,9 @@ type Block struct {
 	// ElectionProof is the vrf proof giving this block's miner authoring rights
 	ElectionProof *crypto.ElectionProof
 
-	// DrandEntries contain the verifiable oracle randomness used to elect
+	// BeaconEntries contain the verifiable oracle randomness used to elect
 	// this block's author leader
-	DrandEntries []*drand.Entry
```
anorth, line 38: I'd prefer the nomenclature less tightly coupled to DRAND throughout. I played with renaming the `drand` package to `beacon` but it was too much noise in this change.

ZenGround0: Yes this is a good idea. I'll file an issue

internal/pkg/drand/drand_grpc.go
```diff
@@ -213,6 +212,8 @@ func (d *GRPC) StartTimeOfRound(round Round) time.Time {
 // 1. RoundsInInterval is called on an interval extending into the future
 // 2. There is an ongoing drand outage causing a gap in drand rounds
 // 3. There is a network partition blocking this GPRC from the latest rounds
+// TODO remove gap-handling logic now that the protocol has no gaps
+// https://github.com/filecoin-project/go-filecoin/issues/4053
```
ZenGround0, line 216: 🎉

ZenGround0: #4055

PR #4056, opened 2020-05-01
update ffr after live config

ZenGround0:
### Motivation
If you call `drand configure` on a daemon and do not restart it the first filecoin round property of the in memory drand grpc construct does not get properly reset so the first filecoin round value gets set to a very high number which makes mining fail. 

### Proposed changes
To fix this we do the calculation along with other resets after fetching the config.  We take advantage of this PR to abstract that ffr calculation and setting to a subfunction of drand grpc.

Closes #

### DIFF
--- a/internal/pkg/drand/drand_grpc.go
+++ b/internal/pkg/drand/drand_grpc.go
@@ -35,6 +35,8 @@ type GRPC struct {
 
 	// The time of the 0th round of the DRAND chain
 	genesisTime time.Time
+	// the time of genesis block of the Filecoin chain
+	filecoinGenesisTime time.Time
 	// The DRAND round first included in the filecoin blockchain
 	firstFilecoin Round
 	// Duration of a round in this DRAND network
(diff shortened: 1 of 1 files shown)


internal/pkg/drand/drand_grpc.go
```diff
@@ -56,26 +58,33 @@ func NewGRPC(addresses []Address, distKeyCoeff [][]byte, drandGenTime time.Time,
...
-	return &GRPC{
-		addresses:     addresses,
-		client:        core.NewGrpcClient(),
-		key:           distKey,
-		genesisTime:   drandGenTime,
-		roundTime:     rd,
-		firstFilecoin: ffr,
-		cache:         make(map[Round]*Entry),
-	}, nil
+	d.firstFilecoin = ffr
```
acruikshank, line 86: nit: there doesn't seem any reason to assign results[0] to `ffr` here.

PR #4059, opened 2020-05-01
rand bytes should match

ZenGround0:
### Motivation
Randomness is underspecced and ambiguous

Through agreement with lotus gfc is moving to lotus's current strategy of taking undigested ticket bytes as input to the randomness mixer.  Tests against latest lotus interopnet confirm this passes validation.

### Proposed changes

Slight refactor to sample ticket rather than mixed ticket.  Turned out to be pretty simple because of the previous change giving the ticket machine direct access to chain sampler.

Closes #

### DIFF
--- a/internal/app/go-filecoin/plumbing/cst/chain_state.go
+++ b/internal/app/go-filecoin/plumbing/cst/chain_state.go
@@ -167,7 +167,7 @@ func (chn *ChainStateReadWriter) SampleChainRandomness(ctx context.Context, head
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


anorth:
I'm landing this for short-term interop needs. However I'm pretty sure this is not what the spec wants. The randomness spec is currently https://github.com/filecoin-project/specs/pull/883, which states:

> any use of a ticket's value, is that of its VRFDigest's bytes

anorth:
After looking at this lots, I can see that this usage of randomness is outside what the spec issue covers, but still the inconsistency is worrying.

sternhenri:
Yes, it sseems like we're missing a hash here to use the digests rather than proof. This would help should we ever choose to change the VRF (to one in which the proof would not work as it does now). 

Arguably this is fine, but it'll mean a larger change in spec should we change VRF...

PR #4063, opened 2020-05-04
Restore faucet test as an in-process integration test

anorth:
### Motivation
I decided to bring back some tests. I chose this one because it's the last thing pinning `testhelpers/iptbtester`, which we can delete after this PR in favour of in-process or daemon tests.

### Proposed changes
Rewrite faucet test as in-process.

- [x] Rebase on #4062

### DIFF
--- a/functional-tests/faucet_test.go
+++ b/functional-tests/faucet_test.go
@@ -1,8 +1,10 @@
 package functional
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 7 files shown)


functional-tests/faucet_test.go
```diff
@@ -12,129 +14,111 @@ import (
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/drand"
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 )
 
 var faucetBinary = "../tools/faucet/faucet"
 
 func TestFaucetSendFunds(t *testing.T) {
 	tf.FunctionalTest(t)
-	t.Skip("Dragons: fake proofs???")
```
acruikshank, line 27: ❤️

functional-tests/mining_chain_test.go
```diff
@@ -24,21 +22,17 @@ import (
 )
 
 func TestSingleMiner(t *testing.T) {
-	tf.FunctionalTest(t)
+	tf.IntegrationTest(t)
```
acruikshank, line 25: This is good. We should should reserve the functional test flag for tests that take over 5 seconds or so. But it no longer makes sense to keep it in this file in the functional tests director. It should probably live in `miner_integration_test.go` and be called `TestMinerMineOnce`.

internal/app/go-filecoin/node/block_propagate_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func TestBlockPropsManyNodes(t *testing.T) {
-	tf.UnitTest(t)
+	tf.IntegrationTest(t)
```
acruikshank, line 30: 🙏

functional-tests/mining_chain_test.go
anorth: I agree that we should probably re-arrange the tests a bit, but leaving that for later.

PR #4064, opened 2020-05-04
Improve message wait to fix miner create integration test flakiness

### LINKED ISSUE #1518: message/waiter.go is not scalable
It walks the chain. It should use an index.

acruikshank:
### Motivation

The miner create test is hanging intermittently. Some investigation revealed a problem in message wait that prevents the miner create command from seeing the create message land on chain in certain circumstances and that failure results in a hang.

The circumstances are that the node receiving tipsets from the miner misses a few tipsets and is forced to catch up. The message wait only reports messages found in the head tipset. If the message lands in one of the catch-up tipsets, that tipset is never reported as the head, so the message wait never returns.

The solution is to fix message wait to check all tipsets for messages back to the last common ancestor of the last tipset it saw.

Additionally, this PR addresses a long standing problem with the message wait functionality that causes it looks back in the chain back to genesis for the message before beginning to wait. Some amount of previous chain examination is required to avoid race conditions where the message is processed and lands on chain before the wait begins. This race is unlikely to involve more than a single round. This PR adds a lookback interval to configure exactly how far to look back, just in case.

### Proposed changes

1. Message wait only examines a small configurable number of tipsets prior to waiting.
2. When handling a new tipset, message wait checks previous tipsets back to the greatest common ancestor of the previous head (this accomodates re-orgs and catch-up intervals).
3. Reinstate TestMinerCreateIntegration test.
4. Update TestMinerCreateIntegration to support new proof type parameter.
5. Remove unneeded and inefficient message.Finds.

Closes #1518

### DIFF
--- a/cmd/go-filecoin/message.go
+++ b/cmd/go-filecoin/message.go
@@ -239,7 +239,6 @@ type MessageStatusResult struct {
 	PoolMsg   *types.SignedMessage
 	InOutbox  bool // Whether the message is found in the outbox
 	OutboxMsg *message.Queued
-	OnChain   bool // Whether the message is found on chain
 	ChainMsg  *msg.ChainMessage
 }
 
--- a/cmd/go-filecoin/message.go
+++ b/cmd/go-filecoin/message.go
@@ -277,11 +277,6 @@ var msgStatusCmd = &cmds.Command{
 			}
 		}
 
-		// Look on chain
-		result.ChainMsg, result.OnChain, err = api.MessageFind(req.Context, msgCid)
-		if err != nil {
-			return err
-		}
 		return re.Emit(&result)
 	},
 	Type: &MessageStatusResult{},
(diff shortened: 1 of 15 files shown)


cmd/go-filecoin/message.go
```diff
@@ -277,11 +277,6 @@ var msgStatusCmd = &cmds.Command{
 			}
 		}
 
-		// Look on chain
```
ZenGround0, line 280: If we don't want this in the message status command anymore we should remove the `OnChain` bool from the status reporting type.

internal/app/go-filecoin/plumbing/msg/waiter.go
```diff
@@ -100,23 +108,24 @@ func (w *Waiter) WaitPredicate(ctx context.Context, pred WaitPredicate, cb func(
...
 
-	return w.WaitPredicate(ctx, pred, cb)
+	return w.WaitPredicate(ctx, lookback, pred, cb)
 }
 
 // findMessage looks for a matching in the chain and returns the message,
 // block and receipt, when it is found. Returns the found message/block or nil
 // if now block with the given CID exists in the chain.
-func (w *Waiter) findMessage(ctx context.Context, ts block.TipSet, pred WaitPredicate) (*ChainMessage, bool, error) {
+// The lookback parameter is the number of tipsets this method will check before giving up.
```
ZenGround0, line 125: nit: the number of tipsets in the past this method will check before giving up

in my understanding the lookback shouldn't limit the total number of not yet synced blocks we wait for the message.

PR #4077, opened 2020-05-05
Upgrade markets to 0.1.3

hannahhoward:
### Motivation

Maintain compatibility for Graphsync with LOTUS, which is now on v0.2.0 of go-fil-markets.

### Proposed changes

This upgrades the go-fil-markets module for v0.1.3, a hotfix release that updates transitive dependencies go-ipld-prime, go-graphsync, and go-data-transfer.

It also solves at least one potential graphsync bug that @acruikshank encountered when running TestMinerCreateIntegration at count 20.

Under the hood, `go-ipld-prime` has received a significant rewrite for performance and memory usage.

go-data-transfer now also maintains it's transfer counter on disk to persist between restarts

I am concerned 0.2.0 and 0.1.1 (current GFC version) will not be compatible for graphsync, which would prevent chain sync.


go.mod
```diff
@@ -21,9 +21,9 @@ require (
 	github.com/filecoin-project/go-amt-ipld/v2 v2.0.1-0.20200131012142-05d80eeccc5e
 	github.com/filecoin-project/go-bitfield v0.0.0-20200416002808-b3ee67ec9060
 	github.com/filecoin-project/go-crypto v0.0.0-20191218222705-effae4ea9f03
-	github.com/filecoin-project/go-data-transfer v0.0.0-20200408061858-82c58b423ca6
+	github.com/filecoin-project/go-data-transfer v0.2.1
```
acruikshank, line 24: ❤️

PR #4080, opened 2020-05-06
Fixing state accessors and updating chain val

ZenGround0:
### Motivation
Updating v0.4.0 to match lotus and join interopnet

### Proposed changes
- Fix state view
- Update chain val
- Fix storage connector to get provider deal ids without using now gone deals by party multimap.

### DIFF
--- a/internal/pkg/state/view.go
+++ b/internal/pkg/state/view.go
@@ -403,19 +403,11 @@ func (v *View) MarketHasDealID(ctx context.Context, addr addr.Address, dealID ab
 	}
 
 	found := false
-	byParty, err := market.AsSetMultimap(StoreFromCbor(ctx, v.ipldStore), marketState.DealIDsByParty)
+	dealStates, err := v.asDealStateArray(ctx, marketState.States)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/pkg/state/view.go
```diff
@@ -403,19 +403,11 @@ func (v *View) MarketHasDealID(ctx context.Context, addr addr.Address, dealID ab
 	}
 
 	found := false
-	byParty, err := market.AsSetMultimap(StoreFromCbor(ctx, v.ipldStore), marketState.DealIDsByParty)
+	dealStates, err := v.asDealStateArray(ctx, marketState.States)
```
ZenGround0, line 406: @anorth I think that this will achieve the same result but I'm going off of an uninformed interpretation of market actor state names, so you should double check that this is ok

anorth: I don't know why this was done in such a complicated way in the first place.

anorth:
I wrote half this PR myself, but am going to invoke maintainer privilege to land it so we can deploy a new network with actors v0.4

PR #4085, opened 2020-05-07
Max mpool count *= 100

ZenGround0:
### Motivation
Logs spam error `message pool is full` regularly because our default is below what typically occurs on a devnet.

### Proposed changes
Increase the limit

Closes #


ZenGround0:
Yeah this might be masking a problem we can fix.  I will file an issue to investigate before merging

ZenGround0:
#4087

PR #4090, opened 2020-05-08
Bug/max tx size error

ZenGround0:
### Motivation
gfc halts with a badger datastore error when executing messages during sync

### Proposed changes
Fix badger datastore issue by limiting the blocks we pass to PutMany when writing actor state

Closes #

### DIFF
--- a/internal/pkg/vm/internal/storage/storage.go
+++ b/internal/pkg/vm/internal/storage/storage.go
@@ -131,12 +131,22 @@ func (s *VMStorage) Flush() error {
 	for _, nd := range s.writeBuffer {
 		blks = append(blks, nd)
 	}
-	// write objects to store
-	if err := s.blockstore.PutMany(blks); err != nil {
-		return err
-	}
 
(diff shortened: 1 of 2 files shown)


internal/pkg/vm/internal/storage/storage.go
```diff
@@ -131,9 +131,21 @@ func (s *VMStorage) Flush() error {
...
+		index := i * maxBatchSize
+		if index+maxBatchSize < len(blks) {
+			if err := s.blockstore.PutMany(blks[index : index+maxBatchSize]); err != nil {
+				return err
+			}
+		} else if index != len(blks) {
+			if err := s.blockstore.PutMany(blks[index:]); err != nil {
+				return err
+			}
+		}
```
anorth, line 148: The indexing might come out simpler if you slice the slice iteratively, like
```
remaining := blks
for len(rest) > 0 {
   last = min(len(remaining), maxBatchSize)
  PutMany(remaining[:last])
  remaining = remaining[last:]
}
```

PR #4097, opened 2020-05-11
enforce a message limit on blocks

arajasek:
This is a thing Lotus does, and so is needed for painnet interop.

### DIFF
--- a/internal/pkg/block/block.go
+++ b/internal/pkg/block/block.go
@@ -19,7 +19,7 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 )
 
-// TODO: If this is gonna stay, it should move to specs-actors
+// BlockMessageLimit is the maximum number of messages in a block
 const BlockMessageLimit = 512
 
 // Block is a block in the blockchain.
--- a/internal/pkg/block/block.go
+++ b/internal/pkg/block/block.go
@@ -19,6 +19,9 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 )
 
+// TODO: If this is gonna stay, it should move to specs-actors
+const BlockMessageLimit = 512
+
 // Block is a block in the blockchain.
 type Block struct {
 	// control field for encoding struct as an array
--- a/internal/pkg/chainsync/fetcher/graphsync_fetcher.go
+++ b/internal/pkg/chainsync/fetcher/graphsync_fetcher.go
@@ -3,7 +3,6 @@ package fetcher
 import (
 	"context"
 	"fmt"
-	"golang.org/x/xerrors"
 	"sync"
 	"time"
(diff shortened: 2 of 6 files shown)


internal/pkg/mining/block_generate.go
```diff
@@ -55,6 +55,9 @@ func (w *DefaultWorker) Generate(
 	mq := NewMessageQueue(pending)
 	candidateMsgs := orderMessageCandidates(mq.Drain())
 	candidateMsgs = w.filterPenalizableMessages(ctx, candidateMsgs)
+	if len(candidateMsgs) > block.BlockMessageLimit {
```
ZenGround0, line 58: We should pass the size limit as a parameter to Drain so that we don't take messages out of the pool and then drop them.  It's ok to merge as is, I will file an issue.

internal/pkg/block/block.go
```diff
@@ -19,6 +19,9 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/encoding"
 )
 
+// TODO: If this is gonna stay, it should move to specs-actors
```
anorth, line 22: No, it wouldn't go there because that code won't reference or enforce it. Here is ok for now.

internal/pkg/chainsync/fetcher/graphsync_fetcher.go
```diff
@@ -436,8 +440,13 @@ func (gsf *GraphSyncFetcher) loadAndVerifyFullBlock(ctx context.Context, key blo
 				return err
 			}
 
+			messageCount += len(messages)
 			return nil
 		})
+
+	if messageCount > block.BlockMessageLimit {
+		return block.UndefTipSet, nil, xerrors.New("Number of messages in block exceeds block message limit")
+	}
```
anorth, line 449: I don't think this is the right place to do this. If we had a non-graphsync fetcher, or obtained blocks in any way, we still need to check this. We should be able to validate blocks produced locally.

This should be part of either the syntax or semantic validator (depending whether we give the former access to a store so it can look up the AMT root and `Count` it.) Neither have store access yet, but the semantic one will definitely need it. We currently do that kind of thing in `Expected.ValidateMining`.

anorth:
We're probably better off with this than without it. Please put a TODO in the graphsync fetcher where this is being done and point to #4098.

internal/pkg/chainsync/fetcher/graphsync_fetcher.go
ZenGround0: I'll push a commit moving this to expected to group it with everybody else.

internal/pkg/mining/block_generate.go
ZenGround0: I'm going to take a stab at doing this.  Will back out if it gets too crazy.

PR #4100, opened 2020-05-12
update specs-actors, ffi, chain-val

frrist:
### Motivation
Compatibility with lotus

### Proposed changes
Update filecoin-ffi, specs-actors and chain-validation

### DIFF
--- a/fixtures/setup.json
+++ b/fixtures/setup.json
@@ -12,12 +12,12 @@
   ],
   "miners": [{
     "owner": 5,
-    "sealProofType": 3,
+    "sealProofType": 1,
     "committedSectors": [{
       "commR": {"/":"bafk4ehzaqugph7z7yajugw6wz6nn7x3akr3e32wecwexllfgqndnjt3nsaya"},
       "commD": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
       "sectorNum": 3,
-      "proofType": 3,
+      "proofType": 1,
       "dealCfg": {
         "commP": {"/":"bafk4chzavlx2s3zsomp6beohyevpbql477qyr7qnvgkvfwbbcqzfijo62arq"},
         "pieceSize": 2048,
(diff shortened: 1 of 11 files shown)


fixtures/setup.json
```diff
@@ -12,12 +12,12 @@
   ],
   "miners": [{
     "owner": 5,
-    "sealProofType": 3,
+    "sealProofType": 1,
```
anorth, line 15: No, don't do this. The sectors are 2KiB.

anorth: See https://github.com/filecoin-project/specs-actors/pull/363/files/217948049d2bf4c7a0ef660836d0ee9574911aa2#diff-2ecd9908929bba03f96f2c075d8d7ac5 for what we should be doing (in the test entry point)

internal/app/go-filecoin/node/builder.go
```diff
@@ -124,6 +128,32 @@ func JournalConfigOption(jrl journal.Journal) BuilderOpt {
...
+				miner.SupportedProofTypes[abi.RegisteredProof(proofType)] = struct{}{}
+			}
+		}
+		return nil
+	}
+}
+
+// MonkeyPatchAddProofTypeOption returns a function that sets package variable
+// SuppurtedProofTypes to include the given registered proof type
+func MonkeyPatchAddProofTypeOption(proofType abi.RegisteredProof) BuilderOpt {
```
anorth, line 150: I would replace rather than add here too, but AFAICT this is fine in the places it's used.

We'll likely get a chain validation test one day that asserts that an invalid proof type is rejected, where that invalid type is actually valid in testnet.

tools/gengen/util/generator.go
```diff
@@ -98,6 +98,13 @@ func (g *GenesisGenerator) Init(cfg *GenesisCfg) error {
 	}
 	g.vrkey = &vrKey
 
+	// Monkey patch all proof types into the specs-actors package variable
+	for _, mCfg := range cfg.Miners {
+		if _, ok := miner.SupportedProofTypes[mCfg.SealProofType]; !ok {
+			miner.SupportedProofTypes[mCfg.SealProofType] = struct{}{}
```
anorth, line 104: lolsigh, this caused concurrent map modification in tests. I changed it to replace the map fully.

PR #4102, opened 2020-05-12
WIP implement chain-val rpc server & match interfaces

frrist:
### Motivation
demonstrate an RPC server against chain-validation


ZenGround0:
Might be useful to come back to but closing down for now fyi @frrist

PR #4115, opened 2020-05-14
Configuration values for interop-net bootstrap and drand

### LINKED ISSUE #3863: Avoid importing fixtures into main binary
The `fixtures` package is imported to the go-filecoin binary and all the `init()` methods in `constants.go`  execute when running go-filecoin. This package should not be imported into the binary!

anorth:
### Motivation
Adding bootstrap and DRAND config for long-running networks. The values all come from @travisperson's work on interop and testnet deployments.

Changed the default sector size configuration to enable 2KiB sectors so that local nodes can be set-up without hand-editing the conf file. The long-running networks will likely have an option similar to `--devnet-interop` that sets the right value for each network.

### Proposed changes
Closes #3863 (but I'll follow up to make it more robust)


functional-tests/network-deployment/relay_check_test.go
```diff
@@ -303,12 +303,8 @@ func networkBootstrapPeers(network string) []string {
...
 
 	switch network {
-	case "nightly":
-		return fixtures.DevnetNightlyBootstrapAddrs
-	case "staging":
-		return fixtures.DevnetStagingBootstrapAddrs
-	case "user":
-		return fixtures.DevnetUserBootstrapAddrs
+	case "interop":
+		return networks.InteropNet.Bootstrap.Addresses
```
anorth, line 307: @travisperson we should consider whether these network deployment tests are worth keeping in this form - I'd appreciate your advice (to follow-up).

internal/app/go-filecoin/connectors/retrieval_market/provider_test.go
```diff
@@ -49,7 +50,12 @@ func TestRetrievalProviderConnector_UnsealSector(t *testing.T) {
 	tf.UnitTest(t)
 	ctx := context.Background()
 	sectorID := rand.Uint64()
-	fixtureFile := "../../../../../fixtures/constants.go"
```
anorth, line 52: This blew my mind. I think I've captured the intent of the test without referring to this file now.

anorth:
@travisperson do you have the values for testnet-2 that we can add to a similar network configuration in a follow-up PR?

functional-tests/network-deployment/relay_check_test.go
travisperson: This test can definitely go. We are not using relays.

travisperson:
@anorth Not at the moment.

Min miner is 1TiB, supported sectors will be 32, 64, and there will be different peers, drand will be the same.

I'll have the peer ids later.

PR #4119, opened 2020-05-14
FSM connector ignores missing deal state

anorth:
### Motivation
The FSM ignores the deal state, and calls `StateMarketStorageDeal` before the state is actually present on chain. Lotus does [something like this](https://github.com/filecoin-project/lotus/blob/d124d10f7c9542d80bcecdf427c78b25f449f3e7/chain/stmgr/utils.go#L271-L309).

### Proposed changes
Return a dummy deal state if not found. I didn't avoid looking it up altogether in case the FSM resolution ends up keeping this method for some case when it _does_ expect the state to be found.

FYI @hannahhoward


hannahhoward:
@anorth Can I go ahead and merge this? (after rebase)

PR #4131, opened 2020-05-14
Update go-fil-markets & solve single host data transfer issues

hannahhoward:
### Motivation

Update to latest go-fil-markets and make sure we have no issues arising from having a single host

### Proposed changes

- Update fil-markets to v0.2.5
- Use facilities in v0.2.5 to share a data-transfer instance between mining and client (I suspect this would have caused making deals as a client from a node running as a miner to break -- not sure that anyone has run this)
- Add some nice event logging for what's going on with deals (storage module has a subscribable interface for getting events on deals -- this could be hooked into an actual command line or HTTP api but here it's just used to make some nice logs)
- Remove an expired TODO (this was addressed a long time ago)

Closes #

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
@@ -8,6 +8,7 @@ import (
 	"github.com/filecoin-project/go-storedcounter"
 
 	"github.com/filecoin-project/go-address"
+	datatransfer "github.com/filecoin-project/go-data-transfer"
 	graphsyncimpl "github.com/filecoin-project/go-data-transfer/impl/graphsync"
 	"github.com/filecoin-project/go-fil-markets/filestore"
 	"github.com/filecoin-project/go-fil-markets/piecestore"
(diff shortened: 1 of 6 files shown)


internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -138,3 +149,19 @@ func (sm *StorageProtocolSubmodule) PieceManager() (piecemanager.PieceManager, e
...
+
+func (sm *StorageProtocolSubmodule) DataTransfer() datatransfer.Manager {
+	return sm.dataTransfer
+}
+
+func (sm *StorageProtocolSubmodule) RequestValidator() datatransfer.RequestValidator {
+	return sm.requestValidator
+}
+
+func (sm *StorageProtocolSubmodule) clientEventLogger(event iface.ClientEvent, deal iface.ClientDeal) {
```
acruikshank, line 161: These should be implemented in the `StorageClientNodeConnector`. Implementing them here adds complexity to the nodes relationship to the storage market.

internal/app/go-filecoin/internal/submodule/storage_protocol_submodule.go
```diff
@@ -138,3 +149,19 @@ func (sm *StorageProtocolSubmodule) PieceManager() (piecemanager.PieceManager, e
 	}
 	return sm.pieceManager, nil
 }
+
+func (sm *StorageProtocolSubmodule) DataTransfer() datatransfer.Manager {
```
acruikshank, line 153: This and the following `RequestValidator` don't appear to be used. If they are needed outside this submodule it would be better to move data transfer into the networking submodule. Otherwise, they should be removed.

hannahhoward:
@acruikshank addressed comments

PR #4136, opened 2020-05-15
make inaccurate prooftype look less wrong

### LINKED ISSUE #4127: Propose Storage Deal command appears to hard code 2K sectors
**Describe the bug**:

https://github.com/filecoin-project/go-filecoin/blob/master/cmd/go-filecoin/client.go#L194

It appears when you propose a deal it always uses the Dev2K proof value.

**Expected behavior**:

The value of the proof for the propose deal would either be manual or taken from the miner's information

acruikshank:
### Motivation

The propose-deal command had been passing in a test registered proof type. This proof only specified the proof version (StackedDRG) so the piece commitment could be calculated correctly. This looks wrong, and it would be wrong once the network supports different proof versions.

### Proposed changes

Use the miner's registered proof type for piece commitment calculation, which ensures they will be in sync.

Closes #4127

### DIFF
--- a/cmd/go-filecoin/client.go
+++ b/cmd/go-filecoin/client.go
@@ -189,9 +189,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 			abi.ChainEpoch(end),
 			price,
 			collateral,
-			// This constant represents StackedDRG encoding for the piece commitment computation.
-			// It's sector size and proof type are irrelevant.
-			abi.RegisteredProof_StackedDRG64GiBSeal,
+			status.MinerInfo.SealProofType,
 		)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


cmd/go-filecoin/client.go
```diff
@@ -190,8 +189,7 @@ be 2, 1 hour would be 120, and 1 day would be 2880.
 			abi.ChainEpoch(end),
 			price,
 			collateral,
-			// proof version (not circuit or size) should be all that is important here
-			constants.DevRegisteredWindowPoStProof,
+			status.MinerInfo.SealProofType,
```
ZenGround0, line 192: Looks good.  One thing to check is that you are now passing a seal proof instead of a post proof.  ProposeStorageDeal probably handles this internally but could be a problem.

acruikshank: It actually should be a piece commitment proof type (which doesn't exist). Hopefully it's at least less confusing than it was.

PR #4139, opened 2020-05-15
mine block before epoch

acruikshank:
### Motivation

Our scheduling for block mining is a bit off. Miners are allowed to send blocks as soon as the epoch begins, which means they should mine the block in the previous epoch. Since mining requires parent blocks to arrive from other miners, we need a configurable propagation delay to trade off between waiting long enough for all the blocks to arrive and providing sufficient time to mine the block

We also had a problem where mining in the scheduler was run in a go routine. This would allow overlapping attempts to mine when the timing is off. This potentially causes a panic.

### Proposed changes

1. Schedule mining synchronously.
2. Add propagation delay parameter alongside related block delay parameter
3. Give chain clock ability to wait for epoch + propagation delay
4. Modify scheduler to wait for propagation delay, then fetch the head, then mine, then wait for the start of the target epoch, then send the block.
5. Fix some tests and remove others that are no longer relevant.

### DIFF
--- a/functional-tests/drand_test.go
+++ b/functional-tests/drand_test.go
@@ -21,6 +21,7 @@ func TestDrandPublic(t *testing.T) {
 	ctx := context.Background()
 	genTime := int64(1000000000)
 	blockTime := 30 * time.Second
+	propDelay := 6 * time.Second
 	// The clock is intentionally set some way ahead of the genesis time so the miner can produce
 	// catch-up blocks as quickly as possible.
 	fakeClock := clock.NewFake(time.Unix(genTime, 0).Add(4 * time.Hour))
@@ -29,7 +30,7 @@ func TestDrandPublic(t *testing.T) {
 	// Future code could decouple the whole setup.json from the presealed information.
 	genCfg := loadGenesisConfig(t, fixtureGenCfg())
 	seed := node.MakeChainSeed(t, genCfg)
-	chainClock := clock.NewChainClockFromClock(uint64(genTime), blockTime, fakeClock)
+	chainClock := clock.NewChainClockFromClock(uint64(genTime), blockTime, propDelay, fakeClock)
 
 	nd := makeNode(ctx, t, seed, chainClock, nil)
 
--- a/functional-tests/mining_chain_test.go
+++ b/functional-tests/mining_chain_test.go
@@ -5,28 +5,29 @@ import (
 	"testing"
 	"time"
 
-	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node/test"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/drand"
-	"github.com/filecoin-project/go-filecoin/internal/pkg/proofs"
-
 	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
 	"github.com/stretchr/testify/require"
 	"gotest.tools/assert"
 
 	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node"
+	"github.com/filecoin-project/go-filecoin/internal/app/go-filecoin/node/test"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/consensus"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/drand"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/proofs"
 	th "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers"
 	tf "github.com/filecoin-project/go-filecoin/internal/pkg/testhelpers/testflags"
 )
+
+const defaultBlockTime = builtin.EpochDurationSeconds
+const defaultPropDelay = 6 * time.Second
 
 func TestSingleMiner(t *testing.T) {
 	tf.IntegrationTest(t)
 	ctx := context.Background()
 	genTime := int64(1000000000)
-	blockTime := 30 * time.Second
-	propDelay := 6 * time.Second
 	// The clock is intentionally set some way ahead of the genesis time so the miner can produce
 	// catch-up blocks as quickly as possible.
 	fakeClock := clock.NewFake(time.Unix(genTime, 0).Add(4 * time.Hour))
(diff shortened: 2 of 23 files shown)


functional-tests/mining_chain_test.go
```diff
@@ -26,6 +26,7 @@ func TestSingleMiner(t *testing.T) {
 	ctx := context.Background()
 	genTime := int64(1000000000)
 	blockTime := 30 * time.Second
+	propDelay := 6 * time.Second
```
ZenGround0, line 29: nit: it feels like we should be referencing constants throughout these tests.

internal/app/go-filecoin/node/block_propagate_test.go
```diff
@@ -213,7 +214,8 @@ func makeNodesBlockPropTests(t *testing.T, numNodes int) (address.Address, []*No
 	genTime := time.Unix(genUnixSeconds, 0)
 	fc := clock.NewFake(genTime)
 	blockTime := 30 * time.Second
-	c := clock.NewChainClockFromClock(1234567890, blockTime, fc)
+	propDelay := 5 * time.Second
```
ZenGround0, line 217: any reason to be inconsistent here?  if not the standard 6 would be a little simpler

internal/pkg/clock/chainclock.go
```diff
@@ -77,18 +86,34 @@ func (cc *chainClock) StartTimeOfEpoch(e abi.ChainEpoch) time.Time {
...
+	cc.waitForEpochOffset(ctx, e, 0)
+}
+
+// WaitForEpochPropDelay returns propDelay time after the start of the epoch, or when ctx is done.
+func (cc *chainClock) WaitForEpochPropDelay(ctx context.Context, e abi.ChainEpoch) {
+	cc.waitForEpochOffset(ctx, e, cc.propDelay)
+}
+
+// waitNextEpochOffset returns when time is offset past the start of the epoch, or ctx is done.
+func (cc *chainClock) waitForEpochOffset(ctx context.Context, e abi.ChainEpoch, offset time.Duration) {
```
ZenGround0, line 108: this interface and decomposition is very nice

internal/pkg/clock/chainclock.go
```diff
@@ -77,18 +86,34 @@ func (cc *chainClock) StartTimeOfEpoch(e abi.ChainEpoch) time.Time {
...
+	epochStart := cc.StartTimeOfEpoch(e).Add(offset)
```
ZenGround0, line 109: nit: the name is not exactly right int the case there is an offset.  Hard to think of a better one though.

internal/pkg/mining/scheduler.go
```diff
@@ -51,56 +61,77 @@ func (s *timingScheduler) Start(miningCtx context.Context) (<-chan Output, *sync
...
 }
 
-func (s *timingScheduler) mineLoop(miningCtx context.Context, outCh chan Output, doneWg *sync.WaitGroup) {
-	// mineLoop is the main event loop for the timing scheduler.  It waits for
-	// a new epoch to start, polls the heaviest head, includes the correct number
+func (s *timingScheduler) mineLoop(ctx context.Context, outCh chan Output) error {
+	mineCh := make(chan Output, 1)
+
+	// start on epoch boundary
+	targetEpoch := s.chainClock.WaitNextEpoch(ctx)
```
ZenGround0, line 79: Fine for now but it looks like this is not setup to handle mining block 1 during the genesis epoch.  Probably best to catch that edge case with work to do "catchup mining"

internal/pkg/mining/scheduler.go
```diff
@@ -51,56 +61,77 @@ func (s *timingScheduler) Start(miningCtx context.Context) (<-chan Output, *sync
...
+		// block time validation should prevent us from seeing a base height than we expect
+		if baseHeight >= targetEpoch {
+			log.Errorf("Scheduled target epoch %d is not greater than base height %d", targetEpoch, baseHeight)
+		}
+
+		// null count is the number of epochs between the one we are mining and the one now.
+		nullCount := uint64(targetEpoch-baseHeight) - 1
+
+		// mine now
+		s.worker.Mine(ctx, base, nullCount, mineCh)
```
ZenGround0, line 124: Since we are now synchronously calling `Mine` we should file an issue to just turn it into a regular function call and cut away all of non-idiomatic and verbose error handling.

ZenGround0:
@acruikshank another issue we should file for following up is simplifying the mine once code.  We can get rid of the bad error swallowing we uncovered a few weeks ago by abstracting out some pieces of this new scheduler.

functional-tests/drand_test.go
```diff
@@ -21,6 +21,7 @@ func TestDrandPublic(t *testing.T) {
 	ctx := context.Background()
 	genTime := int64(1000000000)
 	blockTime := 30 * time.Second
```
anorth, line 23: Since you're touching all these files, use `builtin.EpochDurationSeconds` ?

functional-tests/plege_sector_test.go
```diff
@@ -24,6 +24,7 @@ func TestMiningPledgeSector(t *testing.T) {
 
 	genTime := int64(1000000000)
 	blockTime := 1 * time.Second
+	propDelay := 200 * time.Millisecond
```
anorth, line 27: Hmm, this smells flaky. Given the fake clock I would be much happier if this were set to zero but everything still worked, meaning that the MineOnce underneath here didn't return until the block had been consumed and synced by the local node.

I guess we don't quite have the hookups to achieve that yet.

internal/pkg/clock/chainclock.go
```diff
@@ -26,22 +31,26 @@ type chainClock struct {
 	genesisTime time.Time
 	// The fixed time length of the epoch window
 	epochDuration time.Duration
+	// propDelay is the time between the start of the epoch and the start
+	// of mining to wait for parent blocks to arrive
```
anorth, line 35: "... the start of mining for the subsequent epoch"

internal/pkg/mining/scheduler.go
```diff
@@ -51,56 +61,77 @@ func (s *timingScheduler) Start(miningCtx context.Context) (<-chan Output, *sync
...
 	}()
 	s.isStarted = true
 
 	return outCh, &doneWg
 }
 
-func (s *timingScheduler) mineLoop(miningCtx context.Context, outCh chan Output, doneWg *sync.WaitGroup) {
-	// mineLoop is the main event loop for the timing scheduler.  It waits for
-	// a new epoch to start, polls the heaviest head, includes the correct number
+func (s *timingScheduler) mineLoop(ctx context.Context, outCh chan Output) error {
```
anorth, line 75: Just FYI I am hoping to get rid of this channel and replace it with a synchronous block consuming function.

internal/pkg/mining/scheduler.go
anorth: Yes please. Please follow up with this right away.

internal/pkg/mining/scheduler_test.go
```diff
@@ -17,33 +17,34 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 )
 
+const epochDuration = 1 * time.Second
```
anorth, line 20: With a fake clock we should be able to use long durations here but still ahve the test be fast. If that's not true, please mark as an integration test, and file an issue or follow up to make it true.

anorth:
Thanks. Please remind me to redeploy our interop mining node with this after it lands.

internal/pkg/mining/scheduler.go
acruikshank: #4145

Issue #4141, opened 2020-05-18
Exclude new sectors from proving set for Winning PoSt


anorth:
The set of sectors we compute challenge indices from in winning post includes "NewSectors", which should not be included until they are assigned to a deadline (which happens on proving period boundary).

Lotus does the same right now, so we replicate this for interop.

PR #4142, opened 2020-05-18
Compute winning post proving set with faults

### LINKED ISSUE #4018: Don't load entire sector metadata when sampling few
### Description
During winning post creation and validation (and likely other places) we rely on the entire sector set being passed around.  We predict this will be a scalability bottleneck and need to adjust our design to accommodate this.  

### Acceptance criteria
State viewer must allow selective reading of sector info.  Constructs wrapping state view need to be updated to support selective sector loading.  The election machine type bundling election methods needs to be adjusted to either (1) expose the challenge generate ffi call to mining and validation, or (2) take ownership of state viewing and subsume power table operations.

### Risks + pitfalls
Currently we have messy over complex chain state access management among mining (worker), validation (expected) and the election machine.  However this might be a great opportunity to simplify.

anorth:
### Motivation
We were failing to exclude faulty sectors from the set available to be challenged by Winning PoSt. Behaviour diverged from Lotus in the presence of faulty sectors on testnet.

@arajasek proposed a quick workaround in a8541a78dd300535e0abeedbc7a60030b8e7fc09 but I've taken the opportunity to avoid the inefficient loading of every sector into from storage when we're only going to challenge one of them.

### Proposed changes
Re-arrange election machine logic to accept a state view so the challenge logic can load only the sector/s being challenged. Compute the proving set in the election machine. Exclude faulty sectors but include "new" sectors to replicate existing testnet behaviour (see #4141).

Closes #4018

FYI @ZenGround0


anorth:
I've confirmed this is syncs on testnet (prior halted around epoch 4600)

internal/pkg/consensus/ticket.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
+)
+
+type ChainSampler interface {
+	SampleTicket(ctx context.Context, head block.TipSetKey, epoch abi.ChainEpoch) (block.Ticket, error)
+}
+
+// TicketMachine uses a VRF and VDF to generate deterministic, unpredictable
+// and time delayed tickets and validates these tickets.
+type TicketMachine struct {
```
acruikshank, line 26: Nice refactor

PR #4144, opened 2020-05-18
Support building with Go tip

iand:
The minimum version of Go required to build is Go 1.13.1. This is
checked by explicitly parsing the version string. However when
using Go built from tip the version string has a different format:

devel +bb59a1360a Mon May 18 09:13:38 2020 +0000 linux/amd64

This change uses build tags to always pass the version check if
Go 1.14 or later is being used which also includes Go tip.

### Motivation

Support development by users who track the Go repository

### Proposed changes

Closes #


iand:
> Nice, and thanks!
> 
> Could we use the same trick to replace the semver-parsing logic that hands 1.13?

Unfortunately not. The build tags only exist for major versions and that parsing code required 1.13.1, presumably due to security fixes.

PR #4146, opened 2020-05-18
remove channel from worker mine

### LINKED ISSUE #4145: worker.Mine should return rather than emit output
### Description

The mining loop is not synchronous so the worker no longer needs to communicate the result of mining via a channel. Returning the output of mining will permit better error handling.

### Acceptance criteria

worker.Mine returns the mined block or a error and no longer communicates via channel to its caller.

acruikshank:
### Motivation

Block mining used to occur within nested goroutines. This was causing problems, so it has been simplified to run synchronously. This makes the channel communication it used to use to communicate errors and output unnecessary. This PR removes that channel and replaces it with simple return statements.

### Proposed changes

1. Remove channel argument from worker Generate function and return block header, messsages and error directly.
2. Move `Output` struct to scheduler (which still needs to communicate via channel) and rename to `FullBlock`.
3. Remove channel argument from `worker.MIne` and return header, messages and errors instead.
4. Have scheduler log its own errors and signal its shutdown by closing its output channel (which will trigger a shutdown of mining).
5. __Important__: The scheduler no longer treats an error in mining as unrecoverable. Instead it logs the error and continues in hopes that it will be able to recover by the next epoch. The upside is a lower chance of missing mining opportunities because due to a one off error. The downside is receiving a new error per epoch due to an unrecoverable error.

Closes #4145

### DIFF
--- a/internal/app/go-filecoin/node/node.go
+++ b/internal/app/go-filecoin/node/node.go
@@ -237,15 +237,15 @@ func (node *Node) handleNewMiningOutput(ctx context.Context, miningOutCh <-chan 
 				log.Errorf("scheduler stopped. stopping mining.")
 				node.StopMining(context.Background())
 				return
-			} else {
-				node.BlockMining.MiningDoneWg.Add(1)
-				go func() {
-					if node.IsMining() {
-						node.BlockMining.AddNewlyMinedBlock(ctx, output)
-					}
-					node.BlockMining.MiningDoneWg.Done()
-				}()
 			}
+
+			node.BlockMining.MiningDoneWg.Add(1)
+			go func() {
+				if node.IsMining() {
+					node.BlockMining.AddNewlyMinedBlock(ctx, output)
+				}
+				node.BlockMining.MiningDoneWg.Done()
+			}()
 		}
 	}
 
--- a/internal/app/go-filecoin/node/node.go
+++ b/internal/app/go-filecoin/node/node.go
@@ -224,7 +224,7 @@ func (node *Node) setIsMining(isMining bool) {
 	node.BlockMining.Mining.IsMining = isMining
 }
 
-func (node *Node) handleNewMiningOutput(ctx context.Context, miningOutCh <-chan mining.Output) {
+func (node *Node) handleNewMiningOutput(ctx context.Context, miningOutCh <-chan mining.FullBlock) {
 	defer func() {
 		node.BlockMining.MiningDoneWg.Done()
 	}()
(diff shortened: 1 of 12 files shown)


internal/app/go-filecoin/node/node.go
```diff
@@ -234,20 +234,18 @@ func (node *Node) handleNewMiningOutput(ctx context.Context, miningOutCh <-chan
...
+				return
 			}
+
+			node.BlockMining.MiningDoneWg.Add(1)
+			go func() {
+				if node.IsMining() {
+					node.BlockMining.AddNewlyMinedBlock(ctx, output)
+				}
+				node.BlockMining.MiningDoneWg.Done()
+			}()
```
anorth, line 248: Removing this goroutine later will probably be necessary for us to be able to do rush catch-up mining in a busy loop.

internal/pkg/mining/block_generate.go
```diff
@@ -30,7 +30,7 @@ func (w *DefaultWorker) Generate(
 	nullBlockCount abi.ChainEpoch,
 	posts []block.PoStProof,
 	drandEntries []*drand.Entry,
-) Output {
+) (*block.Block, []*types.SignedMessage, []*types.SignedMessage, error) {
```
anorth, line 33: Use `FullBlock`, it comes from the same package.

internal/pkg/mining/scheduler.go
```diff
@@ -9,15 +9,28 @@ import (
 	"context"
 	"sync"
 
+	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
```
anorth, line 12: Organise imports

internal/pkg/mining/scheduler.go
```diff
@@ -54,16 +67,17 @@ type timingScheduler struct {
 // It returns a channel for reading mining outputs and a waitgroup for teardown.
 // It is the callers responsibility to close the out channel only after waiting
 // on the waitgroup.
-func (s *timingScheduler) Start(miningCtx context.Context) (<-chan Output, *sync.WaitGroup) {
+func (s *timingScheduler) Start(miningCtx context.Context) (<-chan FullBlock, *sync.WaitGroup) {
```
anorth, line 70: I think the channel returned here can be replaced by a callback function/handler passed to `Start` (follow-up).

acruikshank: It is true that this channel could be replaced with a callback, but I disagree that it would be preferable. this channel is used to communicate multiple values from goroutine. It's as good a use for a channel as I can imagine. A callback will need to replicate the close behavior and probably introduce locking. I'm strongly in favor of reducing complexity where we can, but switching to a callback seems a lateral move at best and is less idiomatic. If you have reasons why callbacks would be superior in these situations, we should discuss. It represents a shift in pattern that either should be replicated across the code base or left alone.

anorth: Good points. I wish for a mechanism where a new block emitted by our miner can be processed by the chain sync system synchronously before the miner then starts the next cycle, so it can deterministically build on top of that block. Especially for testing. In production maybe we would have the callback just plop the block on a channel, but the point is that the caller gets to decide whether to introduce async.

Issue #2195, opened 2019-03-07
Don't color logs sent to a non-tty


anorth:
**Describe the bug**
Filecoin node logs, which are printed to stderr, include tags for coloring in a TTY. These make the logs hard to read in a non-TTY context.

**To Reproduce**
1. Run `./go-filecoin daemon 2>stderr.txt` for 30 seconds then kill it
2. `cat stderr.txt`, observe color because your shell is interpreting the color codes
3. `open stderr.txt` (or windows equivalent) to observe codes messing up the log in TextEdit

**Expected behavior**
Logs printed to an output that is not a TTY are not colored.

**Examples**
```
badger 2019/03/07 14:25:44 INFO: All 0 tables opened in 0s
badger 2019/03/07 14:25:44 INFO: Replaying file id: 0 at offset: 0
badger 2019/03/07 14:25:44 INFO: Replay took: 28.565µs
[0;37m14:25:44.085  INFO [0;34mchain.stor: [0mstart loading chain at tipset: { zDPWYqFCzyRooJH4TCQMFnrae4kjWLPDHRZjdYSLAYijK8s5rGsn }, height: 6005 [0;37mdefault_store.go:119[0m
[0;37m14:25:44.087  INFO [0;34mchain.stor: [0mload tipset: zDPWYqFCtGdMCWH1HCjwuHC7J7GRkaJqwzanATLTdsTyCrCAa8cp, height: 6000 [0;37mdefault_store.go:126[0m
[0;37m14:25:44.215  INFO [0;34mchain.stor: [0mload tipset: zDPWYqFCtvXuRmdLYjFwDCGzvaGLUV5rCgvtmEasXs8TqsgqViKL, height: 5400 [0;37mdefault_store.go:126[0m
[0;37m14:25:44.339  INFO [0;34mchain.stor: [0mload tipset: zDPWYqFCwyrCxW8rahkXSZrYeUTsjtkQ5n4isZPEdTEFZojYFpZs, height: 4800 [0;37mdefault_store.go:126[0m
[0;37m14:25:44.618  INFO [0;34mchain.stor: [0mload tipset: zDPWYqFCsFaNUKdTCZzmcLwsyxbWoDFCgRTdj65YYirMcUxqg9je, height: 3600 [0;37mdefault_store.go:126[0m
[...]

PR #4154, opened 2020-05-20
Upgrade to github.com/ipfs/go-log/v2

iand:
For the most part the upgrade is a mechanical replacement of the
package import.

This change fixes log tail which was broken due to the earlier changes
in go-log.

The --elstdout daemon flag is now marked as deprecated since go-log v2
only allows writing to stderr currently. A possible workaround is to
redirect stderr to stdout when invoking go-filecoin.

Fixes https://github.com/filecoin-project/go-filecoin/issues/2195
since colors can be disabled by setting GOLOG_LOG_FMT=nocolor

### Proposed changes

Closes #


cmd/go-filecoin/log.go
```diff
@@ -38,14 +36,12 @@ Outputs event log messages (not other log messages) as they are generated.
 	},
 
 	Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
-		r, w := io.Pipe()
+		r := logging.NewPipeReader()
```
frrist, line 39: I think this creates a reader for leveled log messages produced by go-log, right? If so the `ShortDescription` and `Tagline` fileds in the `cmdkit.HelpText` struct above should be updated to state this.

iand: Done

PR #4155, opened 2020-05-20
Use latest go-fil-markets

shannonwells:
### Motivation
Upgrade to latest go-fil-markets

### Proposed changes
Upgrade to latest go-fil-markets and make any needed changes to get it working.

Changes include:
- We've now moved prefixing datastores largely out of the markets modules, to allow node implementations to customize how they want to prefix their datastores - the module constructors have been updated minimally to stay compatible with the previous release (any greater customization such as setting up in the config file is left to node maintainers)
- The stored ask is now constructed seperately from the storage provider. We've update to constructors with a minimal implementation here, though node maintainers want to interact directly with the ask it's an option
- Per https://github.com/filecoin-project/go-fil-markets/blob/master/storagemarket/impl/provider.go#L92 there is now a mechanism to add custom decision logic to accepting deals. However, we've left it up to node maintainers to implement this, if they want

Closes #

### DIFF
--- a/internal/app/go-filecoin/internal/submodule/retrieval_protocol_submodule.go
+++ b/internal/app/go-filecoin/internal/submodule/retrieval_protocol_submodule.go
@@ -18,6 +18,15 @@ import (
 	"github.com/filecoin-project/go-filecoin/internal/pkg/piecemanager"
 )
 
+// RetrievalProviderDSPrefix is a prefix for all datastore keys related to the retrieval provider
+const RetrievalProviderDSPrefix = "retrievalmarket/provider"
+
+// RetrievalCounterDSKey is the datastore key for the stored counter used by the retrieval counter
+const RetrievalCounterDSKey = "retrievalmarket/client/counter"
+
+// RetrievalClientDSPrefix is a prefix for all datastore keys related to the retrieval clients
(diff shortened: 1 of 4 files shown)


internal/app/go-filecoin/internal/submodule/retrieval_protocol_submodule.go
```diff
@@ -36,22 +46,23 @@ func NewRetrievalProtocolSubmodule(
...
 
-	retrievalDealPieceStore := piecestore.NewPieceStore(ds)
+	retrievalDealPieceStore := piecestore.NewPieceStore(namespace.Wrap(ds, datastore.NewKey(PieceStoreDSPrefix)))
 
 	netwk := network.NewFromLibp2pHost(host)
 	pnode := retmkt.NewRetrievalProviderConnector(netwk, pieceManager, bs, pchMgrAPI, nil)
 
-	// TODO: use latest go-fil-markets with persisted deal store
-	marketProvider, err := impl.NewProvider(providerAddr, pnode, netwk, retrievalDealPieceStore, bs, ds)
+	// TODO: use latest go-fil-markets with persisted deal store\
```
acruikshank, line 54: Can this TODO go away? It shouldn't end in a slash at any rate.

Issue #934, opened 2018-09-19
Create "reward" message for mining rewards


lkowalick:
### Description

We add a "reward" message to provide a reward to a block miner. This is semanticly meaningful message from the filecoin network to the Miner Actor that mined the block.  The processing of this message then transfers the funds to the miner's owner.

### Acceptance criteria

1. Mine a block

1. See a "reward" message from the filecoin network to the miner actor along with the transference of the block reward to the miner actor.

1. See a transfer of funds from the miner actor to to the miner actor owner's account.

### Risks + pitfalls

A lot of tests have addresses for miners that are not associated to a MinerActor's address in the system.

### Where to begin

Replace the block reward in `block_generate` with a message that has a method of `"reward"`

PR #4158, opened 2020-05-21
add block message validation to semantic validation

acruikshank:
### Motivation

We need to check call sequence numbers and a few other things on incoming blocks before we declare the tipset ready for processing. This falls under the category of semantic validation so add a new `ValidateMessagesSemantic` function to be run on each block after messages have been fetched for the tipset. 

partially addresses #934

### DIFF
--- a/internal/pkg/consensus/block_validation.go
+++ b/internal/pkg/consensus/block_validation.go
@@ -4,10 +4,27 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm/actor"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-filecoin/internal/pkg/vm"
+	"github.com/ipfs/go-cid"
+	"github.com/pkg/errors"
+
 	"github.com/filecoin-project/go-filecoin/internal/pkg/block"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/clock"
 	"github.com/filecoin-project/go-filecoin/internal/pkg/types"
 )
+
+type messageStore interface {
+	LoadMessages(context.Context, cid.Cid) ([]*types.SignedMessage, []*types.UnsignedMessage, error)
+	LoadReceipts(context.Context, cid.Cid) ([]vm.MessageReceipt, error)
+}
+
+type chainState interface {
+	GetActorAt(ctx context.Context, tipKey block.TipSetKey, addr address.Address) (*actor.Actor, error)
+}
 
 // BlockValidator defines an interface used to validate a blocks syntax and
 // semantics.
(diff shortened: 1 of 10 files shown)


internal/pkg/consensus/block_validation.go
```diff
@@ -104,6 +126,81 @@ func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *bl
...
+		if err != nil {
+			return errors.Wrapf(err, "message %s of block %s invalid", msgCid, child.Cid())
+		}
+	}
+
+	return nil
+}
+
+func (dv *DefaultBlockValidator) getAndValidateFromActor(ctx context.Context, msg *types.UnsignedMessage, parents block.TipSetKey) (*actor.Actor, error) {
+	actor, err := dv.cs.GetActorAt(ctx, parents, msg.From)
```
anorth, line 176: This needs to resolve pubkey addresses to ID addresses via the init actor state, which I don't think the underlying chain store does. It may be worth doing state access here via state view methods, which could do the resolution.

internal/pkg/consensus/block_validation.go
```diff
@@ -104,6 +126,81 @@ func (dv *DefaultBlockValidator) ValidateSemantic(ctx context.Context, child *bl
...
+
+func (dv *DefaultBlockValidator) validateMessage(msg *types.UnsignedMessage, expectedCallSeqNum map[address.Address]uint64, fromActor *actor.Actor) error {
+	callSeq, ok := expectedCallSeqNum[msg.From]
+	if !ok {
+		callSeq = fromActor.CallSeqNum
+	}
+
+	// ensure message is in the correct order
+	if callSeq != msg.CallSeqNum {
+		return fmt.Errorf("callseqnum (%d) out of order (expected %d)", msg.CallSeqNum, callSeq)
```
anorth, line 197: Include `msg.From` in the error message

internal/pkg/consensus/block_validation_test.go
```diff
@@ -29,33 +33,163 @@ func TestBlockValidSemantic(t *testing.T) {
...
+			blsMessages: []*types.UnsignedMessage{msg2},
+		}, &fakeChainState{
+			actor: newActor(t, 0, 2),
+		})
+
+		err := validator.ValidateMessagesSemantic(ctx, c, parents.Key())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "out of order")
+	})
+}
```
anorth, line 183: Lovely cases. Could you add one with SECP sequence < BLS sequence for same actor (fails validation) and the reverse (passes)?

internal/pkg/consensus/block_validation.go
acruikshank: `GetActorAt` actually does the resolution internally, and finds it at the tipset specified by parents. It has to since actors are keyed in the state tree by ID address. I've confirmed that this code is covered by integration tests that fetch actual actors.

I'm wondering if we should create an actor IDAddress type definition in GFC (or possibly specs-actors) that would signal when an address is not going to get resolved. I foresee a lot more bugs and confusion around this.

PR #4164, opened 2020-05-28
Remove some commented out code better to leave behind

ZenGround0:
### Motivation
General cleanup so that go-filecoin will be easier to pick back up at a later date.

### Proposed changes
Delete old code and commented out code for clarity.
Change some skip comments to make it clear we are ready to reinstate tests with a few simple changes to node setup allowing fake proofs.

Closes #


internal/app/go-filecoin/plumbing/msg/previewer.go
```diff
@@ -43,32 +43,5 @@ func NewPreviewer(chainReader previewerChainReader, cst cbor.IpldStore, bs bstor
 
 // Preview sends a read-only message to an actor.
 func (p *Previewer) Preview(ctx context.Context, optFrom, to address.Address, method abi.MethodNum, params ...interface{}) (gas.Unit, error) {
```
acruikshank, line 45: We should either delete this method altogether or panic. Probably no one is going to believe the zero, but it's better to be broken than wrong.

internal/pkg/testhelpers/core.go
```diff
@@ -117,25 +58,13 @@ func RequireNewFakeActor(t *testing.T, vms vm.Storage, addr address.Address, cod
...
 // that its steps succeed.
 func RequireNewFakeActorWithTokens(t *testing.T, vms vm.Storage, addr address.Address, codeCid cid.Cid, amt types.AttoFIL) *actor.Actor {
-	// act := actor.NewActor(codeCid, amt)
-	// var err error
-	// rawHead, err := (&actor.FakeActor{}).InitializeState(vms, &actor.FakeActorStorage{})
-	// require.NoError(t, err)
-	// act.Head = e.NewCid(rawHead)
-	// require.NoError(t, vms.Flush())
-	// return act
+
```
acruikshank, line 61: I doubt these methods are being referenced. They should be deleted.

ZenGround0: I hear you but they're used in skipped tests I'd rather leave around to be reinstated.

PR #4166, opened 2020-06-01
restore or delete tests for chain ls

acruikshank:
### Motivation

* Chain ls tests had been skipped. Restore the ones that make sense and delete the rest.
* Deals show wasn't implement, but is easy to implement.
* Add a test for deals show in the integration file and delete the deals_daemon tests which were old-fashioned and broken.


cmd/go-filecoin/chain_integration_test.go
```diff
@@ -80,61 +77,33 @@ func TestChainLs(t *testing.T) {
 		assert.True(t, b[0].Parents.Empty())
 	})
 
-	t.Run("chain ls with text encoding returns only CIDs", func(t *testing.T) {
```
acruikshank, line 83: Chain ls no longer has a text encoding.

PR #4170, opened 2020-06-17
out of gas error

jackoelv:
when I debug the localnet, I have found that is the AOLimit set to 1, localnet would be fail because of out of gas.
when set to 300, ok.

### Proposed changes

Closes #


jackoelv:
hei, I am still testing the localnet. till now, I cannot run the localnet because of errors.
just finished fix some part of the code.

PR #4231, opened 2020-11-23
update readme

hunjixin:
1. rm localnet
2. update go version requirement


README.md
```diff
@@ -166,14 +166,6 @@ Open a new terminal to interact with your node:
 
 To see a full list of commands, run `./venus --help`.
 
-### Advanced usage
```
hunjixin, line 169: add back when miner is ok

PR #4243, opened 2020-11-30
upgrade fork & sync chain with lotus in net-2k mode

simlecode:
upgrade fork & sync chain with lotus in net-2k mode


hunjixin (approved):
LGTM

PR #4323, opened 2021-01-24
Small change: Update README.md

RobQuistNL:
### Motivation
There was too much that was replaced.


RobQuistNL:
Why did you close this exactly?

Or am I wrong and was `venus` previously called `venus`?

PR #4489, opened 2021-07-07
refactor remote jwt verify

zl03jsj:
### Motivation
service in venus cluster have same logical on remote jwt verification,
there are lotus of code in each service for same function.

### Proposed changes
extract those code into venus-auth as public module, reduce repetitive code.

Closes #


app/submodule/apiface/jwtauth.go
```diff
@@ -5,9 +5,9 @@ import (
 	"github.com/filecoin-project/go-jsonrpc/auth"
 )
 
-type IJwtAuthAPI interface {
+type IJwtAuthAPI = interface {
```
simlecode, line 8: `type IJwtAuthAPI = interface ` 该成 `type IJwtAuthAPI  interface `

simlecode (approved):

PR #4503, opened 2021-07-20
set default jwt-auth log level to info

zl03jsj:
### Proposed changes

Closes #


zl03jsj:
no need to set default log level, we can set log level through command ./venus log set-level

PR #4586, opened 2021-11-06
Fix/consensus fault error

hunjixin:
### Proposed changes

Closes #

### DIFF
--- a/pkg/consensusfault/check.go
+++ b/pkg/consensusfault/check.go
@@ -5,10 +5,16 @@ import (
 	"context"
 	"fmt"
 	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/network"
 	"github.com/filecoin-project/venus/pkg/fork"
 	"github.com/filecoin-project/venus/pkg/types"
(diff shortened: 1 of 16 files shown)


pkg/consensusfault/check.go
```diff
@@ -138,15 +144,36 @@ func (s *ConsensusFaultChecker) VerifyConsensusFault(ctx context.Context, h1, h2
 }
 
 // Checks whether a block header is correctly signed in the context of the parent state to which it refers.
-func verifyBlockSignature(ctx context.Context, view FaultStateView, blk types.BlockHeader, nv network.Version) error {
-	minerInfo, err := view.MinerInfo(ctx, blk.Miner, nv)
+func verifyBlockSignature(ctx context.Context, blk types.BlockHeader, nv network.Version, curEpoch abi.ChainEpoch, receiver address.Address, gasIpld cbornode.IpldStore, view FaultStateView, getter vmcontext.LookbackStateGetter) error {
+	if nv >= network.Version7 && blk.Height < -policy.ChainFinality {
```
Kubuxu, line 148: Here you are comparing reported fault epoch with negative Finality which will be always false.
It should probably be ‘currentEpoch - blk.Height < policy.ChainFinality’

PR #4640, opened 2021-12-24
tool for auto generate api document.

zl03jsj:
auto generate json-rpc document:
```shell
make api-docs
```


Fatman13:
fix https://github.com/filecoin-project/venus/issues/4608

PR #5114, opened 2022-07-20
fix: error message: correct error message 

StefanGajic:
fix return error message in slashfilter so if the error occurs correct message will be displayed


Fatman13:
Thanks for the PR!

PR #6538, opened 2026-05-09
feat: `eth_baseFee`

wjmelements:
Adds the nonstandard `eth_baseFee` eth jsonrpc method.
This method returns the calculated base fee of the next epoch.
See https://github.com/filecoin-project/lotus/pull/13615
#### Changes
* add `eth_baseFee`
* doc gen


LinZexiao (approved):
LGTM

